Issue #3, opened 2017-06-20
Dev Environment


mbarnett:
I'm currently using HN2's `hydra:server` rake task and jupiter on a non-standard port. We should figure out how we want to provide local Fedora/Solr for dev going forward (possibly a Docker container? or port the rake task code?)

mbarnett:
We can just continue w/ the rake task route for the time being, but if we have the bandwidth to get something else in place quickly, now would be a great time. It just needs a stock Fedora and Solr on accessible ports that we can point the app at, the solr config from the repo, and a way to create the right cores

murny:
Docker would be very easy to provide for this. Especially if we are targeting latest versions. (we can do a lightweight docker-compose which only provides the datastores...mysql/solr/fedora/etc, but lets you keep using your local ruby/rails environment). I can do this once I start the User Auth work epic.

Big question to think about is whatever we do, it needs to work cleanly on travisCI as well.

PR #26, opened 2017-07-11
Port over docker work from hydranorth and add lightweight version for local development

murny:
* Add docker-compose(dockers everything) and docker-compose.lightweight (just dockers fedora/solr)
* Added travisCI which builds the docker-compose and runs minitest and rubocop in the docker containers

resolves https://github.com/ualbertalib/jupiter/issues/3


config/database.yml
```diff
@@ -28,6 +28,13 @@ test:
   <<: *default
   database: jupiter_test
 
+  # TODO: Review this later
```
murny: As comment says this is a bug upstream with how rails provisions dev/test databases using `ENV['DATABASE_URL']`. Will have to keep an eye out for this. This works for now by providing the `url`.

Other option is providing your own env vars for user/password/etc. Here is an example for this: https://github.com/tootsuite/mastodon/pull/3909/files

config/solr.yml
```diff
@@ -0,0 +1,7 @@
+# Needed and used by active fedora, to specify where to find solr
+development:
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/jupiter' %>
+test:
+  url: <%= ENV['SOLR_URL'] || 'http://localhost:8983/solr/jupiter' %>
```
mbarnett: Do you know if something in ActiveFedora is configured to use a separate core or w/e when the RAILS_ENV is test? Concerned that tests clearing out the test solr instance would clobber dev data

murny: This has been fixed. Seperated the cores out for both environments.

PR #28, opened 2017-07-17
Fix cleaning up server pids on container stops

murny:
When stopping containers rails doesn't cleanup properly and leaves a pid file. This prevents docker from starting back up cleanly.

Uses solution from here: https://github.com/docker/compose/issues/1393#issuecomment-99988242

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,7 @@ services:
   web:
     build: .
     image: ualbertalib/jupiter
+    command: bash -c 'rm -f tmp/pids/server.pid && bundle exec rails s'
     environment:
       RAILS_ENV: development
       DATABASE_URL: 'mysql2://root:mysecretpassword@mysql/'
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -63,19 +64,5 @@ services:
...
-  #     SOLR_URL: http://solr:8983/solr/jupiter
-  #     REDIS_URL: 'redis://redis/1'
-  #   volumes:
-  #     - .:/app
-  #   links:
-  #     - mysql
-  #     - fcrepo
-  #     - redis
-  #     - solr
+  #   <<: *rails
```
murny: Some yaml magic, assigns the web yaml into a variable, then reuse the variable for the worker. Since everything is the same except for the command, which we override here

docker-compose.yml
```diff
@@ -41,9 +41,10 @@ services:
     volumes:
       - redis:/data
 
-  web:
+  web: &rails
     build: .
     image: ualbertalib/jupiter
+    command: bash -c 'rm -f tmp/pids/server.pid && bundle exec rails server'
```
murny: Solution is to simply clean the server.pid file everytime we start the server

PR #29, opened 2017-07-18
Add lightweight visibility, embargos and tests

mbarnett:


test/models/jupiter_core/locked_ldp_object_test.rb
```diff
@@ -228,18 +231,22 @@ def test_basic_solr_finds
     end
 
     search_results.each_facet_with_results do |facet|
```
murny: Feel like this could be written better? Instead of just looping over the objects and if checking what they are, if they are the thing i want then assert a bunch of things. Maybe just like pull out the object you want then assert a bunch of things? Not sure though. Very minor regardless, will approve PR without this.

mbarnett: Yeah. Right now the results API is designed to make it easy to render the facet sidebar (where you don't know what facets will be in a result a priori), but it feels awkward to test. Not sure if I should add a second directly-access-the-facet-by-name to ease testing, when it won't be valuable in real code, and this approach will need testing too...

Still trying to find a balance between testability and not throwing a bunch of random just-for-test stuff into APIs.

murny: That makes sense then. No worries about this

Issue #22, opened 2017-07-07
Set up SAML integration for User Auth


mbarnett:
As it says on the tin. If we can get away with not needing apache shibboleth plugin, especially for dev machines, and just use omniauth and saml, we should

murny:
Closing ticket, code for this is in master.

 (However we should create a new ticket for actually doing the work of talking to IST and getting our stuff working with their IdP?)

PR #43, opened 2017-07-24
Omniauth authentication

murny:
https://github.com/ualbertalib/jupiter/issues/22

This implements omniauth integration. For the SAML piece, we need to work with IST IdP and provide the proper metadata for their login/login-uat servers. Should work other then that. Comes with a developer oauth provider for use in development mode until that time comes.

Some improvments? 

- Use cookies and auth_token instead of session
- Add logic to handle redirects (if i navigate to a protected page, should be redirected to login page, then redirected back to protected page after successfully logging in)


Gemfile
```diff
@@ -6,14 +6,20 @@ git_source(:github) do |repo_name|
...
 
 gem 'bcrypt', '~> 3.1.7'
-gem 'bootstrap'
+
+gem 'bootstrap', '~> 4.0.0.alpha6'
+
+# Review this after and update once bootstrap V4.0.0.alpha6+ come out,
+#  latest versions use popper instead of tether
+source 'https://rails-assets.org' do
+  gem 'rails-assets-tether', '>= 1.3.3'
```
murny: Was getting js errors in the console and on acceptance tests. Looks like the code is still looking for tether instead of popper. So replaced the libraries accordingly.

Gemfile
```diff
@@ -6,14 +6,20 @@ git_source(:github) do |repo_name|
...
+end
+
 gem 'haml'
 gem 'hydra-works'
 gem 'jquery-rails'
 gem 'mysql2', '>= 0.3.18', '< 0.5'
-# TODO: this is a missed bootsrap requirement, now fixed on their master  twbs/bootstrap-rubygem@8927eff
-# review and remove when that makes it into a release
-gem 'popper_js'
+gem 'puma', '~> 3.7'
```
murny: Acceptance tests were looking for puma? Added this back in (ships by default by `rails new`)

README.md
```diff
@@ -14,6 +18,13 @@
 
 `$ bundle exec rails test`
 
+# Run system tests
```
murny: Added section about system tests. By default they do not run with the `rails test` command. More info here: http://guides.rubyonrails.org/testing.html#system-testing
(theres a blob in a warning div down the page regarding this)

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,26 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
+  protected
+
+  # Returns the current logged-in user (if any).
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
```
murny: Currently theses are done in the session. We could do this smarter by using cookies (and as well, using a auth_token instead of looking up via user_id, so you can log out a user from admin panel, or have a user log out of all devices, etc etc.)

murny: Also using find_by instead of find...as we want a nil to be returned if user doesnt exist (instead of raising an exception)

app/views/application/_navbar.html.erb
```diff
@@ -0,0 +1,68 @@
+<header class="navbar navbar-inverse fixed-top navbar-toggleable-md bg-primary">
```
murny: Heavily refactored the navbar using bootstrap 4. We can bikeshed this later if you want

Issue #23, opened 2017-07-07
Set up Roles & Authorization Framework


mbarnett:
We'll need some way to bucket users into groups and authorize those users/groups (cancan etc). Pick something, set it up, and establish the basic patterns for others.

PR #46, opened 2017-07-26
Authorization using pundit

murny:
#23

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -1,6 +1,8 @@
 class Admin::DashboardController < Admin::AdminController
+
   def index
     @works = Work.all # TODO: AF has no concept of limits? Needs to solr this?
     @users = User.limit(10)
   end
+
 end
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -1,7 +1,7 @@
 class Admin::DashboardController < Admin::AdminController
 
   def index
-    @works = Work.all # TODO: AF has no concept of limits? Needs to solr this?
+    @works = Work.all # TODO: LockedLDPObject has no concept of limits? Needs to solr this?
     @users = User.limit(10)
   end
 
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -1,0 +1,6 @@
+class Admin::DashboardController < Admin::AdminController
+  def index
+    @works = Work.all # TODO: AF has no concept of limits? Needs to solr this?
+    @users = User.limit(10)
+  end
+end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -13,18 +13,12 @@ class CollectionsController < ApplicationController
   end
 
   def create
-    @collection = Collection.new_locked_ldp_object(collection_params)
+    @collection = Collection.new_locked_ldp_object(permitted_attributes(Collection))
     authorize @collection
     @community = Community.find(params[:community_id])
     @collection.unlock_and_fetch_ldp_object(&:save!)
 
     redirect_to community_collection_path(@community, @collection)
-  end
-
-  protected
-
-  def collection_params
-    params.require(:collection).permit(policy(Collection).permitted_attributes)
   end
 
 end
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -2,17 +2,20 @@ class CollectionsController < ApplicationController
 
   def show
     @collection = Collection.find(params[:id])
+    authorize @collection
     @community = Community.find(@collection.community_id)
   end
 
   def new
-    @community = Community.find(params[:community_id])
     @collection = Collection.new_locked_ldp_object(community_id: params[:community_id])
+    authorize @collection
+    @community = Community.find(params[:community_id])
   end
 
   def create
-    @community = Community.find(params[:community_id])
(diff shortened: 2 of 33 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Admin::DashboardController < Admin::AdminController
+
+  def index
+    @works = Work.all # TODO: AF has no concept of limits? Needs to solr this?
```
mbarnett: Just to clarify: ActiveFedora (kinda) has a concept of that, but Work isn't an ActiveFedora object, it's our own Solr layer (LockedLDPObject) so we need to add that.

I want to get away from thinking in terms of all of the ActiveFedora since ideally we're never dealing with it directly.

app/policies/collection_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def show?
+    true
+  end
+
+  def permitted_attributes
+    [:visibility, :owner, :title, :community_id] if user.admin?
+  end
+
+  class Scope < ApplicationPolicy::Scope
```
mbarnett: Scope isn't going to work with anything derived from LockedLDPObject. Stuff like `.where` and `.find` isn't lazy, so you can't chain multiple scopes together.

app/policies/work_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+    create?
+  end
+
+  def owned?
+    user.present?
+    # TODO: Fix this, how to get owner of works? Currently record (work) has no relationship to a user/creator
+    # record && user && record.creator == user.email
+  end
+
+  def permitted_attributes
```
mbarnett: The intention behind `safe_attributes` was that it would let the model mask out any unsafe attributes: https://github.com/ualbertalib/jupiter/blob/master/app/models/jupiter_core/locked_ldp_object.rb#L126 by overriding the method with, like example:

```ruby
def safe_attributes
  super - [:visibility]
end
```

Which would result in ever attribute but `:id` and `:visibility` being returned.

Can we come up with a pattern to combine that with this, so that we don't have to go around declaring attributes in multiple places?

app/controllers/collections_controller.rb
```diff
@@ -21,7 +24,7 @@ def create
   protected
 
   def collection_params
-    params[:collection].permit(Collection.attribute_names)
+    params.require(:collection).permit(policy(Collection).permitted_attributes)
```
mbarnett: It'd be nice to come up with a shorter pattern for this, so we're not constantly repeating this ceremony in a bunch of controllers

app/policies/work_policy.rb
murny: Could just call it here instead of redefining the list once again like so:

```
def permitted_attributes
  Work.safe_attributes
end
```

app/policies/collection_policy.rb
murny: Can remove. We currently not using this stuff anyways. You would add it to your view/controller to make sure the list you are using only has the propery permissions around it (eg, admin gets to see all, whereas users only get to see non private items)

https://github.com/elabs/pundit#scopes

app/controllers/collections_controller.rb
murny: Pundit does solve this for you. It provides a nice permitted_attributes helper for you. So instead of doing the above...you can replace collection_parms with just  
```
if @collection.update_attributes(post_params)
```

```
if @collection.update_attributes(permitted_attributes(@collection))
```
Then you done need this protected/private method for `<resource>_params`

More info here:
https://github.com/elabs/pundit#strong-parameters

app/policies/work_policy.rb
mbarnett: Could probably push this up to the ApplicationPolicy w/ something like `model.class.safe_attributes`?

app/controllers/collections_controller.rb
mbarnett: Ah nice

app/policies/work_policy.rb
murny: Good call, we could assuming its defined on everything being inherited by it (activerecord objects too? or type check what it is and do something sane for both)

app/policies/collection_policy.rb
mbarnett: We might be able to reintroduce it once I tackle pagination at the Solr end, since I think I can implement most of the lazy enumeration approach at that point

mbarnett: (the flip side is I don't want to turn this into another reinvent-an-ORM project, I just want enough for us to isolate the LDP interactions and drive the site for phase one, and then work on kicking the shims out and moving to AR)

app/policies/work_policy.rb
mbarnett: LDPPolicy isn't a half bad idea

app/controllers/admin/dashboard_controller.rb
murny: Fixed, changed to proper terminology

app/policies/collection_policy.rb
murny: Removed all scope blocks on LockedLdpObjects

app/policies/work_policy.rb
murny: Created a new LockedLdpObjectPolicy which updated all LockedLdpObjects to inherited from. This policy by default will defined the permitted_attributes from the LockedLdpObject safe_attributes

app/controllers/collections_controller.rb
murny: Cleaned all this up to use the helper approach instead.

Issue #52, opened 2017-07-27
Need some way of getting `.count` or `.limit` on LockedLDPObject objects?


murny:
It would be nice for example on the admin dashboard, to show how many Works/Communites/Collections we have in our system (using something like `Work.count`). 

Also be nice to just show the last 10 deposited works or something to the admin as well (using something like `Work.limit(10)` or maybe providing an order clause as well?)

Doesn't seem like you can currently do this via AF. Probably have to query solr or something instead?

mbarnett:
Related to #47, for future reference

mjgiarlo:
ActiveFedora does already support calling `#count` on either specific models or on `ActiveFedora::Base`. It uses Solr to make this calculation. (It does not support `#limit`, though.)

mbarnett:
Thanks for the clarification, Mike. Shane actually meant we need to add it to our home-brewed Solr abstraction that we're introducing to start breaking our dependency on ActiveFedora

mjgiarlo:
@mbarnett Gotcha! :cool:

PR #57, opened 2017-08-11
Search refactor and initial pagination support

### LINKED ISSUE #47: Pagination adaptor
It looks like the correct pattern to follow here is embodied in https://github.com/kaminari/kaminari-mongoid.

Will probably need to consult sunspot and bring in a lazy enumerable interface

mbarnett:
Adds initial pagination support to most scopes (currently compositions of `where`, `limit`, `offset`, and `all`). Adds support for `count`. Lazily defers query execution until results are needed.

Does not bring this to faceted cross-model search (yet)

Closes #47 and #52


app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -174,17 +176,24 @@ def self.solr_name_to_attribute_name(solr_name)
     # Accepts a string id of an object in the LDP, and returns a +LockedLDPObjects+ representation of that object
     # or raises <tt>JupiterCore::ObjectNotFound</tt> if there is no object corresponding to that id
     def self.find(id)
-      results_count, results, = perform_solr_query(%Q(_query_:"{!raw f=id}#{id}"), '', false)
+      results_count, results, = JupiterCore::Search.perform_solr_query(q: %Q(_query_:"{!raw f=id}#{id}"),
```
murny: is the extra comma on results needed?

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -193,38 +202,41 @@ def self.all
...
+    # attr, a string attribute name and sort order
+    def self.sort(attr)
+      all.sort(attr)
     end
 
     def self.valid_visibilities
       [:public, :private, :authenticated]
     end
 
+    # Used to dynamically turn an arbitrary Solr document into an isntance of its originating class
```
murny: typo on instance :trollface:

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: it's a destructuring, as `perform_solr_query` returns a 3-tuple. I'll add an `_` to make that clear.

mbarnett: Nevermind. Replaced it with

```ruby
results_count, results, _ = ...
```

and rubocop told me to correct it to 

```ruby
results_count, results, = ...
```

murny: Seems like it will allow a named underscore variable?

So if the third var is `_facets` or something this should be allowed

http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TrailingUnderscoreVariable
https://github.com/bbatsov/rubocop/blob/d87cea2cbaef494df11375900503784b5bc34859/config/enabled.yml#L1010-L1015

mbarnett: I'd rather turn the cop off than start naming things? `_` or the weird comma do a better job of communicating that the value is irrelevant to what follows, to my mind.

murny: I vote for turning it off then? I'd prefer an underscore var compared to a trailing comma?

mbarnett: Sure, will do

PR #63, opened 2017-08-16
Community and Collection and Notifications Admin

mbarnett:


app/views/collections/show.haml
```diff
@@ -9,3 +9,5 @@
 					= link_to w.title, work_path(w), class: 'list-group-item list-group-item-action'
 			- else
 				There are no works in this Collection
+
+= link_to 'Edit', edit_community_collection_path(@community, @collection), class:'btn btn-secondary'
```
murny: This is failing cause houndci has a special linter for haml:

https://houndci.com/configuration#haml

So its using the default haml-lint config

app/controllers/collections_controller.rb
```diff
@@ -21,4 +21,41 @@ def create
     redirect_to community_collection_path(@community, @collection)
   end
 
+  def edit
+    @community = Community.find(params[:community_id])
+    @collection = Collection.find(params[:id])
+    authorize @community
```
murny: Tiny suggestion: you could move this up before `@collection`? That way if you dont have access to `@community` you dont waste a SQL lookup/hit looking for a `@collection`

mbarnett: Solr, in this case, but good idea

app/controllers/communities_controller.rb
```diff
@@ -27,4 +32,33 @@ def create
...
+        flash[:alert] = I18n.t('application.communities.not_empty_error')
+      end
+
+      redirect_to admin_communities_and_collections_path
+    end
+  end
+
+  protected
+
+  def community_params
```
murny: You dont need these community_params method if you are using pundit correctly. Take a look at the `create` action here. It's using the `permitted_attributes` helper which will go lookup the pundit policy for Community and will return this exact same thing here.

This also applies for the collection controller above

app/models/collection.rb
```diff
@@ -4,6 +4,13 @@ class Collection < JupiterCore::LockedLdpObject
 
   has_attribute :title, ::RDF::Vocab::DC.title, solrize_for: [:search, :facet]
   has_attribute :community_id, ::VOCABULARY[:ualib].path, solrize_for: :pathing
+  has_attribute :tag, ::VOCABULARY[:ualib].temp_tag, solrize_for: :facet
```
murny: This tag stuff is before leah said no to the tag/category stuff? or this something diff?

app/controllers/communities_controller.rb
mbarnett: Missed it in the huge merge. 

Side-note but it would be great if we didn't have to constantly pass the class name in, in the CollectionsController, etc. It infers everything else.

app/models/collection.rb
mbarnett: Yeah I'll rip that out

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -8,6 +8,10 @@ class MultipleIdViolationError < StandardError; end
   class AlreadyDefinedError < StandardError; end
   class LockedInstanceError < StandardError; end
 
+  VISIBILITY_PUBLIC = 'public'.freeze
```
murny: Guessing activefedora doesnt have enums (http://api.rubyonrails.org/classes/ActiveRecord/Enum.html)  :trollface:?

app/models/user.rb
```diff
@@ -1,6 +1,7 @@
 class User < ApplicationRecord
 
   has_many :identities, dependent: :destroy
+  has_many :site_notifications, dependent: :destroy
```
murny: should all site notifications get deleted if the user who created them gets removed? Not sure how notifications are working currently or this is a good solution. Either way I'd imagine this be very rare I suppose ¯\_(ツ)_/¯

mbarnett: They join on an enforced foreign key, so if the user row gets nuked the DB is inconsistent unless you nuke the notification rows. I don't think there's going to be much call for "I want to delete a super user but I want to see past notifications from them"

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: 3 guess. First 2 don't count

app/models/user.rb
murny: K cool just making sure you put some thought into it. :+1:

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1 +1,26 @@
...
+<div class="col my-4" >
+  <ul id="admin-collection-list" class="list-group">
+
+    <% @communities.each do |community| %>
+
+      <li class="list-group-item">
+        <%= link_to community.title, community_path(community)%>
+
+        <div class="ml-auto">
+          <%= link_to "Collections #{fa_icon('chevron-down')}".html_safe, admin_communities_and_collection_path(community), data: {community_id: community.id }, class: 'btn btn-secondary collection-btn', remote: true %>
```
murny: Could pass a block to the link_to so you dont need the .html_safe for these? 

```erb
<% link_to path, options() do %>
  Collections <%= fa_icon 'chevron-down' %>
<% end %>
```

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1 +1,26 @@
...
+          <%= link_to "Close #{fa_icon('chevron-up')}".html_safe, '', class: 'btn btn-secondary close-btn hidden-xs-up', data: {community_id: community.id } %>
```
murny: If theres no href, use a button element instead here?

app/views/admin/site_notifications/new.html.erb
```diff
@@ -0,0 +1,45 @@
+<h4>Manage Notifications</h4>
+  <div class="col" >
```
murny: can you have a col just by itself or does it need to be nested in a row?

mbarnett: Yep, that's a straggler

app/views/layouts/admin.html.erb
```diff
@@ -1,38 +1,48 @@
+<% content_for(:head) do %>
+  <%= javascript_include_tag 'admin' %>
```
murny: Does this need the turoblinks stuff? like `'data-turbolinks-track': 'reload'` (dont know nothing about turbolinks ¯\_(ツ)_/¯)

mbarnett: Only if we want it to reload that file when the user changes a page, which I don't think is necessary

app/views/layouts/application.html.erb
```diff
@@ -14,8 +14,12 @@
   </head>
 
   <body>
-    <%= render partial:'navbar' %>
-    <div class="container">    
+
+    <div class="container">
```
murny: This looks like a merge error. You got multiple containers and this one has no end div tag. This container should be removed.

config/routes.rb
```diff
@@ -16,7 +16,8 @@
   namespace :admin, constraints: AdminConstraint.new do
     root to: 'dashboard#index'
 
-    resources :communities_and_collections, only: [:create, :new, :index]
+    resources :communities_and_collections, only: [:show, :create, :new, :index]
+    resources :site_notifications, only: [:new, :destroy, :create]
```
murny: Can you not edit/update notifications? Or expected behavior is they close (destroy them) and create a new one?

mbarnett: Close and create a new one. I checked w/ Leah and that's her jam

leahvanderjagt:
If you think it's better practice to edit/update, by all means.

I just like having a list of old notifications very handy so I can copy
paste and not re-compose new ones every time.

___________________________________________________________




*Leah VanderjagtDigital Repository Services Coordinator University of
Albertat. 780.492.3851 *
*leahv@ualberta.ca <leahv@ualberta.ca> *

leahvanderjagt:
Rethinking: having said what I said -  if this is done, let's proceed with
what we have and move on

___________________________________________________________




*Leah VanderjagtDigital Repository Services Coordinator University of
Albertat. 780.492.3851 *
*leahv@ualberta.ca <leahv@ualberta.ca> *

app/views/layouts/application.html.erb
```diff
@@ -14,8 +14,9 @@
   </head>
 
   <body>
-    <%= render partial:'navbar' %>
-    <div class="container">    
+    <div class="container">
+      <%= render partial:'navbar' %>
```
murny: Navbar should be on the outside of this main container. As you want it to stretch all across the entire width of the page/screen. (navbar itself has a container for positioning the contents inside the navbar)

mbarnett: I can move it, but oddly it does not actually matter -- the navbar's container supercedes this one

PR #66, opened 2017-08-17
Fix sorting

mbarnett:
@murny Here's the fix. Tests are coming but if you need it you can grab this now. Whatever you want to sort by needs to explicitly be sorlized for sorting (see the mod to Work.rb here)


app/models/jupiter_core/deferred_solr_query.rb
```diff
@@ -27,8 +27,17 @@ def offset(num)
     self
   end
 
-  def sort(attr)
-    criteria[:sort] = attr
+  def sort(attr, order = :asc)
```
murny: I assume you deliberately trying to be different then ActiveRecord right?

Like sort instead of order and takes two args instead of a hash/string etc.

mbarnett: Unless we want to litter "title_tesim"-type solrizer junk all over the codebase, I need to munge attr (:title) into its solr-named sort-index equivalent (title_ssi), which in the ActiveRecord-style "title desc" feels like unclean string splitting versus splitting up the arguments?

PR #73, opened 2017-08-21
Set the owner when creating collections/communities.

cwant:
As per comments here:

https://github.com/ualbertalib/jupiter/commit/7356b86b836e4ad0166bf33f706936d3dca52c84

(Might be worth adding tests to make sure owner gets set ...)

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -13,14 +13,13 @@ class CollectionsController < ApplicationController
   end
 
   def create
-    @collection = Collection.new_locked_ldp_object(permitted_attributes(Collection))
+    @collection =
+      Collection.new_locked_ldp_object(permitted_attributes(Collection)
+                                        .merge(owner: current_user.name))
     authorize @collection
-    @community = Community.find(params[:community_id])
-    @community.unlock_and_fetch_ldp_object do |uo|
-      uo.owner = current_user.name
-      uo.save!
-    end
+    @collection.unlock_and_fetch_ldp_object(&:save!)
 
+    @community = Community.find(params[:community_id])
     redirect_to community_collection_path(@community, @collection)
   end
 
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,5 +1,7 @@
 class CollectionsController < ApplicationController
 
+  before_action -> { authorize :application, :admin? }, except: [:index, :show]
+
   def show
     @collection = Collection.find(params[:id])
     authorize @collection
@@ -15,7 +17,7 @@ class CollectionsController < ApplicationController
   def create
     @collection =
       Collection.new_locked_ldp_object(permitted_attributes(Collection)
-                                        .merge(owner: current_user.name))
+                                        .merge(owner: current_user&.id))
     authorize @collection
     @collection.unlock_and_fetch_ldp_object(&:save!)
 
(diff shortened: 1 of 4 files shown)


app/controllers/collections_controller.rb
```diff
@@ -16,7 +16,10 @@ def create
     @collection = Collection.new_locked_ldp_object(permitted_attributes(Collection))
     authorize @collection
     @community = Community.find(params[:community_id])
-    @collection.unlock_and_fetch_ldp_object(&:save!)
+    @community.unlock_and_fetch_ldp_object do |uo|
```
mbarnett: Should this be @collection? You're (re-)setting the Community's owner every time you create a new collection, right now...

murny: Should be `@collection` here not `@community`

app/controllers/communities_controller.rb
```diff
@@ -27,7 +27,10 @@ def edit
   def create
     @community = Community.new_locked_ldp_object(permitted_attributes(Community))
     authorize @community
-    @community.unlock_and_fetch_ldp_object(&:save!)
+    @community.unlock_and_fetch_ldp_object do |uo|
```
murny: I'd put this assignment in the new_locked_ldp_object call if you can? it doesnt matter for collections/communities as only admins can create them (as defined in pundit policy) but for works and other objects, you will want to authorize that only the owner (or admin) are the only ones that can do things to works. So pundit will check that the owner of the record and current user are the same user.

app/controllers/collections_controller.rb
cwant: Whelp, yeah, that's what happens when you do a quick copy/paste before a meeting!

app/controllers/communities_controller.rb
cwant: Okay, hopefully have addressed this in latest commit.

app/controllers/collections_controller.rb
```diff
@@ -13,11 +13,13 @@ def new
   end
 
   def create
-    @collection = Collection.new_locked_ldp_object(permitted_attributes(Collection))
+    @collection =
+      Collection.new_locked_ldp_object(permitted_attributes(Collection)
+                                        .merge(owner: current_user.name))
```
mbarnett: This isn't a permanent thing, but for the time being we've been tracking ownership by user.id, not name, which is completely non-unique

app/controllers/communities_controller.rb
```diff
@@ -25,7 +25,9 @@ def edit
   end
 
   def create
-    @community = Community.new_locked_ldp_object(permitted_attributes(Community))
+    @community =
+      Community.new_locked_ldp_object(permitted_attributes(Community)
+                                       .merge(owner: current_user.name))
```
mbarnett: ditto

app/controllers/collections_controller.rb
cwant: Okay thanks, will update.

cwant: Now that I look again, this would go kaboom if the person isn't logged in (I assume current_user is nil in that case)

mbarnett: Non-logged-in users (and non-admin users) shouldn't even be able to hit this action

mbarnett: (which isn't being enforced at the moment, so if you want to throw a before action on here, please do)

cwant: That's fine and dandy, but the authorize statement comes after this.

mbarnett: Yeah, the authorize statement is not a replacement for something like https://github.com/ualbertalib/jupiter/blob/master/app/controllers/admin/admin_controller.rb#L3. This whole method needs to be guarded, ditto new, edit, update, and destroy

cwant:
Okay, I think this one has addressed the various issues that have been discussed. Threw in some controller tests too so hopefully will catch some similar probs in the future.

PR #74, opened 2017-08-21
Upgrade to bootstrap beta

murny:
Upgraded bootstrap alpha6 to the new bootstrap beta release.

Some minor changes and fixes.

### DIFF
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -25,43 +25,39 @@
 
 <p class="text-center"><%= t('.display_users_count', users_count: @users.count, total_users_count: User.count) %></p>
 
-<%# Bug in current bootstrap with table-responsive not being 100%,
-wrapping in div is a workaround https://github.com/twbs/bootstrap/issues/21577 %>
-<div class="table-responsive">
-  <table class="table table-bordered">
-    <thead>
+<table class="table table-responsive table-bordered">
+  <thead>
+    <tr>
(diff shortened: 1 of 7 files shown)


app/views/application/_navbar.html.erb
```diff
@@ -1,22 +1,24 @@
-<header class="navbar navbar-inverse fixed-top navbar-toggleable-md bg-primary">
+<header class="navbar navbar-dark fixed-top navbar-expand-md bg-primary">
```
murny: I would almost recommend changing the navbar to a different color? Maybe `navbar-light bg-light`? Which will make it a nice light grey like the old github navbar. Or `navbar-dark bg-dark` will make it like new github navbar? The `bg-primary` doesn't look bad, just the search button (secondary color is now a much darker color then previous) looks a bit off

app/views/admin/users/index.html.erb
```diff
@@ -25,43 +25,39 @@
 
 <p class="text-center"><%= t('.display_users_count', users_count: @users.count, total_users_count: User.count) %></p>
 
-<%# Bug in current bootstrap with table-responsive not being 100%,
```
murny: This bug is fixed now in latest

app/views/application/_navbar.html.erb
mbarnett: If we're going to change the color we might want to bring it closer to the current ERA theme?

PR #77, opened 2017-08-22
Record created at

mbarnett:
Adds a proper record_created_at time that will persist across index loss. Uses record_created_at as default sort order. Also adds intelligent `.first` and `.last` methods.


test/test_helper.rb
```diff
@@ -10,6 +10,10 @@
 
 class ActiveSupport::TestCase
 
+  def freeze_time(&block)
+    travel_to DateTime.now, &block
```
murny: Two suggestions here: 
  - Be consistent, If you are using `Time` class elsewhere, use it here too instead of `DateTime` class? Time is better in performance then DateTime among other reasons
 - You should always handle time with the correct time zones. So Instead of using `Time.now` use `Time.zone.now`. Or better yet rails has a smart helper called `Time.current` which will  make the smart choice between the two: https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/time/calculations.rb#L37-L40

So basically what I am saying is use `Time.current` instead of `DateTime.now` or `Time.now`

mbarnett: Yeah, solid points.

Unfortunately right now I am running into more ActiveFedora Patented Stupidity (it's silently throwing out DateTimes with more precision than one second, and saving the record as blank _after_ validating it as non-blank), so the full solution to this issue is going to end up being very ugly.

Issue #38, opened 2017-07-20
Modify Community Create/Edit To Include Description & Logo Upload


mbarnett:
Mockups:

* https://jupiter.mybalsamiq.com/projects/jupiter/Admin%20-%20create%20new%20community
* https://jupiter.mybalsamiq.com/projects/jupiter/Admin%20-%20community%20detail%20page

DONE WHEN:
- [x] Add description to community. Current ERA or the data dictionary should have information on what predicate to use.
- [x] (My suggestion is to try moving up to the latest #Rails 5.2 beta and investigate [ActiveStorage](https://github.com/rails/activestorage) for logos, as it seems well suited to our needs and might easily bridge the fact that Community isn't an ActiveRecord class, which most asset gems expect).
- [x] Add logo upload to community. Ideally we'd like to avoid the Hydra approach of storing the logo in Fedora and streaming it from there for display. ActiveStorage seems promising in that it can store the asset info in its own table rather than adding columns to an ActiveRecord table, since Collections aren't ActiveRecord. You may need to play with a few approaches and talk them through, but if we have to start creating, say, shadow collection records in the DB, I'm ok with that.
- [x] Display logo on community show page
- [x] Display logo on community index page

cwant:
Some notes on the ActiveStorage investigation:

* The now-obsolete repository for ActiveStorage is here: https://github.com/rails/activestorage
  The last version of code before the merge into Rails is in the 'archive' branch, but there is also a release v0.1 which is what you get when you install ActiveStorage as a gem.
* Release v0.1 mostly works with Rails 5.1 (there is a class variable that needs to be set via an initializer that would not need to be done with Rails 5.2).
* The 'archive' branch depends strongly on Rails 5.2. It has some nice features though: serving up blobs via a rails engine, ability to resize images

So our options are:

* Accept v0.1 as it is -- it might be good enough for our current purposes; If we need to resize images, do it some other way (current ActiveStorage uses MiniMagick gem for that).
* Fork the ActiveStorage gem (from the 'archive' branch) and massage it to work with Rails 5.1
* We become a Rails 5.2 project.

cwant:
The current class does not have a validation on 'title' -- should there be one?
Related: should there be a validation on this new description property I'm adding? (presence, minimum characters ...)

I am getting some interesting behaviours with community, I don't know if this is how it's supposed to behave:

```ruby
pry(main)> community = Community.all.first
[...]

PR #79, opened 2017-08-22
Add description to community, and other fixes

cwant:
This PR is part of issue #38. The community logo will be a separate PR.

Also within this PR:

* Add validation for presence of community title
* Make community#show view more closely resemble the mockup:
  https://jupiter.mybalsamiq.com/projects/jupiter/Admin%20-%20community%20detail%20page
* Fix for incorrectly-nested translations (also use lazy lookup)

### DIFF
--- a/app/views/communities/show.html.erb
+++ b/app/views/communities/show.html.erb
@@ -1,6 +1,6 @@
 <h2 class="mt-3"><%= t('communities.title', title: @community.title) %></h2>
 
-<% if policy(@community).edit? || policy(@community).delete? %>
+<% if policy(@community).edit? || policy(@community).destroy? %>
   <div class="row mt-3" >
     <div class="col text-right" >
       <% if policy(@community).edit? %>
@@ -25,7 +25,7 @@
 
 <div class="row mt-3">
   <div class="col">
-    <h4>
+    <h4 class="collections-header">
       <% if @community.member_collections.count > 0 %>
         <%= t('.collections_list_header') %>
       <% else %>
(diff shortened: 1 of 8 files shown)


app/views/communities/show.html.erb
```diff
@@ -1,25 +1,52 @@
...
-        <% end %>
+        <%= t('.collections_list_header') %>
       <% else %>
         <%= t('.no_collections') %>
       <% end %>
-    </div>
+    </h4>
+  </div>
+  <div class="col text-right">
+    <%= link_to t('.create_new_collection'), new_community_collection_path(@community),
```
murny: Wrap this with an authorize check? I assume only admins should get to see the new collection button?

test/integration/community_show_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+class CommunityShowTest < ActionDispatch::IntegrationTest
+
+  setup do
+    admin = users(:admin_user)
+
+    # TODO: setup proper fixtures for LockedLdpObjects
+
+    # A community with two collections
+    @community1 = Community.new_locked_ldp_object(title: 'Two collection community',
```
mbarnett: Take a look at https://github.com/ualbertalib/jupiter/pull/82 and refactor this all into an `before_all` method. `setup` runs before every single test, so this wastes a lot of time with unnecessary writes.

test/integration/community_show_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+    # A community with no collections
+    @community2 = Community.new_locked_ldp_object(title: 'Empty community',
+                                                  owner: admin.id)
+    @community2.unlock_and_fetch_ldp_object(&:save!)
+  end
+
+  teardown do
+    ActiveFedora::Cleaner.clean!
+  end
```
mbarnett: Once you've moved all the creates into a `before_all`, you can delete the `teardown`, as Fedora/Solr are already cleaned between each file.

test/integration/community_show_test.rb
murny: Another option here is if you want to setup some stuff for only certain tests, you can nest those certain tests and the setup block under a `context` block. This way only things in the context block are affected and not everything else. 

But in this case probably matt's suggestion is better

cwant: Yeah, I had originally set it up to use a context block, but for some reason the tests were complaining about "missing argument v" or such.

Anyways, will explore the options.

cwant: Tests updated -- thanks!

PR #81, opened 2017-08-23
Add page title helper and page title to all views

murny:
Create a page title helper to populate page title for all views


mbarnett:
I think we could be a bit cleverer with helpers or layouts and save ourselves the repetition of  ` | #{t('admin.header')}` or `t('.header'))` everywhere?

murny:
Agreed, will revisit this and come up with something smarter

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% provide(:page_title, "Manage Communities | #{t('admin.header')}") %>
+
```
cwant: i18n?

app/views/admin/site_notifications/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% provide(:page_title, "Manage Notifications | #{t('admin.header')}") %>
```
cwant: i18n?

app/views/collections/edit.html.erb
```diff
@@ -1,2 +1,4 @@
+<% provide(:page_title, 'Edit Collection') %>
```
cwant: i18n?

app/views/collections/new.html.erb
```diff
@@ -1,2 +1,4 @@
+<% provide(:page_title, 'Create Collection') %>
```
cwant: i18n?

app/views/works/edit.haml
```diff
@@ -1,2 +1,4 @@
-%h2.mt-3 Edit
-= render partial: 'form', locals: {work: @work}
\ No newline at end of file
+- provide(:page_title, 'Edit Work')
```
cwant: i18n?

app/views/works/new.haml
```diff
@@ -1,2 +1,4 @@
+- provide(:page_title, 'Create a new work')
```
cwant: i18n?

app/views/works/search.haml
```diff
@@ -1,3 +1,6 @@
+- provide(:page_title, 'Search Work')
```
cwant: i18n?

app/views/works/search.haml
```diff
@@ -12,11 +15,11 @@
...
+
 		.col-md-10
 			- @results.results.each do |result|
 				.row
 					.col
 						= link_to result.title, work_path(result.id)
 - else
-	%h2.mt-3 No results found
\ No newline at end of file
+	%h2.mt-3 No results found
```
cwant: i18n?

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,10 @@
 module ApplicationHelper
+  def page_title(page_title)
+    base_title = 'ERA'
```
cwant: It might be overkill, but I wonder if the application name should be in config, or in i18n.

app/views/works/show.haml
```diff
@@ -1,31 +1,35 @@
+- provide(:page_title, @work.title)
```
cwant: And yeah, more i18n stuff needed through out this file. Perhaps there should be a separate ticket to clear out the reset of the HAML and i18n all views.

murny: Yeah ill ticket something up and work on it :+1: 

Haml stuff is brutal for houndci...its killing me

murny: Ticket created here: https://github.com/ualbertalib/jupiter/issues/83

murny:
Matt's issue is now addressed. Will write some unit tests for the helper tomorrow and then I think this is good to be merged.

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% page_title(t('.header')) %>
```
mbarnett: Any reason not to move this into the layout, and then just call `page_title` in the few places we override this as a default?

murny: The lazy lookup doesn't work outside in the layouts. 

We could drive this all off i18n however? If its not found in i18n it falls back to the default unless you override it in the view or something?

mbarnett: Hmm true. Maybe leave it as is for the time being until we can think of something less ugly.

PR #82, opened 2017-08-23
Hook up .last & fix test performance regressions

mbarnett:
Hooked up the `.last` tests that @cwant stubbed out.

Also if y'all weren't aware, `setup` blocks run before *every* test in the file, so Chris was re-creating and re-saving objects needlessly, which tripled the run time of the test suite (ActiveFedora performance, the tire-fire that never stops burning...).

Changing `setup` to a `before_all` method (from [Minitest Hooks](https://github.com/jeremyevans/minitest-hooks)) saves a bunch of time. The only modifications this required were creating specific sacrificial communities and collections for the delete action tests, and not using fixtures in `before_all`, because that blows up for some reason.


test/controllers/collections_controller_test.rb
```diff
@@ -2,22 +2,20 @@
...
-  setup do
-    @admin = users(:admin)
-    sign_in_as @admin
-
-    # TODO: setup proper fixtures for LockedLdpObjects
+  def before_all
+    super
     @community = Community.new_locked_ldp_object(title: 'Nice community',
-                                                 owner: @admin.id)
+                                                 owner: 1)
```
murny: You cant guartee the ID of admin is `1` unless you add this to the fixtures as fixtures will generate seemingly random ID's

mbarnett: It doesn't matter. None of the tests or logic depend on the id corresponding to anything. 1 could be replaced with `generate_random_string` to no effect, if you think that is clearer.

murny: This is true. Your right considering all that matters is if your an admin. However hate for someone to copy these specs for works and find out some weird behaviour happening ¯\_(ツ)_/¯

mbarnett: It's probably possible to get fixtures loaded in time for before_all to work properly, but it will take some digging. Might be worth ticketing that out separately?

test/controllers/communities_controller_test.rb
```diff
@@ -51,8 +47,10 @@ class CommunitiesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should destroy community' do
+    community = Community.new_locked_ldp_object(title: 'Delete me',
```
cwant: I think we do this enough times that it would be worth creating either a LockedLdpObject method or a test helper for this. That would at least help us having to fix a lot of code if things change down the road. Perhaps something like:

```ruby
community = Community.create_locked_ldp_object(title: 'Delete me', ...)
```
or test helper:

```ruby
community = create_locked_ldp_object(Community,  title: 'Delete me', ...)
```

mbarnett: Yeah, agreed. I might start looking at what it would take to get FactoryGirl working with this, as it's a bit more flexible than fixtures for controlling how these things get instantiated.

murny: Only issue with FactoryGirl (at least on hydranorth) is it doesn't cleanup after itself. So you still need to teardown/ActiveFedora.clean! everytime you use Factorygirl ¯\_(ツ)_/¯

mbarnett: Well, we either do it manually everywhere, and clean up, or use FactoryGirl, and clean up. Fixtures is too dependent on ActiveRecord semantics to make it work properly w/ any of the AF-derived stuff.

Right now ActiveFedora.clean! runs after every test file anyways. It's just a matter of being responsible about not creating more things than you absolutely need to, and keeping mutations of example objects contained, to keep performance from choking on writes to Fedora.

Issue #37, opened 2017-07-20
Modify existing Collection/Community Show pages


mbarnett:
DONE WHEN:

- [x] Reuse the collection partial from the Admin page, with conditionally enabling edit and delete buttons when user is admin.
- [x] Conditionally enable delete and edit buttons for the viewed collection/community when user is admin

cwant:
This one was done in a recently merged PR.

PR #84, opened 2017-08-24
Issue #37, uniting collection partial for community views.

cwant:
Also make presence of collection edit/delete buttons based on policy in community show view.


cwant:
This one ended up having a merge conflict as a result of @murny's i18n/haml work going into master. I have fixed the conflict and pushed another commit, but as a result it need to be approved again.

PR #92, opened 2017-08-25
add docker for deployment with nginx in uat env

weiweishi:
This PR captures what I am working on right now, and comments are welcome. 

To do:
- [x] Try to use community Nginx box instead of building ourselves
- [x] Fixing the style issues


config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+ 
+  location ^~ /assets/ {
+    gzip_static on;
+    expires max;
+    add_header Cache-Control public;
+  }
+ 
+  location = /50x.html {
+    root html;
+  }
```
mbarnett: We can get rid of this.

config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+ 
+  location = /404.html {
+    root html;
+  }
```
mbarnett: We can also get rid of this.

config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+ 
+server {
+ 
+  listen 80;
+ 
+  client_max_body_size 4G;
+  keepalive_timeout 10;
+ 
+  error_page 500 502 504 /500.html;
+  error_page 503 @503;
```
mbarnett: We can get rid of this line, as I don't think out infrastructure is set up to forward a 503 when under maintenance.

config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+    root html;
+  }
+ 
+  location @503 {
+    error_page 405 = /system/maintenance.html;
+    if (-f $document_root/system/maintenance.html) {
+      rewrite ^(.*)$ /system/maintenance.html break;
+    }
+    rewrite ^(.*)$ /503.html break;
+  }
```
mbarnett: Ditto the 503 above, we can remove this section

config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+ 
+  if ($request_method !~ ^(GET|HEAD|PUT|PATCH|POST|DELETE|OPTIONS)$ ){
+    return 405;
+  }
+ 
+  if (-f $document_root/system/maintenance.html) {
+    return 503;
+  }
```
mbarnett: And this part about re-sending the 503 if the file doesn't exist.

config/nginx.conf
```diff
@@ -0,0 +1,64 @@
...
+ 
+  location ~ \.(php|html)$ {
+    return 405;
+  }
```
mbarnett: We should get rid of this and let Rails handle the complaining about not knowing what asdf.php is.

solr/scripts/solr-precreate-jupiter-deployment.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+#
+# This is for docker! Customized script for our jupiter project.
```
murny: You shouldn't need to copy this file for a staging/production/etc deployment. Solr docker does precreating of a solr container just fine. This was needed to provision two different containers (one of dev and one for test which isn't needed here).

docker-compose.deployment.yml
```diff
@@ -0,0 +1,90 @@
...
+
+  solr:
+    image: solr
+    # Uncomment this if you want to access SOLR's web interface
+    # ports:
+    #   - "8983:8983"
+    volumes:
+      - solr:/opt/solr/server/solr/mycores
+      - ./solr/config:/config
+      - ./solr/scripts/solr-precreate-jupiter-deployment.sh:/docker-entrypoint-initdb.d/solr-precreate-jupiter-deployment.sh
```
murny: You can instead use an Entrypoint namespace and provide the creation of the staging/jupiter container instead like so:

```
 entrypoint:
      - docker-entrypoint.sh
      - solr-precreate
      - jupiter
      - /config
```

Dockerfile.deployment
```diff
@@ -0,0 +1,53 @@
+FROM ruby:2.3.4
+MAINTAINER Murny
```
murny: We can change this. `MAINTAINER` is deprecated anyways. They recommend using `LABEL`

We can use ualberta libraries here?

`LABEL maintainer="University of Alberta Libraries"`

https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Gemfile
```diff
@@ -68,3 +68,7 @@ group :test do
   gem 'minitest-hooks'
   gem 'shoulda', require: false
 end
+
+group :staging, :production do
+  gem 'tzinfo-data'
```
murny: Shouldn't need this? Debian should have tzinfo packages? (if your using alpine or something then yeah you will need this?)

docker-compose.deployment.yml
```diff
@@ -0,0 +1,90 @@
...
+      - "3000"
+    depends_on:
+      - mysql
+      - fcrepo
+      - redis
+      - solr
+  nginx:
+    build:
+      context: .
+      dockerfile: Dockerfile.nginx
```
murny: Do you need a seperate dockerfile for nginx? what doe this give over the default?

You can just do this and use the official build image:

```  
nginx:
    image: nginx
    depends_on: 
      - web
    volumes:
      - ./config/nginx.conf:/etc/nginx/conf.d/default.conf
      - uploads:/srv/www/uploads/system
    ports:
      - "80:80" 
```

Here I'm copying and overwritting the default.conf with my own conf found in /config/ngix.conf

config/puma/deployment.rb
```diff
@@ -0,0 +1,25 @@
+workers Integer(ENV['WEB_CONCURRENCY'] || 2)
```
murny: Was there a reason to move this from config/puma.rb to config/puma/deployment.rb? This will also be used in development so you should be able to just leave this in the default location of config/puma.rb?

weiweishi: I kept a version of the puma configuration file for the development environment in config/puma/development.rb, and tested with the dev environment. I think Puma is able to pick up the configuration file based on the environment in config/puma/$environment.rb files. If the file doesn't exist, it will default to config/puma.rb. I don't know much about puma to see if I can consolidate the different configuration files.

docker-compose.deployment.yml
weiweishi: Yes. my next step is to use nginx official build.

docker-compose.deployment.yml
weiweishi: Okay. Thanks!

Gemfile
weiweishi: The image this is based on is ruby:2.3.4, but it was complaining about tzinfo-data when compiling assets. I will spin up a barebone ruby 2.3.4 container to test/verify.

Gemfile
```diff
@@ -41,6 +41,7 @@ gem 'pundit'
 
 # Misc Utilities
 gem 'kaminari'
+gem 'sdoc'
```
murny: Was this breaking on staging/production? Or how come this is moved out here?

Dockerfile.deployment
weiweishi: Okay. sounds good

Gemfile
weiweishi: This was under "development"/"test" group, but it was needed for compiling assets, so I moved it to the general group and have it installed for all envs.

Gemfile
murny: Other option to maybe try is to `apt-get` this if its missing from the debian box they are giving you in the dockerfile. On L7ish add a `tzdata` line so that it gets installed.

`sudo apt-get install tzdata`

weiweishi:
made changes according to suggestions above. @mbarnett @murny 
Remaining question: how to config the saml in the uat env? especially the non-ccid login is now hidden on the login page?

Dockerfile.nginx
```diff
@@ -0,0 +1,31 @@
+# Base image:
+FROM nginx
```
murny: This dockerfile can be removed now? Now that you are using the image in your docker-compose?

Gemfile.lock
```diff
@@ -383,6 +385,7 @@ DEPENDENCIES
   simplecov
   solrizer!
   turbolinks (~> 5)
+  tzinfo-data
```
murny: tzinfo gem is gone from the gemfile now?

Do another bundle install and check in the Gemfile.lock?

Gemfile
murny: Any idea about this @mbarnett ? Does it make sense for Rdoc to be a production dependency? 

Assume this needs to be improved: https://github.com/ualbertalib/jupiter/blob/master/lib/tasks/rdoc.rake So it's only avaiable in dev/test  and then you shouldnt get these compiling asset errors?

config/puma.rb
```diff
@@ -51,6 +51,14 @@
 #   ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
 # end
 #
-
+if ENV['RACK_ENV'] == 'uat' || ENV['RAILS_ENV'] == 'uat'
```
murny: Probably can just check RAILS_ENV? Or do you need to check RACK_ENV as well? (doubt we will ever set RACK_ENV instead of RAILS_ENV?)

Gemfile
mbarnett: My bet would be that when running *any* Rake task, it loads all of the task files, which hits the require at the top of that file, which then blows up. It's probably fine to move the gem up so that tasks enumerate properly.

PR #94, opened 2017-08-30
WIP: Admin cleanup

murny:
WIP

### DIFF
--- a/app/views/admin/communities_and_collections/_collection.html.erb
+++ b/app/views/admin/communities_and_collections/_collection.html.erb
@@ -5,7 +5,7 @@
     <div class="btn-group">
       <% if policy(collection).edit? %>
         <%= link_to t('edit'), edit_community_collection_path(community, collection),
-            class:'btn btn-secondary edit-collection' %>
+            class:'btn btn-outline-primary edit-collection' %>
       <% end %>
       <% if policy(collection).destroy? %>
         <%= link_to t('delete'), community_collection_path(community, collection),
--- a/app/views/admin/communities_and_collections/_collection.html.erb
+++ b/app/views/admin/communities_and_collections/_collection.html.erb
@@ -1,20 +1,17 @@
-<li class="list-group-item list-group-item-info">
-  <div class="row">
-    <div class="col pl-4" >
-      <%= link_to collection.title, community_collection_path(community, collection) %>
-    </div>
-    <div class="col ml-auto text-right">
-      <div class="btn-group">
-        <% if policy(collection).edit? %>
-          <%= link_to t('edit'), edit_community_collection_path(community, collection),
-              class:'btn btn-secondary edit-collection' %>
-        <% end %>
-        <% if policy(collection).destroy? %>
-          <%= button_to t('delete'), community_collection_path(community, collection),
-              class: 'btn btn-danger delete-collection', method: :delete,
-              data: { confirm: t('.delete_confirm', title: collection.title) } %>
-        <% end %>
-      </div>
+<li class="list-group-item list-group-item-action">
+  <div class="d-flex justify-content-between">
+    <%= link_to collection.title, community_collection_path(community, collection) %>
+
+    <div class="btn-group">
+      <% if policy(collection).edit? %>
+        <%= link_to t('edit'), edit_community_collection_path(community, collection),
+            class:'btn btn-secondary edit-collection' %>
+      <% end %>
+      <% if policy(collection).destroy? %>
+        <%= link_to t('delete'), community_collection_path(community, collection),
+            class: 'btn btn-danger delete-collection', method: :delete,
+            data: { confirm: t('.delete_confirm', title: collection.title) } %>
+      <% end %>
     </div>
   </div>
 </li>
--- a/app/views/admin/communities_and_collections/_collections_list.html.erb
+++ b/app/views/admin/communities_and_collections/_collections_list.html.erb
@@ -1,0 +1,7 @@
+<% if @community.member_collections.count > 0 %>
+  <%= render partial: 'admin/communities_and_collections/collection', collection: @community.member_collections, locals: {community: @community}  %>
+<% else %>
+  <li class="list-group-item list-group-item-action">
+    <%= t('communities.show.no_collections') %>
+  </li>
+<% end %>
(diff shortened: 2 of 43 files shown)


app/views/admin/communities_and_collections/_collection.html.erb
```diff
@@ -1,20 +1,17 @@
-<li class="list-group-item list-group-item-info">
```
murny: info is a very bright teal in bootstrap 4 beta...lets just use less color in general and not have anything (changed it to `list-group-item-action` which means when you hover over it it darkens under your mouse)

In the future we should drop the list groups for displaying these lists and use media panels (so theres nice white spacing between each item)

app/views/admin/communities_and_collections/_collection.html.erb
```diff
@@ -1,20 +1,17 @@
...
-              class:'btn btn-secondary edit-collection' %>
-        <% end %>
-        <% if policy(collection).destroy? %>
-          <%= button_to t('delete'), community_collection_path(community, collection),
-              class: 'btn btn-danger delete-collection', method: :delete,
-              data: { confirm: t('.delete_confirm', title: collection.title) } %>
-        <% end %>
-      </div>
+<li class="list-group-item list-group-item-action">
+  <div class="d-flex justify-content-between">
```
murny: use flex to position these elements instead of cols/margins/pull-rights/etc

app/views/admin/communities_and_collections/_collection.html.erb
```diff
@@ -1,20 +1,17 @@
...
-      <%= link_to collection.title, community_collection_path(community, collection) %>
-    </div>
-    <div class="col ml-auto text-right">
-      <div class="btn-group">
-        <% if policy(collection).edit? %>
-          <%= link_to t('edit'), edit_community_collection_path(community, collection),
-              class:'btn btn-secondary edit-collection' %>
-        <% end %>
-        <% if policy(collection).destroy? %>
-          <%= button_to t('delete'), community_collection_path(community, collection),
```
murny: `button_to` actually is an embed button wrapped in a form element. This breaks bootstraps `btn-group` and makes this look extremely weird. Let's just use links here? All still works

app/views/admin/communities_and_collections/_collections_list.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if @community.member_collections.count > 0 %>
+  <%= render partial: 'admin/communities_and_collections/collection', collection: @community.member_collections, locals: {community: @community}  %>
+<% else %>
+  <li class="list-group-item list-group-item-action">
+    <%= t('communities.show.no_collections') %>
```
murny: I wanted a way to show that theres no collections. 

Before when you click `collections` button on a community with no collections, absolutely nothing happned, which is bad UX. Now you get a single row which states there's no collections in this commuity.

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1,29 +1,34 @@
 <% page_title(t('.header')) %>
 
+<ol class="breadcrumb mb-5">
```
murny: Keep consistent and add breadcrumbs everywhere in the admin backend

app/views/admin/communities_and_collections/index.html.erb
```diff
@@ -1,29 +1,34 @@
...
-        <div class="ml-auto">
+        <div class="d-flex justify-content-between">
           <%= link_to admin_communities_and_collection_path(community), data: {community_id: community.id }, class: 'btn btn-secondary collection-btn', remote: true do %>
-            <%= t('collections.header') %>
+            <%= t('collections.index.header') %>
             <%= fa_icon('chevron-down') %>
           <% end %>
 
-          <%= button_tag type: 'button', class: 'btn btn-secondary close-btn hidden-xs-up', data: {community_id: community.id } do %>
+          <%= button_tag type: 'button', class: 'btn btn-secondary close-btn d-none', data: {community_id: community.id } do %>
```
murny: hidden-* classes and others are all removed in beta: https://getbootstrap.com/docs/4.0/migration/#responsive-utilities

They are now display none (d-none) and the various sizes (d-md-none, etc)

app/views/communities/show.html.erb
```diff
@@ -28,30 +28,21 @@
...
     </h4>
-  </div>
   <% if policy(:collection).create? %>
-    <div class="col text-right">
       <%= link_to t('.create_new_collection'), new_community_collection_path(@community),
           class: 'btn btn-primary' %>
-    </div>
   <% end %>
 </div>
-<% if @community.member_collections.count > 0 %>
```
murny: Improved this, we were checking this array multiple times and doing different things like changing the header. Keep the UI consistent and instead of listing the items, list an item with "no collections found" etc

app/views/communities/show.html.erb
```diff
@@ -28,30 +28,21 @@
...
-    <div class="col">
-      <div class="list-group">
-        <%= render partial: 'admin/communities_and_collections/collection',
-            collection: @community.member_collections, locals: { community: @community } %>
-      </div>
+
+<div class="row mt-3">
+  <div class="col">
+    <div class="list-group">
+      <%# should be reversed, admin should be using communities partial not vice versa %>
```
murny: as comment says, be smarter for admin to consume collections partial instead of the other way around.

app/views/admin/communities_and_collections/_collection.html.erb
cwant: Agreed, the teal was horrible.

app/views/admin/communities_and_collections/_collection.html.erb
cwant: That's great -- I didn't know that the new bootstrap had flexbox support. The spacing looks a lot better.

config/locales/en.yml
```diff
@@ -95,7 +95,8 @@ en:
     communities_and_collections:
       title: Communities and Collections
       index:
-        header: 'Manage Communities / Collections'
+        header: 'Manage Communities & Collections'
+        delete_confirm: 'Are you sure you want to delete community "%{title}"?'
```
cwant: Watch out for this delete confirm -- this translation is defined elsewhere in master (under the 'communities' controller/view translations).

app/views/admin/communities_and_collections/_collection.html.erb
mbarnett: Yeah seriously why is "disgusting teal" the colour of "info", that was brutal to look at.

Ewwwww

PR #95, opened 2017-08-31
Community logo with ActiveStorage

cwant:
This is mostly done. Could use another set of eyes though.

Things to watch out for:

* ActiveStorage monkey patched
* GlobalID added to Community to allow ActiveStorage to attach things to it
* `remove_logo` virtual attribute in Community -- not sure how I feel about this, but makes the controller/form easy

Still needed:

* More tests

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,8 +1,9 @@
-//= require communities
 @import "bootstrap";
 
 // TODO: weird bug? had to .css this
 @import "font-awesome.css";
+
+@import "communities";
 
 // TODO: move this stuff below to a general jupiter.scss
 // padding for fixed header.
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,3 +1,4 @@
+//= require communities
 @import "bootstrap";
 
 // TODO: weird bug? had to .css this
--- a/app/assets/stylesheets/communities.scss
+++ b/app/assets/stylesheets/communities.scss
@@ -1,0 +1,16 @@
+$no_logo: #888888;
+
+.community-logo-small {
+  height: 100px !important;
+  width: 100px !important;
+}
+
+.community-logo-small.no-logo {
+  color: $no_logo;
+
+  p {
+    font-size: 8pt;
+    font-style: italic;
+    margin-bottom: 0px;
+  }
+}
--- a/app/assets/stylesheets/communities.scss
+++ b/app/assets/stylesheets/communities.scss
@@ -1,6 +1,6 @@
 .community-logo-small {
-  height: 100px !important;
-  width: 100px !important;
+  height: 100px;
+  width: 100px;
 }
 
 .community-logo-small.no-logo {
--- a/app/assets/stylesheets/communities.scss
+++ b/app/assets/stylesheets/communities.scss
@@ -1,12 +1,10 @@
-$no_logo: #888888;
-
 .community-logo-small {
   height: 100px !important;
   width: 100px !important;
 }
 
 .community-logo-small.no-logo {
-  color: $no_logo;
+  color: gray;
 
   p {
     font-size: 8pt;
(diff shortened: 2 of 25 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -1,3 +1,4 @@
+//= require communities
```
murny: Same commet here: https://github.com/ualbertalib/jupiter/pull/104/files#r137047688

Use Sass `@import` over sprockets `require`

More info here: https://github.com/rails/sass-rails#important-note

Also put this stuff after vendor imports ( so you can reuse their variables among other reasons)

app/assets/stylesheets/communities.scss
```diff
@@ -0,0 +1,16 @@
+$no_logo: #888888;
```
murny: If you do the above, you could use bootstraps "gray" color instead. Just use $gray everywhere. It's defined as 
`#868e96` which is pretty close to your `#888888`

app/assets/stylesheets/communities.scss
```diff
@@ -0,0 +1,16 @@
...
+
+.community-logo-small {
+  height: 100px !important;
```
murny: Do you need to use `!important`?  Should be used sparingly and only as a last resort. Chances are you don't need it

app/models/community.rb
```diff
@@ -8,8 +11,30 @@ def member_collections
     Collection.where(community_id: id)
   end
 
+  def logo
+    @active_storage_attached_logo ||
+      (@active_storage_attached_logo = ActiveStorage::Attached::One.new(:logo, self))
```
murny: couldn't this be rewritten as:

`@active_storage_attached_logo ||= ActiveStorage::Attached::One.new(:logo, self)`?

app/assets/stylesheets/communities.scss
cwant: Hrmm, just using `gray` works, but using `$gray` throws an error. Using `gray` gives `#808080`, which is also very close.

app/models/community.rb
cwant: Yeah, I guess so. I was basically replicating what was happening here:

https://github.com/rails/activestorage/blob/v0.1/lib/active_storage/attached/macros.rb#L4

app/assets/stylesheets/communities.scss
cwant: Yep, not needed.

app/assets/stylesheets/communities.scss
murny: You should be  able to use `$gray` now after sass-rails is being used when #108 gets merged :confused:

cwant: Hrmm, merged in from master now that #108 is in, still $gray doesn't work (do I need to run a rake task or something?).

murny: Mmm wouldn't worry about it. I'll revisit this once this is merged. Probably something funky happening.

murny: Yeah looks like they changed it: https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L11-L62

So you can use one of the $gray-### ones or use the methods provided for accessing the  variable maps like: `colors("gray")` or `grays(600)`

cwant:
Model tests added.

mbarnett:
Whoops, thought of one more thing. We'll need to display the logos on the Community index page and in the admin section ala https://jupiter.mybalsamiq.com/projects/jupiter/Admin%20-%20Manage%20Communities%20and%20Collections

cwant:
Okay, updated the "Done when" for the issue. This should be pretty straightforward.

mbarnett:
Thanks!

cwant:
Branch updated for logo on index page, this is what it looks like now:

community-index-logo

(Clicking the logo or "no logo" takes user to community, same as clicking the title link for the community.)

cwant:
... and I just noticed that the "Has No Logo" in the screenshot means that the translation wasn't found. Will fix.

cwant:
Okay, here we go:

community-index-logo2

Please kindly press the "Approve" button now!

PR #105, opened 2017-09-05
updated readme, and added deploy.sh for ssh deployment of jupiter master

weiweishi:
The deploy.sh allows devs to do 
`ssh deploy@uatserver /home/deploy/jupiter/deploy.sh`

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,5 +1,8 @@
 cd /home/deploy/jupiter && git fetch --all && git reset --hard origin/master
-docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml stop web
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml down web
 docker rmi $(docker images -f "dangling=true" -q)
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml build web
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml up -d
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml run --rm web rake db:migrate
+#Recompile assets is optional, and can be commented out if not needed
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml run --rm web rake assets:precompile
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,5 +1,5 @@
-cd /home/deploy/jupiter && git pull origin master
-docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml rm -f -s -v
+cd /home/deploy/jupiter && git fetch --all && git reset --hard origin/master
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml stop web
 docker rmi $(docker images -f "dangling=true" -q)
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml build web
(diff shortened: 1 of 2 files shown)


deploy.sh
```diff
@@ -0,0 +1,5 @@
+cd /home/deploy/jupiter && git pull origin master
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml rm -f -s -v
```
murny: Random question...Do you have to remove all the containers for this? Or is this just a safer/more foolproof way? 

Should be just be able to just down the web container, rebuild web container  (or pull from dockerhub), and up the web container?

murny: Your also removing the volumes here (your using `-v` flag) ? This will blank the postgres/redis/fedora/solr/etc databases between migration. Is this desired? I would assume not unless you want this to be a true sandbox server (that doesn't guartee data)

weiweishi: These two lines just a precaution to avoid accumulating excessive numbers of unused/dangling images and containers that might potentially consume disk space, or interfere with the next deployment, if the script was running unsuccessfully. Not sure if it's necessary. Got the idea from http://paislee.io/how-to-automate-docker-deployments/

weiweishi: True. Will modify it.

murny: The `docker rmi $(docker images -f "dangling=true" -q` can stay and will probably catch these things. But here your force stopping and removing all docker-compose containers (including their volumes), if I'm reading this right: 
https://docs.docker.com/compose/reference/rm/

I'd just do what the article said here with regards to the deployment script: 
http://paislee.io/how-to-automate-docker-deployments/#deploymentscript

Except instead of pulling the image from dockerhub, we just going to `down` the web app, rebuild it, and start it again etc.

weiweishi:
@murny updated the script for your review.

deploy.sh
```diff
@@ -0,0 +1,5 @@
+cd /home/deploy/jupiter && git fetch --all && git reset --hard origin/master
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml stop web
```
murny: Stop will just stop the container? Is that enough? Or should you also remove the container as well? (use `down` to do both at the same time)

deploy.sh
```diff
@@ -0,0 +1,5 @@
...
+docker rmi $(docker images -f "dangling=true" -q)
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml build web
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml up -d
```
murny: Should we always rake db:migrate in the deploy script? `docker-compose run --rm web rake db:migrate`

murny: Or at least put it as a comment or somewhere in the readme? Maybe recompiling assets too?

```(optional) docker-compose run --rm web rake db:migrate to perform database migrations. Does nothing if your database is up to date.
(optional) docker-compose run --rm web rake assets:precompile to compile new JS and CSS assets.```

deploy.sh
murny: Is this even needed? Might be able to just build and then up the container?

For example here: https://docs.docker.com/compose/production/#deploying-changes

> $ docker-compose build web
> $ docker-compose up --no-deps -d web

PR #106, opened 2017-09-05
Temp fixes for work upload

mbarnett:
The form as it exists is temporary pending other sprints, but this at least lets it work somewhat sensibly for testing purposes. Mostly it excludes various "internal" fields from the page and adds something to make visibility usable by humans.

### DIFF
--- a/app/views/works/_form.html.erb
+++ b/app/views/works/_form.html.erb
@@ -1,7 +1,13 @@
 <%= simple_form_for work, html: {multipart: true} do |f| %>
   <% work.display_attributes.each do |prop, _| %>
+    <% next if ['visibility', 'owner', 'record_created_at', 'doi', 'embargo_end_date'].include? prop %>
     <%= f.input prop %>
   <% end %>
+  <%= f.input :visibility,
+  as: :unread_select,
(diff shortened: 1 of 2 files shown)


app/views/works/_form.html.erb
```diff
@@ -1,7 +1,13 @@
 <%= simple_form_for work, html: {multipart: true} do |f| %>
   <% work.display_attributes.each do |prop, _| %>
```
murny: Should this be `work.safe_attributes` instead? Or does this also have the same below attributes as well? Let's make note and fix this? Cause even though these aren't on the form, people can still submit a `curl`/etc to update/add these.

I'll approve for now though

app/views/works/_form.html.erb
```diff
@@ -1,7 +1,13 @@
...
+    <% next if ['visibility', 'owner', 'record_created_at', 'doi', 'embargo_end_date'].include? prop %>
     <%= f.input prop %>
   <% end %>
+  <%= f.input :visibility,
+  as: :unread_select,
+  collection: ['public', 'private', 'authenticated'],
```
murny: This array should be coming from somewhere instead of being hardcoded in the view

PR #108, opened 2017-09-06
Font awesome issues

### LINKED ISSUE #107: Font Awesome isn't loading on UAT
As an example, down chevrons are not appearing on the "Collections" button on /admin/communities_and_collections

murny:
Shoud fix #107


Gemfile
```diff
@@ -10,13 +10,13 @@ gem 'puma', '~> 3.7'
 gem 'rails', '~> 5.1.1'
 
 # Assets (CSS/JS) stuff
+gem 'bootstrap', '~> 4.0.0.beta'
 gem 'font-awesome-rails'
 gem 'jquery-rails'
+gem 'sass-rails', '~> 5.0'
```
murny: We were missing the beautiful gem... :cry:

lib/tasks/rdoc.rake
```diff
@@ -1,11 +1,15 @@
-require 'sdoc'
-require 'rdoc/task' # ensure this file is also required in order to use `RDoc::Task`
+if Rails.env.development? || Rails.env.test?
```
murny: Lets move this out of production and only have the rake task available in development/test?

.editorconfig
```diff
@@ -1,3 +1,17 @@
-[*.rb]
-indent_style = space
+
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
```
murny: Fix this and make more general, this was making it super hard for me to edit css/sass files/etc

mbarnett: What does root=true control, out of curiosity?

lib/tasks/rdoc.rake
mbarnett: It's that or just more the require inside the task? IDK I can't think of a reason we'd want it in other environments, but we've been a bit too restrictive in other places (seeds should be available in more environments than just dev, say)

.editorconfig
murny: Root means this file is the top-most EditorConfig file. So it will stop looking recursivley pass this editorconfig. You can have sub editorconfigs in different sub folders etc. And it will continue to look up etc.

mbarnett: Gotcha

lib/tasks/rdoc.rake
murny: Require is already inside this task. I also made it `require: false` in the gemfile so its not loading with the application so this could be an option as well. If we find we need this in other environments then for sure we can remove these if statements.

For seeds, we just have to be careful. I agree theres cases where seeds are good for production etc. But right now I got drop/destroy_all statements in the seed.rb which shouldn't be ran on production. Also this is a good usecase for something more powerful like https://github.com/mbleigh/seed-fu. To handle seeds in different environments

mbarnett: No, sorry, I meant, you could shove it literally inside the task:

```ruby
RDoc::Task.new do |rdoc|
  require 'sdoc'
  rdoc.rdoc_dir = 'docs/rdoc'
```

That would stop it from throwing load errors until it's actually run, instead of when it's first parsed. The require false should fix it too, though

murny: ah gotcha, makes sense

PR #109, opened 2017-09-06
Use all the same headings for page headers (and fix broken system tests)

murny:
We using h1's in some places and h2's in other places. Lets use the same everywhere.

Top heading of the page should be an H1 lesser headers should be h2-h4. (Why? Good for SEO and good for accessibility, etc)

If theses are too big we can change the base font to a smaller size or do something smarter like responsive fonts (or use some css to change it, etc.)


cwant:
Looks good here (I'm not selected as a reviewer for this one, but can approve if needed).

mbarnett:
Feel free to approve things you're not "selected" on, imo. Unless I hate it, then please read my mind and reject on my behalf.

PR #110, opened 2017-09-06
[WIP] First stab at treating FileSets as LockedLdpObjects

cwant:
Needs a sanity check before proceeding.

### DIFF
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -1,10 +1,29 @@
 class FileSet < JupiterCore::LockedLdpObject
 
   ldp_object_includes Hydra::Works::FileSetBehavior
-  has_attribute :is_member_of, ::VOCABULARY[:ualib].path, solrize_for: :pathing
+  solr_index :member_of_collections, solrize_for: :exact_match,
+             as: -> { member_of_collections&.to_a.map { |item| item.id} }
+
+  # see also https://github.com/samvera/hydra-works/wiki/Lesson%3A-Add-attached-files
+  def self.add_new_to_work(file, unlocked_work)
+    fileset = self.new_locked_ldp_object
+    fileset.unlock_and_fetch_ldp_object do |unlocked_fileset|
+      unlocked_fileset.owner = unlocked_work.owner
+      unlocked_fileset.visibility = unlocked_work.visibility
+      Hydra::Works::AddFileToFileSet.call(unlocked_fileset, file, :original_file,
+                                          update_existing: false, versioning: false)
+      unlocked_fileset.member_of_collections += [unlocked_work]
+      unlocked_fileset.save!
+      unlocked_work.members << unlocked_fileset
+      # pull in hydra derivatives, set temp file base
+      # Hydra::Works::CharacterizationService.run(fileset.characterization_proxy, filename)
+    end
+  end
 
   def item
-    Work.find(is_member_of)
+    item_id = read_solr_index(:member_of_collections)
+    return nil unless item_id.present?
+    Work.find(item_id.first)
   end
 
 end
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -1,5 +1,10 @@
-class FileSet < ActiveFedora::Base
+class FileSet < JupiterCore::LockedLdpObject
 
-  include Hydra::Works::FileSetBehavior
+  ldp_object_includes Hydra::Works::FileSetBehavior
+  has_attribute :is_member_of, ::VOCABULARY[:ualib].path, solrize_for: :pathing
+
+  def item
+    Work.find(is_member_of)
+  end
 
 end
--- a/app/models/jupiter_core/locked_ldp_object.rb
+++ b/app/models/jupiter_core/locked_ldp_object.rb
@@ -40,7 +40,7 @@ module JupiterCore
     # inheritable class attributes (not all class-level attributes in this class should be inherited,
     # these are the inheritance-safe attributes)
     class_attribute :af_parent_class, :attribute_cache, :attribute_names, :facets,
-                    :solr_name_cache, :reverse_solr_name_cache, :solr_calc_attributes
+                    :reverse_solr_name_cache, :solr_calc_attributes
 
     # Returns the id of the object in LDP as a String
     def id
@@ -118,6 +118,14 @@ module JupiterCore
     def errors
       return super unless ldp_object.present?
       ldp_object.errors
+    end
+
+    def read_solr_index(name)
+      raise PropertyInvalidError unless name.is_a? Symbol
+      type = self.solr_calc_attributes[name]
+      raise PropertyInvalidError unless type.present?
+      solr_name = Solrizer.solr_name(name, :symbol, type: type)
+      solr_representation[solr_name]
     end
 
     # Use this to create a new +LockedLDPObjects+ and its underlying LDP instance. attrs populate the new object's
@@ -425,62 +433,13 @@ module JupiterCore
       #
       #    solr_index :downcased_title, solrize_for: :exact_match, as: -> { title.downcase }
       #
-      def solr_index(name, solrize_for:, as:, type: :string)
+      def solr_index(name, solrize_for:, as:)
         raise PropertyInvalidError unless as.respond_to?(:call)
-        raise PropertyInvalidError unless solrize_for.is_a?(Symbol)
-
-        add_to_solr_index(name, solrize_for: solrize_for, type: type)
+        raise PropertyInvalidError unless name.present?
+        raise PropertyInvalidError unless solrize_for.present? && solrize_for.is_a?(Symbol)
 
         self.solr_calc_attributes ||= {}
-        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for],
-                                            callable: as }
-      end
-
-      def add_to_solr_index(name, solrize_for:, multiple: false, type: :string)
-        raise PropertyInvalidError unless name.is_a? Symbol
-        raise PropertyInvalidError unless solrize_for.present?
-
-        solrize_for = Array(solrize_for)
-
-        # Index should contain only some combination of:
-        #      :search, :sort, :facet, :symbol, and :path
-        # this isn't an exhaustive layering over this mess
-        # https://github.com/mbarnett/solrizer/blob/e5dd2bd571b9ebdb8a8ab214574075c28951e53e/lib/solrizer/default_descriptors.rb
-        # but it helps
-        raise PropertyInvalidError if
-          solrize_for.count { |item| !SOLR_DESCRIPTOR_MAP.keys.include?(item) } > 0
-
-        # TODO: type validation
-
-        self.solr_name_cache ||= {}
-        self.solr_name_cache[name] ||= []
-        solrize_for.each do |descriptor|
-          solr_name = Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[descriptor], type: type)
-          self.solr_name_cache[name] << solr_name
-          self.reverse_solr_name_cache[solr_name] = name
-        end
-
-        if solrize_for.include?(:facet)
-          self.facets <<
-            Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[:facet], type: type)
-        end
-        if solrize_for.include?(:path)
-          self.facets <<
-            Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[:path], type: type)
-        end
-
-        # Define getter
-        define_method name do
-          val = if ldp_object.present?
-                  ldp_object.send(name)
-                else
-                  solr_representation[self.class.solr_name_cache[name].first]
-                end
-          return if val.nil?
-          return DateTime.parse(val).freeze if type == :date
-          return val.freeze if val.nil? || multiple || !val.is_a?(Array)
-          return val.first.freeze
-        end
+        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for], callable: as }
       end
 
       def has_multival_attribute(name, predicate, solrize_for: [], type: :string)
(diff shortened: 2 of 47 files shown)


app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -40,7 +40,7 @@ class LockedLdpObject
     # inheritable class attributes (not all class-level attributes in this class should be inherited,
     # these are the inheritance-safe attributes)
     class_attribute :af_parent_class, :attribute_cache, :attribute_names, :facets,
-                    :reverse_solr_name_cache, :solr_calc_attributes
+                    :solr_name_cache, :reverse_solr_name_cache, :solr_calc_attributes
```
mbarnett: This is going to break on subclassing (adding a solr_name to the subclass will also cause it to be added to the parent class) unless you also update `inherited` to dup this.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
       #
       #    solr_index :downcased_title, solrize_for: :exact_match, as: -> { title.downcase }
       #
-      def solr_index(name, solrize_for:, as:)
+      def solr_index(name, solrize_for:, as:, type: :string)
         raise PropertyInvalidError unless as.respond_to?(:call)
-        raise PropertyInvalidError unless name.present?
-        raise PropertyInvalidError unless solrize_for.present? && solrize_for.is_a?(Symbol)
+        raise PropertyInvalidError unless solrize_for.is_a?(Symbol)
```
mbarnett: Why are we replacing the check for name? The problem with kicking the can down to `add_to_solr_index` is that now the backtrace doesn't show you that the error is coming from the public-facing method you called in the model (oh, I passed the wrong argument), it shows it as originating deeper down the stack.

Even if it's redundant I'd rather sanitize the arguments at the point of contact with the public, rather than invite them further down the stack.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
 
-        self.solr_calc_attributes ||= {}
-        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for], callable: as }
-      end
+        add_to_solr_index(name, solrize_for: solrize_for, type: type)
```
mbarnett: Method name is misleading. Nothing is added to any solr index after calling this.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
       #
       #    solr_index :downcased_title, solrize_for: :exact_match, as: -> { title.downcase }
       #
-      def solr_index(name, solrize_for:, as:)
+      def solr_index(name, solrize_for:, as:, type: :string)
```
mbarnett: My preference would be to list arguments in "conventional call order" or "how I'd like to see these laid out in a model". It would be really awkward to define the type after the lambda. `name, type:, solrize_for:, as:` would probably be preferable. The nice thing about keywords args is that order doesn't matter when it comes to defaults.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
-      def has_multival_attribute(name, predicate, solrize_for: [], type: :string)
-        has_attribute(name, predicate, multiple: true, solrize_for: solrize_for, type: type)
+        self.solr_calc_attributes ||= {}
+        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for],
+                                            callable: as }
       end
 
-      # a utility DSL for declaring attributes which allows us to store knowledge of them.
-      def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :string)
+      def add_to_solr_index(name, solrize_for:, multiple: false, type: :string)
```
mbarnett: I think we can get rid of this, see other notes about `attribute_cache`

app/models/work.rb
```diff
@@ -29,6 +29,10 @@ def self.valid_visibilities
     super + [VISIBILITY_EMBARGO]
   end
 
+  def member_file_sets
```
mbarnett: "member" is probably redundant?

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
-        self.facets << Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[:path], type: type) if solrize_for.include?(:path)
-
-        self.attribute_cache[name] = {
-          predicate: predicate,
-          multiple: multiple,
-          solrize_for: solrize_for,
-          type: type,
-          solr_names: solr_name_cache
-        }
+        if solrize_for.include?(:facet)
```
mbarnett: I'd really prefer single line conditionals where possible

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
+          self.facets <<
+            Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[:facet], type: type)
+        end
+        if solrize_for.include?(:path)
+          self.facets <<
+            Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[:path], type: type)
+        end
```
mbarnett: Turning these one-liners into 3 line blocks is a net decrease in readability in my opinion. The linebreak after the operator is particularly unpleasant.

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: In fact, reviewing further along, let's not create another redundant cache (I'm already juggling too many).

You can already use `attribute_cache[name].solr_names` to get this information without all of the changes related to this.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
+        solrize_for = Array(solrize_for)
 
-        # index should contain only some combination of :search, :sort, :facet, :symbol, and :path
+        # Index should contain only some combination of:
+        #      :search, :sort, :facet, :symbol, and :path
         # this isn't an exhaustive layering over this mess
         # https://github.com/mbarnett/solrizer/blob/e5dd2bd571b9ebdb8a8ab214574075c28951e53e/lib/solrizer/default_descriptors.rb
         # but it helps
-        raise PropertyInvalidError if solrize_for.count { |item| !SOLR_DESCRIPTOR_MAP.keys.include?(item) } > 0
+        raise PropertyInvalidError if
```
mbarnett: Remove this linebreak please.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
+        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for],
+                                            callable: as }
       end
 
-      # a utility DSL for declaring attributes which allows us to store knowledge of them.
-      def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :string)
+      def add_to_solr_index(name, solrize_for:, multiple: false, type: :string)
         raise PropertyInvalidError unless name.is_a? Symbol
-        raise PropertyInvalidError unless predicate.present?
+        raise PropertyInvalidError unless solrize_for.present?
```
mbarnett: `solrize_for` wasn't actually required in calls to has_attribute, previously.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
 
-        # TODO: keep this conveinience, or push responsibility for [] onto the callsite?
-        solrize_for = [solrize_for] unless solrize_for.is_a? Array
+        solrize_for = Array(solrize_for)
```
mbarnett: Don't use `Array()`, it has terrible (and not-equivalent) semantics:

```ruby
 Array({a: 1, b: 2})
 => [[:a, 1], [:b, 2]] 
```
Nobody expects it to work the way it actually works.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
 
-        self.attribute_names << name
-
-        solr_name_cache ||= []
+        self.solr_name_cache ||= {}
+        self.solr_name_cache[name] ||= []
         solrize_for.each do |descriptor|
           solr_name = Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[descriptor], type: type)
-          solr_name_cache << solr_name
+          self.solr_name_cache[name] << solr_name
```
mbarnett: Gathering similar notes in one spot: you don't need to do any of 455, 456, this, 477, or 506 if you use `attribute_cache`

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
+          solr_names: self.solr_name_cache[name]
+        }
 
         define_method "#{name}=" do |*_args|
-          raise LockedInstanceError, 'The Locked LDP object cannot be mutated outside of an unlocked block or without'\
-                                     'calling unlock_and_fetch_ldp_object to load a writable copy (SLOW).'
+          raise LockedInstanceError,
+                'The Locked LDP object cannot be mutated outside of an '\
+                'unlocked block or without calling unlock_and_fetch_ldp_object '\
+                'to load a writable copy (SLOW).'
```
mbarnett: Both to reiterate and clarify, the rubocop linebreak settings are at 120. Please adjust your guidelines to 120 (I'm guessing you're adding all these linebreaks at 80?), and don't make things unnecessarily vertical on my nice long-on-the-horizontal-axis short-on-the-vertical-access screen.

app/models/file_set.rb
```diff
@@ -1,5 +1,10 @@
-class FileSet < ActiveFedora::Base
+class FileSet < JupiterCore::LockedLdpObject
 
-  include Hydra::Works::FileSetBehavior
+  ldp_object_includes Hydra::Works::FileSetBehavior
+  has_attribute :is_member_of, ::VOCABULARY[:ualib].path, solrize_for: :pathing
```
mbarnett: Just sanity checking that we're still planning on using `member_of_collection` instead of this?

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: Actually on further review I think this method becomes unnecessary if you just use `attribute_cache`.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
+
+      def has_multival_attribute(name, predicate, solrize_for: [], type: :string)
+        has_attribute(name, predicate, multiple: true, solrize_for: solrize_for, type: type)
+      end
+
+      # a utility DSL for declaring attributes which allows us to store knowledge of them.
+      def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :string)
+        raise PropertyInvalidError unless predicate.present?
+
+        solrize_for = Array(solrize_for)
```
mbarnett: Array() is bad mojo, as above.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -425,70 +425,92 @@ def derived_af_class
...
         end
+      end
+
+      def has_multival_attribute(name, predicate, solrize_for: [], type: :string)
+        has_attribute(name, predicate, multiple: true, solrize_for: solrize_for, type: type)
+      end
+
+      # a utility DSL for declaring attributes which allows us to store knowledge of them.
+      def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :string)
+        raise PropertyInvalidError unless predicate.present?
```
mbarnett: Bring back the check for name.

app/models/jupiter_core/locked_ldp_object.rb
cwant: The main reason (actually the only reason) I'm doing this is so that I can create `attribute_cache[name][:solr_names]` on line 506. (Chicken/Egg problem). I could use the `reverse_solr_name_cache` to generate this info, or use a separate method to generate the name cache when needed

app/models/file_set.rb
cwant: Yep, that is the plan so far (I have some work on this that I haven't committed yet, wanted to get some feedback on the LockedLdpObject changes first).

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: I'm not quite following you. You don't need to generate `attribute_cache[name][:solr_names]`, because it already existed prior to your changes.

mbarnett: ie on master:

```ruby
2.4.1 :002 > Work.first.attribute_cache[:title][:solr_names]
 => ["title_tesim", "title_sim", "title_ssi"] 
```

already works.

app/models/jupiter_core/locked_ldp_object.rb
cwant: This method also creates the getter. Previously the getter was only created for `has_attribute`, not for `solr_index`, so this method deals with that and the other common parts. I am toying with the method name `register_solr_name` instead of `add_to_solr_index` ...

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: Thinking some more about this PR, I'm also not sure we want to intermix "real" metadata about ActiveFedora properties with metadata about "virtual" properties about solr indexes, because code elsewhere makes various assumptions about things it finds in these caches.

Really, from my original understanding of what you wanted to do, all this PR should be doing is something like change https://github.com/ualbertalib/jupiter/blob/master/app/models/jupiter_core/locked_ldp_object.rb#L428-L435

to 

```ruby
      def solr_index(name, type: :string, solrize_for:, as:)
        raise PropertyInvalidError unless as.respond_to?(:call)
        raise PropertyInvalidError unless name.present?
        raise PropertyInvalidError unless solrize_for.present? && solrize_for.is_a?(Symbol)

        self.solr_calc_attributes ||= {}
        self.solr_calc_attributes[name] = { type: SOLR_DESCRIPTOR_MAP[solrize_for], callable: as }
        solr_name = Solrizer.solr_name(name, SOLR_DESCRIPTOR_MAP[solrize_for], type: type)
       define_method name do
          val = if ldp_object.present?
                  ldp_object.instance_exec(&as)
                else
                  solr_representation[solr_name]
                end
          return if val.nil?
          return DateTime.parse(val).freeze if type == :date
          return val.freeze if val.nil? || multiple || !val.is_a?(Array)
          return val.first.freeze
       end
      end
```

No?

cwant: Okay, will go down this path. I had thought there might be an opportunity to DRY this up (since `has_attribute` and `solr_index` do some similar things), but it might be at the cost of keeping things simple.

mbarnett: Their deceptively similar with subtle differences. I don't actually know that those last 4 lines are all applicable, since there's no real meaningfulness to an attribute that only lives in solr being "single-valued" (all solr values are array, only Fedora understands the concept of a single-valued item).

So it might make more sense to just return val.freeze. I'm not even sure if it makes a lot of sense to give them types other than string.

app/models/work.rb
cwant: Regarding that, just following the convention found in other models, e.g.,

`community.member_collections`
`collection.member_works`

This is why I went with `work.member_filesets`.

cwant:
Closing this PR. Will push a new branch today that has this code reworked to fit in with the Work->Item rename.

PR #112, opened 2017-09-07
Rename works

### LINKED ISSUE #93: Change "Works" to "Items" as discussed in meeting
Need to update at very least all text saying works to items.

Could take this step further and update the model and everything else referencing works to items as well?

murny:
Closes #93 

Requires a migration of data...aka you delete all fedora/solr data (rerun the seeds)


app/assets/javascripts/jupiter/items.js
```diff
@@ -1,30 +1,27 @@
...
+	id =  $("option:selected", '#new_item #community').val();
+
 	$.getJSON('/communities/' + id + '.json').done(function(data) {
 		var items = "";
 		$.each(data, function(idx, item) {
 			items += '<option value="' + item.id + '">' + item.title + '</option>';
 		});
-		$('#new_work #collection').empty().removeAttr('disabled').append(items);
-	}).fail(function() {
-		console.log('dang');
```
murny: just dont handle the fail case?

app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Admin::DashboardController < Admin::AdminController
 
   def index
-    @works = Work.limit(10).sort(:date_created, :desc)
+    @items = Item.limit(10).sort(:date_created, :desc) # TODO: Should use record_created_at?
```
murny: Probably should transfer all this stuff over to the `record_created_at` date now that it implemented and working

app/assets/javascripts/jupiter/items.js
```diff
@@ -1,30 +1,27 @@
 $(document).on('turbolinks:load', function() {
-	$("#new_work #community").change(function() {
+	$("#new_item #community").change(function() {
 		update_collections();
 	});
 
-	$("#new_work .add-file").click(function(event){
-		console.log('asdf');
```
murny: toss this

murny:
Starting to think this isnt possible or extremely hard to do. I got as close as I can. If you look at the individual commits, this conforms to how you want it. But now that it is being merged into this branch it goes to back how git sees it (because there are so many changes, they are basically new files (treated as a rm/add instead of a rename)

Maybe it will be fine if you merge this as is (without squashing the commits/etc)

app/assets/javascripts/jupiter/items.js
cwant: In master there is also a `console.log('dang')` that might get pulled in on your next merge.

app/assets/javascripts/jupiter/items.js
cwant: Oops, re: comment above, looks like you got it here.

cwant:
Other than potential merge conflicts, this looks great -- can't find any reference of `work` anywhere, other than related to HydraWorks or usage in English sentences (e.g. in the test code).

mbarnett:
I think our best bet might be to land all of our in-flight work touching work_* files, then stop, pull this in. The merge is going to be a huge pain otherwise.

cwant:
I could go either way. I have a couple of branches that touch works, but I don't mind cherry-pick + restructuring them on different branches after the fact. (Or hand crafting diff files, if need be.)

mbarnett:
Let's land this and then worry about fixing the rest of the PRs

PR #113, opened 2017-09-08
Fix a11y errors on wrong markup being nested in <ul> elements

murny:
Found small issue when playing around with a11y tools

Only `<li>` elements should be nested within `<ul> or <ol>` elements.

Also `TODO` a weird behavior in communties/collections and admin community and collection pages when editing/deleting stuff

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -41,6 +41,12 @@ class CollectionsController < ApplicationController
       unlocked_collection.update!(permitted_attributes(Collection))
     end
     flash[:notice] = t('.updated')
+
+    # TODO: Needs to be rethinked (and destroy action for both communties and collections)
+    # You can access this from both admin communties and collections index as well as communities show page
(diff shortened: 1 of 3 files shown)


app/controllers/collections_controller.rb
```diff
@@ -41,6 +41,12 @@ def update
       unlocked_collection.update!(permitted_attributes(Collection))
     end
     flash[:notice] = t('.updated')
+
+    # TODO: Needs to be rethinked (and destroy action for both communties and collections)
+    # You can access this from both admin communties and collections index as well as communities show page
+    # When your in communities show page and go to edit the collection this shouldn't be redirecting to
+    # the admin communities and collections index page, it should go back to the communities show page...no?
+    # Perhaps better to only allow this behaviour from admin communities and collections?
```
mbarnett: Yeah, this isn't ideal at all. We need to be cleverer about sending them back to where they were, somehow, I think.

Issue #19, opened 2017-07-06
Derivatives Generation Setup 


mbarnett:
Need to actually set up our derivatives infrastructure

-  Add sidekiq to project
- Spin off derivative job after item creation

PR #114, opened 2017-09-08
Add sidekiq to jupiter

murny:
Sidekiq needed for #19


Dockerfile.deployment
```diff
@@ -36,14 +36,11 @@ RUN mkdir -p $APP_ROOT
 WORKDIR $APP_ROOT
 
 # Preinstall gems in an earlier layer so we don't reinstall every time any file changes.
-COPY Gemfile  ./
-COPY Gemfile.lock ./
-RUN bundle install --without development test 
-
-COPY config/puma.rb config/puma.rb
```
murny: This gets copied over when we copy the codebase over already

Dockerfile.deployment
```diff
@@ -36,15 +36,12 @@ RUN mkdir -p $APP_ROOT
...
+COPY Gemfile  $APP_ROOT
+COPY Gemfile.lock $APP_ROOT
+RUN bundle install --without development test
 
 # *NOW* we copy the codebase in
-
-ADD . /app
+COPY . $APP_ROOT
```
murny: If we specifing a APP_ROOT var lets use it explictly everywhere

docker-compose.deployment.yml
```diff
@@ -12,14 +12,15 @@ volumes:
 
 services:
   mysql:
+    restart: always
```
murny: Add restart always everywhere?

docker-compose.deployment.yml
```diff
@@ -32,66 +33,48 @@ services:
...
       context: .
       dockerfile: Dockerfile.deployment
     image: ualbertalib/jupiter
+    command: bundle exec sidekiq
     env_file: .env_deployment
-    environment:
-      RAILS_SERVE_STATIC_FILES: enabled
-    command: bundle exec puma -e uat
-    volumes:
-      - .:/app
```
murny: We shouldnt be syncing the files in this repo with whats in the container. This is only needed for development when your actively working on files and want to see them changed in the container

docker-compose.deployment.yml
```diff
@@ -32,66 +33,48 @@ services:
...
-    expose:
-      - "3000"
```
murny: expose is already happening in the dockerfile, not needed

docker-compose.deployment.yml
```diff
@@ -32,66 +33,48 @@ services:
...
   redis:
+    restart: always
     image: redis:alpine
     volumes:
       - redis:/data
 
-  web:
-    build: 
+  # Sidekiq
+  worker: &rails
```
weiweishi: I assume &rails is a reference refers back to "web"? why this is needed?

murny: Its to reduce code and duplication. It has nothing to do with docker-compose. It is a YAML variable. You probably seen this in database.yml and other places. 

Since worker and web are basically the same just different commands run and port mappings (since sidekiq needs all the same goodies as rails), so instead of duplicating: 
```
 restart: always
    build:
      context: .
      dockerfile: Dockerfile.deployment
    image: ualbertalib/jupiter
    command: bundle exec sidekiq
    env_file: .env_deployment
    depends_on:
      - mysql
      - fcrepo
      - redis
      - solr```

PR #126, opened 2017-09-11
added dummy values to allow assets:precompile to run

weiweishi:
and updated the docker-compose to pull autobuild images from dockerhub.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,7 +1,7 @@
 cd /home/deploy/jupiter && git fetch --all && git reset --hard origin/master
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml down web
 docker rmi $(docker images -f "dangling=true" -q)
-docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml build web
+docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml pull web
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml up -d
 docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml run --rm web rake db:migrate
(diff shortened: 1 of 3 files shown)


deploy.sh
```diff
@@ -1,8 +1,7 @@
-cd /home/deploy/jupiter && git fetch --all && git reset --hard origin/master
-docker-compose -f /home/deploy/jupiter/docker-compose.deployment.yml down web
+cd /home/deploy/jupiter-master && git fetch --all && git reset --hard origin/master
```
murny: if your pulling the container from dockerhub and the container contains the entire code packaged up do you need this?

I guess the only thing you need is the actual `docker-compose.deployment.yml` file right? 

So I suppose theres a few ways of doing this. Could just curl/grab/git pull this single file from github, or since this deploy.sh script will be running from someones desktop machine, you can rsync the file right from this same directory onto the box. Or perhaps this is a good use case for a seperate repo just for jupiter deployment or something?

¯\_(ツ)_/¯

weiweishi: we still need config/nginx.conf for nginx setup? will need to rsync/grab that file as well.

weiweishi: And we currently only autobuild for master, so for a feature branch, we still need to get the source from git and build?

murny: Yeah anything your docker-compose file will need. So nginx.conf and .env_deployment etc.

murny: You can autobuild branches too if you go down that road. If we got a feature branch called `epic_sprint` then you can setup an autobranch for that and have it autobuild a `jupiter:epic_sprint` version of the container which you have to pull down accordingly in your script.

weiweishi:
I will close this PR and open a new one for additional fixes.

Issue #120, opened 2017-09-08
Collection detail pages missing delete button / breadcrumbs


sfbetz:
No Delete button on collection detail page: 
http://arsenal.library.ualberta.ca/communities/12c09af9-4a21-48a6-aae3-61e2fca89bb2/collections/d84b9b6c-fc64-481d-aead-8b81735b565f

Can replace link to "Parent Community: XXX" with standard breadcrumb above page title: 

community > collection

cwant:
Done in PR #120

Issue #118, opened 2017-09-08
Buttons are wrapping on mobile


sfbetz:
Buttons are tall!
![2017-09-08_15-17-19](https://user-images.githubusercontent.com/12801010/30231655-2a90e74a-94a9-11e7-8ef2-3ab0199e1771.png)

cwant:
Done in PR #132

PR #128, opened 2017-09-11
Issue #120, collection detail breadcrumbs and links.

cwant:
Adding some missing details from the collection detail wireframe, as per Issue #120:

https://jupiter.mybalsamiq.com/projects/jupiter/Admin%20-%20collection%20detail%20page

The before/after:

collection-detail

Still missing from wireframe, but not in scope of Issue #120:

* Add item button
* Collection description


app/views/collections/show.html.erb
```diff
@@ -1,30 +1,50 @@
...
 
-<% if policy(@collection).edit? %>
-  <div class="d-flex justify-content-between mt-3">
-    <h5>
-      <%= link_to t('.parent_community', title: @community.title), community_path(@community) %>
-    </h5>
-    <%= link_to t('edit'),
-                edit_community_collection_path(@community, @collection),
-                class:'btn btn-outline-primary' %>
+  <div>
```
murny: empty div? Any reason for this?

app/views/collections/_breadcrumbs.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="collection-breadcrumb">
```
murny: It's fine to partialize stuff but in this case is it needed? I ask this as this will never be reused anywhere else execept for the collections show page? So why not just put these 8 lines into the show.html.erb (its only around 50 lines right now)?

app/views/collections/show.html.erb
cwant: This makes the buttons from expanding in height. See for example here:

https://github.com/ualbertalib/jupiter/issues/118

Without the div, the contained buttons get tall and ugly -- guess it's a property of the d-flex thingy.

app/views/collections/_breadcrumbs.html.erb
cwant: Mostly personal preference -- I like to have a quick glance at the show.html.erb and see the sections broken down to get an overall roadmap of the page.

I also wonder if these breadcrumbs will be present on things like edit community page, and add work page (the wire frame shows a button to initiate it from this page).

app/views/collections/show.html.erb
murny: Perhaps read a bit about flex box bootstrap utilities? They have examples of not having them expand height to the entire thing? 

https://getbootstrap.com/docs/4.0/utilities/flex/

¯\_(ツ)_/¯

An div with nothing on it is bad? If you cant get any other thing to work put comments around this, so no one comes in the future and removes this. But I'm sure theres a better solution here

app/views/collections/_breadcrumbs.html.erb
murny: But if you keep going at this your going to have a ton of breadcrumbs partials everywhere which will be very hard to refactor/maintain down the road.  

Perhaps a better solution might be to create a generic breadcrumb partial/helper component that can be reused everywhere where we might want breadcrumbs as your right we going to be using breadcrumbs in a few places.

app/views/collections/show.html.erb
cwant: Cheers, style added to prevent the button height from stretching.

app/views/collections/_breadcrumbs.html.erb
cwant: Breadcrumbs moved from own partial.

cwant:
Thanks, will look at merging after #112 is merged (will likely need a re-approve at that time).

cwant:
This one has been updated for #112 renaming, needs re-approval.

Issue #124, opened 2017-09-08
Need to clear search box on Manage Users page


sfbetz:
Need some way to clear search box when it's filled in (x)

cwant:
Merged via PR #129

PR #129, opened 2017-09-12
Issue #124, add button (X) to user search to clear text from input.

cwant:
Button only shows up when text is present in the search input:

clear-search


murny:
This looks good but this is actually covering up a bug in bootstrap. 

When you use the rails helper `search_field_tag` this creates a `input` field with the `type="search"`. This is a special html5 input field that provides a few things over a traditional `type="text` input field. Mainly the browser will automatically store search terms previously entered and show those. And also this exact functionality you just coded up. It will display a "X" button so you can clear the text when you hover or focus on the input box.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search


How to fix the bug? This seems to do it:
```
input[type="search"]::-webkit-search-cancel-button {
  /*Bootstrap 4 override*/
  -webkit-appearance: searchfield-cancel-button;
}
```

Image with this implemented:
image

This input `type="search"` is supported in most browsers but not all. So i guess its up to us if we want to put a bandaid on this bug and use the default functionality or use what you created. If we are using what you created (i say change the input back to `type=text` so we dont get a weird double X or something down the future if bootstrap ever decides to fix this)

cwant:
Thanks, the HTML5 version will have lower support costs down the road -- I'll go down that route.

cwant:
This is now converted to use the CSS fix (tested here).

cwant:
This one is getting stale, please approve.

Issue #127, opened 2017-09-11
Crank seeds up to ~200 mock objects


mbarnett:
DONE WHEN:
- [x] we can seed several hundred dummy objects into Dev/UAT
    - [x] File attachments not required
- [x] Dummy generated titles
- [x] Lorem ipsum some abstracts

cwant:
This was merged in as PR #131 , closing.

PR #131, opened 2017-09-12
Issue #127, many seeds.

cwant:
Adds 5 communities, 2 collections per community, and about 25 items per collection:

Each collection has:

* A Multi-collection item
* A private item
* An item owned by non-admin
* A currently embargoed item
* A formerly embargoed item
* 20 regular, public, admin-owned items

These have obvious names:

seeds1

Setting env variable `DOWNLOAD_COMMUNITY_LOGOS` will grab some logos off the web using Faker gem (if they don't exist already in the `tmp` folder):

seeds2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,8 +40,8 @@ gem 'omniauth-saml'
 gem 'pundit'
 
 # Misc Utilities
-gem 'kaminari'
 gem 'faker'
+gem 'kaminari'
 
 group :development, :test do
   gem 'sdoc', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,6 @@ gem 'sidekiq', '~> 5.0'
 gem 'sinatra' # used by sidekiq/web
 
 # Misc Utilities
-gem 'faker'
 gem 'kaminari'
 
 group :development, :test do
@@ -57,6 +56,9 @@ group :development, :test do
   gem 'pry-rails'
 
   gem 'rubocop', require: false
+
+  # Seeds
+  gem 'faker', require: false
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -41,6 +41,7 @@ gem 'pundit'
 
 # Misc Utilities
 gem 'kaminari'
+gem 'faker'
 
 group :development, :test do
   gem 'sdoc', require: false
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -40,7 +40,7 @@ if Rails.env.development?
     # Attach logos, if possible
     filename = File.expand_path(Rails.root + "tmp/#{thing}.png")
     unless File.exist?(filename)
-      if ENV['DOWNLOAD_COMMUNITY_LOGOS'].present?
+      unless ENV['SKIP_DOWNLOAD_COMMUNITY_LOGOS'].present?
         set = (thing == 'cat') ? 'set4' : 'set1'
         url = Faker::Avatar.image(thing, "100x100", "png", set)
         File.open(filename, 'wb') do |fo|
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,6 +8,7 @@
 
 if Rails.env.development?
   require 'active_fedora/cleaner'
+  require "open-uri"
 
   puts 'Starting seeding of dev database...'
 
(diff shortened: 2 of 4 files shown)


cwant:
It strikes me that the "count always equals 10" issue might be a result of the pagination work.

murny:
Yeah I ran into this issue too. Among other weird things (for example if you assign a bunch of works to a single user, these works will be displayed on every users page under their works...).

Think its a caching issue? If limit is set at 10 and cached this then affects every other query done (`.all`/`.count`)

¯\_(ツ)_/¯

cwant:
I think there is a limit hardcoded somewhere and applied everywhere by default. Check this out:

```ruby
[10] pry(main)> Item.count
=> 10
[11] pry(main)> Item.limit(200).count
=> 172
```

cwant:
Yeah, I think I agree that it's a caching issue (or Solr / ActiveFedora mis-configuration).

For example, if I have a query for all items in specific collection:

```ruby
params = {:q=>"(_query_:\"{!raw f=has_model_ssim}IRItem\") AND _query_:\"{!field f=member_of_paths_dpsim}c622eb61-9912-41ef-845a-9f21de3b0f58/28529e11-0ec0-4911-8157-0c1c4abb2dcf\"", :fq=>"", :facet=>false, :"facet.field"=>[], :sort=>"record_created_at_dtsi desc"}
```

Then:

```
[4] pry(main)> ActiveFedora::SolrService.instance.conn.get('select', params: params)['response']["numFound"]
=> 22
[5] pry(main)> ActiveFedora::SolrService.instance.conn.get('select', params: params)['response']["docs"].count
=> 10
```

So Solr says there are 22 found, but only returns 10 of them.

cwant:
Further result: issuing the same query in the solr web interface gives similar results (num found:22, num returned: 10).

cwant:
Likely this is going on: https://stackoverflow.com/questions/9313111/solr-number-of-results-returned

cwant:
Taking WIP off of this now that the "only fetching 10 objects" issue is resolved.

db/seeds.rb
```diff
@@ -8,45 +8,142 @@
...
-  ).unlock_and_fetch_ldp_object(&:save!)
-
-  Item.new_locked_ldp_object(
-    owner: admin.id,
-    visibility: JupiterCore::VISIBILITY_PUBLIC,
-    title: 'Random images',
-    date_created: Time.zone.now.to_s
-  ).unlock_and_fetch_ldp_object(&:save!)
+  # Seed a non-admin user
+  non_admin = User.create(name: 'Admin', email: 'non_admin@ualberta.ca', admin: false)
```
murny: Change the name perhaps? So people aren't confused if they login into this user and see `Admin` as the top of the navbar for their user menu

cwant: Cheers, thanks for catching that (fixed).

Gemfile
```diff
@@ -44,6 +44,7 @@ gem 'sidekiq', '~> 5.0'
 gem 'sinatra' # used by sidekiq/web
 
 # Misc Utilities
+gem 'faker'
```
murny: If your having this out here (which maybe you could put in development/test group since these seeds run there instead), I'd make this `require: false` and then require it inside the seeds.rb. So this doesn't hurt bootup time for rails and everything else just impacts seeds.rb rake task

db/seeds.rb
```diff
@@ -8,45 +8,142 @@
...
+    community = Community.new_locked_ldp_object(
+      owner: admin.id,
+      title: title,
+      description: Faker::Lorem.sentence(20, false, 0).chop
+    ).unlock_and_fetch_ldp_object(&:save!)
+
+    # Attach logos, if possible
+    filename = File.expand_path(Rails.root + "tmp/#{thing}.png")
+    unless File.exist?(filename)
+      if ENV['DOWNLOAD_COMMUNITY_LOGOS'].present?
```
murny: Any reason to make this the default? Just always download the logos and remove this ENV VAR? (otherwise pop a section in the readme about running seeds and using this ENV VAR but I say just make this the default, unless it's killer for performance or some other reason)

Gemfile
cwant: Thanks for catching that. I'll move it to dev/test section.

db/seeds.rb
cwant: Yeah, in this case, download of logos means that seeding has to have a network connection.

The faker gem actually does very little here, only generates the URL to the logo (technically an avatar, but the web service to grab random images was unacceptably slow) . Once downloaded, the seeding process will use the downloaded file instead in the `tmp` directory of the project, and no longer needs a network connection on future seedings.

I can flip it to be non-default though (SKIP_DOWNLOAD_COMMUNITY_LOGOS). And will add a remark to the README.

murny: Sure that sounds better. Let's flip this and add a remark in the README. :+1:

cwant:
Gemfile reworking, and making logo downloads default (with README note) pushed.

cwant:
Thanks Shane, but I added just one more thing ... since the collection description was added this morning, I pushed a commit to seed that as well. Please re-approve!

PR #132, opened 2017-09-12
Issue #118, buttons get stretched in flexboxes.

cwant:
Also, moved a partial based on one of the comments.

### DIFF
--- a/app/views/communities/show.html.erb
+++ b/app/views/communities/show.html.erb
@@ -33,9 +33,9 @@
 </div>
 
 <div class="d-flex justify-content-between mt-3">
-    <h4 class="collections-header">
-      <%= t('.collections_list_header') %>
-    </h4>
+  <h4 class="collections-header">
+    <%= t('.collections_list_header') %>
+  </h4>
   <% if policy(:collection).create? %>
       <%= link_to t('.create_new_collection'), new_community_collection_path(@community),
           class: 'btn btn-primary align-self-center' %>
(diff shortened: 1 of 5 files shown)


cwant:
This is the typical effect of these "align-self-center" fixes for flexboxes:

stretch1

This fixes some of the badges too, but the "cards" used for the stats on the admin dashboard still looks a little screwy (this could be addressed as part of this PR, or can spin it off to a separate issue):

stretch2

app/views/communities/show.html.erb
```diff
@@ -38,17 +38,15 @@
...
       <%= link_to t('.create_new_collection'), new_community_collection_path(@community),
-          class: 'btn btn-primary' %>
+          class: 'btn btn-primary align-self-center' %>
   <% end %>
 </div>
 
 <div class="row mt-3">
   <div class="col-md-12">
-      <%# should be reversed, admin should be using communities partial not vice versa %>
-      <%# TODO: FIX ALL THIS... %>
```
murny: Should leave this TODO in. Otherwise this PR looks good

As we got lots of if branching and weird markup here. (admins get rendered a collections list which are ul/li list groups and non admins get a div/a list groups etc. The two should be consolidated into one or something smarter here)

cwant: Sounds good, comment restored (also fixed some inconsistent indentation, and removed an extra `</ul>` that got in there somehow.

cwant:
Cheers!

Issue #119, opened 2017-09-08
Remove Community and Collection labels from titles


sfbetz:
Community and Collection detail pages include the collection and community label in the titles: eg here: http://arsenal.library.ualberta.ca/communities/bcd8194d-2c49-4274-905b-634a59e6125c/collections/e8a6cf2d-3b75-46a2-9c24-13b1d0cec8c3

**"Collection:** Presentations (Agricultural, Life & Environmental Sciences)" - This can come out so the title reads "Presentations (Agricultrual, Life & Environmental Sciences)" 

Same for communities.

cwant:
@sfbetz Should I infer that you also want the "Item:" labels removed from the page headers on the item detail pages?

sfbetz:
Yes please! Didn't see those.

cwant:
Merged via PR #134

PR #134, opened 2017-09-13
Issue #119, remove redundant word from headers

cwant:
The labels "Community:", "Collection:" and "Item:" are removed from page headers (but not page titles).
Also did a bit of spacing/indenting tweaks on Collection/Community pages (e.g., community logo). Here are before/after screenshots showing tweaks:

community-tweak

collection-tweak

### DIFF
--- a/app/views/collections/show.html.erb
+++ b/app/views/collections/show.html.erb
@@ -10,9 +10,7 @@
 </div>
 
 <div class="d-flex justify-content-between mt-3">
-  <h1>
-    <%= t('.header', title: @collection.title) %>
-  </h1>
+  <h1><%= @collection.title %></h1>
 
   <%# "align-self-center" stops buttons from growing to size of H1 in this flexbox %>
   <div class="btn-group align-self-center">
@@ -30,7 +28,7 @@
   </div>
 </div>
 
-<h5 class="items-header">
+<h5 class="items-header p-3">
   <%= t('.items_list_header') %>
 </h5>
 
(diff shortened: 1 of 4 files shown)


murny:
Any reason not to change the page title (`<title></title>`) as well? They should be the same

cwant:
I thought it was better keeping it in the title. Even if I changed it, the header and the title wouldn't be the same, since we append " | ERA" to our titles.

@sfbetz, @leahvanderjagt any opinion on this? Should the browser application title bars say:

"Community: The department of Unicorn | ERA"
"Collection: Theses about unicorns | ERA"
"Item: The effects of Nugget Nectar on unicorns | ERA"

or

"The department of Unicorn | ERA"
"Theses about unicorns | ERA"
"The effects of Nugget Nectar on unicorns | ERA"

(This is not the header on the page, but top of the browser window.)

murny:
I am using my opinion from what others do in the wild. If you pick a website in the wild like...github/facebook/youtube/slideshare/linkedin/imdb/etc etc, 

Theres a really good chance you will always see the page title/header being the same in the fashon of `<resource_name> | <website_name>` and not `Resource: <resource_name> | <website_name>`.

sfbetz:
Given limited real estate in places that display page titles (eg tabs), my preference is the latter - otherwise a user with many tabs (guilty) may see only the not-so-useful label Collection: or Item: in their browser, rather than more useful descriptive titles. Let's go with:

"The department of Unicorn | ERA"
"Theses about unicorns | ERA"
"The effects of Nugget Nectar on unicorns | ERA"

cwant:
Shortened page titles.

PR #139, opened 2017-09-14
Fix pagination of defered queries losing some params

### LINKED ISSUE #138: Weird behaviour for User items: Items showing up not belonging to user
**Bug**: `user.items` displays all items even if they dont belong to themselves. This doesn't make sense as `user.items` is defined as the follow:

```ruby
  def items
    Item.where(owner: id)
  end
```

## How to reproduce

So on a fresh instance, with newly ran seeds. Go create a new user, then sign in as admin, you should see the following on the admin dashboard:

image

When you go look at the admin user, you see what you expect, since all these items belong to the admin user:

image

Everything looks good.

But now look at the new non-admin user you just created and you will see something funny:

image

It also shows the 4 items in the system under this user! Which don't belong to this user and shouldn't be showing. However the `user.items.count` is correct which shows 0 for the "Displaying 4 of 0 total items."

Another weird caching/solr issue? Might be due to pagination considering `user.items.count` works fine. 

Related to #137?

mbarnett:
Closes #138

### DIFF
--- a/app/models/jupiter_core/deferred_solr_query.rb
+++ b/app/models/jupiter_core/deferred_solr_query.rb
@@ -60,6 +60,11 @@ class JupiterCore::DeferredSolrQuery
     criteria[:limit]
   end
 
+  # Kaminari integration
+  define_method Kaminari.config.page_method_name, (proc { |num|
+    limit(default_per_page).offset(default_per_page * ([num.to_i, 1].max - 1))
+  })
+
   def total_count
     af_model = criteria[:model].send(:derived_af_class)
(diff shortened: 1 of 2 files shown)


app/models/jupiter_core/deferred_solr_query.rb
```diff
@@ -60,6 +60,11 @@ def limit_value
     criteria[:limit]
   end
 
+  # Kaminari integration
+  define_method Kaminari.config.page_method_name, (proc { |num|
```
murny: Should there be an accomodating unit test for this? Or don't worry about it?

PR #140, opened 2017-09-14
Lots o' results. Closes #137

### LINKED ISSUE #137: `each` only works on 10 objects for descendents of LockedLdpObject
Also, `count` always returns 10 (`total_count` returns the real value though).
This is due to what Solr does by default, see discussion in PR #131.

mbarnett:


app/models/jupiter_core/search.rb
```diff
@@ -1,5 +1,9 @@
 class JupiterCore::Search
 
+  # How dumb is this? Seems pretty dumb, but that's the official recommendation I guess:
+  # https://wiki.apache.org/solr/CommonQueryParameters
+  MAX_RESULTS = 10_000_000
```
mbarnett: !(ten million results)[https://i.giphy.com/media/cCpiKKKTBXFKw/giphy.gif]

PR #141, opened 2017-09-14
Fix collections and communities forms

murny:
Previous communities forms would simply just 500 with an error. And collections would just create an empty collection object without a title. This fixes this.

More PRs coming regarding, reversing the admin communities and collections and how regular community/collection works with regarding to create/update/delete. Also fixing of Items form from crashing like these.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -1,6 +1,6 @@
 class CollectionsController < ApplicationController
 
-  before_action -> { authorize :application, :admin? }, except: [:index, :show]
+  before_action -> { authorize :application, :admin? }, except: [:show]
 
   def show
     @collection = Collection.find(params[:id])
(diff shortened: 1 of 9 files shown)


murny:
Added unit tests, ready to be reviewed now.

Errors on forms now look like:

image


Instead of a 500 error:
image

app/controllers/collections_controller.rb
```diff
@@ -1,6 +1,6 @@
 class CollectionsController < ApplicationController
 
-  before_action -> { authorize :application, :admin? }, except: [:index, :show]
+  before_action -> { authorize :application, :admin? }, except: [:show]
```
cwant: Removing `:index` because it doesn't have an action defined? Might want to mirror this in the routes file too:

```ruby
  resources :communities do
    resources :collections, except: :index
  end
```

murny: Already done 🙂

cwant: Oops, I guess I should read the whole PR before making comments ...

PR #143, opened 2017-09-14
fix precompile with dummy keys and add shared volumes within containers

weiweishi:
-  use dummy keys for assets precompile in Dockerfile
-  allow shared volume in web, nginx and worker containers 
-  remove building images and will use dockerhub autobuilt images instead.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 tmp/*
+!tmp/pids/*
 log/*
 .git/*
 *.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 tmp/*
-!tmp/pids/*
+!tmp/pids/.keep
 log/*
 .git/*
 *.md
--- a/docker-compose.deployment.yml
+++ b/docker-compose.deployment.yml
@@ -9,7 +9,6 @@ volumes:
     driver: local
   redis:
     driver: local
-  approot:
 
 services:
   mysql:
@@ -53,8 +52,6 @@ services:
     image: ualbertalib/jupiter:deployment
     command: bundle exec sidekiq
     env_file: .env_deployment
-    volumes:
-      - approot:/app
     depends_on:
       - mysql
       - fcrepo
@@ -76,6 +73,5 @@ services:
     env_file: .env_deployment
     volumes:
       - ./config/nginx.conf:/etc/nginx/conf.d/default.conf
-      - approot:/app
     ports:
       - "80:80"
--- a/docker-compose.deployment.yml
+++ b/docker-compose.deployment.yml
@@ -9,6 +9,7 @@ volumes:
     driver: local
   redis:
     driver: local
+  approot:
 
 services:
   mysql:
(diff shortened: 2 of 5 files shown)


docker-compose.deployment.yml
```diff
@@ -49,12 +50,11 @@ services:
...
     restart: always
-    build:
-      context: .
-      dockerfile: Dockerfile.deployment
-    image: ualbertalib/jupiter
+    image: ualbertalib/jupiter:deployment
     command: bundle exec sidekiq
     env_file: .env_deployment
+    volumes:
+      - approot:/app
```
weiweishi: This is needed so that the volume can be shared by three different containers. This caused the deployment issue after mounting directory from host was removed.

weiweishi:
The dummy key in assets:precompile has caused the autobuild of the deployment version of the docker image ualbertalib/jupiter:deployment to fail. I will fix the deploy.sh once the autobuild succeeds, and I'll test the process to verify.

docker-compose.deployment.yml
murny: Can you give more information here? 

Worker and Web containers shouldn't have to share the same app directory between the two? They each have their own separate container based off the same image (which contains all the code and everything else they need)

weiweishi: The issue was related to:
```worker_1  | No such file or directory @ rb_sysopen - /app/tmp/pids/sidekiq.pid
worker_1  | /usr/local/bundle/gems/sidekiq-5.0.4/lib/sidekiq/cli.rb:379:in `initialize'
worker_1  | /usr/local/bundle/gems/sidekiq-5.0.4/lib/sidekiq/cli.rb:379:in `open'
worker_1  | /usr/local/bundle/gems/sidekiq-5.0.4/lib/sidekiq/cli.rb:379:in `write_pid'
worker_1  | /usr/local/bundle/gems/sidekiq-5.0.4/lib/sidekiq/cli.rb:43:in `parse'
worker_1  | /usr/local/bundle/gems/sidekiq-5.0.4/bin/sidekiq:11:in `<top (required)>'
worker_1  | /usr/local/bundle/bin/sidekiq:17:in `load'```. worker container was trying to access and create file in /app/tmp/pids/sidekiq.pid and wasn't able to do so. 

Similarly, with Nginx, it was trying to write logs in app/logs, and couldn't access it.  

If the shared directory is not needed, maybe something else need to be reconfigured so that it won't be looking for the web container's app directory?

Comment (murny) [reply docker-compose.deployment.yml]: For the sidekiq one, just create a new file (empty file called `.keep` or `.gitkeep` inside a new folder under `/tmp/pids/`

For nginx, you shouldn't be putting the nginx logs inside the rails container. It should be on the server itself or left in the nginx container. Here is some info:

`The official nginx image creates a symbolic link from /dev/stdout to /var/log/nginx/access.log, and creates another symbolic link from /dev/stderr to /var/log/nginx/error.log, overwriting the log files and causing logs to be sent to the relevant special device instead. See the Dockerfile.` So looks like this is already being done. Maybe have a peak at the official nginx docker and how to manage/setup logs for it?

Comment (weiweishi) [reply docker-compose.deployment.yml]: It's configured here:
https://github.com/ualbertalib/jupiter/blob/master/config/nginx.conf#L26-L27. I can change it to /var/log instead. 

And I was still experiencing issue with the worker container - I will pull the changes out, and tested with the latest success build with ualbertalib/jupiter:deployment then.

File: .dockerignore [hunk 2]
```diff
@@ -1,4 +1,5 @@
 tmp/*
+!tmp/pids/*
```
murny: Mmm not that I am thinking about this maybe this should be `!tmp/pids/.keep` instead of `*` (wildcard). In case somehow you have pid files in here already and this aint doing weird stuff. 

I can fix

PR #144, opened 2017-09-14
Populate the 'visibility' field on Item edit form.

cwant:
App was not filling in visibility correctly when editing an item, now it does. Although the current form doesn't seem to have an 'embargo' visibility -- oversight or feature?

### DIFF
--- a/app/views/items/_form.html.erb
+++ b/app/views/items/_form.html.erb
@@ -5,7 +5,7 @@
   <% end %>
   <%= f.input :visibility,
     as: :unread_select,
-    collection: ['public', 'private', 'authenticated'],
+    collection: Item::VISIBILITIES,
     selected: item.visibility,
     input_html: { name: 'item[visibility]' }
   %>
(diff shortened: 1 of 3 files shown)


app/views/items/_form.html.erb
```diff
@@ -4,9 +4,10 @@
     <%= f.input prop %>
   <% end %>
   <%= f.input :visibility,
-  as: :unread_select,
-  collection: ['public', 'private', 'authenticated'],
-  input_html: {name: 'item[visibility]'}
+    as: :unread_select,
+    collection: ['public', 'private', 'authenticated'],
```
murny: Could be a separate PR, but we should be using the `JupiterCore::VISIBILITY_PUBLIC` etc constants (or if this array list is going to be popular create a array for this in JupiterCore::VISIBILITIES or something) for this instead of hardcoding strings

Otherwise looks good

cwant: I've done this as a bit of a hierarchical thing (`Item` has one extra visibility that `JupiterCore` doesn't).

PR #146, opened 2017-09-15
Improve travis

murny:
Only build docker web container and its dependency containers (no worker container needed)

Also only build test database with no seeds (previous we were doing dev/test with seeds)

Remove cleanup, since don't think this is required (not in official docs and no one in the wild is really doing it)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,18 @@ before_script:
   - docker version
   - docker-compose version
 
-  - docker-compose build
-  - docker-compose up -d
+  - docker-compose build web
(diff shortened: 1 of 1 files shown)


cwant:
Nice -- comparing with a Travis run on a different PR, the DB setup went from 90s to 20s.
If you merge master into the branch, I will approve (branch out-of-date).

murny:
Builds on hold for some reason... (probably unrelated just travis is busy) ¯\_(ツ)_/¯. Will merge once building again

Issue #88, opened 2017-08-25
File downloads


mbarnett:
Files are serialized into Fedora by Hydraworks. This _should_ be fairly straightforward, but we may need to reference Hyrax's implementation.

(I've moved multi-file zipping to its own ticket)

DONE WHEN:
- [x] Files can be downloaded via Jupiter
- [ ] Download permissions are being applied correctly via policy (owner or user is authenticated if visibility is authenticated or anyone if visibility is public)
- [ ] Correct masking of buttons when user does not have correct permissions
- [x] UI "files section sidebar" partial is implemented.

weiweishi:
just to confirm that this issue is not yet done in PR #279, for all the permission handling etc? @mbarnett

mbarnett:
The exact permission logic has not been wired up, but it is stubbed out at the moment.

mbarnett:
More specifically, we just need to decide exactly what the logic in here should be: https://github.com/ualbertalib/jupiter/blob/master/app/policies/item_policy.rb#L28

leahvanderjagt:
Hi Matt, do you need stakeholder input? I'm happy to provide if necessary

mbarnett:
I have a feeling this is already defined somewhere else, but if you want to dump the info in here we can get this finished off.

I think the logic is something like: You can download all of the files in an item if any of the following are true:

1) You're an admin

2) You own the item

3) The item is open access and you're anyone, logged in or not logged in

4) The item is ccid protected and you're logged in with a ccid

Are those all of the scenarios? Are there any others?

Issue #90, opened 2017-08-25
Multi-file zipping for multi-file downloads


mbarnett:
Decision captured from the mockup meeting and comment below:

For phase 1, Leah is okay with just having the "Download all" button to simply just initiate the download for all files (much like google drive etc aka as if the user actually clicked each and every download button themselves).  aka: queue up multiple parallel downloads. 

Wishlisty down the road:

Zipping all files with the license file would be a nice to have. But since we do not store this anywhere we would have to generate this on the fly. This could take a long time to process this request. (we could email them to the user when the zip has been finished but what about users who are not logged into the system etc?)

If zip is very easy to accomplish (without the license) then probably do the zip approach (zip file is probably a better UX anyways)? Looks like dropbox creates a zip file for multiple files. Needs some investigation of the two approaches.

murny:
From mockup meeting, Leah is okay with just having the "Download all" button to simply just initiate the download for all files (much like google drive or dropbox etc).

Zipping all files with the license file would be a nice to have. But since we do not store this anywhere we would have to generate this on the fly. This could take a long time to process this request. (we could email them to the user when the zip has been finished but what about users who are not logged into the system etc?)

mbarnett:
Just so I'm clear, meaning we would queue up like 10 parallel downloads in the browser or w/e?

murny:
Yes exactly

leahvanderjagt:
Yes

___________________________________________________________



*Leah VanderjagtDigital Repository Services Coordinator, University of
Alberta t. 780.492.3851 / **leahv@ualberta.ca <leahv@ualberta.ca> *

*The University of Alberta is situated on traditional Treaty 6
territory and homeland of the Métis peoples.*
Amiskwaciwâskahikan / ᐊᒥᐢᑲᐧᒋᕀᐋᐧᐢᑲᐦᐃᑲᐣ / Edmonton

mbarnett:
Well that's somewhat easier. I don't know if any discussions happened wrt external (to Fedora) file storage happened while I was gone, but the parallel downloads would tend to argue in favor of that, as the performance of starting N different connections with the Rails server to stream the files piecewise from Fedora could be an issue, although we might be able to live with it for the time being.

weiweishi:
Need to confirm if the work is already been done. @murny

PR #148, opened 2017-09-18
[WIP] Issue #88, file download. Almost minimum viable product.

cwant:
This is what we have so far:

files

To do:

* Sanity check of this thing
* Tests


app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -121,6 +121,14 @@ def errors
...
     end
 
+    def read_solr_index(name)
+      raise PropertyInvalidError unless name.is_a? Symbol
+      type = self.solr_calc_attributes[name]
+      raise PropertyInvalidError unless type.present?
+      solr_name = Solrizer.solr_name(name, :symbol, type: type)
+      solr_representation[solr_name]
+    end
+
```
cwant: I still think dynamically writing a getter would be better than the above. To distinguish from a full-fledged attribute, could add `cached_` to the beginning of the method name. E.g., call `cached_original_file_name` instead of `read_solr_index(original_file_name)`.

This would at least mean that Solrizer.solr_name() would only need to be called at method definition time, not every time we want to fetch data.

app/models/item.rb
```diff
@@ -44,6 +48,23 @@ def each_community_collection
...
     end
+
+    def add_communities_and_collections(communities, collections)
+      return unless communities.present? && collections.present?
+      communities.each_with_index do |community, idx|
+        add_to_path(community, collections[idx])
+      end
+    end
+
+    def add_files(params)
```
murny: Good job moving this stuff heavily logic stuff out of the controller into the model :+1: 

One quick suggestion: 
Instead of passing in the entire params object, just pass in the file as it's the only thing your using out of this? 

This also keeps this less tied to a controller (and the request/response lifecycle) and more generic as well

`def add_files(params)` becomes `def add_files(file)` and everythign `params[:item][:file]` becomes `file`

app/controllers/file_sets_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    stream_from_fedora
+  end
+
+  def download_original
+    stream_from_fedora(disposition: 'attachment', content_type: 'application/octet-stream')
+  end
+
+  private
+
+  def stream_from_fedora(disposition: nil, content_type: nil)
```
mbarnett: We might need to consider saner ways to achieve this that give us some kind of useful handle when it comes to downloading multiple items. Streaming has terrible performance characteristics.

app/controllers/file_sets_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    @file_name = params[:file_name]
+    return render status: :not_found unless @file_name == @file_set.original_file_name
+
+    uri = URI(@file_set.original_uri)
+    request = Net::HTTP::Get.new(uri)
+    response = Net::HTTP.start(uri.hostname, uri.port) do |http|
+      http.request(request)
+    end
+    send_data response.body, filename: @file_name, disposition: disposition, content_type: content_type
+  end
```
mbarnett: Whiteboarded some possible approaches to this with Shane this morning -- since we've been talking about sidestepping Hydra's derivatives process in favor of Paperclip anyways, one approach will probably be to continue to ingest files into Fedora, but also create a shadow item ActiveRecord, store files on the filesystem for download, use a background worker for the zip portion of #90, but allow them to directly download the filesystem version of the file here, to avoid the memory issues we've experienced with Fedora streaming.

app/models/item.rb
```diff
@@ -30,6 +30,10 @@ def self.valid_visibilities
     super + [VISIBILITY_EMBARGO]
   end
 
+  def file_sets
+    FileSet.where(member_of_collections: id)
```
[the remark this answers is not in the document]
cwant: Please see discussion on September 7th in slack about `member_of_collections` in this context. (No, not trying to add a file to a 'collection' object.)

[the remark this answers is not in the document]
mbarnett: I remember the discussion, and reading it now, it's a discussion of the suitability of the approach? At no point is the question "is it sane, clear, and good for the codebase for FileSets to refer to the Work they're in as a Collection". Even if we had said "yeah, sure" then, reading it in the context of this PR, it confused the hell out of. 

Do you really think the naming here is clear?

app/controllers/file_sets_controller.rb
mbarnett: Neglected to mention that we still need to talk this approach over w/ Weiwei to determine our short-term and long terms goals for file storage, so this is by far the least critical part of the PR for now. I know Avalon is going to be taking a good deal of your time next week, but if you do get back to this it's probably best to focus on other issues while we sort this out

app/models/item.rb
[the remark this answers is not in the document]
mbarnett: Brain dumping a couple of possible solutions that might let us avoid this kind of confusing name mess:

1) We sitck something like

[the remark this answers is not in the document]
mbarnett: Brain dumping a couple of possible solutions before I completely forget them this week:

1) We sitck something like:

```ruby
# Hydra::PCDM horribly misnames memberOf property, so we introduce in less confusing names here
alias_method : member_of_collections, :member_of_items
alias_method : member_of_collections=, :member_of_items=
```

in FileSet's unlocked block. This avoids duplication, but I don't think it solves the "I want to query back by the work id elsewhere" problem, and I still think that needs more thought. Leaking this `solr_index` stuff all over the place feels messy, so maybe we need a better API around that, and I'm hooping to put some thought into that if I have some downtime at the conference.

2) Shrug and create a better named property. The Works Actor will continue to deal with the works internals, and our code won't be completely disorienting to read. The only cost is the duplication of the data and any potential for them to get out of sync.

app/controllers/file_sets_controller.rb
weiweishi: I've booked a meeting with Peter/Kenton and Leah to talk about the intermediate file storage idea, and we will at least have a short-term decision in the next week or so, if we need more time for a long-term solution, which ties into the investigation we do for Fedora/providing access to digitization materials. @mbarnett @cwant

mbarnett:
Closing this in order to reopen shortly with a new version of the same work by Chris, with master conflicts resolved and alterations to account for #183

PR #150, opened 2017-09-18
Fix admin actions for collections and communities

murny:
Idea behind this PR is remove all the duplication that is happening and the weird UI/UX as a result of this (being bounced around between admin community and collection and regular community/collection views). 

So this improves this by moving all the actions(create/update/destroy) for community/collections a user can do to the admin backend since admins are the only ones that can do this stuff anyways.


app/controllers/admin/communities_and_collections_controller.rb
```diff
@@ -1,18 +0,0 @@
-class Admin::CommunitiesAndCollectionsController < Admin::AdminController
```
murny: Split these up into two separate controllers one for each model

app/controllers/admin/communities_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def index
+    @communities = Community.all
+  end
+
+  def show
+    @community = Community.find(params[:id])
+    respond_to do |format|
+      format.js
+      format.html { render template: 'communities/show' }
```
murny: reusing the communities/show template here (I'm injecting a special admin breadcrumbs partial though for this template)

app/controllers/collections_controller.rb
```diff
@@ -1,74 +1,9 @@
 class CollectionsController < ApplicationController
 
-  before_action -> { authorize :application, :admin? }, except: [:show]
```
murny: Remove all this stuff now. Since admins are the only ones that can do this stuff lets move these actions to the admin collection controller instead.

app/views/admin/communities/_collection.html.erb
```diff
@@ -0,0 +1,17 @@
+<li class="list-group-item list-group-item-action">
```
murny: the show breadcrumbs and collection breadcrumbs are used here to override the respective partials in the non admin communities/collections partial. 

Easy way to reduce the amount of duplication between the two areas.

app/views/communities/show.html.erb
```diff
@@ -37,33 +48,15 @@
...
   <% if policy(:collection).create? %>
-      <%= link_to t('.create_new_collection'), new_community_collection_path(@community),
+      <%= link_to t('.create_new_collection'), new_admin_community_collection_path(@community),
           class: 'btn btn-primary align-self-center' %>
   <% end %>
 </div>
 
 <div class="row mt-3">
   <div class="col-md-12">
-    <%# TODO: FIX ALL THIS... %>
```
murny: Much improved now...the separate logic is now handled with the collection partial. Depending if your in admin or normal you get served a different partial.

test/integration/collection_show_test.rb
```diff
@@ -31,19 +31,17 @@ def before_all
     get community_collection_url(@community, @collection)
 
     # Breadcrumb
-    breadcrumbs = css_select 'div.collection-breadcrumb'
```
murny: Let's not add extra html/css to the views that their only purpose is for being used in tests. This can be written smarter as shown below

murny:
- Instead of a concern which i think would work poorly here since you are only reusing it once and overriding it the second time anyawys. I did a `before_action :fetch_<object>` on the admin communities/collections which reduced alot of duplication

- For the breadcrumbs,I meant the partial to convey the message that it's only being used in the `show.html.erb` template. But I can see how this is confusing....Hmm not sure how to name this better

cwant:
Yeah, as usual I figured out your intention with the `show` about 10 seconds after pressing the comment button. I'm still not convinced that the `show` is necessary, but on the other hand it's not a big deal.

I am also getting random intermittent failures with travis today. Doesn't look like the extra delay fixed it.

PR #151, opened 2017-09-19
allow seeding into uat env

weiweishi:
before we have a script to ingest more realistic data.


murny:
Got a rubocop fail...run `rubocop -a` and push that

weiweishi:
```
rubocop -a
Inspecting 111 files
...............................................................................................................

111 files inspected, no offenses detected
```
nothing to be pushed?

PR #152, opened 2017-09-19
Issue #122, relabel user fields.

### LINKED ISSUE #122: Manage users - labels
Change labels on manage user pages:

Admin? - Role: (admin OR user)
Suspended? -  Status: (active OR suspended) 
Created at - Created: (display date)
Updated at - Updated: (display date)

Id - we don't know what this is, so probably unnecessary to display

cwant:
Fixes #122


app/views/admin/users/index.html.erb
```diff
@@ -30,26 +30,24 @@
...
   <thead>
     <tr>
-      <th><%= sortable 'id' %></th>
-      <th><%= sortable 'email' %></th>
-      <th><%= sortable 'name' %></th>
-      <th><%= sortable 'created_at' %></th>
-      <th><%= sortable 'updated_at' %></th>
-      <th><%= sortable 'admin' %></th>
-      <th><%= sortable 'suspended' %></th>
+      <th><%= sortable t('admin.users.show.email') %></th>
```
murny: This will break the sorting behaviour. Sortable has a special title parameter however. So this can be rewritten as:
```erb
<th><%= sortable, 'email', title: t('admin.users.show.email') %></th>
```

config/locales/en.yml
```diff
@@ -185,6 +185,15 @@ en:
         search_placeholder: 'Search by email or name...'
         search_button: 'Search Users'
         header: 'Manage Users'
+        name: 'Name'
+        role: 'Role'
+        admin_role: 'Admin'
```
murny: Make sure to remove the old translations? Look at the diff above in the show partial

app/views/admin/users/index.html.erb
cwant: Cheers, looks like it is second argument for title (not title parameter).

config/locales/en.yml
cwant: Thanks, done.

app/views/admin/users/index.html.erb
```diff
@@ -30,26 +30,24 @@
...
       <tr>
-        <th scope="row"><%= user.id %></th>
-        <td><%= link_to user.email, admin_user_path(user) %></td>
+        <th scope="row"><%= link_to user.email, admin_user_path(user) %></th>
         <td><%= user.name %></td>
-        <td><%= time_ago_in_words(user.created_at) %></td>
-        <td><%= time_ago_in_words(user.updated_at) %></td>
-        <td><%= user.admin %></td>
-        <td><%= user.suspended %></td>
+        <td><%= time_tag user.created_at, format: '%F' %></td>
```
murny: Just curious did stakeholders ask for absolute dates and this format, etc? Or just personal preference?

No preference either way. 

One thing I will say is more of a future thing but I would like to create our own time_tag helper at some point which includes a `title` attribute or a popover on the time tag. Pretty much every big website does this. So when you hover over it you get the time in the full format. Look at any time stamp on github for an example.

cwant: The wireframes showed dates in this format -- actually, their dates had two digit years, but I think this is more standard (and Y2K compliant!).

I am a big fan of the full format pop over. If it's worthwhile, I could push a version that uses a helper, something like:

```ruby
  def jupiter_time(datetime, options = {})
    time_options = options.dup
    time_options[:format] ||= '%F'
    time_options[:title] ||= datetime
    time_tag datetime, time_options
  end
```

Yes?

murny: I wouldn't worry about it now. Considering you can just add `title: datetime` in the options and pass that to the time_tag helper.

There's some nice gems that do this client side and what not. For example:
https://github.com/basecamp/local_time

cwant: I'll merge as is. The ticket is a good idea -- would be better to establish standards for this sooner rather than later.

PR #156, opened 2017-09-20
serving assets from nginx instead of rails

weiweishi:


docker-compose.deployment.yml
```diff
@@ -73,5 +74,6 @@ services:
     env_file: .env_deployment
     volumes:
       - ./config/nginx.conf:/etc/nginx/conf.d/default.conf
+      - assets:/app/public/assets/
```
murny: You could say

```
volumes_from:
      - web
```

I think here instead instead of duplicating this.

Other then that this looks good. Approving

docker-compose.deployment.yml
```diff
@@ -61,8 +62,8 @@ services:
   # Rails
   web:
     <<: *rails
-    environment:
-      RAILS_SERVE_STATIC_FILES: enabled
+    volumes:
+      - assets:/app/public/assets/
```
murny: You might have to add the file_storage/uploads directory here to be put into a volume?

I can see uploaded images are doing something weird here on UAT (go to communities index page)

docker-compose.deployment.yml
murny: Ah, so it is. Nevermind then! This is fine

docker-compose.deployment.yml
weiweishi: The directory is persistently set up at /app/storage by https://github.com/ualbertalib/jupiter/blob/master/config/storage_services.yml, and it doesn't have any effect to set up a volume in docker-compose. db:seed works when running on the container, but not working properly when run `docker-compose -f docker-compose.deployment.yml run --rm web rake db:seed`

PR #157, opened 2017-09-20
Fix seeding for UAT

mbarnett:


db/seeds.rb
```diff
@@ -138,7 +138,7 @@
       title: "Multi-collection random images of #{thing.pluralize}",
       description: Faker::Lorem.sentence(20, false, 0).chop,
       # will not be required in future, probably will have a created_at field
-      date_created: Time.zone.now.to_s
+      date_created: Time.now
```
murny: should keep the `Time.zone.now` but I'm being difficult :trollface:

mbarnett: Time is always seconds since epoch, and that doesn't change with timezone. `zone` is meaningful when you're going to translate the Time into a string for presentation, but that's not what's happening here.

Internally all Date, DateTime, and Times get changed to UTC prior to saving, so `zone` is redundant.

PR #158, opened 2017-09-20
 Refactor flexboxes and communities/collections design/layout

murny:
Lots of changes here:

- Generalized logos into a reusuable partial and restyled heavily (since items will eventually need them not just communities)
- Use `media object` and `media lists` for places where we have images instead of trying to mark this up ourselves: https://getbootstrap.com/docs/4.0/layout/media-object/
- Changes to flexbox:
  - consistently use flexbox everywhere instead of pull-right/rows/etc
  - instead of aligning a single item, align all the items 
  - align items on the start/top instead of in the middle/center
  - use flex-wrap on things that don't have a containing div, otherwise items will get pushed off the page on mobile as default is no-wrap, etc etc.
- Add descriptions and logos to community indexes and fix dislaying of lists

Before:
image
After:
image

Before:
image

After:
image

Before
image
After:
image

Etc.


app/views/communities/index.html.erb
```diff
@@ -1,37 +1,28 @@
...
-  </div>
+  <% if policy(Community).create? %>
+    <%= link_to t('.create_new_community'), new_admin_community_path,
+      class: 'btn btn-primary align-self-center' %>
+  <% end %>
 </div>
 
 <div class="row mt-3">
-  <div class="col">
-    <% if policy(Community).create? %>
```
murny: Lets put the "New"/"Create" button at the top of the page and inline with the header. Since this list is going to be huge and you dont want users scrolling to the bottom every time they want to create a new communtiy

app/views/communities/index.html.erb
```diff
@@ -1,37 +1,28 @@
...
+<div class="d-flex flex-wrap justify-content-between mt-3">
+  <h1><%= t('.header') %></h1>
 
-<div class="row mt-3">
-  <div class="col">
-    <div class="list-group">
-      <% @communities.each do |community| %>
-        <div class="d-flex justify-content-between">
-
-          <%= link_to community_path(community) do %>
```
murny: We had a double link to the communtiy page...lets just use a media object and link the entire thing

cwant:
This looks great. The one peculiarity I noticed is that on the communities page, the abstract becomes a link (I don't mind it being a link per se, but the underline on the abstract on hover looked a bit off to me).

murny:
Yeah we might have to style this further. I agree with you though :+1: .

Looks like if we put a ` text-decoration: none;` on the anchor tag this should fix it. (removes the underline for everything, including the H1)

PR #160, opened 2017-09-20
Issue #130, item validations of paths. Bonus: collection validation of community.

### LINKED ISSUE #130: Add a "Work must belong to at least one community/collection" validation
DONE WHEN
- [ ] Validation prevents items from being created without being in any Communities/Collections
- [ ] Seeds & tests working w/ validation

cwant:
Fixes #130


app/controllers/admin/collections_controller.rb
```diff
@@ -15,8 +15,8 @@ def new
 
   def create
     @collection =
-      Collection.new_locked_ldp_object(permitted_attributes(Collection).merge(owner: current_user&.id))
-
+      Collection.new_locked_ldp_object(permitted_attributes(Collection)
+                .merge(owner: current_user&.id, community_id: @community.id))
```
mbarnett: I'm wondering if `current_user&.id` isn't just making it easier for future bugs to "work". Wouldn't it be better to crash if we are in a situation in which somehow a Collection is being created but there's no current user it's being created by?

The system's in an inconsistent state at that point AFAICT

app/models/item.rb
```diff
@@ -40,10 +40,29 @@ def each_community_collection
...
+      member_of_paths.each do |path|
+        community_id, collection_id = path.split('/')
+        begin
+          Community.find(community_id)
+        rescue JupiterCore::ObjectNotFound
+          errors.add(:member_of_paths, :community_not_found, id: community_id)
+        end
+        begin
+          Collection.find(collection_id)
+        rescue JupiterCore::ObjectNotFound
```
mbarnett: Same comment as above

app/models/collection.rb
```diff
@@ -40,6 +41,15 @@ def can_be_destroyed?
                         list_of_items: member_items.map(&:title).join(', ')))
       throw(:abort)
     end
+
+    def community_validations
+      return unless community_id
+      begin
+        Community.find(community_id)
+      rescue JupiterCore::ObjectNotFound
```
mbarnett: Maybe

```ruby
community = Community.where(id: community_id).first
errors.add(:community_id, :community_not_found, id: community_id) unless community.present?
community
```

It just seems like we're going the long way around to get where's semantics.

test/models/collection_test.rb
```diff
@@ -2,16 +2,41 @@
...
+  end
+
+  test 'community must exist' do
+    community_id = generate_random_string
+    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular_user).id,
+                                                  community_id: community_id)
+    collection.unlock_and_fetch_ldp_object {}
+
+    assert_not collection.valid?
+    assert_includes collection.errors[:community_id], "is missing community with ID \"#{community_id}\""
```
mbarnett: I'd suggest that what we care about in this test is that there's an error *on* `community_id`, not the specific text that constitutes that error?

If Sonya comes to us and says, hey could we word this error differently, we'll have to change both the implementation of the validation and the test, which is a sign that the test is overly coupled to the implementation's internals IMO. 

(you could make the argument that this is also true of the "must be present" checks, and I'd be inclined to agree, but there's at least the excuse there that the message is standard, whereas there's a reasonable chance this is going to need to change).

test/models/item_test.rb
```diff
@@ -65,4 +79,25 @@ class ItemTest < ActiveSupport::TestCase
...
+    item = Item.new_locked_ldp_object
+    community_id = generate_random_string
+    collection_id = generate_random_string
+    item.unlock_and_fetch_ldp_object do |unlocked|
+      unlocked.add_to_path(community_id, collection_id)
+    end
+
+    assert_not item.valid?
+    assert_includes item.errors[:member_of_paths], "is missing community with ID \"#{community_id}\""
+    assert_includes item.errors[:member_of_paths], "is missing collection with ID \"#{collection_id}\""
```
mbarnett: Same observation as above.

test/models/collection_test.rb
```diff
@@ -2,16 +2,41 @@
...
-    assert c.valid?
-    assert_equal c.visibility, JupiterCore::VISIBILITY_PUBLIC
+    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular_user).id)
+    collection.valid?
+    assert_equal collection.visibility, JupiterCore::VISIBILITY_PUBLIC
+  end
+
+  test 'a community_id must be present' do
+    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular_user).id)
+    collection.unlock_and_fetch_ldp_object {}
```
mbarnett: This isn't necessary.

test/models/collection_test.rb
```diff
@@ -2,16 +2,41 @@
...
+
+    assert_not collection.valid?
+    assert_includes collection.errors[:community_id], "can't be blank"
+  end
+
+  test 'community must exist' do
+    community_id = generate_random_string
+    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular_user).id,
+                                                  community_id: community_id)
+    collection.unlock_and_fetch_ldp_object {}
```
mbarnett: Ditto

test/models/item_test.rb
```diff
@@ -65,4 +79,25 @@ class ItemTest < ActiveSupport::TestCase
     assert_not_includes Item.display_attribute_names, :member_of_paths
   end
 
+  test 'a community/collection path must be present' do
+    item = Item.new_locked_ldp_object
+    item.unlock_and_fetch_ldp_object {}
```
mbarnett: Again, unnecessary.

app/models/item.rb
mbarnett: Or below, I guess

app/controllers/admin/collections_controller.rb
murny: Should be impossible to even get here if current_user is nil, since you being authorized twice to see if your an admin (in routes and controller). So definitely agree. Let it crash and crash hard if its here.

test/models/collection_test.rb
murny: Or instead of hard coded string...use the I18n translation: `I18n.t('activemodel.errors.models.ir_collection.attributes.community_id.community_not_found')`

which therefore your test still testing the string of the error but no longer overly coupled

mbarnett: Yeah, that's probably an even better idea

app/models/collection.rb
cwant: Hrmm, `where` seems to be broken (at least when combined with `first`):

```ruby
[9] pry(main)> community = Community.where(id: 'huh').first
NoMethodError: undefined method `[]' for nil:NilClass
from /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:117:in `block in where_clause'
[10] pry(main)> wtf?
Exception: NoMethodError: undefined method `[]' for nil:NilClass
--
0: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:117:in `block in where_clause'
1: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:116:in `each'
2: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:116:in `map'
3: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:116:in `where_clause'
4: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:101:in `reified_result_set'
5: /home/cwant/gitwork/jupiter/app/models/jupiter_core/deferred_solr_query.rb:47:in `each'
6: (pry):7:in `first'
7: (pry):7:in `__pry__'
8: /home/cwant/gitwork/jupiter/vendor/bundle/ruby/2.4.0/gems/pry-0.10.4/lib/pry/pry_instance.rb:355:in `eval'
9: /home/cwant/gitwork/jupiter/vendor/bundle/ruby/2.4.0/gems/pry-0.10.4/lib/pry/pry_instance.rb:355:in `evaluate_ruby'
```

mbarnett: Yikes. I'll take a look.

mbarnett: If you merge master it should work as intended now.

app/controllers/admin/collections_controller.rb
cwant: Done.

app/models/collection.rb
cwant: Thanks, restructured validation to use `where` in the latest push.

test/models/collection_test.rb
cwant: Gone.

test/models/collection_test.rb
cwant: Restructured to use I18n output in lastest push (one border case, if the translation is missing, the test passes, but likely not too much of an issue).

cwant:
The PR has been updated to address all of the comments.

test/models/collection_test.rb
murny: You added the raise exception in the test.rb environment, right? So if the translation is missing it now crashes, no longer an issue

cwant: I think only ActionView crashes (that is, the rails `t` helper is affected by the config setting). This is a model test, so no crash.

murny: Oooh interesting? Good to know! :+1:

PR #163, opened 2017-09-21
Fix communuities list items underlining everything on hover

murny:
Issue from here as noted by chris: https://github.com/ualbertalib/jupiter/pull/158#issuecomment-331276992

This fixes this.

### DIFF
--- a/app/views/communities/index.html.erb
+++ b/app/views/communities/index.html.erb
@@ -13,7 +13,7 @@
   <div class="col-md-12">
     <ul class="list-group">
     <% @communities.each do |community| %>
-      <li class="list-group-item list-group-item-action era-community-list-item">
+      <li class="list-group-item list-group-item-action jupiter-community-list-item">
       <%= link_to community_path(community), class: 'media' do %>
         <%= render partial: 'logo', locals: {object: community} %>
         <div class="media-body">
--- a/app/views/communities/index.html.erb
+++ b/app/views/communities/index.html.erb
@@ -13,7 +13,7 @@
   <div class="col-md-12">
     <ul class="list-group">
     <% @communities.each do |community| %>
-      <li class="list-group-item list-group-item-action">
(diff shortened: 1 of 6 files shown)


app/views/communities/index.html.erb
```diff
@@ -13,7 +13,7 @@
   <div class="col-md-12">
     <ul class="list-group">
     <% @communities.each do |community| %>
-      <li class="list-group-item list-group-item-action">
+      <li class="list-group-item list-group-item-action era-community-list-item">
```
murny: Good practice to prefix your css classes

- `.js-*` classes to denote behavior (as opposed to style) for anything classes being used by javascript
-  `era-*` for custom stuff for the project,
- everything else is safe to assume its from bootstrap

¯\_(ツ)_/¯

cwant: I wonder if this should be `jupiter-*` instead (we have very few references to 'ERA' in the source, beyond the i18n stuff)

murny: We can do this. Whats @mbarnett opinions? 

I sometimes get confused. So is it fair to assume: 

Code is exclusively jupiter referencing (everything is jupiter labelled in the code) and front facing stuff that will eventually be ERA is controlled via i18n stuff?

mbarnett: I don't have a strong preference either way?

murny: Well I'm going to say yes to my assumption then. Since one of the ideals was to allow other people to use Jupiter (which we will say is the software platform), so lets use jupiter everywhere in the code, and only use ERA when it comes to i18n (as this is the current brand and can be overridden by other people wanting to adopt jupiter) etc.

murny: PR has been updated to reflect this. Only place we were using ERA outside of i18n was for an alt tag on the navbar brand logo which is now fixed.

PR #170, opened 2017-09-25
Fix admin communitity list and remove the underlining on hover issue

### LINKED ISSUE #169: Fix hover underlining on admin communities & collections list page
Issue to remind myself to fix this. If you got to `/admin/communities` you will see this same bug happening

Forgot to be fixed in https://github.com/ualbertalib/jupiter/pull/163 (only did the external communities list pages)

Should be able to easily port the same fix over.

murny:
fixes #169 

Ended up doing some yak shaving:

- Brought in scss-lint as an attempt to tame houndCI. Good tool to bring in regardless. Essentially it's rubocop for your sass files.
- Upgrading all gems as well (as I had to for sass gem to fix dependencies). This includes a new rubocop version which has some changes (prefers safenavigator operator now)
- Enable rails cops for rails best practices


.scss-lint.yml
```diff
@@ -0,0 +1,23 @@
+# Linter Documentation:
+# https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md
+
+scss_files: 'app/assets/stylesheets/**/*.scss'
```
murny: We can wire this up on TravisCI is this is something you guys want to be incorporated into our CI workflow. For now it's there and configured. You can see what HoundCI will complain about before it hits your PR.

Simply run `scss-lint` in your console to run this.

.scss-lint.yml
```diff
@@ -0,0 +1,23 @@
...
+  # explaining why these linters are being disabled for this file.
+  # See https://github.com/brigade/scss-lint#disabling-linters-via-source for
+  # more information.
+  DisableLinterReason:
+    enabled: true
+
+  # TODO: Might enable this if this gets annoying
+  # Avoid qualifying elements in selectors (also known as "tag-qualifying").
+  # QualifyingElement:
+  #   enabled: false
```
murny: So far I am agreeing with this. But might get annoying

.rubocop.yml
```diff
@@ -105,10 +109,6 @@ Style/ClassVars:
   Exclude:
     - !ruby/regexp /_test\.rb$/
 
-Style/PredicateName:
```
murny: Deprecated name space, new namespace is `Naming`

Gemfile
```diff
@@ -60,7 +60,10 @@ group :development, :test do
   gem 'pry'
   gem 'pry-rails'
 
-  gem 'rubocop', require: false
+  # TODO: Pointing rubocop at master to resolve this bug: https://github.com/bbatsov/rubocop/pull/4749
+  # Once 0.50.1 or something lands, point back to the gem
+  gem 'rubocop', github: 'bbatsov/rubocop', require: false
```
murny: As comment says need to point this to master for an issue when running autocorrect (`rubocop -a`) and the new safenavigator cop. Once a new gem is cut, will point this back.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -294,6 +294,8 @@ def ldp_object=(obj)
       # NOTE: it's important to establish the owning object PRIOR to calling to_solr, as solr_calc_properties
       # could need to call methods that get forwarded to the owning object
       @solr_representation = @ldp_object.to_solr
+
+      # TODO: rubocop Lint/Void throwing cop violation here Fix this? @mbarnett
       @ldp_object
```
murny: Doesn't like this returning of the object: Raising a 
```
app/models/jupiter_core/locked_ldp_object.rb:298:7: W: Lint/Void: Variable @ldp_object used in void context.
      @ldp_object
      ^^^^^^^^^^^
```

http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/Void

mbarnett: I can't figure out what this cop is trying to do.

```ruby
# bad

def some_method
  some_num * 10
  do_something
end
```

```ruby
# good

def some_method
  do_something
  some_num * 10
end
```

...those return completely different things? Seems weird. Maybe just adding a `return` there will fix it?

murny: Doesn't like a return, new cop gets raised lol.

Yeah no idea. I disabled it on the file. I can remove the TODO comment here if you cant think of a better way to resolve this

murny: Removed this TODO, since it seems silly

mbarnett: It seems like it must be buggy or something. The method needs to return `@ldp_object` to be chainable, and removing that line fundamentally changes what it returns.

I feel like I'm missing something because it seems like whoever wrote this cop didn't understand that the last thing evaluated in a method == what the method returns

mbarnett: Seems like this cop has been around forever, so this must be some kind of regression bug

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -193,6 +193,8 @@ def self.find(id)
     end
 
     # find with "return nil if no object with that ID is found" semantics
+    # Note: This behaves differently then AR find_by.
```
murny: Added a note here as I made this mistake. Should we check the raise an error if the argument is a hash or something?

murny:
Thanks! Merging as is. 

Will add sass-lint to travis at the same of when I am adding system tests :+1:

PR #172, opened 2017-09-25
Lock solr version, version 7 came out this weekend, seems to be break…

murny:
…ing builds


murny:
See if this fixes builds, If so this is probably the issue and we can pull this in.

PR #173, opened 2017-09-29
Better approach to solrization issues

mbarnett:
Adds a class-level method to hoist already-defined ActiveFedora associations into the LockedLDPObject by indexing its IDs into Solr. Also allows for giving the association a new (better) name than the awful/misleading names some existing associations have. Also renames `solr_index` to `additional_search_index` in an effort to give it a name that indicates its purely adjunct nature/discourage use for non-primary use-cases -- we want to reserve this for cases where it's intended to transform data to better match a users' search.

 I'm thinking areas where we were trying to index attributes from associations onto an object for display are better served by either adding LockedLDPObjects representing the association, or denormalizing the join-happy Work-FileSet-File relationship and getting some of that data into Fedora on the Work or FileSet

### DIFF
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -1,5 +1,6 @@
-class FileSet < ActiveFedora::Base
+class FileSet < JupiterCore::LockedLdpObject
 
-  include Hydra::Works::FileSetBehavior
+  ldp_object_includes Hydra::Works::FileSetBehavior
+  use_existing_association :member_of_collections, using_name: :member_of_works
 
 end
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -1,6 +1,6 @@
 class FileSet < JupiterCore::LockedLdpObject
 
   ldp_object_includes Hydra::Works::FileSetBehavior
-  use_existing_association :member_of_collections, using_name: :member_of_works
+  use_existing_association :member_of_collections, using_name: :items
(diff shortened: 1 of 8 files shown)


app/models/file_set.rb
```diff
@@ -1,5 +1,6 @@
-class FileSet < ActiveFedora::Base
+class FileSet < JupiterCore::LockedLdpObject
 
-  include Hydra::Works::FileSetBehavior
+  ldp_object_includes Hydra::Works::FileSetBehavior
+  use_existing_association :member_of_collections, using_name: :member_of_works
```
cwant: Since `member_of_works` is a name of our choosing, would `member_of_items` be more appropriate? (Since our work model is now called `Item` instead of 'Work`).

cwant: Based on comments in other PRs, we might want to drop `member_of_` too, and since a fileset can only belong to one item, drop the `s` also, leaving just `item`.

mbarnett: Yeah, that's a good point.

Is `file_set.items` too much like the item is in the file_set? Maybe `owning_item` or `belongs_to_item` (that's kind of gross) or something else?

mbarnett: Or just `owner`?

mbarnett:
Will reopen with changes discussed in Slack

Issue #51, opened 2017-07-27
Add system tests to Travis CI


murny:
Currently Travis is only doing `rubocop` and `rails test`. `rails test` does not run system tests by default, you need to run another command called `rails test:system`.

Big concern is the extra infrasture needed. We need chromedriver/selenium installed (or use a different driver). Needs some looking into.

murny:
Found some info about doing this:
https://medium.com/@pacuna/using-rails-5-1-system-tests-with-docker-a90c52ed0648
https://gist.github.com/sanjibukai/a3d9aa38e0b0ae05368a7999fb813b4b

PR #174, opened 2017-10-03
Issue #125. Autocomplete, filter, paginate, sort users via AJAX

### LINKED ISSUE #125: Autocomplete / autosuggest users in search
Can we do an autocomplete / autosuggest search from list of users? there may be hundreds of user accounts with unfamiliar name spelling - this would make life easier for admins.

cwant:
Some notes:

* Branch name says 'typeahead', but typeahead has been discarded
* Added some more User seeds (without identities) so that the dev pages show pagination
* Renamed the default Admin, Non-admin user seeds so that the interface shows e.g., 'Jane Admin' instead of the generic noun 'Admin'.
* Pagination and sort are now done via AJAX on users page

Fixes #125


cwant:
I had thought of the double wildcard, but I think in general people want to match the start of tokens (e.g., if they type in 'no' they probably aren't looking for 'john_snow').

Thanks for the tip about the system tests and the search scope/index action -- will review and refactor accordingly. :+1: (To be honest, I don't run the system tests as often as I should.)

mbarnett:
Tangentially, does CI run the system tests against PRs currently?

murny:
Not yet, Will be working on this soon probably this week or next:https://github.com/ualbertalib/jupiter/issues/51

cwant:
I've addressed most of the comments. After poking around, the one that doesn't quite feel right is merging the autocomplete controller action into #index.

While similar, I think the use cases are different enough that they should be separate. E.g., autocomplete basically just provides a display name (name+email) and a URL, whereas #index provides all attributes for the selected user (good for sorting a table of users -- which down the road we might want to turn into an AJAX activity that uses the #index action).

Anyways, definitely open to suggestions as to how this can be done in a non-kludgy manor.

app/views/admin/users/index.html.erb
```diff
@@ -61,3 +50,6 @@
 </table>
 
 <%= paginate @users %>
+<%= javascript_tag do %>
+  init_autocomplete('#user-autocomplete', '<%= autocomplete_admin_users_path %>', { 'display_key': 'name_email'});
```
murny: Lets not include inline javascript tags? Must be a better way here?

app/assets/javascripts/application.js
```diff
@@ -15,4 +15,5 @@
 //= require jquery3
 //= require popper
 //= require bootstrap
+//= require typeahead.bundle
```
murny: Your PR comment suggest you used a fork? Any documentation here?
```
Using forked typeahead script. The current typeahead doesn't
render all results from AJAX query.
```

What/Why was this forked? When can we upgrade to regular? etc?

app/models/user.rb
```diff
@@ -12,8 +12,15 @@ class User < ApplicationRecord
...
-        .or(where('lower(name) LIKE ?', "%#{query.downcase}%"))
+      sanitized = sanitize_sql_like(query)
+      # Match start of first name, last name, email, hyphenated surnames
+      start_of_string = "#{sanitized}%".downcase
+      after_space = "% #{sanitized}%".downcase
+      after_hyphen = "%-#{sanitized}%".downcase
+      where('lower(name) like ?', start_of_string)
+        .or(User.where('lower(name) like ?', after_space))
+        .or(User.where('lower(name) like ?', after_hyphen))
+        .or(User.where('lower(email) like ?', start_of_string))
```
mbarnett: Should we look at bringing in a full-text SQL solution instead of doing this kind of thing? Or bring in Sunspot on the ActiveRecord side to tie into Solr for that?

murny: I would say Ransack or pg_search (if we had postgres) for sure. 

Also regarding this comment:

``` I had thought of the double wildcard, but I think in general people want to match the start of tokens (e.g., if they type in 'no' they probably aren't looking for 'john_snow').``` If you use the double wildcard, typeahead highlights your search query with the results.. so it be pretty evident that `no` becomes john s**no**w

Comment (mbarnett) [reply app/models/user.rb]: Ransack would probably be less crazy than Sunspot 👍

Comment (cwant) [reply app/models/user.rb]: Regarding the double wildcard, I'll do it your way. The highlighting of matched characters has already been enabled. Lets hope nobody every searches for somebody named Albert (in an organization with ualberta.ca email addresses).

To be honest, I wonder about the utility of bringing in a new gem to avoid doing three lines of `or()` (particularly in a project that otherwise uses a low amount of SQL). But moot point, if I do the double wildcard, the `or()` go away.

File: app/views/admin/users/index.html.erb (hunk 1 above)
Comment (cwant) [reply app/views/admin/users/index.html.erb]: The one thing I like about the inline (and the main reason I did it this way) is that I can use the path helpers to find the autocomplete action (e.g. `autocomplete_admin_users_path` vs using hard-coded string `/admin/users/autocomplete`). I'm open to alternatives though.

Comment (murny) [reply app/views/admin/users/index.html.erb]: Put the autocomplete path on the `text_field_tag` as a data attr? Then when your java script kicks it, it reads the data attr and initializes the autocomplete?

Inline javascript tags are bad as your javascript isnt cacheable, it hurts performance, and this also ties your rails app to close to the client js, etc.

Comment (cwant) [reply app/views/admin/users/index.html.erb]: This is a good solution, I'll do it this way.

File: vendor/assets/javascripts/typeahead.bundle.js [hunk 4]
```diff
@@ -1,7 +1,3 @@
-// Note downloaded from:
```
murny: Ah sorry didn't see this! Github collapses this due to file size being to big. But i guess having it in the `application.js` will be more noticeable

app/assets/javascripts/application.js
```diff
@@ -15,4 +15,11 @@
 //= require jquery3
 //= require popper
 //= require bootstrap
+
+// Note typeahead fork downloaded from:
+// https://raw.githubusercontent.com/coopy/typeahead.js/fix-async-render-dist/dist/typeahead.bundle.min.js
+// Fixes this bad bug: https://github.com/twitter/typeahead.js/pull/1212
+// Typeahead is infrequently updated, so no estimated date for a fix in production script.
```
murny: Yeah no kidding, this project looks abandoned (last commit was 3 years ago) :cry: . Did you ever evaluate another search project? Like Select2 or something?

app/models/user.rb
```diff
@@ -12,8 +12,8 @@ class User < ApplicationRecord
 
   scope :search, lambda { |query|
     if query.present?
-      where('lower(email) LIKE ?', "%#{query.downcase}%")
-        .or(where('lower(name) LIKE ?', "%#{query.downcase}%"))
+      sanitized_query = "%#{sanitize_sql_like(query.downcase)}%"
```
murny: Good call on `sanitize_sql_like` :+1:

test/controllers/admin/users_controller_test.rb
```diff
@@ -89,4 +89,11 @@ class Admin::UsersControllerTest < ActionDispatch::IntegrationTest
     end
   end
 
+  context '#autocomplete' do
+    should 'be able to autocomplete user' do
+      get autocomplete_admin_users_url
+      assert_response :success
```
murny: Should beef this up and bit? Check the response body? Etc?

app/controllers/admin/users_controller.rb
```diff
@@ -77,6 +79,15 @@ def impersonate
     redirect_to root_path, notice: t('admin.users.show.impersonate_flash', user: @user.name)
   end
 
+  def autocomplete
+    return render(json: '') if params[:query].blank?
```
murny: Would this ever be blank from the UI (I assume a blank form submits an empty string ('') not nil? Why not just make `params[:query]` required?

db/seeds.rb
```diff
@@ -18,13 +18,19 @@
...
   admin.identities.create(provider: 'developer', uid: 'admin@ualberta.ca')
 
   # Seed a non-admin user
-  non_admin = User.create(name: 'Non-admin', email: 'non_admin@ualberta.ca', admin: false)
+  non_admin = User.create(name: 'Bill Non-admin', email: 'non_admin@ualberta.ca', admin: false)
   non_admin.identities.create(provider: 'developer', uid: 'non_admin@ualberta.ca')
 
+  # A bunch of non-identity users for to manipulate in the admin interface
+  (0..100).each do
+    name = Faker::GameOfThrones.unique.character
```
murny: Haha `Faker::GameOfThrones` nice! Thats awesome

test/controllers/admin/users_controller_test.rb
murny: I guess thats what your integration tests do.  (controller tests and integration tests are using the exact same API) But integration tests are meant for testing how various parts of your application interact. Aka testing workflows within your application. So probably can move these tests here etc?

test/policies/admin/user_policy_test.rb
```diff
@@ -91,4 +91,14 @@ class Admin::UserPolicyTest < ActiveSupport::TestCase
     end
   end
 
+  context '#autocomplete?' do
+    should 'not be able to unsuspend an unsupended user' do
+      refute Admin::UserPolicy.new(@current_user, [:admin, @user]).unsuspend?
```
murny: Copy paste here? Or what is this?

app/assets/javascripts/application.js
cwant: I tried the jquery-ui one and the typeahead one, did not evaluate others.

test/system/admin_users_test.rb
```diff
@@ -21,12 +21,25 @@ class AdminUsersTest < ApplicationSystemTestCase
...
       assert_selector 'tbody tr:last-child th[scope="row"]', text: admin.email
+    end
 
-      fill_in name: 'query', with: 'admin'
-      click_button I18n.t('admin.users.index.search_button')
+    should 'be able to autocomplete a user' do
+      admin = users(:admin)
 
-      assert_selector 'tbody tr', count: 1
-      assert_selector 'tbody tr:first-child th[scope="row"]', text: admin.email
```
murny: Wait I may be confused here? Does the table not change anymore? It just gives you suggestions when you type and if you click on a suggestion it goes to the user detail page?

murny: We might have to discuss this if this is how its working? And clarify requirements with sonya/leah

cwant: Table does not change. If you click on a suggestion, it goes there.

test/policies/admin/user_policy_test.rb
cwant: Oops

test/system/admin_users_test.rb
murny: But I can still type like 'Joe' and hit enter? And this will go and change the table (aka take me to `/admin/users?query=joe`)?

cwant: As currently implemented, you can click suggestions or browse the full list.

I suppose the requirements do need to be clarified if you want the table to change also as the user types.

A couple of points though: a) the mockup shows no table at all; b) at a standup there was a discussion that once the page had autocomplete it would no longer respond to a query in the query string (I think the context was related to what happens on page reload).

app/policies/admin/user_policy.rb
```diff
@@ -20,6 +20,10 @@ def impersonate?
     regular_user? && not_self?
   end
 
+  def autocomplete?
+    admin?
```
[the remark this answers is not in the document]
cwant: Just to be sure: are any of the user policies being run? (i.e., if the check is elsewhere -- likely the `ensure_admin` in the admin controller -- should the entire file be removed?)

test/system/admin_users_test.rb
murny: As always mockups have never been fleshed out enough to be used as the source of truth of requirements, reason why I have been trying to do exactly that. Going back and fixing/updating them, and raising requirement questions so that maybe they can be. Also mockup now has a table.

Yeah I think theres confusion here. If you hooked up the table and search via AJAX then you would get something that feels exactly like autocomplete/typeahead, without needing this external library or any other library. This would solve their requirements with very little code on your part needing to be written. But I guess it all depends on what they want.

If you want to discuss this in person let me know. But I think we making this more complicated then needed.

app/policies/admin/user_policy.rb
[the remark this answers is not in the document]
murny: Yes everything here is being used. They are used to check the state of things, Like you can't suspend an admin...you have to make them a normal user and then suspend them, etc. Shows/Hides the buttons on the UI as well.

test/system/admin_users_test.rb
cwant: I'm fine with making it the way the mockup is currently presented provided that @leahvanderjagt or @sfbetz sign off on it. And yeah, removing a poorly supported external library would be a plus.

sfbetz: To clarify this one, we were imagining the table would change dynamically as we typed. So if I start typing "son.." I want to see the table show me user results with those letters. As I keep typing I want the table to change reflecting more specific results until I see the user I want in the list.

cwant: Sounds good, I will work towards that (the mockup from six days ago showed a list dropping down from the autocomplete, not a table).

sfbetz: My wireframing did leave much to the imagination - sorry!

db/seeds.rb
cwant: Yeah, we might need to have a 'Killed' column in the user table. :dagger: :smiling_imp: 
Although that gives me the idea that we should probably create some suspended users in the seeds, since we are sorting by that (and likely filtering too).

cwant:
Big rewrite pushed, see updated description above.

Gemfile
```diff
@@ -79,5 +79,7 @@ group :test do
   # Faker added 0.5 seconds to the test suite per call. Haikunator seems much faster for faking strings
   gem 'haikunator'
   gem 'minitest-hooks'
+  # Allows access to controller instance variables ('assigns') in controller/integration tests
+  gem 'rails-controller-testing'
```
murny: Lets' not bring this in. I agree with rails decision to remove this. Check the response body instead here or something smarter. Or make them system tests and check the HTML markup etc.

test/integration/admin/index_test.rb
```diff
@@ -0,0 +1,121 @@
...
+                 { name: 'Trader Joe', email: 'lol@example.com', admin: false, suspended: false },
+                 { name: 'Harland Sanders', email: 'that.joker@example.com', admin: true, suspended: true }])
+  end
+
+  test 'with no query' do
+    user = users(:admin)
+    sign_in_as user
+    get admin_users_path
+    assert_equal assigns[:users].sort, User.all.sort
+    assert_equal assigns[:users].count, 12
```
murny: For example you can change this stuff to checking the response json body:
```
json_response = JSON.parse(@response.body) //Maybe make this a generic helper? 
assert json_response['user'].is_a?(Array)
assert_equal json_response['user'].sort, User.all.sort
assert_equal json_response['user'].count, User.count
```

cwant: Thanks Shane, I've removed the gem and reworked the tests.

Gemfile
```diff
@@ -79,5 +79,6 @@ group :test do
   # Faker added 0.5 seconds to the test suite per call. Haikunator seems much faster for faking strings
   gem 'haikunator'
   gem 'minitest-hooks'
+  # Allows access to controller instance variables ('assigns') in controller/integration tests
```
murny: Can remove this comment now :smile:

app/models/user.rb
```diff
@@ -1,5 +1,11 @@
 class User < ApplicationRecord
 
+  FILTER_MAP = { all: I18n.t(:all),
```
murny: View code like this shouldn't live in the model? Move to helper or just use in the view itself?

app/models/user.rb
```diff
@@ -12,8 +18,25 @@ class User < ApplicationRecord
...
+        where(admin: false)
+      when 'admin'
+        where(admin: true)
+      when 'active'
+        where(suspended: false)
+      when 'suspended'
+        where(suspended: true)
+      else
+        # we just ignore 'all' (means unscoped), fail on everything else
+        raise ArgumentError, 'not a valid filter' unless FILTER_MAP.keys.include?(filter.to_sym)
```
murny: Default should just be `all` here. Not raise an exception. Like if filter='' or someone munged the filter, just give them all the results and don't throw a 500 exception.

Could be an honest mistake like I copied paste the url but i removed/added an extra character. And now that url returns a 500 page

cwant: Done.

app/models/user.rb
cwant: Done.

Gemfile
cwant: Done. (D'oh!)

app/assets/javascripts/jupiter/autocomplete.js
```diff
@@ -0,0 +1,21 @@
+$(document).on('turbolinks:load', function() {
+  // Fetch new table when clicking headers and pagination links
+  $('.jupiter-autocomplete-results').on('click', 'thead a, a.page-link', function() {
```
murny: is this needed if you have `remote: true` on all the links? Think this is exactly what remote: true does

murny: Update the sort links to remote true and I bet this works without this code

cwant: Will give it a go, and remove the JS if things work out.

cwant: Well, I'll be gosh-darned -- it works! :+1: . The helper for the table headers already had 'remote-true'.

app/views/admin/users/index.html.erb
```diff
@@ -1,63 +1,38 @@
...
 
-  <div class="input-group">
-    <%= search_field_tag :query, params[:query], placeholder: t('.search_placeholder'), class: 'form-control' %>
-    <span class="input-group-btn">
-      <%= button_tag name: nil, type: 'submit', class: "btn btn-secondary" do %>
-        <%= fa_icon 'search' %>
-        <%= t('.search_button') %>
-      <% end %>
-    </span>
+  <div id="users" class='jupiter-autocomplete-results'>
```
murny: `jupiter-autocomplete-results` class never being used anywhere

app/assets/javascripts/jupiter/autocomplete.js
```diff
@@ -0,0 +1,16 @@
+$(document).on('turbolinks:load', function() {
```
murny: Matt created an admin javascript folder (https://github.com/ualbertalib/jupiter/tree/master/app/assets/javascripts/admin). This should go there. As this will only be used in the admin users page? Maybe rename this as such as well. As autocomplete is pretty generic.

murny: After that this looks good. I'll approve

cwant: I had envisioned this as being a reusable thing (render a table of results based on autocomplete, filter, and pagination -- no mention of users anywhere in the code). But at this point, I'm beyond caring, and will make it less generic.

murny: Okay we can leave as is then. I just don't see any autocomplete happening in general jupiter. But admin for sure (Announcements for example). Lets just move this to admin folder as is and thats good enough?

cwant: Yeah, no worries. I've already moved to make it user-specific (push coming soon). We can re-generalize it by changing two lines of code in the future.

And sorry if that last comment sounded harsh (upon re-reading it sounds that way) -- I just wanted to express that it's more important for me to get the PR done and move on than to dig in my heels.

Issue #168, opened 2017-09-22
Improvement for predeploy.sh/deploy.sh scripts


murny:
Some quick things I noticed and some suggestions to improve them. Some of these are just food for thought. Suggestions below: 

1.  When we pull down `docker-compose.deployment.yml` onto the server, we should rename this to just `docker-compose.yml` then you no longer have to pass the flag option to the docker-compose command everywhere

2. We could put all docker files or a seperate REPO (including a nice README.md to explain everything) that need to be curled/git pulled/rsynced/copied over to the server in a `docker` folder in jupiter. The script will just copy everything over in this folder. Why? Better way to manage dependencies. You can just add any new files/remove files from this folder without constantly updating the predeploy script. Also simplifies predeploy script to basically a couple commands 

3. If we move things to a folder/repo as suggested above, we can also moving the `Dockerfile.deployment` and ` `docker-compose.deployment.yml`  files into this folder. In this folder/repo we can rename these two files as `Dockerfile` and `docker-compose.yml` (which will solve #1 above as we no longer have to do the rename on the server and no longer have to use `-f` flag everywhere). If doing this we should indicate that these files are for production use, either by commenting well, naming the folder as docker-production or something (if its a repo called it jupiter-deploy or something).

4. Make predeploy script repeatable, should be able to run it as many times as I want. Which means we probably need to include checks for existing files like https://github.com/ualbertalib/jupiter/blob/master/predeploy.sh#L12 or if using `rsync` or another command, you can configure it to ignore existing files or only update files that are have a modiefied timestamp that is newer then the source file. Why? What happens if theres changes to nginx.conf or docker-compose, etc etc? Someone has to manually copy these over/edit them on the server instead of just simply...rerunning the `predeploy.sh` script again 

5. The last step in the `deploy.sh` should include a `touch tmp/restart.txt` which will restart the rails server and refresh any cache/assests. (https://github.com/ualbertalib/di_internal/pull/7)

murny:
`#1` is now complete

PR #176, opened 2017-10-03
Fix having to pass in -f flag to docker-compose

murny:
Also lots of refactors to conform to bash best practices

This fixes Number 1 in this issue https://github.com/ualbertalib/jupiter/issues/168

> When we pull down docker-compose.deployment.yml onto the server, we should rename this to just docker-compose.yml then you no longer have to pass the flag option with the file (-f docker-compose.deployment.yml) to the docker-compose command everywhere


deploy
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
murny: Executables shouldn't have an extension, also provide a shebang and space this script out a bit for readability

deploy
```diff
@@ -0,0 +1,10 @@
...
+
+docker rmi $(docker images -f "dangling=true" -q)
+
+docker-compose pull web
```
murny: No longer need the `-f docker-compose.deployment.yml` everywhere as we are pulling down this yml file and renaming it to the default name

predeploy
```diff
@@ -0,0 +1,25 @@
...
+# This script should be either rsynced or copied to the UAT server when setting up fresh server.
+# For more information, please see the documentation: https://github.com/ualbertalib/di_internal/blob/master/System-Adminstration/UAT-Environment.md
+
+set -o errexit
+set -o pipefail
+set -o nounset
+[[ "${TRACE}" ]] && set -o xtrace
+
+
+readonly branch_name="master"
```
murny: Best practice is to use upper case for Environment variables and also to make these variables immutable

predeploy
```diff
@@ -0,0 +1,25 @@
...
+readonly install_directory="/home/deploy/jupiter"
+
+mkdir -p "${install_directory}"
+mkdir -p "${install_directory}/config"
```
murny: Use quotes and `{}` everywhere as bash can do tricky things when you have spaces, etc.

predeploy
```diff
@@ -0,0 +1,25 @@
...
+curl -o "${install_directory}/deploy" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/deploy"
+curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.deployment.yml"
+curl -o "${install_directory}/config/nginx.conf" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/config/nginx.conf"
+
+if [[ ! -f "${install_directory}/.env_deployment" ]]
```
murny: Previously this if check wasn't doing what it was intended. As it was checking the directory of this script and not the actual install_directory (which very well could be different!)

predeploy
```diff
@@ -0,0 +1,25 @@
...
+[[ "${TRACE}" ]] && set -o xtrace
+
+
+readonly branch_name="master"
+readonly install_directory="/home/deploy/jupiter"
+
+mkdir -p "${install_directory}"
+mkdir -p "${install_directory}/config"
+curl -o "${install_directory}/deploy" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/deploy"
+curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.deployment.yml"
```
murny: Main change of this PR. Let's rename the `docker-compose.deployment.yml` to `docker-compose.yml` when we pull it down from github

predeploy
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# This script will be sitting in /home/deploy directory, and is for setting up directories and pulling required files for docker deployment
+# This script should be either rsynced or copied to the UAT server when setting up fresh server.
+# For more information, please see the documentation: https://github.com/ualbertalib/di_internal/blob/master/System-Adminstration/UAT-Environment.md
+
+set -o errexit
```
murny: Good practice to set these

weiweishi:
LGTM. Learnt something new.
I totally missed the ticket #168 but agreed that many of the points should be addressed to make the script/process more clear to read and more streamlined to execute.

PR #178, opened 2017-10-03
Add System Test to docker/travisCI

### LINKED ISSUE #51: Add system tests to Travis CI

murny:
fixes #51 

- system tests are now passing on travisCI in this PR (adds an additional ~30 seconds to CI build time)
- Also added scss-lint to travisCI as well (adds a couple seconds to CI build time)
- Use bundler with jobs/retry params (cut down a good 20-40 seconds from CI build time)
- Fix up dev login logic and removed dummy data for dev/test for saml (as I was spamming uat-login hard by accident)
- Jobs take around ~5 minutues now


Dockerfile
```diff
@@ -27,16 +25,13 @@ RUN apt-get update -qq \
 #     && chmod a+x /usr/local/fits/fits-1.0.6/fits.sh \
 #     && ln -s /usr/local/fits/fits-1.0.6/fits.sh /usr/bin/fits
 
-# install phantomjs for capybara as we are using poltergeist
```
murny: Don't need this stuff as we are using selenium in a seperate container (removed npm/nodejs-legacy packages above too)

Dockerfile
```diff
@@ -27,16 +25,13 @@ RUN apt-get update -qq \
...
-# RUN npm install -g phantomjs-prebuilt
-
 RUN mkdir -p /app
 WORKDIR /app
 
 # Preinstall gems in an earlier layer so we don't reinstall every time any file changes.
 COPY Gemfile /app
 COPY Gemfile.lock /app
-RUN bundle install
+RUN bundle install --jobs=3 --retry=3
```
cwant: Just out of curiosity, do you know the impact on bundle install time of the parallelization? (likely not 3x faster, but hopefully greater than 1x).

murny: On our CI, `docker-compose build`  previously was hovering around 150-160 seconds to build. Now its around 120-130 seconds.

From the official bundle docs when they introduced `jobs` they said they say anywhere between 40-60% speedup on fresh bundle installs: http://bundler.io/v1.5/whats_new.html

Issue #166, opened 2017-09-21
Index date_created into a year created


mbarnett:
DONE WHEN:

- [ ] date_created changed to string to allow for free-formed entry
- [ ] year_created solr index somehow intelligently populating from date-like strings (Peter suggests 4 digit regexes) (Shane suggested Chronic, which might give better results in some cases, but we might need to fall back to a regex for some proposed inputs like "before 1989").

pbinkley:
Here's the [corresponding issue](https://github.com/ualbertalib/metadata/issues/109) on the data dictionary.

weiweishi:
migrated object will have sort_year generated. This needs to be populated during item deposit.

cwant:
Is there anything else needed for this ticket now that `sort_year` is implemented?

weiweishi:
feel this is completed. Close it for now. If I miss anything, please feel free to reopen.

PR #179, opened 2017-10-04
Issue # 164. Some format wrangling for dates encoded as strings.

cwant:
This allows entering information about Items in our forms (create/update)
to work again.

This definitely needs a sanity check. An alternative approach would be to convert string dates coming from forms to dates in the controller (and hence can't use methods like `#update_attributes`).

Before:

```
uitem.date_created = '2017-09-29T20:29:25+00:00'
=> "2017-09-29T20:29:25+00:00"
uitem.save!
RSolr::Error::Http: RSolr::Error::Http - 400 Bad Request
Error: {"responseHeader":{"status":400,"QTime":0},"error":{"metadata":["error-c\
lass","org.apache.solr.common.SolrException","root-error-class","org.apache.sol\
r.common.SolrException"],"msg":"Invalid Date String:'2017-09-29T20:29:25+00:00'\
","code":400}}

uitem.date_created = '2017-09-29'
=> "2017-09-29"
uitem.save!
RSolr::Error::Http: RSolr::Error::Http - 400 Bad Request
Error: {"responseHeader":{"status":400,"QTime":0},"error":{"metadata":["error-c\
lass","org.apache.solr.common.SolrException","root-error-class","org.apache.sol\
r.common.SolrException"],"msg":"Invalid Date String:'2017-09-29'","code":400}}

uitem.date_created = 'dog'
=> "dog"
```

After:

```
uitem.date_created = '2017-09-29T20:29:25+00:00'
=> "2017-09-29T20:29:25+00:00"
uitem.save!
=> true

uitem.date_created = '2017-09-29'
=> "2017-09-29"
uitem.save!
=> true

uitem.date_created = 'dog'
TypeError: dog is not a Date type
```

### DIFF
--- a/app/models/jupiter_core/locked_ldp_object.rb
+++ b/app/models/jupiter_core/locked_ldp_object.rb
@@ -1,4 +1,5 @@
 # coding: utf-8
+
 # JupiterCore::LockedLdpObject classes are lightweight, read-only objects
 # TODO: this file could benefit from some reorganization, possibly into several files
 module JupiterCore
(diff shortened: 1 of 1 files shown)


mbarnett:
I thiiiink where we landed on this is we're going to make date_created a straight up string rather than date, to allow for free-form input, and then add some logic somewhere based on regexes or chronic or chronic falling back to regexes to tease out a year, and index that into a separate property that has an actual "date" type. The related ticket there is https://github.com/ualbertalib/jupiter/issues/166

cwant:
Thanks. I did see that ticket and noted the overlap.

The primary issue right now is that the item form is currently unusable since #147 was merged (no matter what is put in the form, the app throws a 500 due to `date_created` on save). To work on the item deposit features, it might be worth merging in some stop gap fix until #166  is sorted out.

mbarnett:
Good point. I wonder if we shouldn't just rip it off the form, or straight out of the model, for the time being, until someone's in a position to do it the right way?

There's a bunch of other attributes missing on the model anyways, and it would force us to clean up a few spots like https://github.com/ualbertalib/jupiter/blob/03e198f5ee6574e23fd18e60a1c94630737cc650/app/controllers/admin/dashboard_controller.rb#L4 where we used date_created early on but probably want to be using record_created_at anyways?

cwant:
Sure, lets close this one and I can submit a separate PR removing `date_created` for now.

PR #181, opened 2017-10-04
Issue #164: remove `date_created` from the item model.

### LINKED ISSUE #164: Items won't save via the web form
This is due to the fact that the Item model no longer takes a string as input. Need to convert the form input to either a nil or a date type before save (either in model, or in controller).

cwant:
This will allow the item form to work without crashing. Please note that it may be worth rerunning the seeds after this change (probably fine without, but who knows ...)

Fixes #164


app/models/item.rb
```diff
@@ -11,7 +11,6 @@ class Item < JupiterCore::LockedLdpObject
   has_attribute :contributor, ::RDF::Vocab::DC.contributor, solrize_for: [:search, :facet]
   has_attribute :description, ::RDF::Vocab::DC.description, type: :text, solrize_for: :search
   has_attribute :publisher, ::RDF::Vocab::DC.publisher, solrize_for: [:search, :facet]
-  has_attribute :date_created, ::RDF::Vocab::DC.created, type: :date, solrize_for: [:search, :sort]
```
murny: Do we want to remove this entirely? I think we will have some type of "date_created" in item still. Whatever it will be called.

mbarnett: I suggested ripping it out for the time being on https://github.com/ualbertalib/jupiter/pull/179

Essentially, in the long term the desire is, it's a string, plus another index populated by additional logic. Since the form is broken right now, and Chris is working on other things on the form, it seemed easier to rip it out and then do it correctly with #166

app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Admin::DashboardController < Admin::AdminController
 
   def index
-    @items = Item.limit(10).sort(:date_created, :desc) # TODO: Should use record_created_at?
+    @items = Item.limit(10) # Default order: record_created_at (desc)
```
mbarnett: We probably don't need the comment? We're not really leaving comments anywhere else we're relying on defaults.

app/models/item.rb
murny: Okay sounds good. Approving this PR then

PR #182, opened 2017-10-05
Facet linking

mbarnett:
Pagination, and tests are following shortly behind, but we might as well get this PR review started


app/controllers/items_controller.rb
```diff
@@ -47,8 +47,9 @@ def update
   end
 
   def search
-    @results = Item.search(q: params[:q])
-    authorize @results, :index?
+    params[:facets].permit! if params[:facets].present?
+    @results = JupiterCore::Search.search(q: params[:q], facets: params[:facets], models: Item)
+    authorize Item.new_locked_ldp_object
```
murny: Why creating a new object to authorize on? Should be able to just do `authorize Item` aka authorize on the class instead of the instance

app/helpers/application_helper.rb
```diff
@@ -4,4 +4,14 @@ def page_title(title)
     @page_title.push(title) if title.present?
     @page_title.join(' | ')
   end
+
+  def facetable_query_params(facet_name, value)
+    query_params = { q: params[:q] }
+    active_facets = params[:facets] if params[:facets]
+    active_facets ||= {}
```
murny: Couldn't you combine these two lines?
```ruby
active_facets = params[:facets] || {}
```

app/models/jupiter_core/search.rb
```diff
@@ -71,6 +75,11 @@ def self.perform_solr_query(q:, fq: '', facet: false, facet_fields: [],
     params[:start] = start if start.present?
     params[:sort] = sort if sort.present?
 
+    # puts "*" * 40
```
murny: remove debug comments here?

app/controllers/items_controller.rb
mbarnett: Yeah, this is a quick hack so I could debug the interface. I'll rip it back out. This whole method probably belongs in its own controller dedicated to search

app/controllers/items_controller.rb
```diff
@@ -46,10 +46,13 @@ def update
...
 
+  # put this in its own controller
   def search
     params[:facets].permit! if params[:facets].present?
-    @results = JupiterCore::Search.search(q: params[:q], facets: params[:facets], models: Item)
-    authorize Item.new_locked_ldp_object
+
+    @results = JupiterCore::Search.faceted_search(q: params[:q], facets: params[:facets], models: [Item, Collection, Community])
+    @results.sort(:title, :asc).page params[:page]
+    authorize Item, :search?
```
murny: Is this needed? I think pundit is smart enough to say... Okay I'm in the search action, go look up the search policy in the ItemPolicy?

Just needs `authorize Item` here I believe

PR #184, opened 2017-10-05
Deposit form: collection/community selection

### LINKED ISSUE #180: Collection/Community element in Item form
The goal is to match the mockup here:

https://jupiter.mybalsamiq.com/mockups/6148333.png

DONE WHEN:
- [x] The form allows entering multiple community/collection pairs
- [x] The form allows removing community/collection pairs
- [x] The form allows modifying each community/collection pair
- [x] When updating an item, the form correctly populates the existing community/collection pairs in the edit form
- [x] The form does not lose any entered information about collections/communities on error
- [x] The form identifies "Community/Collection" as a required field
- [x] The form element for community/collection has a tooltip

cwant:
Making community/collection item form elements properly handle multiple values, populates correctly in edit form, etc.

Here is the screenshot:

c_c_screenshot

Here is the mockup:

Mockup

Fixes #180


app/assets/stylesheets/items.scss
```diff
@@ -0,0 +1,14 @@
+$cc-background: $gray-100;
+
+.communities-collections {
+  .invalid-feedback {
+    display: block;
```
murny: Any real reason for this? Bootstrap provides a `.d-block` utility class if you need: https://getbootstrap.com/docs/4.0/utilities/display/#common-display-values

cwant: The `.invalid-feedback` span is generated by simple_form's `errors` method, rather than via the magic of the `input` method. It shows up as `display: none` by default:

https://github.com/twbs/bootstrap-rubygem/blob/master/assets/stylesheets/bootstrap/_forms.scss#L258

The error messages that are generated via the `input` method have some additional classes added so that they show up:

https://github.com/twbs/bootstrap-rubygem/blob/master/assets/stylesheets/bootstrap/mixins/_forms.scss#L38

This surprised me, so I brute forced it.

murny: Should we make this a bit more generic? and put this in a simple_form.scss or something so it works for all forms using simple form? Or is this specific to communities-collections?

cwant: Yeah sure, done.

app/helpers/application_helper.rb
```diff
@@ -21,4 +21,8 @@ def facetable_query_params(facet_name, value)
 
     query_params
   end
+
+  def qmark_tooltip(text)
```
murny: nitpicky but can we change the `qmark` to something more descriptive?  Use `question_mark_tooltip` or better yet just call it `help_tooltip` etc?

app/views/items/_community_collection_input.html.erb
```diff
@@ -0,0 +1,30 @@
+<% community_id = local_assigns[:community_id] %>
```
murny: Hmm these should all be required and not optional? Is using `local_assigns` needed here then? Can we just 500/fail hard if they are not supplied? etc?

app/views/items/_form.html.erb
```diff
@@ -9,19 +9,34 @@
...
+    <div class="card-block">
+      <%= f.label :member_of_paths, label: t('admin.communities.title'),
+          class: 'card-title', required: true %>
+      <%= qmark_tooltip(t('.community_collection_tooltip')) %>
+      <%= f.error(:member_of_paths) %>
+      <div class="communities-collections-list">
+        <% if @item.member_of_paths.any? %>
+          <% @item.member_of_paths.each do |path| %>
+            <% community_id, collection_id = path.split('/') %>
+            <% collections = Collection.where(community_id: community_id) %>
```
murny: This stuff (variable assignment/Database queries) shouldn't be done in a view? Is there a better way of doing these things? In the controller?

app/helpers/application_helper.rb
cwant: Done.

app/assets/javascripts/jupiter/items.js
```diff
@@ -1,27 +1,60 @@
 $(document).on('turbolinks:load', function() {
-	$("#new_item #community").change(function() {
-		update_collections();
-	});
+  $('#item-form')
```
murny: Is the ID needed here? Can we use a class or something instead?

app/assets/javascripts/jupiter/items.js
```diff
@@ -1,27 +1,60 @@
...
+});
 
-	$("#new_item .add-file").click(function(event){
-		event.preventDefault();
-		new_file_input();
-	});
+function update_collection_select($community_select) {
+  $collection_select = collection_select($community_select);
+  id =  $community_select.find('option:selected').val();
+  // Bad value? Disable the collection select
```
murny: How does a user correct this if you disable the select box? Or what does this do? Will this ever happened even?

app/views/items/_community_collection_input.html.erb
cwant: Actually no. When creating a new item, these aren't set so the form needs to handle that (not have a default selected). Also. if a community isn't selected, can't load the list of collections.

app/assets/javascripts/jupiter/items.js
murny: Oh sorry I see now. You disable it when its being submitted, and then you reenable this after a successful api call

murny: So I guess this is for if you choose a community with no collections or something? It will disable the collections? And then you can just change the community to removed the disabled eh?

app/views/items/_community_collection_input.html.erb
murny: But your setting them right here? Only way this partial gets rendered I think? Maybe I'm confused?

https://github.com/ualbertalib/jupiter/blob/63f9a173137ebc0d9e717eda354466a127189c4c/app/views/items/_form.html.erb#L23-L26

cwant: A couple of lines after that, there is an example where they are not set:

https://github.com/ualbertalib/jupiter/blob/63f9a173137ebc0d9e717eda354466a127189c4c/app/views/items/_form.html.erb#L28

murny: Ah I see. So why not just pass in the locals here? Instead of these three variable assignments in this file? 
```
<%= render partial: 'community_collection_input', 
                 locals: { f: f, community_id: nil, collection_id: nil, collections: [] } %> ```

Comment (cwant) [reply app/views/items/_community_collection_input.html.erb]: Done.

File: app/views/items/_form.html.erb (hunk 4 above)
Comment (cwant) [reply app/views/items/_form.html.erb]: Got the query into the controller. Turns out there was a community model method to get child collections (called `member_collections` ... we should probably change it to `collections` at some point). I'll just point out that the situation is a bit more complicated than what you outlined, since an item can be in multiple community/collection pairs.

File: app/controllers/items_controller.rb [hunk 7]
```diff
@@ -1,6 +1,8 @@
 class ItemsController < ApplicationController
 
   before_action :load_item, only: [:show, :edit, :update]
+  before_action :load_communities, only: [:new, :create, :edit, :update]
```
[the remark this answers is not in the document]
murny: Thats not true on second look, need the create/update in case of the render of errors. 

But still theres some duplication here. Could just use `initialize_communities_and_collections` and remove `load commmunities` or we need do something else here to remove this setting of instance variable twice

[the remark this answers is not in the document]
cwant: The thing is though, if the item fails validation for :new or :edit, then these form valiables need to be set for :update or :create when it needs to render the form again (there is no redirect back to :new or :edit).

Yeah, I accidentaly left `@communities` in `initialize_communities_and_collections`, I had intended to pull that out.

Every form (whether it's a new item or an existing one) needs to load all of the communities. This could be tweaked so it's done a bit smarter. (e.g., only load the communities in :update, :create when the item validation fails).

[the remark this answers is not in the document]
murny: Yeah I looked closer and saw that. Added another comment but you probably didn't see.

Ah okay, if you pull it out then this is fine how it is. :+1: 

Mhmm, sure we can do something smarter as well. But thats okay.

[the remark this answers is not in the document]
cwant: Reduced it to one `before_action`, this should work a bit better.

app/assets/javascripts/jupiter/items.js
```diff
@@ -1,27 +1,60 @@
...
+    });
+});
 
-	$("#new_item .add-file").click(function(event){
-		event.preventDefault();
-		new_file_input();
-	});
+function update_collection_select($community_select) {
+  $collection_select = collection_select($community_select);
+  id =  $community_select.find('option:selected').val();
```
murny: need to `var` this or something? id is never defined.

app/assets/javascripts/jupiter/items.js
cwant: An alternative is to replace `$('#item-form')` with `$(document)` -- some containing element is needed so that the events get added to new DOM elements created via JS as the "+ Add Another" button is clicked. I figured that making that container only as big as it needed to be would be better.

murny: Well I think a class can be used instead of an ID right? `$('.js-item-form')`

cwant: Just pushed, I changed it to just 'form' (pretty much same effect).

app/assets/javascripts/jupiter/items.js
cwant: `var`-ed up a bunch of stuff in that file. Javascript is too forgiving some times.

PR #185, opened 2017-10-06
Facet admin communities

mbarnett:
Adds pagination to search results page, adds pagination and facets (unstyled) to Admin Communities page, renames search method to more clearly indicate it's only relevant when you want faceted results, adds various comments about future directions or other potentially needed tweaks.


app/controllers/admin/communities_controller.rb
```diff
@@ -3,7 +3,11 @@ class Admin::CommunitiesController < Admin::AdminController
   before_action :fetch_community, only: [:show, :edit, :update, :destroy]
 
   def index
-    @communities = Community.all
+    # anybody have a better idea for doing this? Lame to have to litter it everywhere
+    params[:facets].permit! if params[:facets].present?
```
murny: Don't think you need the if statement or the bang (which returns itself)?

Think you can just do this:

`params.permit [:facets]`

But I am maybe 80% sure.

As for a better way? Not sure... Looks fine for me, but yeah if this starts to repeat everywhere then you can encapsulate some of this in a `application_controller` helper method (which maybe does the params permit and makes the call to the search/page  and just returns the results object etc)

mbarnett: Let me try that, that might be a better way.

The problem is AFAIK they want to add facet selection in a few different places, so all of those actions would need this. BUT IDK maybe it's only 4 or 5 in practice.

mbarnett: No joy. Still throws `unable to convert unpermitted parameters to hash`.

I think the problem is, params[:facets] is itself a hash, and you don't know ahead of time what keys it will contain, so you can either use ! to permit anything in there, individually permit any possible facet name, or it explodes.

murny: Alright this is fine for now. We can revisit later

Issue #188, opened 2017-10-06
Refactor search into its own controller/style results


mbarnett:
Right now the main search runs through the items controller, which is misleading as it also searches Collections and Communities. This should be split out to its own controller.

Also the results page is mostly unstyled, and only titles are displayed. This needs UI love before anyone would consider it finished.

The upshot is that it all currently works, and any Item properties solrized for faceting will appear in the facet results on the side when there are results for that facet.

https://github.com/ualbertalib/jupiter/blob/8282084496e5c7274f76207e893e481960e5a9b5/app/controllers/items_controller.rb#L50-L57

PR #190, opened 2017-10-11
Part of Issue #188: move search to own controller.

cwant:
Move search to own controller. Restyling of seach/facets (the other part of #188) will be in a separate PR.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -8,7 +8,6 @@ class SearchController < ApplicationController
     @results = JupiterCore::Search.faceted_search(q: params[:search], facets: params[:facets],
                                                   models: [Item, Collection, Community], as: current_user)
     @results.sort(:title, :asc).page params[:page]
-    authorize Item
   end
 
 end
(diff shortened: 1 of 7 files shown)


app/controllers/search_controller.rb
```diff
@@ -2,4 +2,13 @@ class SearchController < ApplicationController
 
   skip_after_action :verify_authorized
 
+  def index
+    params[:facets].permit! if params[:facets].present?
+
+    @results = JupiterCore::Search.faceted_search(q: params[:search], facets: params[:facets],
+                                                  models: [Item, Collection, Community], as: current_user)
+    @results.sort(:title, :asc).page params[:page]
+    authorize Item
```
murny: Have to be careful here. This actually looks in the `ItemPolicy` class and looks at the `index?` action. 

There is no `index` action in the Items controller so I say we can just remove the ItemPolicy index action and also remove this line here for `authorize Item`?

Considering we got `skip_after_action :verify_authorized` at the top and search should always be true to everyone

cwant: Oops, yeah missed that. I'm surprised that the search still works despite the missing policy.

murny:
I wouldn't close the main #188 issue yet however? We need tests for this search controller and everything?

But that can be a seperate PR

Issue #153, opened 2017-09-19
Investigate: TravisCI being flaky with attempting to create database


murny:
Travis is randomly failing sometimes when it attempts to create/migrate the test database. Sometimes it succeeds, other times it fails. A quick restart fixes it...but it's getting silly. Let's investigate this:

Example build: https://travis-ci.org/ualbertalib/jupiter/builds/277470610

```
$ docker-compose run --rm -e RAILS_ENV=test web rails db:create
Starting jupiter_redis_1 ... 
ysql_1 ... 
ysql' (111)>
Couldn't create database for {"adapter"=>"mysql2", "encoding"=>"utf8", "pool"=>5, "username"=>"root", "password"=>"mysecretpassword", "host"=>"mysql", "database"=>"jupiter_test"}, {:charset=>"utf8"}
(If you set the charset manually, make sure you have a matching collation)
Created database 'jupiter_test'
```

```
$ docker-compose run --rm -e RAILS_ENV=test web rails db:migrate
Starting jupiter_redis_1 ... 
Starting jupiter_solr_1 ... 
ysql_1 ... 
ActiveRecord::NoDatabaseError: Unknown database 'jupiter_test'
....<< followed by a long stack trace and failed build >>....

```
```
The command "docker-compose run --rm -e RAILS_ENV=test web rails db:migrate" failed and exited with 1 during 
```

Probably due with the `rails db:create` instead of the `rails db:migrate`

Healthy builds `rails db:create` looks like this:
```
$ docker-compose run --rm -e RAILS_ENV=test web rails db:create
Starting jupiter_solr_1 ... 
Starting jupiter_redis_1 ... 
Starting jupiter_mysql_1 ... 
Created database 'jupiter_test'
```

Tried to up the wait time from `sleep 10` to `sleep 15` in thinking it was the database container not finish initializing but this doesn't seem to be helping.

Linking this one to #51

murny:
Closing issue for time being. Going to chalk this one up as hiccups with TravisCI (there were some delays/outages around this time: https://www.traviscistatus.com/incidents/4qj8t3220jy8)

Haven't seen a failed build in awhile

murny:
Reopening as this still seems to be an ongoing issue. Fails on the migration as test database wasn't found? Going to try to add a sleep between the create and migrate....

murny:
[...]

PR #196, opened 2017-10-12
Try to fix travis from tanking on db migrate?

murny:
Travis still failing on db migrate, says test database doesnt exist. DB create fails as well. Guessing DB container isnt fully initialized yet.

Maybe some extra sleep will help?

#153

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,13 @@ before_script:
   - docker-compose build web
   - docker-compose up -d web
 
-  - sleep 15 # wait for containers to finish setting up
+  - sleep 10 # wait for containers to finish setting up
   - docker ps
 
   - docker-compose run --rm -e RAILS_ENV=test web rails db:create
+
+  - sleep 5 # wait for database to finish setting up
+
   - docker-compose run --rm -e RAILS_ENV=test web rails db:migrate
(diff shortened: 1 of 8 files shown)


murny:
Hack for now...seems to fix it. If continues to happen, i say we rip this out and replace it with `docker-compose.lightweight`

As you said, better waiting an extra 15 secs then to manually restart the build all the time

murny:
Proper solution is now done. Brought in a wait-for-it script as suggested by official docs:
https://docs.docker.com/compose/startup-order/

screen shot 2017-10-12 at 11 21 17 am

Seems to fix the issue.

.travis.yml
```diff
@@ -10,10 +10,9 @@ before_script:
   - docker-compose build web
   - docker-compose up -d web
 
-  - sleep 15 # wait for containers to finish setting up
   - docker ps
 
-  - docker-compose run --rm -e RAILS_ENV=test web rails db:create
+  - docker-compose run --rm -e RAILS_ENV=test web ./wait-for-it mysql:3306 --timeout=0 --strict -- rails db:create
```
cwant: Maybe five minute timeout rather than no timeout?

PR #208, opened 2017-10-17
[WIP] Restructure/add user activity columns

### LINKED ISSUE #54: Update user stat columns after they log in
I left a bunch of stat columns for the user table here:

https://github.com/ualbertalib/jupiter/blob/master/db/migrate/20170714174108_create_users.rb#L11-L16

(functionality is exactly like devise's trackable functionality:
http://www.rubydoc.info/github/plataformatec/devise/master/Devise/Models/Trackable)

Need to implement this. Essentially pretty simple, when a user logs in successfully, you background a job (using something like sidekiq) that goes and updates all these columns (signin count gets incremeneted, dates get updated, ip address updated from the users request, etc)

This information is very informative to an admin user can be displayed on the users admin dashboard ( #40 )

cwant:
Update via ActiveJob when the user navigates Jupiter.

Fixes #54


app/models/user.rb
```diff
@@ -21,4 +21,19 @@ def items
     Item.where(owner: id)
   end
 
+  def update_activity!(remote_ip, now, sign_in: false)
+    raise InvalidParameter, :remote_ip if remote_ip.blank?
```
murny: is InvalidParameter defined? You typically have to define this (inherrited from StandardError or something) as this isnt a standard exception.
```
# When the input is somehow bad
  class Jupiter::InvalidParameters < StandardError; end
end
```

cwant: Interesting -- I wonder why it didn't barf on this then? (maybe because I've never run it without the correct args ...?)

cwant: I think `ArgumentError` should suffice.

app/controllers/application_controller.rb
```diff
@@ -31,6 +38,7 @@ def current_user
   def sign_in(user)
     @current_user = user
     session[:user_id] = user.try(:id)
+    UpdateUserActivityJob.perform_later(@current_user.id, Time.now.utc.to_s, request.remote_ip, sign_in: true)
```
murny: Might have to think about this. When you sign in, this job will get fired. Then the very next page view, current_user method will be called, which will fire off another job. Both will update the user table simultaneously.

Is this fine behavior? It's alright i guess? Can we improve this? Not sure? Perhaps doing this update synchronously (no job or use `perform_now` which uses a synchronous/blocking call instead)?  Then that second job won't get fired off (and no future jobs get fired off until after 5 minutes has passed)

cwant: I've switched that one to `perform_now`. Should be fine, we can reevaluate if it's not.

test/jobs/update_user_activity_job_test.rb
```diff
@@ -0,0 +1,76 @@
+require 'test_helper'
+
+class UpdateUserActivityJobTest < ActiveJob::TestCase
```
murny: Is a job test needed? The job itself is extremely simple. It simply just checks if params are correct, grabs the current user calls a model method `update_activity!` to do the updating. Majority of this stuff is probably already being tested in your integration tests. Or I think some good modelling testing of this `update_activity!` method is needed in the `test/models/user_test.rb` instead

cwant: Sure, can move to model tests. Ideally I would have liked to: 1) test in the model; 2) mock the model method (or the job) in the integration test to ensure that the model method (or the job) gets run. The mocking in minitest isn't as mature as it is in RSpec, so I find this sort of setup to be difficult.

murny: Shouldnt need to mock I think? Tests will run the jobs inline (perform_now). So jobs should be running and columns should be updating when you login/view pages etc? (may need to set the job queue adapter in the test.rb? or some other configuratio needed. But think thats done)

murny: Also if your doing things that will jobs and you want to see if jobs are enqueued and what not, activejob in test uses a simple mock array (instead of an actual redis queue). You can then use ActiveJob::TestHelpers to play around with this mock array

cwant: This is what I was doing for the integration test, but the 'perform_now' used when signing-in doesn't go through the queuing system at all (invisible to the test helpers).

murny: I think that's cause `perform_now` will not use the queue at all even in non test environment. It's ran synchronously.

cwant: Yes, absolutely -- it just means that there will be a lot of overlap between the model test and the integration test.

app/jobs/update_user_activity_job.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  queue_as :default
+
+  def perform(current_user_id, now, last_ip_address, sign_in: false)
+    raise ArgumentError, :now if now.blank?
+    raise ArgumentError, :current_user_id if current_user_id.blank?
+    raise ArgumentError, :last_ip_address if last_ip_address.blank?
+
+    current_user = User.find_by(id: current_user_id)
+    current_user.update_activity!(last_ip_address, now, sign_in: sign_in)
```
cwant: This is nasty: the job and the model method it wraps have their arguments in a different order.

test/jobs/update_user_activity_job_test.rb
murny: And that's okay. Model test is testing that the correct things are being updated according. Integration/controller test is testing more so the job is being enqueued only when last seen hasn't been updated in 5 minutes (that work flow). As well as signing in triggers the full update of all the sign in columns etc.

cwant: Okay, restructured tests.

PR #209, opened 2017-10-18
Update tooltip on deposit form (text from Leah via slack).

cwant:


config/locales/en.yml
```diff
@@ -254,7 +254,9 @@ en:
     create:
       created: 'Item was successfully created!'
     form:
-      community_collection_tooltip: 'TODO: add tooltip text'
+      community_collection_tooltip:
+        'Choose the Community and Collection that best match your item(s).
+        If you need a new collection, email erahelp@ualberta.ca'
```
murny: Wonder if making the email address a url (mailto link) or stand out with quotes or something be nice?

cwant: Currently the tooltip is via a title attribute on a span tag, so links/html don't render. We'd need to adopt a different solution to get clickable links to work.

murny: Ah this is true. This is fine how it is. Probably a bigger UX decision needed here. Like you can't even copy and paste this Email address in a tooltip.... As soon as you stop hovering its gone. Plus mobile? Can't hover on mobile..

Have a feeling if all this help text is this tiny. We should just display them underneath the form inputs as help text instead. 

http://getbootstrap.com/docs/4.0/components/forms/#help-text

Issue #216, opened 2017-10-23
Consolidate all the env variables to be managed by secrets.yml


weiweishi:
Similarly to HN, we will consolidate all the environment variables to be managed by secrets.yml for staging/production environments, so that secrets.yml will be the SSOT for variable values that need to be managed/overwritten by Ansible/deployment process.

PR #218, opened 2017-10-24
Use secrets.yml to manage env var in deployment

weiweishi:
Address issue #216

### DIFF
--- a/config/fedora.yml
+++ b/config/fedora.yml
@@ -1,26 +1,16 @@
 # Needed and used by active fedora, to specify where to find fedora
+default: &default
+  user: <%= Rails.application.secrets.fedora_user %>
+  password: <%= Rails.application.secrets.fedora_user %>
+  url: <%= Rails.application.secrets.fedora_user %>
+  base_path: <%= Rails.application.secrets.fedora_user %>
 development:
-  user: fedoraAdmin
-  password: fedoraAdmin
-  url: <%= ENV['FCREPO_URL'] || 'http://localhost:8080/fcrepo/rest' %>
-  base_path: /dev
+  <<: *default
 test:
-  user: fedoraAdmin
-  password: fedoraAdmin
-  url: <%= ENV['FCREPO_URL'] || 'http://localhost:8080/fcrepo/rest' %>
-  base_path: /test
+  <<: *default
 uat:
-  user: <%= ENV['FCREPO_USER'] %>
-  password: <%= ENV['FCREPO_PASSWORD'] %>
-  url: <%= ENV['FCREPO_URL'] %>
-  base_path: /uat
+  <<: *default
 staging:
-  user: <%= ENV['FCREPO_USER'] %>
-  password: <%= ENV['FCREPO_PASSWORD'] %>
-  url: <%= ENV['FCREPO_URL'] %>
-  base_path: /staging
+  <<: *default
 production:
-  user: <%= ENV['FCREPO_USER'] %>
-  password: <%= ENV['FCREPO_PASSWORD'] %>
-  url: <%= ENV['FCREPO_URL'] %>
-  base_path: /prod
+  <<: *default
--- a/config/fedora.yml
+++ b/config/fedora.yml
@@ -1,9 +1,9 @@
 # Needed and used by active fedora, to specify where to find fedora
 default: &default
-  user: <%= Rails.application.secrets.fedora_user %>
-  password: <%= Rails.application.secrets.fedora_user %>
-  url: <%= Rails.application.secrets.fedora_user %>
-  base_path: <%= Rails.application.secrets.fedora_user %>
+  user: <%= Rails.application.secrets.fcrepo_user %>
+  password: <%= Rails.application.secrets.fcrepo_password %>
+  url: <%= Rails.application.secrets.fcrepo_url %>
+  base_path: <%= Rails.application.secrets.fcrepo_basepath %>
 development:
   <<: *default
 test:
--- a/config/fedora.yml
+++ b/config/fedora.yml
@@ -3,7 +3,7 @@ default: &default
   user: <%= Rails.application.secrets.fcrepo_user %>
   password: <%= Rails.application.secrets.fcrepo_password %>
   url: <%= Rails.application.secrets.fcrepo_url %>
-  base_path: <%= Rails.application.secrets.fcrepo_basepath %>
+  base_path: <%= Rails.application.secrets.fcrepo_base_path %>
 development:
   <<: *default
 test:
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,7 +1,7 @@
 require 'yaml'
 require 'erb'
-secrets = YAML.load_file("config/secrets.yml")
-redis_url = ERB.new(secrets["shared"]["redis_url"]).result(binding)
+secrets = YAML.load_file('config/secrets.yml')
+redis_url = ERB.new(secrets['shared']['redis_url']).result(binding)
 Sidekiq.configure_server do |config|
   config.redis = { url: redis_url }
 end
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,11 +1,7 @@
-require 'yaml'
-require 'erb'
-secrets = YAML.load_file('config/secrets.yml')
-redis_url = ERB.new(secrets['shared']['redis_url']).result(binding)
 Sidekiq.configure_server do |config|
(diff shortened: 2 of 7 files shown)


config/fedora.yml
```diff
@@ -1,26 +1,16 @@
 # Needed and used by active fedora, to specify where to find fedora
+default: &default
+  user: <%= Rails.application.secrets.fcrepo_user %>
+  password: <%= Rails.application.secrets.fcrepo_password %>
+  url: <%= Rails.application.secrets.fcrepo_url %>
+  base_path: <%= Rails.application.secrets.fcrepo_basepath %>
```
murny: Keep it `base_path` in the secrets? aka `Rails.application.secrets.fcrepo_base_path`

config/secrets.yml
```diff
@@ -17,7 +17,10 @@
 
 # Environmental secrets are only available for that specific environment.
 
-development:
+shared:
```
murny: Nice find and usage of `shared` :+1:

config/secrets.yml
```diff
@@ -69,3 +106,12 @@ production:
...
   saml_idp_sso_target_url: 'https://login.ualberta.ca/saml2/idp/SSOService.php'
+
+  database_url: <%= ENV['DATABASE_URL'] %>
+  fcrepo_user: <%= ENV['FCREPO_USER'] %>
+  fcrepo_password: <%= ENV['FCREPO_PASSWORD'] %>
+  fcrepo_url: <%= ENV['FCREPO_URL'] %>
+  fcrepo_basepath: /prod
+  solr_url: <%= ENV['SOLR_URL'] %>
+  redis_url: <%= ENV['REDIS_URL'] %>
```
murny: usage of ENV['REDIS_URL']l hasn't been consolidated to this file yet? There's a redis and sidekiq initializer that are still using the env var.

config/secrets.yml
```diff
@@ -69,3 +106,12 @@ production:
...
+  sidekiq_logfile: log/sidekiq_prodution.log
```
murny: spelling mistake on `production`

config/secrets.yml
weiweishi: In docker-compose, sidekiq workers are running in a different container - so it doesn't have access to Rails.application.secrets. wonder what would be the best way to handle that?

weiweishi: keep getting 
```
worker_1    | uninitialized constant Rails
worker_1    | (erb):3:in `<main>'
worker_1    | /usr/local/lib/ruby/2.3.0/erb.rb:864:in `eval'
worker_1    | /usr/local/lib/ruby/2.3.0/erb.rb:864:in `result'
worker_1    | /usr/local/bundle/gems/sidekiq-5.0.5/lib/sidekiq/cli.rb:388:in `parse_config'
worker_1    | /usr/local/bundle/gems/sidekiq-5.0.5/lib/sidekiq/cli.rb:225:in `setup_options'
worker_1    | /usr/local/bundle/gems/sidekiq-5.0.5/lib/sidekiq/cli.rb:39:in `parse'
worker_1    | /usr/local/bundle/gems/sidekiq-5.0.5/bin/sidekiq:11:in `<top (required)>'
worker_1    | /usr/local/bundle/bin/sidekiq:17:in `load'
worker_1    | /usr/local/bundle/bin/sidekiq:17:in `<top (required)>'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli/exec.rb:74:in `load'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli/exec.rb:74:in `kernel_load'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli/exec.rb:27:in `run'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli.rb:362:in `exec'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli.rb:22:in `dispatch'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/cli.rb:13:in `start'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/exe/bundle:30:in `block in <top (required)>'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/lib/bundler/friendly_errors.rb:121:in `with_friendly_errors'
worker_1    | /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.15.4/exe/bundle:22:in `<top (required)>'
worker_1    | /usr/local/bin/bundle:23:in `load'
worker_1    | /usr/local/bin/bundle:23:in `<main>'
jupiter_worker_1 exited with code 1
```

murny: Mmm yeah this might be a problem. Not sure a good workaround is. Other then using ENV vars for sidekiq :neutral_face:. Can at least use it for the redis intializer i think?

The config file for sidekiq `config/sidekiq.yml` had no issues though? I'm sure this will crash as well.

Sidekiq is also new and needs requirements for deployment itself and this should be taken into consideration.

config/initializers/sidekiq.rb
```diff
@@ -1,7 +1,11 @@
+require 'yaml'
```
weiweishi: I guess we could do something like this. It's a bit of clunky but at least it will allow us to continue to use secrets.yml as our SSOT.

config/secrets.yml
weiweishi: I've modified config/initializers/sidekiq.rb to read values from secrets.yml and I think the config/sidekiq.yml can stay as-is .  
Question for sidekiq: 1. does it need its own separate server? 2. if it's on a separate server, what do we need on the server? And I assume we need to run sidekiq on the server as this:  `RAILS_ENV={rails_env} bundle exec sidekiq`?

config/initializers/sidekiq.rb
murny: Yes this works but is pretty hacky. Sidekiq shouldn't need to know all this. 

Maybe jupiter is a good time to restart this conversation of using environment variables. You pass in one ENV var to sidekiq... instead of tip toeing and tripping all over this mess by doing something like this. Ansible supports ENV extremely well right out of the box. They are using ENV vars in other places anways. Why are we continue doing silly stuff ? (we shouldnt be overwritting any code from ansible, even if its this one single file)? 

Also I should mention secrets.yml is going away. This will force us down a very different path. 

Rails 5.1 already offers encrypted secrets.yml. This is the rails version we are on currently. So why we still overwriting this file? Lets encrypt it instead and leave it as is. More info here: https://www.engineyard.com/blog/encrypted-rails-secrets-on-rails-5.1

In Rails 5.2 the scene gets even more different.
https://github.com/rails/rails/pull/30067

Some key take aways:

- `config/secrets.yml` becomes `config/credentials.yml.enc`
- This new file is a flat format, not divided by environments, like secrets.yml has been.
- RAILS_MASTER_KEY and then simply rely on a single master key that can live in that ENV or in config/master.key

weiweishi:
now replaced the initialiers/sidekiq.rb and I think at least this will enable us to move forward with the staging env setup. will bring up the credentials.yml.enc in Rails 5.2 with @henryzhang87 so that we need to keep that in our radar to replace our current practices to set all the env vars.

murny:
Looks good. 

Only one left I think is the [config/initializers/redis.rb](https://github.com/ualbertalib/jupiter/blob/master/config/initializers/redis.rb)

Can just do something like:

```
Redis.current = Redis.new(url: Rails.application.secrets.redis_url)
```

I think?

weiweishi:
yes, thank you. it was changed and tested, and then removed as I thought it was related to the sidekiq config issue. Put it back now.

PR #225, opened 2017-10-30
Issue #188 restyle search

cwant:
Note: Facets pulled out of Admin community page (wireframe shows autocomplete, not faceted search).
Note: how facets behave on mobile will be handled in #230.

Includes filtering, pagination and sorting of search results. System tests.

Seeds added to add creators, contributors, language to the Item seeds to demonstrate faceting for those things.

Adding sorting for titles to Community/Collection model, so you may been to reindex at a console:

Community.all.each { |o| o.unlock_and_fetch_ldp_object {} }
Collection.all.each { |o| o.unlock_and_fetch_ldp_object {} }

(Probably won't need to do that if you re-seed the database.)

Screenshots:

localhost_3000_search_search cats

localhost_3000_search_direction desc search of sort title


app/assets/javascripts/jupiter/search.js
```diff
@@ -0,0 +1,8 @@
+// Toggle between 'Show' and 'Hide' link for collapsible
+$(document).on('turbolinks:load', function() {
+  $('.hideshow-control')
```
murny: Namespace this class to indicate javascript usage `.js-*`

http://codeguide.co/#css-classes

app/controllers/search_controller.rb
```diff
@@ -1,13 +1,34 @@
...
-    @results = JupiterCore::Search.faceted_search(q: params[:search], facets: params[:facets],
-                                                  models: [Item, Collection, Community], as: current_user)
-    @results.sort(:title, :asc).page params[:page]
+    @max_facets = MAX_FACETS
+    @active_tab = params[:tab]&.to_sym || :item
+    @results = {}
+
+    # TODO: Likely we want to do one search and segregate the results by model
+    [:item, :collection, :community].each do |model|
+      # Only facet for the current tab or the result count will be wrong on the tab header
```
murny: We could also separate this into different views if easier. For example:

- `/search` is the default (which searchs items)
- `/search/communities` is the community tab which search for communities
- `/search/collections` is the collection tab which search for collections

Then you can clean up all this logic and move them into their own views/actions/etc.

As searching each model in the index would probably be a performance issue im sure once we start having a ton of records. Also this preserves state nicely. You can send a user to `/search/communities` for a list of communites etc.

cwant: Could do, but you don't get the hit counts in the tab if you do that (which is the only reason the search on the non-active model is being run). I'm fine either way, but I think from a UX standpoint it's useful to see the counts.

test/system/search_test.rb
```diff
@@ -0,0 +1,325 @@
+require 'application_system_test_case'
+
+class SearchTest < ApplicationSystemTestCase
+
+  def before_all
+    super
+    @community = Community.new_locked_ldp_object(title: 'Fancy Community', owner: 1)
+                          .unlock_and_fetch_ldp_object(&:save!)
+    @collections = (0..1).map do |i|
```
murny: Might be a style thing but why the range stuff everywhere? Why not just do `2.times` instead of `(0..1)`? Same result just don't have to guess...does 0..9 give me 10 or 9 results? etc.

app/controllers/search_controller.rb
murny: Yeah we will see once we start getting records in here and how the search performs. Future investigation needed.

test/system/search_test.rb
cwant: I agree, where reasonable. Unfortunately, our ol' pal Batsov says:

```
test/system/search_test.rb:9:20: C: Performance/TimesMap: Use Array.new(2) with a block instead of .times.map.
    @collections = 2.times.map do |i| ...
```
I would rather use a range than use `Array.new(2)`.

app/controllers/search_controller.rb
mbarnett: Yeah this is probably fine as is but maybe add another TODO just to come back and check the performance of this when we have a load more models?

test/system/search_test.rb
mbarnett: I vote we just turn this cop off, because that's a pretty awful recommendation

murny: Note that the rubocop is a performance instead of a style recommendation? If your N is a big number I'm guessing this performs horribly compared to intializing a empty array. But yeah turn this of for now.

mbarnett: Yeah if you're doing it for 10_000.times.do or something, you've definitely gone wrong somewhere. 2.times.do seems clearer than Array.new(2).map, which is just confusing looking.

Basically this cop needs to distinguish between small numbers and big ones or something. Kind of goofy as is.

app/assets/javascripts/jupiter/search.js
cwant: Done.

cwant:
If you mean the bug I was discussing with Weiwei, that's something else currently in master, not part of this PR. I can create a ticket for it (on the admin communities page, the collection collapse/expand is throwing 500 here).

mbarnett:
Ah gotcha. Is this good to go, then? I can give it a final lookover now if so

cwant:
Yep, good to go.

PR #226, opened 2017-10-31
Support for headless system tests via PhantomJS.

cwant:
Enabled via the env variable CAPYBARA_PHANTOMJS.
Does not disrupt work space while running and runs a lot faster too (13s for phantom vs 23s for selenium).

One issue though: during the run, the terminal is littered with the text "Viewport argument key "shrink-to-fit" not recognized and ignored."

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - docker-compose run --rm -e RAILS_ENV=test web rails test
-  - docker-compose run --rm -e RAILS_ENV=test -e CAPYBARA_NO_HEADLESS=true web rails test:system
+  - docker-compose run --rm -e RAILS_ENV=test web rails test:system
   - docker-compose run --rm web rubocop
   # - docker-compose run --rm web scss-lint
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - docker-compose run --rm -e RAILS_ENV=test web rails test
-  - docker-compose run --rm -e RAILS_ENV=test web rails test:system
+  - docker-compose run --rm -e RAILS_ENV=test CAPYBARA_NO_HEADLESS=true web rails test:system
   - docker-compose run --rm web rubocop
   # - docker-compose run --rm web scss-lint
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - docker-compose run --rm -e RAILS_ENV=test web rails test
-  - docker-compose run --rm -e RAILS_ENV=test CAPYBARA_NO_HEADLESS=true web rails test:system
+  - docker-compose run --rm -e RAILS_ENV=test -e CAPYBARA_NO_HEADLESS=true web rails test:system
   - docker-compose run --rm web rubocop
   # - docker-compose run --rm web scss-lint
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - docker-compose run --rm -e RAILS_ENV=test web rails test
-  - docker-compose run --rm -e RAILS_ENV=test web rails test:system
+  - docker-compose run --rm -e RAILS_ENV=test -e CAPYBARA_NO_HEADLESS=true web rails test:system
   - docker-compose run --rm web rubocop
   # - docker-compose run --rm web scss-lint
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,4 +75,4 @@ services:
     ports:
       - "3000:3000"
     environment:
-      CAPYBARA_NO_HEADLESS: true
+      CAPYBARA_NO_HEADLESS: 'true'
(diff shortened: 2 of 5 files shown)


murny:
I think phantomjs is dead no? Did you look into Headless Chrome or anything?  Majority of projects have already switched already

murny:
Apparently headless chrome can be added just with a simple config change:

```ruby
require "test_helper"

class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
  driven_by :selenium, using: :chrome, screen_size: [1400, 900], options: { args: %w[headless disable-gpu] }
end```

**Comment by cwant** [comment]
Merde, okay, seems I need to keep track of what projects are dying each month.

Will try the chrome flag for headless.

**Comment by murny** [comment]
I'd just switch everything over to headless (as the default etc). If you can get the headless chrome thing to work fine

**Comment by cwant** [comment]
Unfortunately the config change was not that easy. Pushed a revised version using chrome headless as default (can enable the on-screen browser with env variable 'CAPYBARA_NO_HEADLESS'). About 17s per run here.

File: test/application_system_test_case.rb [hunk 1]
```diff
@@ -1,12 +1,26 @@
 require 'test_helper'
+require 'selenium-webdriver'
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
+  if ENV['CAPYBARA_NO_HEADLESS']
```
murny: Is there any reason to run this not headless? There's really no gains here other then your browser being highjacked and seeing screens go by extremely quick which is silly.

cwant: Currently travis doesn't like headless (don't know why).

test/application_system_test_case.rb
```diff
@@ -1,12 +1,26 @@
...
+    driver = :selenium
+  else
+    Capybara.register_driver :selenium_chrome_headless do |app|
```
murny: Do you have to create this yourself?

Capybara already registers this exact driver itself with the same config:
https://github.com/teamcapybara/capybara/blob/6517d2e59b50b4270f82cdf349e09cb63bf81a5a/lib/capybara.rb#L489-L494

cwant: Works here now. I'll send another test to see if travis likes it (maybe it was the gpu stuff that was causing the issues there).

cwant:
Looks like Travis doesn't like the non-headless. There is some documentation for getting it to work with the Travis images, but we are using docker. Some info here:

https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-the-Chrome-addon-in-the-headless-mode

test/application_system_test_case.rb
```diff
@@ -1,12 +1,18 @@
 require 'test_helper'
+require 'selenium-webdriver'
```
murny: Is this still needed? now that you removed the custom driver implementation?

test/application_system_test_case.rb
```diff
@@ -1,12 +1,18 @@
...
+           else
+             :selenium_chrome_headless
+           end
+
   # Set options if you have a special selenium url (like if your running selenium in a docker container)
   # Otherwise just use the defaults by providing empty hash
   options = ENV['SELENIUM_URL'].present? ? { url: ENV['SELENIUM_URL'] } : {}
-
-  driven_by :selenium, using: :chrome, screen_size: [1400, 1400], options: options
+  driven_by driver, using: :chrome, screen_size: [1400, 1400], options: options
```
murny: Let's wrap this in the `if ENV conditon`?

```ruby
if ENV['CAPYBARA_NO_HEADLESS']
  options = ENV['SELENIUM_URL'].present? ? { url: ENV['SELENIUM_URL'] } : {}
  driven_by :selenium, using: :chrome, screen_size: [1400, 1400], options: options
else
  driven_by :selenium_chrome_headless
end
```
Or something?

test/application_system_test_case.rb
cwant: Yes. Believe it or not (and it surprised me), when I take it out, the thing fails.

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_script:
 
 script:
   - docker-compose run --rm -e RAILS_ENV=test web rails test
-  - docker-compose run --rm -e RAILS_ENV=test web rails test:system
+  - docker-compose run --rm -e RAILS_ENV=test -e CAPYBARA_NO_HEADLESS=true web rails test:system
```
murny: Could throw this in the` docker-compose.yml` instead ¯\\_(ツ)_/¯?

docker-compose.yml
```diff
@@ -74,3 +74,5 @@ services:
     command: bash -c 'rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b "0.0.0.0"'
     ports:
       - "3000:3000"
+    environment:
+      CAPYBARA_NO_HEADLESS: 'true'
```
murny: This is overwriting the environment variables from worker/rails. Just move this up there and all should work

cwant: Ah, okay, I thought maybe it was due to the YAML formatting.

PR #227, opened 2017-10-31
Revert "Support for headless system tests via PhantomJS."

cwant:
Reverts ualbertalib/jupiter#226


cwant:
Naw, let's not do this.

PR #233, opened 2017-11-03
Adding sort, pagination to communities & collections

cwant:
Pagination and sort links added for both user and admin community and collections pages.
Faceting removed out of admin communities page.

Seeds added to supply enough communities to properly demonstrate pagination. The community 'The Everything Department' had a bunch of empty collections to demonstrate pagination there.

Adding sorting for titles to Community/Collection model, so you may been to reindex at a console:

Community.all.each { |o| o.unlock_and_fetch_ldp_object {} }
Collection.all.each { |o| o.unlock_and_fetch_ldp_object {} }

NOTE: Pagination of items within a collection will be handled elsewhere (this involves faceting, so it is more closely aligned to the Facet restyling PR).

System tests done.


app/controllers/concerns/paginate_and_sort_methods.rb
```diff
@@ -0,0 +1,13 @@
+module PaginateAndSortMethods
+  extend ActiveSupport::Concern
```
mbarnett: How do you feel about moving these methods into ApplicationController and sharing them via inheritance instead of by composing concerns?

cwant: Could do, it's not a big deal either way (although the "I" in "SOLID" says I shouldn't).

mbarnett: I can see that too. I think it's a frequency thing -- if almost every controller is sorting and paginating, than there's no gain to SOLID dogmatism because nearly every controller is expected to exhibit the same interface -- all we gain is a magic incantation that gets copied and pasted everywhere. If it's a narrower concern, than a concern is the logical fit. 

I'm not going to die on this hill or anything but to me it looks like the only non-paginating/sorting holdout right now is likely `sessions_controller`, which is always going to be a weird case.

cwant: No worries, I moved them.

app/helpers/application_helper.rb
```diff
@@ -25,4 +25,11 @@ def facetable_query_params(facet_name, value)
   def help_tooltip(text)
     content_tag(:span, fa_icon('question-circle'), title: text)
   end
+
+  def results_range(results)
```
murny: Nice! Does this work with active record kaminari? Or just ActiveFedora?

test/system/collections_pagination_and_sort_test.rb
```diff
@@ -0,0 +1,107 @@
+require 'application_system_test_case'
+
+class CollectionsPaginationAndSortTest < ApplicationSystemTestCase
+
+  def before_all
+    super
+    @community = Community.new_locked_ldp_object(title: 'Community', owner: 1).unlock_and_fetch_ldp_object(&:save!)
+    # For sorting, creation order is 'Fancy Collection 00', 'Nice Collection 01', 'Fancy Collection 02', etc. ...
+    (0..25).each do |i|
```
murny: Is creating 25 communities super slow or anything?  Or not so bad?

As an aside I'm sure we will tweak 25 per page limit to something lower. You can also do this in here just to show the pagination with a smaller number

test/system/collections_pagination_and_sort_test.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  context 'Non-logged in user (or anybody really)' do
+    should 'be able to sort and paginate collections' do
+      visit community_path(@community)
+      assert_selector('div', text: '1 - 25 of 26')
+      # Default sort is by title. First 13 say 'Fancy', last 12 say 'Nice'
+      # Below: /Fancy Collection 00.*Fancy Collection 02.*<SNIP>.*Nice Collection 21.*Nice Collection 23/
+      a_to_z_match = Regexp.new(((0..12).map { |i| format('Fancy Collection %02i', 2 * i) } +
```
murny: I assuming this is checking if every single title is on the page and in the correct order? That a bit extreme?  Can probably just check if the first element is Fancy Collection 00 and the last element is Nice collection 23. Instead of all this looping/mapping.

Can use the selectors `:first-child` and `:last-child` to help you find the first and last element in the list.

app/helpers/application_helper.rb
cwant: Works with either (just tried it on the User page, works).

test/system/collections_pagination_and_sort_test.rb
cwant: The difference for and extra 26 communities and 26 collections is about 5 seconds (40s seed on master, 45s seed on this branch).

I'll tone down the page limit to 10 for both communities and collections (this will likely be closer to the final value than the kaminari default of 25).

cwant: Oops, I was referring to seeding, not the test suite. Will tone down the pagination anyways.

db/seeds.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
murny: editor added this or something?

This check makes sense only in Ruby 1.9, since in 2.0+ utf-8 is the default source file encoding.

test/system/communities_pagination_and_sort_test.rb
```diff
@@ -0,0 +1,102 @@
...
+    super
+    # For sorting, creation order is 'Fancy Community 00', 'Nice Community 01', 'Fancy Community 02', etc. ...
+    (0..25).each do |i|
+      Community.new_locked_ldp_object(title: format("#{['Fancy', 'Nice'][i % 2]} Community %02i", i), owner: 1)
+               .unlock_and_fetch_ldp_object(&:save!)
+    end
+  end
+
+  context 'Non-logged in user (or anybody really)' do
+    should 'be able to sort and paginate communities' do
```
murny: The other comment i got for your test is to break them up into smaller sub tests? This test is almost 80 lines?

Have one for sorting and one for paginate? Or Break up each of the sorts? etc?

test/system/collections_pagination_and_sort_test.rb
cwant: Reworked it (using the selectors too). I check the first two/last two, since I don't think just first/last is enough for validating sorting.

db/seeds.rb
cwant: Yeah, emacs annoyance. Will remove.

app/models/collection.rb
```diff
@@ -1,5 +1,7 @@
 class Collection < JupiterCore::LockedLdpObject
 
+  COLLECTIONS_PER_PAGE = 10
```
murny: This might be a good default (10 or 15) in general for kaminari. You can create an intializer so everything gets this default instead of configuring this per model.

https://github.com/kaminari/kaminari#configuring-kaminari

`rails g kaminari:config`
`default_per_page 10`

cwant: Done.

PR #239, opened 2017-11-10
[TODO] Collection communities dropdown

murny:
This PR is a WIP for adding communties dropdown to the new/edit collection screens.

Currently parking this functionality as is. Will need to come back and revisit.

This works. You can change a community for a collection without problem in the new/edit screens. However there is a big issue here and that is in regard to how items are dependent on both the community and collection. Currently when making this change, all previous items depending to a collection become orphaned, as they are still referencing the previous community/collection pair. So we need someway of updating all items in collection to the new community/collection pair. This needs figuring out. Commented out section in PR indicates this.

### DIFF
--- a/app/controllers/admin/collections_controller.rb
+++ b/app/controllers/admin/collections_controller.rb
@@ -1,6 +1,9 @@
 class Admin::CollectionsController < Admin::AdminController
 
+  # TODO: shouldn;t need to fetch community on every action?
+  # Should be able to retrieve this via `Collection.community` in most places
   before_action :fetch_community
+
   before_action :fetch_collection, only: [:show, :edit, :update, :destroy]
 
   def show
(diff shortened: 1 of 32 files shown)


app/controllers/admin/collections_controller.rb
```diff
@@ -10,29 +13,46 @@ def show
...
     @collection.unlock_and_fetch_ldp_object do |unlocked_collection|
       if unlocked_collection.update(permitted_attributes(Collection))
-        redirect_to admin_community_collection_path(@community, @collection), notice: t('.updated')
+        # TODO: Need to update every item in the collection to the new community and collection pair
+        # How to do this? Remember items can belong to multiple collections.
+        # if @collection.community.id != @community.id
+        #   @collection.member_items.each do |item|
+        #     item.unlock_and_fetch_ldp_object do |unlocked_item|
+        #       # Not the right call. Need to figure this out
+        #       # unlocked_item.update_communities_and_collections(@collection.community.id, @collection.id)
```
murny: Need to figure out this logic.  Need to go through every item in the collection and replace the preexisting community/collection with the new community/collection. You can't just replace the entire member path as items can belong to multiple collections.

app/controllers/admin/collections_controller.rb
```diff
@@ -10,29 +13,46 @@ def show
...
+    @communities = Community.all
+  end
 
   def update
     @collection.unlock_and_fetch_ldp_object do |unlocked_collection|
       if unlocked_collection.update(permitted_attributes(Collection))
-        redirect_to admin_community_collection_path(@community, @collection), notice: t('.updated')
+        # TODO: Need to update every item in the collection to the new community and collection pair
+        # How to do this? Remember items can belong to multiple collections.
+        # if @collection.community.id != @community.id
```
murny: Only need to do this if community has changed

mbarnett:
Closing this since it's older than

PR #249, opened 2017-11-10
Fix for #246 -- make admin communities 'Collections' buttons work again.

### LINKED ISSUE #246: Button to expand collection on admin community index page broken
The work is related to #229 and once it's fixed the function can be ported over

cwant:
The pagination stuff for collections broke the lightweight collapsible collection list on the admin communities page. This PR fixes that by breaking the collection lists into two partials (with/without pagination). System test added.

Fixes #246


app/views/communities/show.html.erb
```diff
@@ -49,7 +49,7 @@
 <div class="row mt-3">
   <div class="col-md-12">
     <ul class="list-group">
-      <%= render partial: 'communities/collections_list' %>
+      <%= render partial: 'communities/collections_list_sort_paginate' %>
```
murny: I don't know if creating yet another partial is the solution here? 

Previous partial was used in multiple places (show.js and show.html)? Now we are replacing those multiple places with each having their own partial. 

So question is? Does partial still make sense here? They never being reused. Or can we somehow pull the paginate/sort/search out of this partial so that it can be reused once again between show.js and show.html?

cwant: I'm personally indifferent, although I think sometimes adding a partial as a semantic break point can help a cluttered view, whether it's reused or not.

If it's important to you, I can remove the partials and move the markup to the page templates (no biggie either way, really).

PR #264, opened 2017-11-15
Item view in search. Includes:

### LINKED ISSUE #244: Expanding how items are presented in search results.
Currently each matching item is a list item with just the item title. The current wireframe looks like this:

Search Mockup
https://jupiter.mybalsamiq.com/projects/jupiter/Search

DONE WHEN
- [ ] Each item search result has a thumbnail (a placeholder at least)
- [ ] Each item search result has a download button (a placeholder at least)
- [ ] Each item search result lists the title, author(s), year and abstract

cwant:
* Each item hit includes thumbnail, list of creators, description, year, and download link
  * Each creator listed in result links to search for that creator
  * Change creator/contributor to multi-value in item model
  * Add `created` attribute predicate to item model, and derived `sort_year`
  * Convert search results to bootstrap `list-group` and update tests
  * Non-functional download link with icon (real link after fileset work)
  * Update seeds so some have multiple creators, some have long descriptions.
  * Change item policy so that only admin can delete items.
  * Add edit/delete buttons to item results that satisfy policy
  * Make item thumbnail use same partial as community/logo
  * Make multi-valued creator/contributor work on item form
  * Make download button in search results only appear if files are present and point it to the correct download link.

Here is a screenshot:

new_item_search_screenshot

Fixes #244


app/helpers/items_helper.rb
```diff
@@ -0,0 +1,6 @@
+module ItemsHelper
+  def creator_search_path(creator)
+    # TODO: the search path may need to be revisited
+    search_path(search: "creator_tesim:#{creator}")
```
mbarnett: Right after coffee, let me slap together a method so that you can fish this out of the `Item` model without having to hardcode the solr mangling. I've had to do this in a couple of places already, so we're past due for a proper API

app/assets/stylesheets/items.scss
```diff
@@ -1,8 +1,23 @@
...
   }
 }
+
+// For item thumbnails on the search results pages
+.item-thumbnail-small {
+  height: 150px;
+  width: 120px;
+}
+
+// TODO: remove me after real thumbnails are sorted out
```
murny: Probably out of scope but we should have a fallback image though just like communities (which just shows the font awesome icon for a picture). As some items may not generate a thumbnail for whatever reason and will need to fallback to something. So we should take what community did and generalize this for placeholder thumbnails in jupiter type of thing etc.

weiweishi: yes, agreed. I will add a ticket to track this.

cwant: Removed the CSS thingy and have a placeholder partial for a thumbnail that reuses the font-awesome icon. (The partial assumes activestorage for the thumbnail, but it doesn't matter much for now as that part of the code will currently never get hit.)

cwant:
Screenshot updated to reflect the fallback image.

This PR is currently blocked by PR #273 (or this one could be approved as-is, and that PR can be redone against master).

This PR is currently blocking PR #271.

PR #266, opened 2017-11-15
Add MIT license

pbinkley:
MIT license is approved for DI work on Jupiter.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 University of Alberta Libraries
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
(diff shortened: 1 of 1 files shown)


cwant:
I would remove the "MIT License" line from the top of the file, as I think it affects how Github parses the file.  Compare for example the view of your license file, and an MIT Licence from another project:

https://github.com/ualbertalib/jupiter/blob/eafc5b839f3fb447ac3f89fd21fb2b21e4c7f5ff/LICENSE

https://github.com/mitmproxy/mitmproxy/blob/master/LICENSE

Github has added a summary of rights permissions/restrictions for the other file that isn't present in your file (and the only difference I can see is that first "MIT License" line).

cwant:
Actually, it's quite possible the following is needed to add a license after the fact:

https://help.github.com/articles/adding-a-license-to-a-repository/

pbinkley:
I used the method Chris linked to, so that first line is part of the template Github provides. Maybe that summary only gets displayed when the license is in master? Github wants to display "how repositories on GitHub are licensed", so a license file in a non-master branch might not trigger their display. https://github.com/blog/2252-license-now-displayed-on-repository-overview . Why don't we merge and see how it looks, and fix it if it still doesn't show.

cwant:
Yeah, merging into master did the trick.

Issue #229, opened 2017-11-02
Community page (users)


cwant:
The wireframes are here:

![Browse Mockup](https://jupiter.mybalsamiq.com/mockups/6035142.png)
https://jupiter.mybalsamiq.com/projects/jupiter/Communities%20and%20Collections

A lot could be taken from (or shared with the) admin community page:

https://jupiter.mybalsamiq.com/projects/jupiter-admin/Communities%20and%20Collections

DONE WHEN:

- [x] The list of communities paginates
- [x] Sorting collections A-Z and Z-A works
- [x] A button toggles the display of a collection list for each community shown #246 

Near future work:
- [x] Auto-complete partial matches communities/collections as shown #242

cwant:
The third item in the list is blocked by PR #246 (after that is merged, the collection dropdown can be copied from admin communities).
The last item (autocomplete) can be worked on separately (covered in Issue #242)

cwant:
Last checkmark filled in, closing.

PR #272, opened 2017-11-17
Port over the collection list toggle from admin/communities to communities

cwant:
A bunch of 'admin' stuff was moved to regular jupiter.
This fills in a checkbox for Issue #229


app/controllers/admin/communities_controller.rb
```diff
@@ -9,7 +9,9 @@ def index
   def show
     respond_to do |format|
       format.js do
+        # Used for the collapsable dropdown to show member collections
         @collections = @community.member_collections
+        render template: 'communities/show'
```
murny: Other option is just consume the non-admin js action? Then this format.js isn't needed. 

But I guess this also potentially future proofs us in case maybe one day they diverge or admin collection show has something special that the regular non community doesn't.
¯\\_(ツ)_/¯

app/views/communities/index.html.erb
```diff
@@ -36,6 +36,23 @@
                  <p class="text-muted"><%= community.description %></p>
```
murny: We should remove the admin/communities/index and just consolidate everything into the communities/index? With an overridable breadcrumb just like community details/collection details. As they should be pretty much the same

cwant: Yeah, would need to block a couple of buttons via policy, but otherwise they are basically the same page. Will do.

cwant: The one issue I'm having is that for regular communities page, the community paths all point to `community_path(community)` but for the admin communities page they should all point to `admin_community_path(community)`.

Will likely need to replace those with `url_for(action: :show, id: community)`, which isn't as pretty but gets the namespacing right.

murny: Another option is to use what we did in community show page. There's a reusable partial called "collection" which is the actual content for the list item. The url in this partial that links to the collection show page changes depending on the url (admin vs not admin). This all works from view inheritance much like the breadcrumbs on this page.

app/controllers/admin/communities_controller.rb
cwant: The other issue is the "collection vs admin/collection" links in the dropdown break when only consuming the non-admin js action.

PR #273, opened 2017-11-17
Add an API for looking up the solr name of an attribute

mbarnett:
Adds an API so you don't have to hardcode the solr name of anything:

`search_path(search: "creator_tesim:#{creator}")`

Should become:

`search_path(search: "#{Item.solr_name_for(:creator, role: :search)}:#{creator}")`

Or something similar.

### DIFF
--- a/app/models/jupiter_core/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/deferred_faceted_solr_query.rb
@@ -38,16 +38,7 @@ class JupiterCore::DeferredFacetedSolrQuery
 
   def sort(attr, order = :desc)
     raise ArgumentError, 'order must be :asc or :desc' unless [:asc, :desc].include?(order.to_sym)
-
-    # Right now we're just going to look this up on the first model, but for this to make sense,
-    # as something results can be sorted by, all models should have the same attribute name, solrized_for_sorting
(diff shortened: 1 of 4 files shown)


cwant:
I'm wondering if we can hide the Solr stuff from the outside world by going one step further: make your `solr_name_for` method private, and have a public method that uses it like:

```
def search_term_for(attribute, value)
  "#{solr_name_for(attribute, role: :search)}:#{value}"
end
```

The path then becomes:

```search_path(search: Item.search_term_for(:creator, creator))```

Looks a bit more encapsulated and cleaner to me.

**Comment by mbarnett** [comment]
Yeah, that makes a lot of sense. Will add shortly.

**Comment by mbarnett** [comment]
Ok, I split the difference between your idea and the more monkey-patch intensive idea I had in Slack the other day, and came up with:

```ruby
obj.search_term_for(:creator)
=> "creator_tesim:\"John Henry\""
```

so you can use it without having to essentially repeat the attribute twice:

`search_path(search: item.search_term_for(:creator))`

cwant:
I've just hit another wall with this: multi-valued fields aren't covered by this design (e.g., if an item has three subjects, and you want three different links from an item page to search for each subject).

I'm going to open a new issue.

PR #278, opened 2017-11-22
Use select2 for communities/collections typeahead

### LINKED ISSUE #242: Typeahead search for communties/collections on communties and collections index view
As per mockup:

https://jupiter.mybalsamiq.com/projects/jupiter-admin/Communities%20and%20Collections

cwant:
Does what it's supposed to, as per the mockup here:

https://jupiter.mybalsamiq.com/projects/jupiter/Communities%20and%20Collections

Added to both communities and admin communities index pages.
Added a system test.

Here is a screenshot:

communities_typeahead

Fixes #242 
This fills in the last checkbox of #229

### DIFF
--- a/app/assets/javascripts/jupiter/communities_collections_searchbox.js
+++ b/app/assets/javascripts/jupiter/communities_collections_searchbox.js
@@ -1,0 +1,26 @@
+$(document).on('turbolinks:load', function() {
+  $('.js-communities-collections-searchbox').each(function() {
+    $(this).select2({
+      theme: 'bootstrap',
+      ajax: {
+        // The '.json' fixes issues with chrome cacheing
+        url: 'communities.json',
+        allowClear: true,
+        dataType: 'json',
+        data: function(params) {
+          var query = {
+            query: params.term
+          }
+          return query;
+        }
+      }
+    }).on('select2:select', function(e) {
+      var data = e.params.data;
+      if (data.path) {
+        // To ensure we get the placeholder when user clicks back (firefox)
+        $(this).val('');
+        window.location.href = data.path;
+      }
+    });
+  });
+});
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/jupiter/communities_collections_searchbox.js
```diff
@@ -0,0 +1,25 @@
+$(document).on('turbolinks:load', function() {
+  $('.js-communities-collections-searchbox').each(function() {
```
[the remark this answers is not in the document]
cwant: Yeah, that was left over for when I was deciding how the urls for the admin vs non-admin would be handled (I thought I would need to have separate `data-thingies` to config each). Now that it 'just works', I can remove.

app/controllers/concerns/communities_collections_typeahead.rb
```diff
@@ -0,0 +1,28 @@
+module CommunitiesCollectionsTypeahead
+  # Note: needs these duck-typed methods: `path_to_community`, `path_to_collection`
```
murny: Perhaps we could define these methods here and either...

- they just raise an exception saying..."overwrite and implement me!" 

- or just define the methods to use this the default urls for community/collection (non admin ones). Which then you can remove these methods in the communities_controller and only overwrite them in the admin ones like you currently are?

cwant: Yeah, both good suggestions. I think I like option 2 since it results in less code :+1:

app/controllers/communities_controller.rb
```diff
@@ -25,4 +35,15 @@ def show
     end
   end
 
+  private
```
murny: Can remove all this now :)

murny: After that i'll approve

cwant: D'oh! Patch pushed now.

PR #279, opened 2017-11-23
File Downloads

mbarnett:
Merges Chris' work on the file download UI with the belongs_to/has_many work in LockedLDPObject and brings the entire thing up to date with current master


app/controllers/items_controller.rb
```diff
@@ -47,7 +39,9 @@ def update
 
     @item.unlock_and_fetch_ldp_object do |unlocked_item|
       unlocked_item.update_attributes(permitted_attributes(@item))
-      unlocked_item.update_communities_and_collections(communities, collections)
+      unlocked_item.add_communities_and_collections(communities, collections)
```
cwant: As far as I can tell, `update_communities_and_collections` and `add_communities_and_collections` do the same thing (one was left over from the original PR, the other got added in master). One should be removed.

app/controllers/file_sets_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def download
+    stream_from_fedora(disposition: 'attachment', content_type: 'application/octet-stream')
+  end
+
+  private
+
+  def load_and_authorize_fileset
+    @file_name = params[:file_name]
+    @file_set = FileSet.find(params[:id])
```
cwant: This line makes `item#edit` crash.

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <div class="d-flex justify-content-center">
+      <div class="p-1">
+        <%= link_to(t('.view'), fileset_uri(@item.file_sets.first, :show), class: 'btn btn-outline-primary') %>
+      </div>
+      <div class="p-1">
+        <%= link_to(t('.download'), fileset_uri(@item.file_sets.first, :download), class: 'btn btn-outline-primary') %>
+      </div>
+    </div>
+  <% else %>
+    <h4><%= t('.header') %></h4>
```
cwant: I get `translation missing: en.items.files_section_sidebar.header` here.

Issue #269, opened 2017-11-16
Item deposit form - Second phase


murny:
Ticket to track whats out of scope for "draft release" deadline for Item Deposit Form

Pre-beta:
- [x] display drafts on admin user detail/user dashboard #98 
- [x] Figure out thumbnail on Item View and ingest this into Fedora accordingly (is it the first file? or something smarter?) #19 
- [x] Multiple community/collection assignment 
  - Need to limit duplicate community/collection pairs (doesn't affect fedora though)? Button placements could use styling?
- Edit Item? #322 
- [x] issue with progress bar letting you click to a uncompleted step on error submission
- Figure out filename issue with regards to TempFile https://github.com/ualbertalib/jupiter/issues/400
- [x] validate link to related items?
  - Decision is to backburn this and wait for requirements. Removed link on item view since it's broken anways
- [x] look into file size limit (256 MB) on dropzone -> bumped to 1GB
- Beef up description/help on dropzone. PDF thumbnail should indicate that something happens

Post-beta?
- [x] Create CCID authenticated items (admin only? this is done already)
- Create private items (this not needed anymore?)
- Thesis Deposit with associated specialized metadata
- Loading... when deposit item button is pressed
- Date created check requirements? Optional month/day? Maybe use String field and cast to date so presence check works and allows for true optional month/day

mbarnett:
I can handle the thumbnail as part of #19

murny:
Most of this is done now. Closing out.

Issue #255, opened 2017-11-14
Item deposit form - Review and confirmation


weiweishi:
Wireframe:
![Image of Search Mockup](https://jupiter.mybalsamiq.com/mockups/6172848.png)
https://jupiter.mybalsamiq.com/projects/jupiter/Deposit%20-%20Review%20and%20add

Done when:
- [x] progress bar updated
- [x] user can complete the deposit
- [x] user can go back to previous step, and item will be saved as draft

Issue #254, opened 2017-11-14
Item deposit form - Upload Item


weiweishi:
Wireframe:
![Image of Upload items Mockup](https://jupiter.mybalsamiq.com/mockups/6172847.png)
https://jupiter.mybalsamiq.com/projects/jupiter/Deposit%20-%20Upload%20files

Done when:
- [x] user can attach file(s) from file browser 
- [x] progress bar updated
- [ ] uploaded file(s) will be characterized
  - Should only happen on the deposit? Not in draft?

murny:
Best file upload libraries for this seem to be:
https://github.com/tors/jquery-fileupload-rails (rails gem for https://github.com/blueimp/jQuery-File-Upload)
https://github.com/ncuesta/dropzonejs-rails (rails gem for https://github.com/enyo/dropzone)

Need to look into both libraries to see what abilities they have for sorting/reordering the downloads etc.

Issue #253, opened 2017-11-14
Item deposit form - license selection


weiweishi:
Wireframe:
![Image of Choose a license Mockup](https://jupiter.mybalsamiq.com/mockups/6172846.png)
https://jupiter.mybalsamiq.com/projects/jupiter/Deposit%20-%20Choose%20a%20license
Done when:
- [x] user will be directed to this page after describing their item
- [x] progress bar changes 
- [x] license is a required field. User can choose one from the list, or enter free text
- [x] regular user can set visibility as public or an embargo date
- [ ] admin user can set visibility as public, authenticated, or draft (what is draft? huh). They can also set intended after embargo visibility to be something other than public.  (from last discussion this isn't needed and isn't changeable from the UI)

PR #280, opened 2017-11-24
Deposit item

murny:
Deposit Item #100 

In Scope:
-  Describe an item #252
-  Assign a license and any embargo necessary #253
-  Upload files and move into desired order #254
-  Agree to terms of use #255
-  Deposit an item into ERA #255

Out of scope:
- Second Phase #269

TODO List:
- [x] Hook up progress bar and wizard forms (setup wicked)
- [x] Add describe item form with all goodies (select 2, etc)
- [x] Add license and visibility with all goodies (links to CC pages, etc)
- [x] Add "You have unsaved changes, are you sure you want to leave?" javascript functionality
- [x] Wire up dropzone for file uploads
- [x] Figure out sortable file uploads with dropzone
- [x] Add mockup page for review and deposit item step
- [x] Clarify requirements with Sonya before moving on
- [x] Figure out data model (active record? need to create draft item and all relational tables for this)
- [x] Hook up wicked with new data model
- [x] Convert all rails form helpers to form_for or simple_form once data model and wicked controller done
- [x] Populate review and deposit screen from data model
- [x] figure out file uploads with regards to editing/removal/reuploading/etc
- [x] figure out select 2 tag creation
- [x] testing
- [x] figure out integration with active fedora (ingesting DraftItem into Fedora as an Item)
- [X] Clean ups and polish
- [ ] figure out editing Items?


db/migrate/20171130222838_create_draft_item_tables.rb
```diff
@@ -48,7 +48,7 @@ def change
       t.timestamps
     end
 
-    create_table :draft_items_languages, id: false do |t|
+    create_table :draft_items_languages do |t|
```
murny: Looks like ActiveRecord doesn't really like this in Rails 5. Removing this makes everything work. 

More info: https://github.com/rails/rails/issues/13893
specifically: https://github.com/rails/rails/issues/13893#issuecomment-311771738 and below comments

app/assets/javascripts/jupiter/item_draft.js
```diff
@@ -0,0 +1,79 @@
+var unsavedChanges = false;
```
murny: This file contains javascript for handling:

- Accordion and Hide/Show functionality for buttons (changing the icons from plus/minus or up/down arrows, etc)
- Preventing users from leaving the form when they made a change to it. Prompting users with a dialog saying, "Are you sure?" etc.
- Prevents clicking on future links (bootstrap will color them disabled, but you can still navigate to them?)
- initializing select2 on all inputs with `.js-select2` class
- Stole some of the community/collection dropdown logic from items for AJAXing options into the select box, etc.

app/assets/javascripts/jupiter/upload_files.js
```diff
@@ -0,0 +1,43 @@
+$(document).on('turbolinks:load', function() {
```
murny: Pretty much just some javascript for intializing dropzone for file uploads.

app/assets/javascripts/jupiter/upload_files.js
```diff
@@ -0,0 +1,43 @@
...
+        // TODO: See decision above regarding `createImageThumbnails` option, if we don't care about thumbnails this all can be removed
+        // this.on('addedfile', function(file) {
+        //   if (!file.type.match(/image.*/)) {
+        //     // This is not an image, so Dropzone doesn't create a thumbnail.
+        //     // Set a default thumbnail:
+        //     filesDropzone.emit("thumbnail", file, '/images/default_todo.jpg');
+        //   }
+        // });
+
+        this.on('success', function(file, response) {
```
murny: On successful upload we will append the file to the upload lists and after a second we will remove the file from the dropzone list

app/assets/stylesheets/item_draft.scss
```diff
@@ -0,0 +1,14 @@
+// TODO: Hacks for select 2 widths being weird when initialized inside a hidden element
```
murny: Comment explains this. Some CSS hacks to try to fix a width issue with Select2 fields.

app/controllers/items/draft_controller.rb
```diff
@@ -0,0 +1,77 @@
+class Items::DraftController < ApplicationController
+
+  include Wicked::Wizard
+
+  steps(*DraftItem.wizard_steps.keys.map(&:to_sym))
+
+  def show
```
murny: Basically just render the wizard for the step that was requested. Did add some logic for handling and preventing users from skipping ahead to uncompleted steps. As well as trying to go to the review step without any files.

app/controllers/items/draft_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    else
+      params[:draft_item][:status] = DraftItem.statuses[:active]
+
+      @draft_item.update_attributes(permitted_attributes(DraftItem))
+      render_wizard @draft_item
+    end
+  end
+
+  # Deposit link
+  def create
```
murny: create method is simply for the Deposit link to get to the wizard, as we want to create a inactive DraftItem to kick off the wizard workflow

app/controllers/items/draft_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    # Handles edge case of removing all files via ajax then attempting to directly view the review step
+    elsif step == steps.last && @draft_item.files.empty?
+      redirect_to wizard_path(:upload_files, item_id: @draft_item.id),
+                  alert: t('.files_are_required_to_continue')
+    else
+      render_wizard
+    end
+  end
+
+  def update
```
murny: Pretty straight forward, updating of the DraftItem, depending on the wizard step you are on.

app/controllers/items/files_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Items::FilesController < ApplicationController
```
murny: This controller handles the ajax actions for file uploads (creating, destroying and setting of thumbnail)

app/views/items/draft/choose_license_and_visibility.html.erb
```diff
@@ -23,187 +23,54 @@
     <small class="form-text text-muted mt-0 mb-3">
       <%= t('.license.help_html') %>
     </small>
-    <div id="js-license-accordion" role="tablist">
```
murny: Reduce all this duplication (100+ lines) by utilizing the hash of license_accordions that we already have from our helper.

app/controllers/items_controller.rb
```diff
@@ -3,12 +3,16 @@ class ItemsController < ApplicationController
   before_action :load_item, only: [:show, :edit, :update]
   before_action :initialize_communities_and_collections, only: [:new, :edit]
 
+  def show; end
+
   def new
+    # TODO: Remove for Draft Item Controller
```
murny: Once this is merge we probably should clean up the old Item CRUD code

test/controllers/admin/dashboard_controller_test.rb
```diff
@@ -9,7 +9,7 @@ class Admin::DashboardControllerTest < ActionDispatch::IntegrationTest
   end
 
   should 'not be able to get to /admin if non admin user' do
-    user = users(:regular_user)
+    user = users(:regular)
```
murny: Dropped all `_user` from the user fixutres. As this was redundant.

test/fixtures/identities.yml
```diff
@@ -5,11 +5,11 @@ admin_saml:
...
-  user: regular_user
+  user: regular
   uid: userccid
   provider: saml
 
 user_facebook:
-  user: regular_user
-  uid: 012345678
+  user: regular
+  uid: abc012345
```
murny: uid is a string field, need to add alpha characters here or else its treated as a integer

.scss-lint.yml
```diff
@@ -3,6 +3,9 @@
 
 scss_files: 'app/assets/stylesheets/**/*.scss'
 
+exclude:
```
cwant: Why? Might be worth a comment or seeing if you can make this comply.

murny: I'll add a comment. It's basically some hack code to fix a bug with select2 (file itself is well commented on). Can I make it comply? Maybe some of it. But I am hoping this is quite temporary and select2 can merge a fix for this (there is 2 PRs out that fix the issue)

app/views/items/draft/choose_license_and_visibility.html.erb
```diff
@@ -0,0 +1,131 @@
...
+
+      <% license_accordions.map do |accordion_name, licenses| %>
+        <div class="card">
+          <div id="js-<%= "#{accordion_name}" %>" class="card-header" role="tab">
+            <h5 class="mb-0">
+              <a data-toggle="collapse"
+                href="#js-collapse-<%= "#{accordion_name}" %>"
+                aria-expanded="<%= license_accordion_open?(accordion_name, @draft_item.license) %>"
+                aria-controls="js-collapse-<%= accordion_name %>">
+                <%= t(".license.#{accordion_name}") %>
```
cwant: I wonder if the i18n could be dried up for the licenses (perhaps via a helper), since the 'choose_license_and_visibility.license' translation keys mostly have information that is from controlled vocabulary? (URL, license name).

To be honest, I'm not 100% sure what the mapping around of the license stuff is for (I assume it's so that the most recent CC version is always used, now and in the future?)

mbarnett: I'm not sure I understand this comment, Chris. I don't see what wrapping a helper around this t call would improve?

murny: I got nothing relating to controlled vocab (my license keys are actually different then the controlled vocab codes as well).  

This is one area where we can consolidate and make them the same down the road (just like matt did for normalising the language codes into actual names which matched my language names, then I ditched all my conversation code and can just go 1 to 1 right into Fedora.)

The mapping here is to reduce a ton of code duplication: See original and code diff here:
https://github.com/ualbertalib/jupiter/pull/280/commits/5d67f097d141cbcd1f01e2f950f989701c498441

Hopefully that makes some sense.

mbarnett: Maybe the confusion here is that license is an ActiveRecord enum instead of the stringly typed URIs we're forced to sling around in ActiveFedora?

mbarnett: Like to be clear, in terms of skating where the puck is going, my medium to long term plan is that if we can ditch ActiveFedora, we move to using enums for all of the controlled vocab stuff in terms of the app's data store. URIs should be a presentation artifact used for things like RDF export serialization, not something we're perpetually dealing with internally. 

So I see this as laying the initial groundwork for future work in this area, even if it's not per-se DRY with the ActiveFedora code, which should ideally become redundant down the road.

cwant: Okay, sounds good -- I'm sure it will work fine. @mbarnett My main point was that the full name of the license is already in I18n and the url is already in the controlled vocabulary. The translation keys that this `t` call points to in the locale file has both of those pieces of information hard coded into it. Probably not a big deal though, not difficult to keep this information in sync.

mbarnett: Ah gotcha. Yeah, we'll need to think about how to square the circle between the enum and the current code, somehow, as this stuff gets closer to unification.

cwant: @mbarnett that sounds good. Yes, storing the URL for these things internally is a pain, for sure.

PR #285, opened 2017-11-28
Issue #251, adding attributes to items, part #1

cwant:
Details in #251

I am breaking this into two (or more) stories so that some of the items don't cause log jams for the others. This will allow work on item view and deposit to more forward.

**INCLUDED WITH THIS PR:**

* [depositor](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#ualdepositor)
  - [x] Single-valued
  - [x] Not required
* [ingestBatch](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#ualingestbatch)
  - [x] Single-valued
  - [x] Not required
* [fedora3handle](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#ualfedora3handle)
  - [x] Single-valued
  - [x] Not required
* [fedora3UUID](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#ualfedora3uuid)
  - [x] Single-valued
  - [x] Not required
* [license](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dctermslicense)
  - [x] Single-valued
  - [x] Required
  - [x] From controlled vocabulary
  - [x] Model tests
* [doi](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#prismdoi)
  - [x] Single-valued
  - [x] ~~Required~~ **NOT REQUIRED** (as per discussions in this story)
  - [x] ~~Model tests~~
* [language](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dctermslanguage)
  - [x] Multi-valued
  - [x] Required
  - [x] From controlled vocabulary
  - [x] Model tests

**INCLUDED IN SEPARATE PR:**

Curve ball: the published vs non-published articles that are stored in two predicates, but faceted in a single one. Will add some more detail later.

* owner
  - [ ] Need to flesh out predicates, etc
* visibility_after_embargo
  - [ ] Need to flesh out predicates, etc
* embargo_history
  - [ ] Need to flesh out predicates, etc
* [type](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dctermstype)
  This one has a tight relationship with publication status, and has special requirements related to indexing and faceting, etc (implementation is blocked by #293)
  - [ ] Single-valued
  - [ ] Required
  - [ ] From controlled vocabulary
  - [ ] Model tests


app/models/item.rb
```diff
@@ -70,11 +107,19 @@ def thumbnail
...
     validate :communities_and_collections_validations
+    validate :language_validations
+    validate :license_validations
 
     before_validation do
       # TODO: for theses, the sort_year attribute should be derived from ual:graduationDate
       self.sort_year = Date.parse(created).year.to_s if created.present?
+
+      # TODO: replace with real DOI ...
+      self.doi = "bogus-doi-#{Time.current.utc.iso8601(3)}" if doi.nil?
```
murny: We will have to think about this or push back on this being a required field. Future discussion needed I guess ¯\\_(ツ)_/¯

Having your creation of a resource tied to a 3rd party resource that is totally out of your control is kinda crazy? (if it's down or something, nothing can get created for example). Also putting in a bogus value as a work around isn't that great either, as this will be displayed and potentially picked up by google etc. (might as well make this not required at this point)

cwant: Very valid concerns. @weiweishi , @ualbertalib/metadata any thoughts about the required status of DOI?

weiweishi: that is a very valid point indeed. Datacite DOI (which is where we will register our doi with) did have a 2.5-day outage in recent weeks: http://status.datacite.org/, so hard validation on the field probably is not a good idea. The queue where the doi creation jobs are pushed to is probably good enough to help manage the failed doi registrations.

app/models/item.rb
```diff
@@ -6,24 +6,42 @@ class Item < JupiterCore::LockedLdpObject
...
   VISIBILITIES = (JupiterCore::VISIBILITIES + [VISIBILITY_EMBARGO]).freeze
 
+  # Dublin Core attributes
   has_attribute :title, ::RDF::Vocab::DC.title, solrize_for: [:search, :sort]
   has_multival_attribute :creator, ::RDF::Vocab::DC.creator, solrize_for: [:search, :facet]
   has_multival_attribute :contributor, ::RDF::Vocab::DC.contributor, solrize_for: [:search, :facet]
   has_attribute :created, ::RDF::Vocab::DC.created, solrize_for: [:search, :sort]
-  has_attribute :sort_year, ::VOCABULARY[:ualib].sort_year, solrize_for: [:search, :sort]
-  has_attribute :subject, ::RDF::Vocab::DC.subject, solrize_for: [:search, :facet]
+  has_attribute :sort_year, ::VOCABULARY[:ual].sortyear, solrize_for: [:search, :sort]
```
weiweishi: I think sort_year should be facet as well? according to the wireframe and data dictionary. 
https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#ualsortyear

app/models/item.rb
```diff
@@ -6,24 +6,42 @@ class Item < JupiterCore::LockedLdpObject
...
   has_attribute :publisher, ::RDF::Vocab::DC.publisher, solrize_for: [:search, :facet]
   # has_attribute :date_modified, ::RDF::Vocab::DC.modified, type: :date, solrize_for: :sort
-  has_attribute :language, ::RDF::Vocab::DC.language, solrize_for: [:search, :facet]
-  has_attribute :doi, ::VOCABULARY[:ualib].doi, solrize_for: :exact_match
-
-  has_multival_attribute :member_of_paths, ::VOCABULARY[:ualib].path,
+  has_multival_attribute :language, ::RDF::Vocab::DC.language,
+                         solrize_for: [:search, :facet],
+                         facet_value_presenter: ->(language) { Item.language_text(language) }
+  has_attribute :embargo_end_date, ::RDF::Vocab::DC.modified, type: :date, solrize_for: [:sort]
```
weiweishi: I think embargo_end_date is mapped to dcterms:available? https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#dctermsavailable

app/models/item.rb
```diff
@@ -6,24 +6,42 @@ class Item < JupiterCore::LockedLdpObject
...
+  has_attribute :license, ::RDF::Vocab::DC.license, solrize_for: [:search]
+  # `type` is an ActiveFedora keyword, so we call it ...
+  has_attribute :item_type, ::RDF::Vocab::DC.type, solrize_for: [:search, :facet]
+
+  # UAL attributes
+  has_attribute :depositor, ::VOCABULARY[:ual].depositor, solrize_for: [:search]
+  has_attribute :fedora3_handle, ::VOCABULARY[:ual].fedora3handle, type: :text, solrize_for: :exact_match
```
weiweishi: why these three fields are specified as type:text? you have :exact_match so I assume it won't tokenize the text - but just checking since other ID like fields like doi are not text fields.

app/models/item.rb
```diff
@@ -6,24 +6,42 @@ class Item < JupiterCore::LockedLdpObject
...
                          type: :path,
                          solrize_for: :pathing,
                          facet_value_presenter: ->(path) { Item.path_to_titles(path) }
 
-  has_attribute :embargo_end_date, ::RDF::Vocab::DC.modified, type: :date, solrize_for: [:sort]
+  # Prism attributes
+  has_attribute :doi, ::VOCABULARY[:prism].doi, solrize_for: :exact_match
+
+  # Not sure about these ...
+  has_attribute :embargo_history, ::VOCABULARY[:ual].embargoHistory, solrize_for: :exact_match
```
weiweishi: I think this, and the next property are not in ual namespace - I think they are in projecthydra namespace.

weiweishi:
I think I missed dcterms:rights on the model. It's another curve ball - as any record will have to have either one rights or license, and only one, not both. 

https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dcrights 

I will add it to the remaining work.

app/models/item.rb
cwant: Done. But it's interesting that the metadata page says "display: false" but on the wireframes it's showing this field on item view (as "Date created")

app/models/item.rb
cwant: Yeah, this one was already in there (but I moved it in the file to group the vocabularies together, so it shows up on the diff). This is one of the fields that will go into the next PR.

app/models/item.rb
cwant: Okay, sounds good, changing. I'm used to traditional databases, so guess I need to think about tokenization more, and less about the underlying data type.

app/models/item.rb
cwant: I'm going to remove these two (and `item_type`) since they aren't supposed to be in this PR. Will deal with them outside of this PR.

app/models/item.rb
```diff
@@ -70,7 +101,11 @@ def thumbnail
     validates :embargo_end_date, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
     validates :member_of_paths, presence: true
     validates :title, presence: true
+    validates :language, presence: true
+    validates :license, presence: true
```
weiweishi: I will summerize the requirement on rights vs license on a different issue. I think this is good for now.

config/vocabularies.yml
```diff
@@ -15,12 +15,23 @@
...
-    - doi
+    - depositor
+    - fedora3handle
+    - fedora3uuid
+    - ingestbatch
     - path
-    - sort_year
+    - sortyear
+    # Just making these up here ...
+    - embargohistory
```
weiweishi: picking on nitty-gritty here. these probably should be removed from this PR too.

cwant: Tsk, so picky! (but yes, you are absolutely correct -- will remove)

test/models/jupiter_core/locked_ldp_object_test.rb
```diff
@@ -6,7 +6,7 @@ class LockedLdpObjectTest < ActiveSupport::TestCase
     ldp_object_includes Hydra::Works::WorkBehavior
     has_attribute :title, ::RDF::Vocab::DC.title, solrize_for: [:search, :facet]
     has_attribute :creator, ::RDF::Vocab::DC.creator, solrize_for: [:search, :facet]
```
weiweishi: curious - do we need to bring the test models inline with the actual model? e.g. creator is still single value here - will that be problematic?

cwant: Technically these don't need to be brought in line (these are hypothetical objects), but I can update the tests to avoid confusion.

weiweishi:
LGTM :+1:

Issue #287, opened 2017-11-28
Collections order on community page


sfbetz:
Collections on a Community detail page are ordered alpha by title. On the community browse page, when you open the list of collections under a community, they appear to be random. Can we make them alpha there too?

cwant:
This one is done now.

PR #289, opened 2017-11-29
Issue #287, sort collapsible collection lists on communities page.

cwant:
Sorted by title by default.

### DIFF
--- a/app/controllers/communities_controller.rb
+++ b/app/controllers/communities_controller.rb
@@ -25,7 +25,7 @@ class CommunitiesController < ApplicationController
       end
       format.js do
         # Used for the collapsable dropdown to show member collections
-        @collections = @community.member_collections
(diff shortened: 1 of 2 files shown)


app/controllers/communities_controller.rb
```diff
@@ -25,7 +25,7 @@ def show
       end
       format.js do
         # Used for the collapsable dropdown to show member collections
-        @collections = @community.member_collections
+        @collections = @community.member_collections.sort(sort_column, sort_direction)
```
murny: Might as well do the admin ones too?

https://github.com/ualbertalib/jupiter/blob/5496ac03eef72f68b3af136bc968477e843707f7/app/controllers/admin/communities_controller.rb#L25

cwant: Oopsie, done now.

PR #290, opened 2017-11-29
Multisorting

mbarnett:
As discussed in Slack:

```ruby
JupiterCore::Search.faceted_search(q: 'title_tesim:theses', models: [Collection], as: user).sort(:community_title, :asc).sort(:title, :asc)
```

Should do what you want now.


cwant:
Works here!

gif probably won't work

PR #291, opened 2017-11-29
Allow presenter to override facet category name.

### LINKED ISSUE #288: Rename facet for "Member of Paths"
Facet for "Member of paths" should be renamed to "Collections"

mbarnett:
Closes #288

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -22,7 +22,7 @@ class Item < JupiterCore::LockedLdpObject
                          type: :path,
                          solrize_for: :pathing,
                          facet_value_presenter: {
-                           name: 'Collections',
+                           name: I18n.t('items.facets.member_of_paths'),
                            value: ->(path) { Item.path_to_titles(path) }
                          }
 
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -21,10 +21,7 @@ class Item < JupiterCore::LockedLdpObject
   has_multival_attribute :member_of_paths, ::VOCABULARY[:ualib].path,
                          type: :path,
                          solrize_for: :pathing,
-                         facet_value_presenter: {
-                           name: I18n.t('items.facets.member_of_paths'),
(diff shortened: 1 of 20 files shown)


app/models/item.rb
```diff
@@ -21,7 +21,10 @@ class Item < JupiterCore::LockedLdpObject
   has_multival_attribute :member_of_paths, ::VOCABULARY[:ualib].path,
                          type: :path,
                          solrize_for: :pathing,
-                         facet_value_presenter: ->(path) { Item.path_to_titles(path) }
+                         facet_value_presenter: {
+                           name: 'Collections',
```
cwant: Should be `I18n`.

cwant: A couple of other things:
1) I wonder if this should now be called `facet_presenter` since it's utility has spread to cover more than just the value?
2) I am finding a need to have a facet presenter for an `additional_search_index` (the item type attribute needs a hybrid facetable thing that merges in the publication status attribute -- e.g., they want to facet on 'Published Articles', 'Draft Articles, etc). Although it's beyond the scope of this issue, any chance you could hook that up while you're in this part of the code?

mbarnett: true

mbarnett: re: #2, yeah, I think it makes sense to tackle that now.

app/models/item.rb
```diff
@@ -21,7 +21,10 @@ class Item < JupiterCore::LockedLdpObject
   has_multival_attribute :member_of_paths, ::VOCABULARY[:ualib].path,
                          type: :path,
                          solrize_for: :pathing,
-                         facet_value_presenter: ->(path) { Item.path_to_titles(path) }
+                         facet_value_presenter: {
+                           name: I18n.t('items.facets.member_of_paths'),
```
cwant: In this case, I think this should be a lambda also, otherwise the user gets the word 'Collection' in whatever language is running on the server (since this is run at init time, not per request)

Issue #220, opened 2017-10-25
accessibility check [ WIP ]


weiweishi:
Error:
On deposit form
missing form labels 
* Visibility field
* Communities and Collections field

Empty link:
* The Remove link for the "Communities and Collections" 

On search result page:
* Thumbnail on the item search result page

Alert:
On item detail page:
* A main element or main landmark is presented (broken same-page link)
On item search page:
* heading level on title is skipped.

murny:
I believe this is all fixed now? Feel free to reopen if it is not

PR #296, opened 2017-11-30
fixes for current WAVE scan result #220

weiweishi:
Address issues that are currently identified by WAVE (Accessibility Scan)
https://github.com/ualbertalib/jupiter/issues/220


app/assets/stylesheets/logo.scss
```diff
@@ -9,3 +9,8 @@
   max-width: inherit;
   width: 100px;
 }
+.logo-text {
```
murny: Looks like your hiding this from display?

There is a bootstrap utility for this called `sr-only` which will only be displayed for screen readers:
https://getbootstrap.com/docs/4.0/utilities/screenreaders/

weiweishi: This is something I'm not sure about. 

This is to address the issue on item result page, when no thumbnail is presented, the thumbnail link becomes an empty link. But I'm not sure if this is the best way to handle it. See also: https://github.com/ualbertalib/jupiter/pull/296/files#diff-9f4ffa558424de7b50ffb065b48727dbR13

weiweishi: Aha, okay, I will try that!

app/views/application/_navbar.html.erb
```diff
@@ -32,9 +32,9 @@
         <li role="presentation" class="nav-item mr-3">
 
           <form class="form-inline" action="<%= search_path %>">
-            <div class="input-group">
+            <div class="input-group" aria-label: "search">
```
murny: this in a html element, so it should be defined with an `=` like:
``` <div class="input-group" aria-label="search">```

File: app/views/application/_logo.html.erb [hunk 3]
```diff
@@ -1,5 +1,5 @@
 <% if object.logo.attached? %>
-  <%= image_tag(object.logo.url, class: "d-flex mr-3 img-thumbnail logo-small") %>
+  <%= image_tag(object.logo.url, class: "d-flex mr-3 img-thumbnail logo-small", alt:"#{object.title}") %>
```
murny: small nitpick but you probably don't need to string interpolate the title value, can simply just be:

` <%= image_tag(object.logo.url, class: "d-flex mr-3 img-thumbnail logo-small", alt: object.title) %>`

app/views/application/_thumbnail.html.erb
```diff
@@ -5,11 +5,13 @@
                   'logo-small'
                 end %>
 <% if object.thumbnail&.attached? %>
-  <%= image_tag(object.thumbnail.url, class: "d-flex mr-3 img-thumbnail #{size_class}") %>
+  <%= image_tag(object.thumbnail.url, alt: "#{object.title}", class: "d-flex mr-3 img-thumbnail #{size_class}") %>
```
murny: Same here

app/views/application/_thumbnail.html.erb
```diff
@@ -5,11 +5,13 @@
...
 <% else %>
   <div class="d-flex justify-content-center align-items-center mr-3 img-thumbnail <%= size_class %>">
     <small class="text-muted">
-      <i class="fa fa-5x fa-picture-o" aria-hidden="true"></i>
+      <i class="fa fa-5x fa-picture-o" aria-hidden="true">
```
murny: `aria-hidden` tells screen readers to ignore this element. So your object title below wouldn't be seen anyways.

So the small tag probably shouldn't be nested inside the `i` tag?

weiweishi: Yes, good catch. Just moved the tag out.

test/system/search_test.rb
```diff
@@ -141,7 +141,7 @@ def before_all
       assert_selector 'a.nav-link', text: 'Communities (1)'
 
       # No community/collection results initially shown
-      assert_selector 'div.jupiter-results-list a', text: 'Item', count: 5
+      assert_selector 'div.jupiter-results-list a', text: 'Item', count: 10
```
murny: No idea why this test changed? Seems weird. But all is working now

weiweishi:
Thank you @murny . I was trying to fix the test too but it was broken on something else for me in Travis.

PR #297, opened 2017-11-30
Improved Faceting & Facet Presenters

### LINKED ISSUE #293: Make additional_solr_index facet compatible

mbarnett:
Introduces "facet presenters" where any given attribute can have a Ryan Bates-esque presenter created for it to customize the way it is displayed to the user. This is an improvement over defining them in models, because a facet can span multiple models (say, a collection path spanning Items, Theses, etc), so there is no one "owning" model for the facet that can be responsible for the presentation. 

This also makes it easier to allow solr-only attributes to share in the presentation logic.

The framework is here for creating presenters for other kinds of objects as well, if we want to pull more logic out of the views in the future.

This PR should solve any other problems related to `additional_search_index` & faceting as well. Just make sure the index is solrized for `:facet` and define a presenter in `presenters/facet_values` named the same as the index, if necessary.

Closes #293


app/models/presenters/facet_values/language.rb
```diff
@@ -0,0 +1,12 @@
+class Presenters::FacetValues::Language < Presenters::FacetValues::DefaultPresenter
+
+  def display
+    CONTROLLED_VOCABULARIES[:language].each do |lang|
+      if lang[:uri] == @value
+        return I18n.t("controlled_vocabularies.language.#{lang[:code]}")
+      end
+    end
+    raise ApplicationError("Language not found for #{language_uri}")
```
cwant: Should likely be ArgumentError (bug introduced by me, this exception doesn't really exist)

app/models/presenters/facet_values/language.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+end
```
cwant: I am a bit surprised not to see a presenter for the license attribute. It follows an identical pattern to language.

cwant: Ah, I see, I didn't hook license up to the previously named `facet_value_presenter`.

mbarnett: Aha. I knew there was an answer to this question, I just couldn't remember what it was...

mbarnett: Do you want me to hook this up for you, or do you want to handle that as part of your work on the data model?

cwant: I'm not sure. It looks like we are working on very similar parts of the code, so it might not be a bad idea to touch base for a couple of minutes.

In my current work, I've refactored some of this stuff so that it can be more abstract and can be reused as part of the controlled vocabulary validations (in current branch I need to deal with language, license, and Frankenstein item type/publication status solr-only field). This is roughly what I have so far:

``` ruby
  # Some URI --> [text|code] functions for controlled vocabularies
  def self.uri_to_code(uri, vocabulary)
    CONTROLLED_VOCABULARIES[:language].each do |term|
      return term[:code] if term[:uri] == uri
    end
    raise ArgumentError("#{uri} not found in controlled vocabulary #{vocabulary}")
  end

  def self.code_to_text(code, vocabulary)
    I18n.t("controlled_vocabularies.#{vocabulary}.#{code}")
  end

  def self.uri_to_text(uri, vocabulary)
    code_to_text(uri_to_code(uri_value, vocabulary), vocabulary)
  end
```
Then presenter for language/license is basically a one liner.

It might be worth introducing a module to include this stuff, or have a Presenters::FacetValues::URIPresenter as a common superclass of language/license/etc. Or keep in Item class as we need to display things like Language/License in the main Item display too, not just facets.

mbarnett: It sounds like what we really want is almost a proper URI type that attributes could be declared as, but that would require major surgery to AF and solrizer at this point.

Maybe a URIPresenter parent class would suffice for the time being, and we can revist when we have more control over the types of things that pop out of the database

cwant:
Whelp, yeah, on top of that, the way I raised the exeption it is fine for python, not so good ruby. Should basically be:

```
raise ArgumentError, "Language not found for #{language_uri}"
```
(with a comma).

mbarnett:
Chris, you reminded me of one of my favorite tweets: 

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">Programming Skills: PRIMARILY RUBY AND PYTHON BUT I CAN USE ANY TYPE OF GEM TO CONTROL ANY TYPE OF SNAKE</p>&mdash; bandit (@UtilityLimb) <a href="https://twitter.com/UtilityLimb/status/126780301211992064?ref_src=twsrc%5Etfw">October 19, 2011</a></blockquote>

app/models/presenters/facet_values/uri_value.rb
```diff
@@ -0,0 +1,15 @@
...
+  protected
+
+  def translate_uri(vocab, uri)
+    raise ArgumentError unless vocab.is_a? Symbol
+    CONTROLLED_VOCABULARIES[vocab].each do |entry|
+      if entry[:uri] == uri
+        return I18n.t("controlled_vocabularies.#{vocab}.#{entry[:code]}")
+      end
+    end
+    raise ArgumentError, "Vocabulary not found: #{vocab}"
```
cwant: This is a little misleading, probably not finding the URI is more important to report than the vocabulary not found. Near the top of the function you might want to include

```
raise ArgumentError, "Vocabulary not found: #{vocab}" unless CONTROLLED_VOCABULARIES.key?(vocab)
```

mbarnett: Ah yeah, good point

mbarnett: fizixed

PR #304, opened 2017-12-04
`ApplicationError` --> `ArgumentError`

cwant:
Although `ApplicationError` is a plausible exception name no such exception is include with Ruby. In this case, ArgumentError is more appropriate anyways.


cwant:
A fix for this will be pulled in via PR #297 . Not merging this one as to minimize merge conflicts with that branch.

PR #306, opened 2017-12-04
Issue #251, item properties part 2.

cwant:
Curve balls: 

1) the published vs non-published articles that are stored in two predicates, but faceted in a single one. Methods added to handle this in solr.
2) License and rights have an exclusive-or relationship: one must be included, not both
3) It's uncertain whether `owner` will be a single-valued or multi-valued property. This can be dealt with outside of the scope of this PR.

To be included:

* [owner](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#biboowner)
  - [x] Change predicate to `bibo` vocabulary
  - [x] ~~Multi-valued!~~
* [visibility_after_embargo](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#aclvisibilityafterembargo)
  - [x] Required if embargoed
  - [x] Model tests
  - [x] From controlled vocabulary
  - [x] Update seeds
* [embargo_history](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#aclembargohistory)
  ~~Thesis only?!?~~
  **Any validations needed for content?**
  - [x] Multi-valued
  - [x] Not required
* [type](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dctermstype)
  This one has a tight relationship with publication status, and has special requirements related to indexing and faceting, etc ~~(implementation is blocked by #293)~~
  - [x] Single-valued
  - [x] Required
  - [x] From controlled vocabulary
  - [x] Model tests
  - [x] Update seeds
* [status](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#bibostatus)
  - [x] Single-valued
  - [x] Required if type above is an article, otherwise must not be present
  - [x] From controlled vocabulary
  - [x] Model tests
  - [x] Update seeds
* [license](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dctermslicense)
  - [x] Required if rights not included (must have one of the two)
* [rights](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_generic.md#dcrights)
  - [x] Single-valued
  - [x] Required if no license (must have one of the two)
  - [x] Model tests
  - [x] Update seeds
* [accessRights](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#dctermsaccessrights)
  This is what we've been calling `visibility` and have been using our home-baked predicate.
  - [x] Single-valued
  - [x] Required
  - [x] Model tests
  - [x] Controlled vocabulary
  - [x] Update seeds


weiweishi:
@cwant @mbarnett @murny Re: multi-value for owner - Leah is going to review the multiple owner report Zach generated, and clean the data as much as ERA team can. She will then make a decision with the consultation to the team, whether we need to migrate the data as-is before we can find a better solution (which means some work needed to make the application to handle multiple owners data, but allow only single owner entering into the system.) or if we can use owner as the single-value property. I will discuss it with the team and open new ticket if it's the former case.

app/models/item.rb
```diff
@@ -16,10 +19,13 @@ class Item < JupiterCore::LockedLdpObject
...
   has_attribute :publisher, ::RDF::Vocab::DC.publisher, solrize_for: [:search, :facet]
   # has_attribute :date_modified, ::RDF::Vocab::DC.modified, type: :date, solrize_for: :sort
-  has_multival_attribute :language, ::RDF::Vocab::DC.language,
-                         solrize_for: [:search, :facet]
-  has_attribute :embargo_end_date, ::RDF::Vocab::DC.modified, type: :date, solrize_for: [:sort]
+  has_multival_attribute :language, ::RDF::Vocab::DC.language, solrize_for: [:search, :facet]
+  has_attribute :embargo_end_date, ::RDF::Vocab::DC.available, type: :date, solrize_for: [:sort]
   has_attribute :license, ::RDF::Vocab::DC.license, solrize_for: [:search]
+  has_attribute :rights, ::RDF::Vocab::DC.rights, solrize_for: :exact_match
+  # `type` is an ActiveFedora keyword, so we call it `item_type`
```
mbarnett: Pretend this is my shouting FEDORA

Simpsons character shouting Mendoza

app/models/item.rb
```diff
@@ -33,10 +39,23 @@ class Item < JupiterCore::LockedLdpObject
...
+
   # Solr only
   additional_search_index :doi_without_label, solrize_for: :exact_match,
                                               as: -> { doi.gsub('doi:', '') if doi.present? }
 
+  # This combines both the controlled vocabulary codes from item_type and published_status above
+  # (but only for items that are articles)
+  additional_search_index :item_type_with_status,
+                          solrize_for: :facet,
+                          as: -> { item_type_with_status_code }
```
mbarnett: You can lose a level of indirection (calling a lambda that calls a method) with:

`as: method(:item_type_with_status_code)` or `:item_type_with_status_code.to_proc`

(that's the intention, anyways. Let me know if not)
(it could probably just take a symbol and to_proc it itself, come to think of it)

config/vocabularies.yml
```diff
@@ -30,10 +30,19 @@
   terms:
```
mbarnett: Not related to your changes in this file, but should we rename this file/the initializer/the global constant to Terms or something?

The name is similar enough to Controlled Vocabularies that I keep confusing the two.

cwant: Seems it's an aggregate of namespaces and properties -- maybe `NAMESPACES[:prism].doi` or `ONTOLOGIES[:bibo].owner`?

mbarnett: I tend to think of the naming in terms of what you get out of the map rather than what you put in -- ie) `doi` or `owner` isn't a Namespace or an Ontology, it's a....term? That's not terribly descriptive. Predicate?

mbarnett: The pattern that file is superseding is this one: https://github.com/ualbertalib/HydraNorth/blob/4480e7f5d95ff85e41d14a55e39a9f046ed02293/lib/rdf_vocabularies/ualid.rb. Basically, it declares terms objects that ActiveFedora understands in some vocabulary namespace.

cwant: Predicate could work. Another thought (maybe not a good one), might be to generate a constant for each namespace. e.g. `PRISM.doi` or `BIBO.owner`. The pluses are that you get shorter lines of code, and you don't have to decide whether it's a term, a predicate, or anything else. I was thinking about something like this for the controlled vocabularies since `LICENSE.attribution_4_0_international` is a bit easier to swallow than  `CONTROLLED_VOCABULARIES[:license].attribution_4_0_international`. The negative is that it may not be clear where the constant was defined.

mbarnett: Yeah, I thought about doing that originally, but pity on the poor fool who tried to hunt down the non-existent definition led me to use a constant you could actually grep the sourcecode and find. Partially as a reaction against the horror that was ever finding the definition of anything in Hydra's twisty mess of gems.

app/models/item.rb
cwant: Yeah, tried these plus a number of combinations involving ampersands and `lambda`. No luck. A lot of errors about either no receiver or undefined class method.

mbarnett: Ah, just leave it as-is and I'll look into making that more friendly at some point.

config/vocabularies.yml
cwant: Just checking: do we have a concrete action item out of this discussion? I can change `VOCABULARY` to `PREDICATES` or `TERMS` -- is that the road we want to take?

weiweishi:
The PR looks good to me :100:

cwant:
This will need re-approval due to some breakage from another merged PR.

PR #307, opened 2017-12-06
WIP - Skip to search link

weiweishi:
This is to add "skip to main content" and "skip to search" links to the application following recommendations from https://webaim.org/techniques/skipnav/. 

Both links work fine independently, but when I use "skip to main content" and then "skip to search" link, the tab focus is incorrect, instead of focusing on the search list, it goes back to the header. 

I was trying to solve it with JS, but doesn't seem to help. The JS doesn't seem to be called after the URL is changed to the anchor link. @murny @mbarnett @cwant Dp you have any idea why that is?

### DIFF
--- a/app/assets/javascripts/jupiter/skip.js
+++ b/app/assets/javascripts/jupiter/skip.js
@@ -1,0 +1,7 @@
+$("#skip-nav").click(function() {
+    $('main').focus();
+});
+$("#skip-to-results").click(function() {
+    $('main').blur();
+    $("#results-anchor").focus();
+});
--- a/app/assets/stylesheets/jupiter.scss
+++ b/app/assets/stylesheets/jupiter.scss
@@ -9,3 +9,50 @@ body {
   font-weight: bold;
   text-decoration: inherit;
 }
+#skiptocontent a {
+	padding:6px;
+	position: absolute;
+	top:-40px;
+	left:0px;
+	color:white;
+	border-right:1px solid white;
+	border-bottom:1px solid white;
+	border-bottom-right-radius:8px;
+	background:transparent;
+	-webkit-transition: top 1s ease-out, background 1s linear;
+    transition: top 1s ease-out, background 1s linear;
+    z-index: 100;
+}
+
+#skiptocontent a:focus {
+	position:absolute;
+	left:0px;
+	top:0px;
+	background:$red;
+	outline:0;	
+	-webkit-transition: top .1s ease-in, background .5s linear;
+    transition: top .1s ease-in, background .5s linear;
+}
+#skiptoresults a {
+        padding:6px;
+        position: absolute;
+        left:-1000px;
+        color: $white;
+        border-right:1px solid $white;
+        border-bottom:1px solid $white;
+        border-bottom-right-radius:8px;
+        background:transparent;
+        -webkit-transition: top 1s ease-out, background 1s linear;
+    transition: top 1s ease-out, background 1s linear;
+    z-index: 100;
+}
+
+#skiptoresults a:focus {
+        position:absolute;
+        left:100px;
+        background:$red;
+        outline:0;
+        -webkit-transition: top .1s ease-in, background .5s linear;
+    transition: top .1s ease-in, background .5s linear;
+}
+
--- a/app/views/application/_navbar.html.erb
+++ b/app/views/application/_navbar.html.erb
@@ -7,7 +7,7 @@
       <%= t('site_name') %>
     <% end %>
     <div id ="skiptocontent">
-      <%= link_to "Skip to Main Content", '#content-anchor', id:"skip-nav" %>
+      <%= link_to t('.skip_to_main_content'), '#main', id:"skip-nav" %>
     </div>
 
     <button type="button"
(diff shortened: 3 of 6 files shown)


app/assets/javascripts/jupiter/skip.js
```diff
@@ -0,0 +1,7 @@
+$("#skip-nav").click(function() {
```
murny: Do you need any javascript at all? the focus is a nice touch but is it needed? 

Even in the wild, most sites i've seen implementing this do not do this at all (such as reddit, github, getbootstrap.com, even the document you linked here https://webaim.org/techniques/skipnav/#creating).

app/assets/stylesheets/jupiter.scss
```diff
@@ -9,3 +9,50 @@ body {
   font-weight: bold;
   text-decoration: inherit;
 }
+#skiptocontent a {
```
murny: I can pretty guarantee much of all this custom CSS can go? Go read into bootstrap and use bootstrap for styling:
http://getbootstrap.com/docs/4.0/getting-started/introduction/

app/views/application/_navbar.html.erb
```diff
@@ -6,6 +6,9 @@
       <%= image_tag 'mc_360.png', width: '30', height: '30', class: 'd-inline-block align-top', alt: t('.logo_alt') %>
       <%= t('site_name') %>
     <% end %>
+    <div id ="skiptocontent">
```
murny: Typically you want this to be the first tab link on the page. Right now the logo/site name will be first. So either positon this better in the dom or use a "tabindex=1" attr

weiweishi:
I don't like the js approach either. The js was added because the focus
can't be set correctly. So it's not really doing what I intended it to do,
and can be removed, if I can figure out how to jump to the second anchor
link with proper tab focus properly.


*Weiwei Shi*

Digital Initiative Applications Librarian
University of Alberta Libraries
2-10L Cameron Library
Edmonton, Alberta, Canada T6G 2J8
Phone:(780)492-7802
Fax: (780)248-1209
Email: weiwei.shi@ualberta.ca

app/views/application/_navbar.html.erb
murny: Also regarding the heavy customization with css and everything, this can be styled and marked up better. 

Want an bootstrap example from the official getbootstrap.com site? Here you go:
```html
    <a id="skippy" class="sr-only sr-only-focusable" href="#content">
      <div class="container">
        <span class="skiplink-text">Skip to main content</span>
      </div>
    </a>
```

```#skippy {
  display: block;
  padding: 1em;
  color: #fff;
  background-color: $bd-purple;
  outline: 0;

  .skiplink-text {
    padding: .5em;
    outline: 1px dotted;
  }
}
```

All this CSS can be tossed too and instead you can use the bootstrap classes instead

app/views/application/_navbar.html.erb
```diff
@@ -6,6 +6,9 @@
...
+      <%= link_to t('.skip_to_main_content'), '#main', id:"skip-nav" %>
```
murny: ID here only being used for the javascript. So namespace it as such: `js-skip-nav`.

Better yet, do you need an ID for this? Or can a class suffice?

app/views/search/_model_results.html.erb
```diff
@@ -1,3 +1,7 @@
+<div id="skiptoresults">
```
murny: Same comments as above apply here too

app/views/search/_model_results.html.erb
```diff
@@ -1,3 +1,7 @@
...
+<%= link_to t('search.skip_to_search', '#results-anchor', id: "skip-to-results" %>
```
murny: Looks like your missing a bracket here? 

`t('search.skip_to_search'` => `t('search.skip_to_search')`

app/views/application/_navbar.html.erb
weiweishi: 👍  I will modify my css accordingly. (need to remember to check bootstrap site for best practices)

app/assets/javascripts/jupiter/skip.js
weiweishi: yes, exactly. The js is recommended by some sites to solve the issue of the anchor links in some browsers. So I will modify it but leave it in. It doesn't solve my two anchor links issue though.

weiweishi:
fixed the issues and modified css accordingly. But still run into the issue that the skip link doesn't work the second time it's clicked. will auto focus back to the nav bar, instead of the proper anchor link.

app/assets/stylesheets/jupiter.scss
```diff
@@ -9,3 +9,17 @@ body {
...
+.skip {
+  background-color: $red;
+  color: $white;
+  display: block;
+  outline: 0;
+  padding: 1em;
+
+  .skiplink-text {
+    outline: 1px dotted;
+    padding: .5em;
```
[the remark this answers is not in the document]
weiweishi: scss-lint apparently doesn't agree with this.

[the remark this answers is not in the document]
murny: It was fine as 0.5 (with a leading zero), it was complaining previously about the order of these properties.

Regardless you can use bootstrap to replace most of this. (using the class `.p-2` is 0.5 rem for example, `.p-3` is 1rem. `background-color: $red` can be replaced with `.bg-danger`, `color: $white` can be replaced with `.text-white`, etc. )

Why am I pushing bootstrap classes? Because their documented very well and are far easier to maintain in the future. Custom CSS should be a last resort.

PR #311, opened 2017-12-08
Add properties to Community and Collection models as per issues #256 and #257

### LINKED ISSUE #256: Additional properties for community's rails data model
These fields need to be added to the community's rails data model:

- [ ] fedora3UUID, predicate ual:fedora3UUID, no need to be on UI, not a required field
- [ ] creator, predicate dcterms:creator, no need to be on UI. (pending answers to the following question)

Question:
- depositor vs creator in https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profiles/community/profile.json. Which one we need to use to keep track of the creator of the community? can depositor be someone other than era admin in the existing data? Do we need to carry that data forward? @ualbertalib/metadata-team 

- how are (or will) write permissions controlled on the collection/community level? (Edit, deposit in, delete) @mbarnett @murny

### LINKED ISSUE #257: Additional properties for collection's rails data model
- [ ] memberOf - currently this membership is handled by community_id on collection, and are treated as a path, can we map community_id to PCDM:memberOf, so our metadata data model is inline with PCDM? @mbarnett This field is multi-value, required field for collection, and will be on collection creation form, and collection view. 
- [ ] creator (similiar question to the one in #256)
- [ ] fedora3UUID, predicate ual:fedora3UUID, single value, will not be displayed on UI, used for redirect. 

Question:
- depositor vs creator, similiar to #256 @ualbertalib/metadata-team

cwant:
Included:
* Community:
   - [x] fedora3UUID, predicate ual:fedora3UUID
   - [x] creator, predicate dcterms:creator
* Collection:
  - [x] fedora3UUID, predicate ual:fedora3UUID
  - [x] creator, predicate dcterms:creator
  - [x] memberOf - predicate PCDM:memberOf, unit test

Fixes #256
Fixes #257

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -12,6 +12,8 @@ class Collection < JupiterCore::LockedLdpObject
                 solrize_for: :pathing
 
   has_attribute :description, ::RDF::Vocab::DC.description, solrize_for: [:search]
+  has_multival_attribute :creator, ::RDF::Vocab::DC.creator, solrize_for: :exact_match
+  has_attribute :fedora3_uuid, ::VOCABULARY[:ual].fedora3uuid, solrize_for: :exact_match
 
   additional_search_index :community_title, solrize_for: :sort,
                                             as: -> { Community.find_by(id: community_id)&.title }
@@ -38,8 +40,15 @@ class Collection < JupiterCore::LockedLdpObject
     validates :title, presence: true
     validates :community_id, presence: true
     validate :community_validations
+
     before_validation do
       self.visibility = JupiterCore::VISIBILITY_PUBLIC
+    end
+
+    before_save do
+      community.unlock_and_fetch_ldp_object do |uo|
+        self.member_of_collections += [uo]
+      end
     end
 
     def can_be_destroyed?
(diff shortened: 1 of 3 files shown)


app/models/collection.rb
```diff
@@ -12,6 +12,8 @@ class Collection < JupiterCore::LockedLdpObject
                 solrize_for: :pathing
 
   has_attribute :description, ::RDF::Vocab::DC.description, solrize_for: [:search]
+  has_multival_attribute :creator, ::RDF::Vocab::DC.creator, solrize_for: :exact_match
```
murny: Random question, any reason why we don't pluralize multi value attributes? 

AKA `creators` vs `creator` (since creators implies its multi value where creator implies a single value)?

cwant: I guess I've just been matching the predicates. Doesn't matter much to me either way. @mbarnett do you have an opinion on this?

mbarnett: I'd probably pluralize it? Personally I'm dumb enough that if something looks singular in name I'll forget that it could be an array.

weiweishi: pluralization makes sense - let's be consistent to make all multiple value properties pluralized. (don't think this change will have any impact on the triples the metadata people is going to get, correct?)

mbarnett: The metadata triples contain the predicate, not the Ruby property name, so AFAIK it shouldn't change anything on their end

weiweishi: Thanks for the confirmation. I'm fine with the change as long as it's the case.

cwant: I will do the pluralization of all multi-valued as a separate issue/PR, since it touches almost everything. (I'll queue it in as up next.)

PR #319, opened 2017-12-11
Truncate long descriptions in search results list.

### LINKED ISSUE #298: Truncate abstracts on results page
Some of our abstracts are reallllllly long and leading to a lot of scrolling. Can we truncate abstracts on the results page? Say something like 300 characters followed by an ellipses?

cwant:
Fixes #298


app/views/search/_item_hit.html.erb
```diff
@@ -37,8 +37,7 @@
                         &.join(', ')&.html_safe %>
     </div>
     <div>
-      <%# TODO: for long description, implement 'more ...' / 'less ...' %>
-      <%= result.description %>
+      <%= truncate result.description, length: SearchHelper::DESCRIPTION_DISPLAY_MAX, separator: ' ' %>
```
murny: I'd say make this more generalized. We going to be using this in different places (communities' description on browse page for example needs something similar to this).

Shadow the truncate method (call it jupiter_truncate or something). Throw this in application helper. Have the method take in text, length (default is 300) and separator (default is ellipsis? `...`) etc. Then most of this can be cleaned up.

The other option you can also try is to do this via CSS instead of ruby. Take a peak at `.text-truncate` class in bootstrap: https://getbootstrap.com/docs/4.0/utilities/text/#text-wrapping-and-overflow

mbarnett: That's actually an interesting point. We might want to render the full description onto the page and then use truncation in CSS to make it look reasonable, so that Google Scholar gets a full look at the description for their indexing?

PR #320, opened 2017-12-11
Add fallback when populating item sort_year

mbarnett:
We can't necessarily parse fuzzy `created` strings, so add a fallback that tries to find the first thing that looks like a year, and stuff it into `sort_year`


mbarnett:
The sort year part got mixed into the other PR because I did this on master. The unique bits in the other PR were just the Policy changes.

PR #327, opened 2017-12-12
Add `pcdm:memberOf` predicate to items (pointing to collections URIs)

cwant:


mbarnett:
(Just leaving a note here to remind myself to check how much PCDM orthodoxy is going to cost us on ingest speed with those cascading collection saves (its not going to be pretty, sadly 😔))

app/models/item.rb
```diff
@@ -119,6 +119,18 @@ def thumbnail
       end
     end
 
+    before_save do
+      # This adds the `pcdm::memberOf` predicate
+      member_of_paths.each do |path|
```
mbarnett: We have access to the standard ActiveModel::Dirty stuff, so if we wrap this in a

```ruby
if member_of_paths_changed?
  <rest of method>
```

we can save ourselves some Fedora traffic in cases where the objects paths weren't updated.

app/models/item.rb
```diff
@@ -119,6 +119,18 @@ def thumbnail
...
+        _community_id, collection_id = path.split('/')
+        collection = Collection.find_by(collection_id)
+
+        collection.unlock_and_fetch_ldp_object do |uo|
+          self.member_of_collections += [uo]
```
mbarnett: I think we have to zero this out first, in case the object was removed from any collections?

mbarnett:
Also we should probably extend the dirty attribute checking and the zeroing to https://github.com/ualbertalib/jupiter/blob/master/app/models/collection.rb#L48. I missed that on the original review.

cwant:
There are some problems with extending to the collection version:

1) Zeroing the `member_of_collection` for community will wipe out any other collections in the community
2) Seems `id_changed?` isn't a thing for collection (I think that's the only time you'd want to update the community)

mbarnett:
1. You'd want to zero it for collection: collection.member_of_collections (ugh) = []. The Community's member_of_collections property would track things that the community itself belonged to, which shouldn't exist afaik?

2. It should be `community_id_changed?`

cwant:
Yeah, I think I've fully confusing myself with point one. (Unlocking the community threw me off.)

mbarnett:
It's *super* confusing. I misread the entire PR the first time through, had it all backwards.

cwant:
I think I'll ditch the `uo` name for the unlocked object and replace with something more illustrative (should reduce 100% confusion to only 75%)

PR #328, opened 2017-12-12
Facet creators and contributors together.

cwant:


app/models/item.rb
```diff
@@ -61,6 +64,11 @@ class Item < JupiterCore::LockedLdpObject
                           solrize_for: :facet,
                           as: -> { item_type_with_status_code }
 
+  # Combine creators and contributors for faceting
+  additional_search_index :all_contributors,
+                          solrize_for: :facet,
+                          as: -> { creators + contributors.to_a }
```
murny: `.to_a` is leaving me confused? Is this needed? 

Whats the order of operations here? converts contributors into an array and then concats whatever creators is? Or does it concat both and then to arrays them?

Isn't contributors an array? and creators is an array already? Or this is in case if one of them are nil, this still works?

mbarnett: I don't think they're arrays. They're likely ActiveTriples::Relation instances.

cwant: Yeah, this was to handle nil. Now that I look at it, I forgot to make creator required, which will impact this line of code. Since this is merged already, will add this to one of the other PRs about item properties.

cwant: (and will add a comment why one is converted to array, the other not)

PR #334, opened 2017-12-18
Issue #294. Item view: matching wireframes.

### LINKED ISSUE #294: Item view - Attribute display
To prepare for the first round of user testing, item view needs to be updated to make it more usable. This includes:
image

- [ ] Move fields: Visibility, owner, and record created at to the bottom of the list
- [ ] make subject and creator show as links to search, and not in array with "[","]"
- [ ] make all links within the attributes point to faceted search results pages

The related wireframe is at:
image

cwant:
The screenshots of what we are trying to match are in Issue #294.

Notes:

* Attributes that do not have a heading (title, alternative_title, description) have the HTML `title` attribute set so that people can know what they are on hover.
* Seeds updated to create a mondo-item per collection with fairly full metadata

Here is a screenshot of default view (some fields hidden, 'More information ...' link shown):

item_view_default_view

Here is a screenshot of full view (all fields shown, after clicking 'More information ...'):

item_view_full_view

Fixes #294


app/assets/stylesheets/items.scss
```diff
@@ -6,3 +6,8 @@ $cc-background: $gray-100;
     background-color: $cc-background;
   }
 }
+
+.item-more-less-link {
```
murny: Can use bootstrap for this stuff:
https://getbootstrap.com/docs/4.0/utilities/text/#font-weight-and-italics

`.font-weight-bold` and `.font-italic`

app/views/items/_more_information.html.erb
```diff
@@ -0,0 +1,65 @@
+<ul class="list-group">
+  <% if @item.languages.present? %>
+    <li class="list-unstyled list-group-item-action">
+      <dl>
+        <dt><%= t('items.show.attributes.languages') %></dt>
+        <dd>
+          <ul class="list-unstyled">
+            <% @item.languages.each do |language| %>
+              <li><%= search_link_for(Item, :languages, value: language,
+                      display: CONTROLLED_VOCABULARIES[:language].uri_to_text(language)) %></li>
```
mbarnett: I'd like to see calls to `CONTROLLED_VOCABULARIES` get hidden in a helper method.

app/helpers/application_helper.rb
```diff
@@ -38,4 +38,13 @@ def results_range(results)
     last = results.offset_value + results.count
     t(:page_range, first: first, last: last, total: results.total_count)
   end
+
+  def search_link_for(model, attribute, value: nil, facet: true, display: nil)
```
mbarnett: Why not take the instance instead of the class? From what I could see every call to this has access to an item instance, and if we passed it, we wouldn't have the `Item` class hardcoded into views.

cwant: This is used in search results too, where you don't have a particular instance. See here: https://github.com/ualbertalib/jupiter/pull/334/files#diff-259482bd68dea53d8e36e5b6e508d927R36

mbarnett: Wouldn't `result` be an item here?

cwant: Hrmm, yeah, I think you're right there -- I'll give it a go.

cwant:
I've addressed all of the recent comments -- this PR could use another look.

app/helpers/items_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+    search_link_for(item, :item_type_with_status,
+                    value: item.item_type_with_status_code,
+                    display: t("controlled_vocabularies.item_type_with_status.#{item.item_type_with_status_code}"))
+  end
+
+  def language_search_link(item, language_uri)
+    search_link_for(item, :languages, value: language_uri,
+                                      display: CONTROLLED_VOCABULARIES[:language].uri_to_text(language_uri))
+  end
+end
```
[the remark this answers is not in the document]
cwant: I fully agree with your assessment, particular about the presenters. Also, I think a lot of this code will shift anyways as you refactor the code/uri/i8n/vocabulary stuff.

Issue #234, opened 2017-11-06
Google Scholar optimization


pbinkley:
To-do: breaking down the tasks

A summary of the Google Scholar work done on HydraNorth is [here](https://docs.google.com/document/d/1mVwcjHuTSg4srNDpVX4f9sIeIFske-MWzOg_oR41KYQ/edit). Our main areas of improvement were:

- [x] Modify download urls with .pdf
- [x]  Modify directory structure to place the file "below" the landing page
- [ ] ~~Browse by date (building on Sufia sitemap)~~
- [x]  Ensure date of publication is present
- [x]  Ensure abstracts are accessible

PR #340, opened 2017-12-20
[WIP] Add a basic cached sitemap. Some work remaining to flesh it out

### LINKED ISSUE #9: Port sitemap work
Caching currently lasts 24 hours from the first request to /sitemap.xml, although we can raise or lower this as desired. My back-of-the envelope math says it will take around 45 seconds to respond to the first request assuming about 45k items in the repository. A nightly cron task to curl the URL would eliminate any lag for most users.

- [x] Get the modified date indexed into Solr
- [x]  Theses.
- [x]  Item visibility (filter out private items)
- [x]  Add a static robots.txt file?
- [x]  Other requirements from Google Scholar requirements: #234 
- [x]  Change SitemapPolicy to simply inherit from ApplicationPolicy.
- [x]  Add an application specific property on fileset to include hash, file size, and mimetype information.

mbarnett:
@weiweishi As discussed, this wouldn't be a bad first place for Tricia to get her feet wet, depending on what other requirements come from Peter.

Caching currently lasts 24 hours from the first request to /sitemap.xml, although we can raise or lower this as desired. My back-of-the envelope math says it will take around 45 seconds to respond to the first request assuming about 45k items in the repository. A nightly cron task to curl the URL would eliminate any lag for most users.

Remaining work:

- [x] Get the modified date indexed into Solr 
- [ ] Theses.
- [x] Item visibility (filter out private items)
- [x] Add a static robots.txt file?
- [x] Other requirements from Peter (I think #234 is the tracking ticket?)
- [x] Change SitemapPolicy to simply inherit from ApplicationPolicy.

Closes #9

### DIFF
--- a/app/policies/sitemap_policy.rb
+++ b/app/policies/sitemap_policy.rb
@@ -1,0 +1,7 @@
+class SitemapPolicy < LockedLdpObjectPolicy
+
+  def index?
+    true
+  end
+
+end
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -1,0 +1,13 @@
+class SitemapController < ApplicationController
+
+  def index
+    authorize :sitemap, :index?
+
+    @cache_key = Time.current
+
+    @communities = Community.all
+    @collections = Collection.all
+    @items = Item.all # should be non-private
+    # TODO: Theses
+  end
+end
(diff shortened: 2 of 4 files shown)


app/policies/sitemap_policy.rb
```diff
@@ -0,0 +1,7 @@
+class SitemapPolicy < LockedLdpObjectPolicy
```
murny: Very minor but this probably doesn't need to inherit from LockedLdpObjectPolicy here, can just use ApplicationPolicy instead

mbarnett: Good point. I'll add that to the list.

mbarnett:
Superseded by #379

PR #342, opened 2017-12-21
Ordered creators hack

### LINKED ISSUE #333: Ordering on Creators
It is critical to have orders in the multivalue field: creators. Fedora currently prevents the values to be saved in a specific order, and creators are presented on the UI in random order. 
This is a high priority and we need to figure out an intermediate solution to work around this issue for our first release.

mbarnett:
Adds a 'json_array' type for single-valued attributes, that appears to be an array for all outside uses, but serializes as a json string into Fedora. Updates the `creators` attribute to use this, so that creators can maintain a stable order.

Closes #333


test/models/jupiter_core/locked_ldp_object_test.rb
```diff
@@ -60,6 +60,29 @@ def unlocked_method_dont_let_locked_methods_mutate(attempted_title)
                  @@klass.attribute_names.sort
   end
 
+  test 'json_array attributess appear to be normal arrays, and maintain order' do
```
cwant: `s/attributess/attributes/`

mbarnett: This is actually a test for Attributesses, the wives of feudal Attributes

cwant:
@ualbertalib/metadata I asked in slack whether it was appropriate to store json in a triple with predicate `dc:creator`. @zschoenb suggested that `bibo:authorList` might be more appropriate for this data. Should we stick with json in `dc:creator` or move it to `bibo:authorList` for ordering the authors?

mbarnett:
Discussion of Chris' question is on #333

mbarnett:
Thoughts around refactoring related to the new requirements on this are in #344

app/models/item.rb
```diff
@@ -5,7 +5,9 @@ class Item < JupiterCore::LockedLdpObject
   ldp_object_includes Hydra::Works::WorkBehavior
 
   # Contributors (faceted in `all_contributors`)
-  has_multival_attribute :creators, ::RDF::Vocab::DC11.creator, solrize_for: [:search]
+  has_attribute :creators, RDF::Vocab::BIBO.authorList, type: :json_array, solrize_for: [:search]
```
pgwillia: There's a bunch of places where `solrize_for: :search` is used when I wonder if `solrize_for: :displayable` is meant. Is this that?

mbarnett: We don't have an equivalent to `displayable` right now (I'm not actually super clear on what that was used for in Sufia/Blacklight, actually). There's only 5 valid values for `solrize_for` currently: `:search, :sort, :facet, :exact_match, :pathing`

This is what they map to in Solrizer:

https://github.com/ualbertalib/jupiter/blob/master/app/models/jupiter_core/locked_ldp_object.rb#L26

pgwillia: FYI: from http://www.solrtutorial.com/basic-solr-concepts.html
> Term Storage
> 
> Well, when we are displaying search results to users, they generally expect to see the original document, not the machine-processed tokens (which may bear very little resemblance to the source text).
> 
> That's the purpose of the stored attribute: to tell Solr to store the original text in the index somewhere.
> Sometimes, there are fields which aren't searched, but need to be displayed in the search results. You accomplish that by setting the field attributes to stored=true and indexed=false.
> 
> So, why wouldn't you store all the fields all the time?
> 
> Because storing fields increases the size of the index, and the larger the index, the slower the search. In terms of physical computing, we'd say that a larger index requires more disk seeks to get to the same amount of data.

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -413,30 +413,9 @@ def ldp_object=(obj)
       @ldp_object
     end
 
+    # make it easier for instances to call the class method
```
pgwillia: I'm not sure I understand the distinction or why `self.class` is needed?

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -630,6 +643,7 @@ def define_cached_reader(name, multiple:, type:, canonical_solr_name:, specializ
                   solr_representation[canonical_solr_name]
                 end
           return if val.nil?
+          return val.freeze if type == :json_array # conversion happened in the overriden AF reader
```
pgwillia: What's the purpose of freeze?

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -769,21 +786,23 @@ def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :stri
         raise PropertyInvalidError if solrize_for.count { |item| !SOLR_DESCRIPTOR_MAP.keys.include?(item) } > 0
 
         raise PropertyInvalidError, "Unknown type #{type}" unless [:string, :text, :path, :bool, :date, :int,
-                                                                   :float].include?(type)
+                                                                   :float, :json_array].include?(type)
 
         self.attribute_names << name
 
+        solr_type = (type == :json_array ? :string : type)
```
pgwillia: I feel like this might need a comment

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -872,14 +891,25 @@ def has_attribute(name, predicate, multiple: false, solrize_for: [], type: :stri
...
+          # To work around a lack of ordering in multivalued attributes in Fedora, we need to serialize the :creators
+          # array into a single-valued string, so that we can maintain a fixed order. We corerce it and other
+          # "json_array" properties back into arrays when reading them.
+          if type == :json_array
+            alias_method :"shadowed_#{name}", :"#{name}"
+
+            define_method name.to_s do
+              owning_class.coerce_value(self.send("shadowed_#{name}"), to: :json_array)
+            end
+          end
```
pgwillia: The comment is good -- otherwise I'd have no idea what was happening here.  But I still don't really get how this works. Could you fill me in?

app/models/jupiter_core/locked_ldp_object.rb
mbarnett: There's an instance method `coerce_value`, and a Class method (static method in Java terminology). The instance method just forwards the call to class method ala https://stackoverflow.com/a/2527911/185104

The answer there doesn't mention it, but using `self.class` instead of the literal class name, Truck in that example, means that if a child class overrides the class method, the instance method on the child class will call the child class' override instead of forwarding the call to the base class' class method

ie given:

```ruby
class A
  def self.foo
     puts "in A's class method"
  end

  def foo
    self.class.foo
  end
end

class B
  def self.foo
     puts "in B's class method"
  end

  def foo
    B.foo
  end
end

class C < A
  def self.foo
    puts "in C's class method"
  end
end

class D < B
  def self.foo
    puts "in D's class method"
  end
end
```

then

```
2.2.1 :032 > C.new.foo
in C's class method
 => nil 
2.2.1 :033 > D.new.foo
in B's class method
 => nil 
```

Using `self.class` results in less surprising behaviour when inherited.

PR #357, opened 2018-01-03
found documentation at a different location

pgwillia:
I think this is the correct location of the generated documentation.


pgwillia:
No problem. I'll ask on Slack first for future issues.

PR #359, opened 2018-01-04
[WIP] Metadata additions and changes

cwant:
Included:

* For `Item`, change attribute `derived_from` to `source`;
* Add attribute `identifiers` to `Item`;
* Move attribute `fedora3_uuid` from `Item`, `Community`, `Collection` to `JupiterCore:LockedLdpObject`.

To do:
- [ ] Add `identifiers` to Item view (decision needed from stakeholders).

### DIFF
--- a/test/models/jupiter_core/locked_ldp_object_test.rb
+++ b/test/models/jupiter_core/locked_ldp_object_test.rb
@@ -56,8 +56,8 @@ class LockedLdpObjectTest < ActiveSupport::TestCase
   end
 
   test 'attribute definitions are working' do
-    assert_equal [:creator, :date_ingested, :hydra_noid, :id, :member_of_paths, :owner, :record_created_at, :title,
-                  :visibility], @@klass.attribute_names.sort
+    assert_equal [:creator, :date_ingested, :fedora3_uuid, :hydra_noid, :id, :member_of_paths, :owner,
+                  :record_created_at, :title, :visibility], @@klass.attribute_names.sort
   end
 
   test 'attribute metadata is being tracked properly' do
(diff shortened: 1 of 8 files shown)


test/models/jupiter_core/locked_ldp_object_test.rb
```diff
@@ -173,19 +173,20 @@ def unlocked_method_dont_let_locked_methods_mutate(attempted_title)
     title = generate_random_string
     obj = @@klass.new_locked_ldp_object(title: title)
     assert_equal '#<AnonymousClass id: nil, visibility: nil, owner: nil, record_created_at: nil, hydra_noid: nil, '\
-                 "date_ingested: nil, title: \"#{title}\", creator: [], member_of_paths: []>", obj.inspect
```
cwant: @mbarnett Since `fedora3_uuid` is in `Item`, `Collection`, `Community` (and soon to be `Thesis`), I have moved the attribute to `LockedLdpObject`.

I am beginning to question this decision, and think it might be best to load some of these things into a shared concern instead of putting them in the base class. At least with these things being in a concern, it will reduce churn on tests such as the one on this line of code (and several other tests in this file). Any opinions about this? (And a similar comment for `hydra_noid` and `date_ingested`)

mbarnett: The churn is annoying but for what this test was originally written for (things expected to be shared by all objects not making it into the debug description string), it's probably valid if we expect these to be shared by all subclasses. 

Putting them into a concern and then assuming they're present everywhere but not testing that by making the synthetic test class here include it feels like kind of a cheat?

cwant: Okay, I'm fine either way. Looks like `title` is also used for all of the classes, so will pull it into `LockedLdpObject` too then (and rework tests, since that attribute is used as an example of an attribute added in a subclass).

cwant: Shoot, after staring at this a while, I've just realized that I can't (or at least shouldn't) put a lot of the stuff into `LockedLdpObject` because of the odd duck in the family: `FileSet`.

mbarnett: If #361 works, we could shim in an intermediate class in the hierarchy? Or fall back on the concern approach if that's overly ugly.

cwant:
I think I'm going to pull the plug on this one because:

* `dcterms:identifier` has been removed from the data dictionary;
* Moving `fedora3_uuid` to `LockedLdpObject` doesn't make sense because `FileSet` doesn't need it.

The only thing left if this PR is renaming `derived_from` to `source`, and I can perhaps slip that in with the thesis modelling PR.

PR #364, opened 2018-01-05
 Custom URI Validator and validation refactors

mbarnett:
Adds a Rails URI validator:

```ruby
```

Also refactors existing validations to simplify them by removing the URI checks, makes the names a little more explicit about what each validation is enforcing, and restructures the controlled vocab yaml files into maps, so that code-uri and uri-code mappings don't incur a linear search-and-compare through arrays.


mbarnett:
(we should probably merge #363 first and then I can accommodate those changes in this branch)

db/seeds.rb
```diff
@@ -100,7 +100,7 @@
                       end
         # Probabilistically about 70% English, 20% French, 10% Ukrainian
         languages = if seed % 10 > 2
-                      [CONTROLLED_VOCABULARIES[:language].eng]
+                      [CONTROLLED_VOCABULARIES[:language].english]
                     elsif seed % 10 > 0
                       [CONTROLLED_VOCABULARIES[:language].fre]
```
cwant: This `fre` didn't get converted over . There is likely a `ukr` in that file somewhere too (maybe a no linguistic content as well)

cwant:
Lots of things to like in this PR. A few things that I am concerned about:

1) I personally don't see value in having the language codes as full strings (e.g., `english`), particularly since the codes that were there (e.g. `eng`) come from `iso639-2`, so they aren't just arbitrary abbreviations. Not a show stopper though.

2) I suspect that we will already see some parts of this PR get stretched to the breaking point. For example, in PR #363 we have the license field populated with licenses that are legacy. The main issue there is that we are getting some licenses that we no longer support as an option coming in from imported records. Maybe this could be handled if the validator took an array? e.g., `validates :license, uri: { in_vocabulary: [:license, :old_license] }`. Another option might be to crunch the `license` and `old_license` controlled vocabularies together in a single YAML, then handle which ones are displayed on forms in the models.

mbarnett:
Yeah, I want to let #363 get merged and then see how this has to change to accommodate the new use-cases -- making it accommodate multiple vocabs should be reasonably straightforward.

re: 1) I've gone back and forth on it, but ultimately I can't get over the fact we're creating method names from this, and best practices for method names basically say "make them meaningful to the reader" and not "you want to know what a call to method `.zxx` means? Just read this ISO on language abbreviations!". We're calling these all over the tests, and I prefer being able to read the tests without having to refer to a supplementary chart.

cwant:
#363 has been merged now.

mbarnett:
Cool I'll tackle merging those changes into this tomorrow.

mbarnett:
Superseded by #386

PR #369, opened 2018-01-09
Issue #367, rework Item `item_type` and `publication_status`.

### LINKED ISSUE #367: Item type `article` and it's (publication) `status`(es)
Current `Item` model has required attribute `item_type` (`dcterms:type`). When that is set to the URI representing `article`, an additional `publication_status` (`bibo:status`) must be included (otherwise, `publication_status` must be excluded).

Currently `publication_status` is single-valued, but should be multi-valued. It should have two states:

* either the status is URI representing `published`;
* or two URIs must be present: one representing `draft`, and other representing `submitted`

Validations should be updated, as should test/seeds.

@weiweishi @ualbertalib/metadata does this match your understanding of how this should work?

cwant:
Now `publication_status` is multivalued and has only two acceptable
configurations for articles.

Fixes #367

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -54,7 +54,8 @@ class Item < JupiterCore::LockedLdpObject
   has_attribute :doi, ::TERMS[:prism].doi, solrize_for: :exact_match
 
   # Bibo attributes
-  has_attribute :publication_status, ::TERMS[:bibo].status, solrize_for: :exact_match
+  # This status is only for articles: either 'published' (alone) or two triples for 'draft'/'submitted'
+  has_multival_attribute :publication_status, ::TERMS[:bibo].status, solrize_for: :exact_match
 
   # Project Hydra ACL attributes
   has_multival_attribute :embargo_history, ::TERMS[:acl].embargoHistory, solrize_for: :exact_match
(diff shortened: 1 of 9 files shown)


app/models/item.rb
```diff
@@ -54,7 +54,8 @@ class Item < JupiterCore::LockedLdpObject
   has_attribute :doi, ::TERMS[:prism].doi, solrize_for: :exact_match
 
   # Bibo attributes
-  has_attribute :publication_status, ::TERMS[:bibo].status, solrize_for: :exact_match
+  # This status is only for articles: either 'published' (alone) or two triples for 'draft'/'submitted'
+  has_multival_attribute :publication_status, ::TERMS[:bibo].status, solrize_for: :exact_match
```
mbarnett: Just checking that it's not going to be a problem that the two triples 'draft'/'submitted' are going to be in an  arbitrary order?

app/models/item.rb
```diff
@@ -92,7 +93,9 @@ def item_type_with_status_code
     item_type_code = CONTROLLED_VOCABULARIES[:item_type].uri_to_code(item_type)
     return item_type_code unless item_type_code == 'article'
     return nil if publication_status.blank?
-    publication_status_code = CONTROLLED_VOCABULARIES[:publication_status].uri_to_code(publication_status)
+    publication_status_code = CONTROLLED_VOCABULARIES[:publication_status].uri_to_code(publication_status.first)
```
mbarnett: like here there's no guarantee that this isn't going to come back out as [submitted, draft] instead of [draft, submitted]

app/models/item.rb
```diff
@@ -253,7 +256,17 @@ def item_type_and_publication_status_validations
         if publication_status.blank?
           errors.add(:publication_status, :required_for_article)
         else
-          uri_validation(publication_status, :publication_status)
+          begin
+            # Complication: need either 'published' alone or 'draft' and 'submitted' together
+            statuses = publication_status.map do |status|
+              CONTROLLED_VOCABULARIES[:publication_status].uri_to_code(status)
+            end.sort
+            if statuses != ['published'] && statuses != ['draft', 'submitted']
```
mbarnett: ditto here

app/models/item.rb
cwant: Here it doesn't matter, thanks to the next line of code.

app/models/item.rb
cwant: There is a .sort there to ensure the order.

app/models/item.rb
cwant: These get combined with `item_type` for discovery (via faceting). I don't thing the order is an issue here.

app/models/item.rb
mbarnett: Ah gotcha, missed the sort at the end there.

PR #370, opened 2018-01-09
Issue #98, user profile/dashboard.

### LINKED ISSUE #98: User Dashboard
TODO

Mockup:
Image of User Dashboard Mockup
https://jupiter.mybalsamiq.com/projects/jupiter/User%20Dashboard

cwant:
Fixes #98


app/policies/application_policy.rb
```diff
@@ -40,6 +40,10 @@ def admin?
     user.try(:admin?)
   end
 
+  def logged_in?
+    !user&.suspended?
```
mbarnett: I think this returns true if user is nil? 

```ruby
user = nil
user&.suspended?
=> nil
!user&.suspended?
=> true
```

Not quite what I'd expect from this method (and I think it just leads to a crash or weird behaviour on the profile when user is nil, after they've been authorized).

I think we want

```ruby
user.present? && !user.suspended?
```

cwant: Yeah, thanks for catching that.

murny: You can probably just drop the suspended clause too as this will never be true here. 

For `user.present?` to be true, user cannot be suspended. If they are they get logged out and handled accordingly. A

ll done in the `current_user` method which is what pundit uses for `user`: https://github.com/ualbertalib/jupiter/blob/master/app/controllers/application_controller.rb#L22

cwant: Yeah, sounds good.

app/policies/profile_policy.rb
```diff
@@ -0,0 +1,7 @@
+class ProfilePolicy < ApplicationPolicy
+
+  def index?
+    logged_in?
```
murny: Pundit deals more with access controls around a resource object. So unless you have a profile object around this looks really weird. As Pundit isn't a direct mapping between controllers but instead models. 

Also what your trying to achieve here is more authentication then authorisation.

So better approach? If you want to keep using Pundit, throw this in the user policy or just use the application policy?
Your ensure login now becomes: 
```ruby
def ensure_logged_in
     authorize :application, :logged_in? # or :user, :logged_in?
   end
```

Or just don't even use pundit and define that ensure_logged_in as simply, `current_user.present?`

cwant: I have to perform a check with pundit, otherwise rails throws `Pundit::AuthorizationNotPerformedError`.

I don't mind ditching the profile policy though (definitely seemed a bit redundant).

app/views/profile/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  </dd>
+
+  <dt class="col-sm-3"><%= t('admin.users.email')%></dt>
+  <dd class="col-sm-9">
+    <%= @user.email %>
+  </dd>
+
+</dl>
+
+<%= render partial: 'admin/users/user_items',
```
murny: should inverse this? admin uses general code, instead of general using admin code. As you may see the view `admin/users/user_items`  and think its only being used in admin context which here this isn't true anymore.

app/controllers/profile_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def index
+    @user = current_user
+    item_search_setup(Item.search_term_for(:owner, @user.id, role: :exact_match))
+  end
+
+  private
+
+  def ensure_logged_in
+    authorize :user, :logged_in?
```
murny: Very minor but what's your thoughts on, considering there is only one action in this file (and probably will only be one ever? Even if we added a settings section or something this controller probably would get refactored anyways), you can reduce 5 lines of code and just stick this authorise call right in the index action?

PR #373, opened 2018-01-11
Thesis modelling, and metadata mixed bag

cwant:
* share properties with `Collection`/`Community`/`Item`/`Thesis` via concern;
* handling shared properties with `Item` via concern;
* create attributes distinct to `Thesis`.
* rename `Item` property `derived_from` to `source`;
* solr index for `item_type_with_status` shared with `Item` for facetting;
* derived property `sort_year` derived from graduation date for `Thesis`;
* `bibo:status` provided by `rdf-vocab`, so remove from `terms.yml`
* move `Community` to `ual` vocabulary namespace (uppercase intentional);
* remove fake `jupiter_core` vocabulary namespace.

[Metadata requirements](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_thesis.md)

[View requirements](https://jupiter.mybalsamiq.com/projects/jupiter/Thesis%20Detail)

# TODO:

- [x] Tests
- [x] Thesis detail view

### Shared with everything except FileSet (i.e., in Collection/Community/Item/Thesis)

- [x] title
  - [x] **Similar to item, collection, community, etc**
  - [x] dcterms:title
  - [x] single value
  - [x] searchable, sortable, not facetable
  - [x] required

- [x] fedora3_uuid
  - [x] same as the property in item (predicate: ual:fedora3UUID)

- [x] ~~hydraNoid (Please also add this for item/collection/community)~~
  - [x] Already in `LockedLdpObject

- [x] depositor
  - [x] predicate: http://terms.library.ualberta.ca/depositor
  - [x] similar to depositor for item
  - [x] not required
  - [x] not sortable, searchable or facetable

### Shared with Item

- [x] memberOf
  - [x] **similar to memberOf for item -> collection, member_of_collections relationship**
  - [x] predicate: PCDM::memberOf
  - [x] ~~must belong to one of two possible collection/community pairs~~ See #378
  - [x] multi-value,  value: URI to parent collections
  - [x] searchable, facetable, not sortable
  - [x] required

- [x] doi
  - [x] **similiar to doi for item**
  - [x] predicate: prism:doi
  - [x] single value
  - [x] not required
  - [x] searchable, not facetable, not sortable

- [x] rights
  - [x] similar to rights for item (**BUT different validation**)
  - [x] predicate: dc:rights
  - [x] single value
  - [x] not required
  - [x] not searchable, sortable or facetable

- [x] subject (can't use `subjects` -- reserved)
  - [x] similar to subjects for item (**BUT different validation**)
  - [x] predicate: dc:subject
  - [x] multi-value
  - [x] not required
  - [x] searchable, facetable, not sortable

- [x] alternative_title
  - [x] **similar to Item**
  - [x] predicate: dcterms:alternative
  - [x] single value
  - [x] not required
  - [x] searchable, not sortable or facetable

- [x] ~~identifier~~ (no longer in data dictionary)
  - [x] predicate: dcterms:identifier
  - [x] single_value
  - [x] not required
  - [x] not sortable or facetable

- [x] is_version_of
  - [x] **similar to is_version_of for Item**
  - [x] dcterms:isVersionOf
  - [x] single value
  - [x] not searable, sortable or facetable
  - [x] not required

- [x] fedora3_handle
  - [x] same as the property in item (predicate: ual:fedora3Handle)

- [x] ingestBatch
  - [x] same as the property in item (predicate: ual:ingestBatch)

- [x] embargo_end_date
- [x] embargo_history
- [x] visibility_after_embargo

### New attributes

- [x] Institution
  - [x] single-value, value: URI
  - [x] predicate: http://ontoware.org/swrc/ontology#institution
  - [x] controlled vocabulary
    - University of Alberta (http://id.loc.gov/authorities/names/n79058482)
    - St. Stephen's College (http://id.loc.gov/authorities/names/n2009054054)
  - [x] not facetable, sortable
  - [x] not required

- [x] language
  - [x] should be handled similairly to languages for item (**BUT single-valued and different validation**)
  - [x] dcterms:language
  - [x] value is from controlled vocabulary and saved as URI
  - [x] not required
  - [x] facetable, not sortable

- [x] abstract
  - [x] predicate: dcterms:abstract
  - [x] single-value
  ~~- [x] should index as "description"~~ (We don't have advance search or facet on this field, so no longer relevant)
  - [x] not required
  - [x] searchable, not sortable or facetable

- [x] date_accepted
  - [x] predicate: dcterms:dateAccepted
  - [x] single value
  - [x] not required
  - [x] not  sortable or facetable
 
- [x] date_submitted
  - [x] predicate: dcterms:dateSubmitted
  - [x] single value
  - [x] not required
  - [x] not sortable or facetable

- [x] degree
  - [x] predicate:  bibo:degree
  - [x] single value
  - [x] not required
  - [x] not facetable, sortable

- [x] committee_member
  - [x] predicate:  http://terms.library.ualberta.ca/committeeMember
  - [x] multi-value
  - [x] not required
  - [x] not searchable, sortable or facetable. 

- [x] department
  - [x] predicate: http://terms.library.ualberta.ca/department
  - [x] multi-value
  - [x] not required
  - [x] not sortable or facetable

- [x] dissertant
  - [x] predicate: http://terms.library.ualberta.ca/dissertant
  - [x] required
  - [x] single value
  - [x] facet together with creator/contributor
 
- [x] graduationDate
  - [x] predicate: http://terms.library.ualberta.ca/graduationDate
  - [x] required
  - [x] single value
  - [x] facet together with item's created property (treated as created for Thesis, will likely to have sort_year, value comes to us in this format: 2017-06)

- [x] specialization
  - [x] predicate: http://terms.library.ualberta.ca/specialization
  - [x] not required
  - [x] multi-value
 
- [x] supervisor
  - [x] predicate: http://terms.library.ualberta.ca/supervisor
  - [x] not required
  - [x] multivalue

- [x] thesis_level
  - [x] predicate: http://terms.library.ualberta.ca/thesisLevel
  - [x] not required
  - [x] single value

- [x] proquest
  - [x] predicate: http://terms.library.ualberta.ca/proquest
  - [x] not required
  - [x] single value

- [x] unicorn
  - [x] predicate: http://terms.library.ualberta.ca/unicorn
  - [x] not required
  - [x] single value

- [x] rdf:type
  - [x] always set to `bibo:thesis`


app/models/concerns/common_item_properties.rb
```diff
@@ -0,0 +1,160 @@
+module CommonItemProperties
```
mbarnett: `CommonItemProperties` and `CommonObjectProperties` are a bit confusingly named (I couldn't figure out why collections had all of this stuff, at first, until I realized I wasn't reading the file I thought I was reading).

I think the problem is that the important word is buried in the middle of the long name. Let's drop the Common prefix.

app/models/concerns/common_object_properties.rb
```diff
@@ -0,0 +1,18 @@
+module CommonObjectProperties
```
mbarnett: Same comment as above.

app/models/concerns/common_object_properties.rb
```diff
@@ -0,0 +1,18 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    ldp_object_includes Hydra::Works::WorkBehavior
```
mbarnett: Collections and Communities should not include Work behaviors.

Actually, a big problem in Sufia development was sorting out the rat's nest of nested includes and included includes, etc, etc, ad infinitum. Let's do this explicitly in each concrete class, rather than burying it in a concern. We can document the requirement that A gets included prior to B (which is something _else_ that Sufia failed to do).

app/models/concerns/common_item_properties.rb
```diff
@@ -0,0 +1,160 @@
...
+  extend ActiveSupport::Concern
+
+  VISIBILITY_EMBARGO = CONTROLLED_VOCABULARIES[:visibility].embargo.freeze
+  VISIBILITIES = (JupiterCore::VISIBILITIES + [VISIBILITY_EMBARGO]).freeze
+  VISIBILITIES_AFTER_EMBARGO = [CONTROLLED_VOCABULARIES[:visibility].authenticated,
+                                CONTROLLED_VOCABULARIES[:visibility].draft,
+                                CONTROLLED_VOCABULARIES[:visibility].public].freeze
+
+  included do
+    ldp_object_includes Hydra::Works::WorkBehavior
```
mbarnett: See comment below about bad experiences with nesting includes.

app/models/item.rb
```diff
@@ -101,49 +63,21 @@ def item_type_with_status_code
...
     validates :creators, presence: true
-    validate :communities_and_collections_validations
     validate :language_validations
     validate :license_and_rights_validations
-    validate :visibility_after_embargo_validations
     validate :item_type_and_publication_status_validations
 
     before_validation do
-      # TODO: for theses, the sort_year attribute should be derived from ual:graduationDate
+      # Note: for Thesis, the sort_year attribute is derived from ual:graduationDate
```
mbarnett: We can lose this whole comment, and document Thesis behaviour in the Thesis model now that it exists.

app/views/items/show.html.erb
```diff
@@ -43,115 +43,10 @@
   </div>
 
   <%# MAIN CONTENT %>
-  <div>
-    <h1 class="mt-3" title="<%= t('.attributes.title') %>"><%= @item.title %></h1>
+  <% if @item.is_a?(Thesis) %>
+    <%= render partial: 'thesis_show' %>
+  <% else %>
+    <%= render partial: 'item_show' %>
+  <% end %>
```
mbarnett: If you leverage partial paths correctly, you don't need to do this.

`render @item` calls `@item.to_partial_path` to figure out which partial to use. By default this will return `items/_item.html.erb` for Item instances, and `theses/_thesis.html.erb` for Thesis instances.

If you override `to_partial_path` in each class to shave the controller name off

(eg. in item.rb:
```ruby
def to_partial_path
  'item'
end
```
 and the same thing with `'thesis'` in thesis.rb)

Then the partials become controller-relative. `render @item` would here use 

 `app/views/items/_item.html.erb` and `app/views/items/_thesis.html.erb`

app/views/search/_model_results.html.erb
```diff
@@ -49,8 +49,10 @@
       <ul class="list-group">
         <% results.each do |result| %>
           <li class="list-group-item list-group-item-action">
-            <% if model == :item %>
+            <% if result.is_a?(Item) %>
               <%= render partial: 'search/item_hit', locals: { result: result } %>
+            <% elsif result.is_a?(Thesis) %>
+              <%= render partial: 'search/thesis_hit', locals: { result: result } %>
```
mbarnett: Same comment here re: leveraging partial_paths. Assuming you override the method as above, `render result` will use `app/views/search/_item.html.erb` and `app/views/search/_thesis.html.erb`

cwant: While this works fine for results fetched through the search controller,  this partial gets included *everywhere* for rendering search results. And partial paths are relative to the controller, not the view

For example when viewing a collection, this partial shows the items contained in the collection. The problem there is that it looks for the partial relative to the controller, so it looks for `app/views/collection/_item.html.erb` instead of `app/views/search/_item.html.erb`.

I don't see an obvious way to tell rails to include a partial relative to the including partial, without including the full partial name (which defeats the purpose of the ruby magic figuring out the partial name based on the class).

mbarnett: Put them under `app/views/application/`. Partial resolution "falls" up the controller inheritance scheme. This way all controllers share a common partial unless there's a more specific implementation.

cwant: That did the trick! I put the `to_partial_path` in `LockedLdpObject` -- let me know if you think that will cause problems.

mbarnett: I _think_ that should be ok (knock on wood)

app/helpers/items_helper.rb
```diff
@@ -5,9 +5,9 @@ def item_type_search_link(item)
                     display: t("controlled_vocabularies.item_type_with_status.#{item.item_type_with_status_code}"))
   end
 
-  def language_search_link(item, language_uri)
-    search_link_for(item, :languages, value: language_uri,
-                                      display: CONTROLLED_VOCABULARIES[:language].uri_to_text(language_uri))
+  def language_search_link(item, language_uri, attribute: :languages)
```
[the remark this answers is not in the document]
cwant: Yeah, wasn't necessary -- I've since removed it.

app/views/items/_thesis_more_information.html.erb
```diff
@@ -0,0 +1,105 @@
+<ul class="list-group">
+  <% if @item.language.present? %>
+    <li class="list-unstyled list-group-item-action">
+      <dl>
+        <dt><%= t('items.show.language') %></dt>
+        <dd><%= language_search_link(@item, @item.language, attribute: :language) %></dd>
```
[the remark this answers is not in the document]
cwant: We are taking care to indicate the singlar/plural distinction. In thesis language is single-valued, in item it's multi-valued. In thesis they do get faceted together by adding an additional solr index for thesis.

PR #374, opened 2018-01-11
Add polymorphic JupiterCore::LockedLdpObject#find

mbarnett:
@cwant as discussed in Slack yesterday:

`JupiterCore::LockedLdpObject.find(id, types: [Item, Thesis])`


app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -275,14 +275,24 @@ def self.facet_term_for(attr_name, value, role: :facet)
 
     # Accepts a string id of an object in the LDP, and returns a +LockedLDPObjects+ representation of that object
     # or raises <tt>JupiterCore::ObjectNotFound</tt> if there is no object corresponding to that id
-    def self.find(id)
+    def self.find(id, types: [])
+      if self == LockedLdpObject
+        raise ArgumentError, 'Must specify at least one type to find' if types.blank?
```
cwant: One thing that might cause confusion: the error you raise asks to specify a 'type', but your keyword argument is called 'types'

mbarnett:
Tweaked the error message a bit

PR #379, opened 2018-01-15
Sitemap & Google Scholar

### LINKED ISSUE #9: Port sitemap work

pgwillia:
Replaces #340 Closes #9 
- [x] Get the modified date indexed into Solr
- [x]  Item visibility (filter out private items)
- [x]  Add a static robots.txt file?
- [x]  Other requirements from Google Scholar requirements: #234 
- [x]  Change SitemapPolicy to simply inherit from ApplicationPolicy.
- [x]  Add an application specific property on fileset to include hash, file size, and mimetype information. 
- [x]  Theses.
- [x] splitting into sitemapindex and sitemap


app/views/items/show.html.erb
```diff
@@ -1,5 +1,7 @@
 <% page_title(@item.title) %>
 
+<%= render partial: 'google_scholar_metadata' %>
```
murny: These meta tags will be in the `<body>`. Should probably inject them into the `<head>` of the page instead? 

Requires some additional work to make this happen. Look into `content_for` and `yield` if you need help figuring out how to do this.

app/controllers/sitemap_controller.rb
```diff
@@ -0,0 +1,27 @@
+class SitemapController < ApplicationController
+
+  skip_after_action :verify_authorized
+
+  def index; end
+
+  def communities
+    @communities = Community.all
+    raise 'sitemap should contain less than 50,000 targets' if @communities.count > 50_000
```
mbarnett: I don't think we'll want to start serving Google a 500 page if/when any of these creeps above 50,000 without us noticing. Maybe replace `raise` with `logger.warn`? (actually even better would be to push a warning to Rollbar, but we don't have that set up yet. Maybe add a TODO: about that as well?)

app/models/concerns/item_properties.rb
```diff
@@ -134,6 +148,10 @@ def display_attribute_names
     def valid_visibilities
       super + [VISIBILITY_EMBARGO]
     end
+
+    def public
+      Item.where(visibility: JupiterCore::VISIBILITY_PUBLIC)
```
mbarnett: Since this concern is shared with Thesis, I don't think this quite works (Thesis.public will return public Items, instead). `self.where(visibility: JupiterCore::VISIBILITY_PUBLIC)` should do the appropriate thing in both classes.

app/views/items/_google_scholar_metadata.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<meta name="dc.identifier" content="<%= @item.doi %>"/>
+<meta name="citation_doi" content="<%= @item.doi %>"/>
+<!-- does not support these metadata -->
+<!--
+  <meta name="citation_journal_title" content=""/>
+  <meta name="citation_volume" content=""/>
+  <meta name="citation_issue" content=""/>
+  <meta name="citation_firstpage" content=""/>
+  <meta name="citation_lastpage" content=""/>
+-->
```
mbarnett: Do we actually want to output this whole set of comments into the HTML, or do we want this comment not to render at all? (Use <%# %> is the latter).

app/models/concerns/item_properties.rb
```diff
@@ -110,6 +110,20 @@ def add_files(files)
...
+                                                      action: :download,
+                                                      id: unlocked_fileset.id,
+                                                      file_name: unlocked_fileset.contained_filename,
+                                                      only_path: true)}\" \
+rel=\"content\" \
+hash=\"#{unlocked_fileset.original_file.checksum.algorithm.downcase}:"\
+"#{unlocked_fileset.original_file.checksum.value}\" \
+length=\"#{unlocked_fileset.original_file.size}\" \
+type=\"#{unlocked_fileset.original_file.mime_type}\"\
+/>"
```
mbarnett: Rather than have to escape all of these newlines, this is probably a great use-case for a heredoc: https://infinum.co/the-capsized-eight/multiline-strings-ruby-2-3-0-the-squiggly-heredoc

app/models/concerns/item_properties.rb
mbarnett: I'd also love to see this get exercised in the test suite. Probably just add another assertion to these two existing test cases: https://github.com/ualbertalib/jupiter/blob/master/test/models/thesis_test.rb#L5 and https://github.com/ualbertalib/jupiter/blob/master/test/models/item_test.rb#L5

mbarnett:
One other thing I'd love to see would be an added integration test exercising the sitemap.

app/models/concerns/item_properties.rb
pgwillia: This is actually more like a run-on sentence situation.  There should be **no newlines** or extra spaces/tabs between the values. My understanding is that heredoc would maintain newlines. Is there a solution to make something like that more readable?

config/environments/production.rb
```diff
@@ -86,4 +86,7 @@
 
   # Which ActiveStorage service to use
   config.active_storage.service = (ENV['ACTIVE_STORAGE_SERVICE'] || :local).to_sym
+
+  # we can tell crawlers to go away during the beta launch phase
+  config.allow_crawlers = ENV['RAILS_ALLOW_CRAWLERS'].present?
```
murny: For future reference:
Could have defined this in `config/application.rb` once instead of doing this in every environment especially when they are the same.

test/controllers/resourcesync.xsd
```diff
@@ -0,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
murny: Why does this resourcesync stuff living in `test/controllers`? Is it being used as a file? Maybe `test/fixtures/file` a better place?

test/controllers/sitemap.xsd
```diff
@@ -0,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
murny: Same for this one? Don't see anything reference these two files in your PR either?

test/controllers/sitemap_controller_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert_select 'loc', /sitemap-collections.xml/
+    end
+  end
+
+  context 'items sitemap' do
+    setup do
+      get items_sitemap_url
+    end
+    should 'be valid sitemap xml' do
+      schema = Nokogiri::XML::Schema(File.open(File.join(File.dirname(__FILE__), 'sitemap.xsd')))
```
murny: Ah your using it here. Yeah should look into file_fixtures and put them in `test/fixtures/files`

test/controllers/sitemap_controller_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert_select 'loc', /sitemap-collections.xml/
+    end
+  end
+
+  context 'items sitemap' do
+    setup do
+      get items_sitemap_url
+    end
+    should 'be valid sitemap xml' do
+      schema = Nokogiri::XML::Schema(File.open(Rails.root + 'test/fixtures/files/sitemap.xsd'))
```
murny: Nokogiri isn't defined in our Gemfile (might be getting this from another gem, but regardless we should add it in case this other gem removes it down the road and now our test suite fails).

test/controllers/sitemap_controller_test.rb
```diff
@@ -58,7 +58,7 @@ def before_all
       get sitemapindex_url
     end
     should 'be valid sitemapindex xml' do
-      schema = Nokogiri::XML::Schema(File.open(File.join(File.dirname(__FILE__), 'siteindex.xsd')))
+      schema = Nokogiri::XML::Schema(File.open(Rails.root + 'test/fixtures/files/siteindex.xsd'))
```
murny: This can get even easier here. You should be able to just call:
```ruby
file_fixture('siteindex.xsd')
```

http://api.rubyonrails.org/v5.1/classes/ActiveSupport/Testing/FileFixtures.html

test/controllers/sitemap_controller_test.rb
```diff
@@ -0,0 +1,162 @@
...
+      assert_select 'url' do
+        assert_select 'loc'
+        assert_select 'lastmod'
+        assert_select 'changefreq'
+        assert_select 'priority'
+      end
+    end
+    should 'show location and last modified' do
+      assert_select 'loc', community_collection_url(@collection.community, @collection)
+      assert_select 'lastmod', @collection.updated_at.to_s\
```
murny: Curious, why the `\` at the end of `updated_at.to_s`? Happens a couple times in this file

pgwillia: Typo?!

config/routes.rb
```diff
@@ -62,5 +62,14 @@
...
   end
 
+  get 'sitemap.xml', to: 'sitemap#index', defaults: { format: :xml }, as: :sitemapindex
+  get 'sitemap-communities.xml', to: 'sitemap#communities', defaults: { format: :xml }, as: :communities_sitemap
+  get 'sitemap-collections.xml', to: 'sitemap#collections', defaults: { format: :xml }, as: :collections_sitemap
+  get 'sitemap-items.xml', to: 'sitemap#items', defaults: { format: :xml }, as: :items_sitemap
+  get 'sitemap-theses.xml', to: 'sitemap#theses', defaults: { format: :xml }
+
+  # Dynamic robots.txt
+  get 'robots.:format' => 'robots#robots'
```
murny: Should we just limit it to `robots.txt` here instead of allowing any format extension? I assume it doesn't matter since in the controller action for this it only has a `responds_to :text` anyways?

test/fixtures/files/resourcesync.xsd
```diff
@@ -0,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
murny: Last comment I have on this PR, this file is never used anywhere? Is this something being kept around for future? Or If it's not needed just remove it?

pgwillia: It is being used.  It's included by the sitemap.xsd.

config/routes.rb
pgwillia: Yeah, doesn't matter but the only thing that makes sense is robots.txt -- I'll change it.

PR #381, opened 2018-01-16
Issue #332, reorder facets so that selected appear first.

### LINKED ISSUE #332: Facet ordering for search results
* On the search page, ideally selected facets (and facet groups) should appear before non-selected ones
* Currently the facets for solr-only attributes appear after the other ones. The facet category that combines creators and contributors (`:all_contributors`) is a fairly important solr-only facet group, so we need to investigate ways to boost it's importance in the ordering.

cwant:
Also: in general, make author/subject appear before other when applicable
(since these are solr-only attributes, they were showing up at the end of the list of facets).

Fixes #332


app/controllers/concerns/item_search.rb
```diff
@@ -16,6 +16,8 @@ def item_search_setup(base_query = nil)
     query.append(params[:query]) if params[:query].present?
     options = { q: query, models: [Item], as: current_user }
     options[:facets] = params[:facets]
+    # Make sure selected facets and solr-only authors/subjects appear first in facet list
+    @first_facet_categories = (params[:facets]&.keys || []) + ['all_contributors_sim', 'all_subjects_sim']
```
mbarnett: We should get the solr suffix goop out of this somehow

app/models/jupiter_core/deferred_faceted_solr_query.rb
```diff
@@ -43,9 +43,25 @@ def sort(attr, order = :desc)
...
+      @facets.each do |facet|
+        if first_facet_categories.include?(facet.solr_index)
+          first_facets << facet
+        else
+          second_facets << facet
+        end
+      end
+    else
+      first_facets = @facets
+    end
```
mbarnett: You could leverage Enumerable#partition to replace lines 48-60. https://apidock.com/ruby/Enumerable/partition

```ruby
facets = @facets.partition {|facet|  first_facet_categories.include?(facet.solr_index)}.flatten
facets.each do |facet|
  yield facet if facet.present?
end
```

That's a bit easier to read.

cwant: Yeah, that's better. What I decided to do instead though was to use `sort_by` to ensure the facets indicated in `first_facet_categories` have their ordering preserved. This was to ensure that the selected ones always appear before unselected things (i.e., author and subject).

app/controllers/concerns/item_search.rb
cwant: Closest thing I can find to get rid of the goop is something like `Item.solr_name_for(:all_contributors, role: :facet)`. Still kind of goopy though. Would it be worth adding a `facet_index(key)` method to `LockedLdpObject` that wraps the above? (And hides the solr-ishness)

app/controllers/concerns/item_search.rb
```diff
@@ -16,6 +16,8 @@ def item_search_setup(base_query = nil)
     query.append(params[:query]) if params[:query].present?
     options = { q: query, models: [Item, Thesis], as: current_user }
     options[:facets] = params[:facets]
+    # Make sure selected facets and solr-only authors/subjects appear first in facet list
+    @first_facet_categories = (params[:facets]&.keys || []) + ['all_contributors_sim', 'all_subjects_sim']
```
mbarnett: So AFAIK the only holdup remaining on this PR is replacing this with a call to `solr_name_for`? I thought there was a discussion related to this, but maybe GitHub hid it...

cwant: Yeah, github hid it: https://github.com/ualbertalib/jupiter/pull/381/files/6d594e4eaee9d1d3c058557ada3740194e22951a#diff-a97ecb6d8d7c2d55dddb4c86731bf839

mbarnett: Aha!

I think for now we could just use Item.solr_name_for(:all_contributors, role: :facet). We need a better way of dealing with this overall, probably by maintaining a global view of all the "Solrized names" -- or just ripping out Solrizer and eliminating this distinction between property names and property names in Solr

app/controllers/search_controller.rb
```diff
@@ -11,6 +11,10 @@ def index
     @active_tab = params[:tab]&.to_sym || :item
     @results = {}
 
+    # Make sure selected facets and solr-only authors/subjects appear first in facet list
+    @first_facet_categories = params[:facets]&.keys || []
+    @first_facet_categories += ['all_contributors_sim', 'all_subjects_sim'] if @active_tab == :item
```
mbarnett: (missed this pair btw)

cwant: Dang it! Thanks for catching that.

PR #385, opened 2018-01-16
Issue #338, change sorting of items from `record_created_at` to `sort_year`

cwant:
Special care was needed because although all of our models have `record_created_at`,
only items (and theses) have `sort_year`.


app/views/search/_item_sort.html.erb
```diff
@@ -0,0 +1,16 @@
...
+      <%= search_sort_label(params[:sort], params[:direction]) %>
+    <% else %>
+      <%= t('search.sort_by') %>
+    <% end %>
+  </button>
+  <div class="dropdown-menu" aria-labelledby="search-sort-button">
+    <%= search_sort_link('title', 'asc') %>
+    <%= search_sort_link('title', 'desc') %>
+    <%= search_sort_link('sort_year', 'desc') %>
+    <%= search_sort_link('sort_year', 'asc') %>
```
mbarnett: Instead of having 2 almost identical partials, could we just pass a local into `_sort.html.erb` to specify the correct value for these last two links (AFAICT this is the only difference between the files)? If we don't DRY this up we'll inevitably update one file and not the other, at some point in the future.

PR #386, opened 2018-01-17
Custom URI Validator and validation refactors, round 2

mbarnett:
Adds a Rails URI validator:

```ruby
```

this also works for validating the the URI is in at least on of multiple vocabularies:

```ruby
    validates :license, uri: { in_vocabularies: [:license, :old_license] }
```

Also refactors existing validations to simplify them by removing the URI checks, splits up kitchen-sink "foo_validations" methods into meaningfully-named, single-responsibility validation methods, and restructures the controlled vocab yaml files into maps, so that code-uri and uri-code mappings don't incur a linear search-and-compare through arrays.


mbarnett:
Great success

Merge conflict resolved, plz reapprove

PR #388, opened 2018-01-18
Allow the composition of deferred query objects across models

mbarnett:
Solves https://github.com/ualbertalib/jupiter/blob/master/app/models/collection.rb#L36-L40 by allowing the composition of queries instead of forcing an expensive reification of both queries full results sets to generate the array. 

Thus given: 

```ruby 
  def member_items
    Item.where(member_of_paths: path)
  end

  def member_theses
    Thesis.where(member_of_paths: path)
  end
```

`member_items + member_theses` produces a single query object with merged search criteria:

```
#<JupiterCore::DeferredSimpleSolrQuery:0x007fe1c1f9e5e8 @criteria={:model=>[Item, Thesis], :limit=>10000000, :sort=>[], :sort_order=>[], :where=>{:member_of_paths=>"9f3bf4b4-284f-4c13-919f-19e58601f316/bd33f488-5d12-4424-8db4-544b293288b7"}}> 
```

In general any deferred queries should be composable, provided they share identical attribute names and solr forms. Pre-existing sorts, limits, and offsets are ignored and should be re-specified on the final query object (it wasn't obvious to me which side's should "win").


app/models/jupiter_core/deferred_simple_solr_query.rb
```diff
@@ -32,11 +33,22 @@ def offset(num)
 
   def sort(attr, order = :desc)
     raise ArgumentError, 'order must be :asc or :desc' unless [:asc, :desc].include?(order.to_sym)
-    criteria[:sort] << criteria[:model].solr_name_for(attr.to_sym, role: :sort)
+    criteria[:sort] << criteria[:model].first.solr_name_for(attr.to_sym, role: :sort)
     criteria[:sort_order] << order
     self
   end
 
+  def +(other)
```
cwant: I think it's pretty clear what this does when the two query terms are the same, but what happens if I do `Item.where(title: 'hamburger') + Thesis.where(title: 'cheeseburger')`? Do I get items with hamburger in the title and theses with cheeseburger in the title? Or do I get both items and theses that have either hamburger or cheeseburger in the title? It's not clear to me without diving in deep and following the code, so maybe add a comment about this (or a test might help).

mbarnett: Yeah, that occurred to me last night. I'm going to raise an error on the same key, different value case, and add some doc comments about the general semantics.

mbarnett: I've made this a lot more flexible, now. Tests still incoming, but take a look at the comment and tell me if that makes sense to you.

PR #408, opened 2018-01-29
Add active draft items to users dashboard/admin user detail

murny:
Removes TODO for user profile/admin user detail for draft items. User's can now come back and finish there saved draft items once they start editing.


app/views/profile/_user_items.html.erb
```diff
@@ -1,15 +1,59 @@
-<div class="d-flex flex-wrap align-items-center justify-content-between mt-3">
```
murny: Alot of our titles have useless flexbox styles. Probably copy and paste from elsewhere. There is only one thing in this div (the header) so no point of making sure its justified-content-between and aligned, etc.

Also this div is quite useless, we can probably move the `mt-3` right onto the header itself

app/views/profile/_user_items.html.erb
```diff
@@ -1,15 +1,59 @@
...
-  <h4><%= "(TODO) #{draft_items_header}" %></h4>
+<div class="mt-3">
+  <h2><%= draft_items_header %></h4>
```
murny: Bumped these to H2's, since H4 looks as big as some of the text on this page

app/views/profile/_user_items.html.erb
```diff
@@ -75,7 +72,5 @@
   <%= render partial: 'search/model_results', locals: { results: @results, model: :item,
                                                         search_term: params[:query] } %>
 <% else %>
-  <div>
```
murny: An div with no css on it is literally useless in most cases (unless you want display block, but probably better ways for achieving this)

PR #415, opened 2018-01-31
[WIP] Issue #413, Static pages

### LINKED ISSUE #413: Static Page setup
- [ ] set up routes/controllers/views for static "About Us" Page
 - [ ] update the content on static pages (@leahvanderjagt and @sfbetz will provide content)

cwant:
* 'deposit_path' is now 'help_path'
  * 'technology_path' points to our github repo.
  * There are no direct links to 'contact', 'help', and 'technology' on our welcome page.

Fixes #413

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,0 +1,5 @@
+class PagesController < ApplicationController
+
+  skip_after_action :verify_authorized
+
+end
--- a/app/views/application/_footer.html.erb
+++ b/app/views/application/_footer.html.erb
@@ -7,7 +7,7 @@
                  'https://www.google.ca/maps/place/University+of+Alberta',
                  class: 'text-dark small' %></p>
         <p class="mb-0"><%= link_to t('application.footer.privacy_policy'),
-                 '#TODO', class: 'text-dark small' %></p>
+                 '/policies', class: 'text-dark small' %></p>
       </div>
       <div class="order-first p-1 py-3">
         <%= image_tag 'ualib-logo.png', alt: 'logo' %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -72,4 +72,11 @@ Rails.application.routes.draw do
   get 'robots.txt' => 'robots#robots'
 
   root to: 'welcome#index'
+
+  # Static pages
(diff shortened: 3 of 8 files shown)


app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,5 @@
+class PagesController < ApplicationController
```
mbarnett: Can we call this `StaticPageController`? There's some vague notion of Admin-controllable content pages down the road that we might want to distinguish this from.

cwant: Done (well, I went with the plural `StaticPagesController` instead of `StaticPageController`).

app/views/application/_footer.html.erb
```diff
@@ -7,7 +7,7 @@
                  'https://www.google.ca/maps/place/University+of+Alberta',
```
murny: Address should probably go to Contact Us (which would have a link to the map/maybe even an embed map/etc and all the other bells and whisltes you want)

config/routes.rb
```diff
@@ -74,4 +74,11 @@
   get 'robots.txt' => 'robots#robots'
 
   root to: 'welcome#index'
+
+  # Static pages
+  get '/about', to: 'static_pages#about'
+  get '/contact', to: 'static_pages#contact'
+  get '/policies', to: 'static_pages#policies'
+  get '/help', to: 'static_pages#help'
+  get '/technology', to: redirect('https://github.com/ualbertalib/jupiter')
```
murny: Can probably just stub in a  technology static page which just has a link to the github page (very dependant on what Leah/sonya comes back with though => for example majority of these pages could be consolidated into 1 etc as noted in mockups)

mbarnett:
So it seems like we can merge this and wait on update the content when we get new copy from Leah? I can approve now if that's the case.

murny:
We can merge, but more work needed here outside of the content, for example we need a nav on the static pages as currently there is no way to discover them except for the 1 in the main navbar and 2 in the footer.

As shown in mockups: https://jupiter.mybalsamiq.com/projects/jupiter/Static%20Pages%20-%20Contact%20Us

Issue #417, opened 2018-01-31
Rewrite "RSpec"y Context/Should tests into before_all/test


mbarnett:
It's a bit disappointing that this split in style crept in at all, but I've been seeing a lot of flakiness with flapping tests, surprising and unintuitive cases of repeated work slowing down tests, etc, stemming from all of the crufty magic this DSL performs (quick: describe the concrete behaviour of nested `context` blocks with multiple `setup`s)

(example weirdness I've seen from this nested context stuff: I've had assertions fail on the line _after_ a `binding.pry` I inserted to find out why it was failing, without the `binding.pry` activating first. This stuff is really, really fragile and hard to work with)

In contrast the `before_all`/`test 'test name' do` behaviour is straightforward and easy to reason about. I think we should standardize on it.

pgwillia:
- [ ] remove contexts
- [ ] change all before_all to setup
- [ ] change all should to test
- [ ] create fixtures for reused setup

murny:
Some extra thoughts. 

- If you can remove all context/should blocks, then we can replace `gem shoulda` for `gem shoulda-matchers` (as gem shoulda contains both shoulda-matchers and shoulda-context (context being the one that gives should/context blocks, and shoulda-context is whats holding back shoulda and isn't maintained so big win)


- `after_all/before_all/before(:all)` is actually from a gem called https://github.com/jeremyevans/minitest-hooks (`setup` is a bit different as this runs before every test). Is this needed? Not really, since minitest is just ruby :tm: you can just remove those blocks and leave the code like:

```ruby
require "test/helper"

class TestBlah < TestWhatever
  @@expensive_fedora_item_creation ||= do_expensive_item_creation

  def test_something
    assert_equal 'Cool title', @@expensive_fedora_item_creation .title
  end
end
```
And places where we using it to wipe out the entire fedora database in test_helper.rb:

```ruby
 # clean ActiveFedora at the end of all tests in a given test class
  def after_all
    # we clean fedora here
  end
```

Minitest  provides native helpers for this as well called:  `before_suites/after_suites` 

So do we need this gem? Probably not. More info here about why minitest doesn't have a before_all/etc:  https://github.com/seattlerb/minitest/issues/61

pgwillia:
Here are some things I've noticed:

* `before_all` is a hook provided by minitest-hooks and a common `after_all` hook in the test_helper.rb to clean up -- trying to move these blocks to use class level variables isn't working like I'd like
[...]

PR #418, opened 2018-01-31
Move sitemap tests into integration and remove setup and repeated get…

murny:
… requests

https://github.com/ualbertalib/jupiter/issues/417

### DIFF
--- a/app/views/profile/index.html.erb
+++ b/app/views/profile/index.html.erb
@@ -1,8 +1,6 @@
 <% page_title(@user.name) %>
 
-<div class="mb-3">
-  <h1><%= @user.name %></h1>
-</div>
+<h1 class="mb-3"><%= @user.name %></h1>
 
 <dl class="row">
   <dt class="col-sm-3"><%= t('admin.users.created') %></dt>
@@ -14,7 +12,6 @@
   <dd class="col-sm-9">
     <%= @user.email %>
   </dd>
-
 </dl>
 
 <%= render partial: 'user_items',
--- a/test/integration/sitemap_test.rb
+++ b/test/integration/sitemap_test.rb
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'open-uri'
 
 class SitemapTest < ActionDispatch::IntegrationTest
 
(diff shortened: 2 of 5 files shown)


test/integration/sitemap_test.rb
```diff
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'open-uri'
 
 class SitemapTest < ActionDispatch::IntegrationTest
```
murny: Should have been SitemapControllerTest here if we were going to keep it the same

Reason to move to integration test? These tests are not testing the controller actions of the sitemap controller but instead are testing much more including interacting with the controller, routes and view (asserting differences in the dom/xml of the presenting page)

test/integration/sitemap_test.rb
```diff
@@ -1,5 +1,4 @@
 require 'test_helper'
-require 'open-uri'
```
murny: Never seeing this being used in this file? Removed

test/integration/sitemap_test.rb
```diff
@@ -103,57 +98,48 @@ def before_all
...
+      assert_select 'loc', { count: 0, text: item_url(@private_item) }, 'private items should not appear in the sitemap'
     end
   end
 
   context 'collections sitemap' do
-    setup do
-      get collections_sitemap_url
-    end
     should 'be valid sitemap xml' do
       get collections_sitemap_url
```
murny: Here is a bad example of how confusing the setup block can get. We are actually doing two `get collections_sitemap_url` in this should block

test/integration/sitemap_test.rb
```diff
@@ -103,57 +98,48 @@ def before_all
         assert_select 'rs|ln[length=?]', uo.original_file.size.to_s
         assert_select 'rs|ln[type=?]', uo.original_file.mime_type.to_s
       end
-    end
 
-    should 'not show private items' do
```
murny: Here and in many other places we doing a full `get request` to the page just for a couple assertions. In my opinion it's okay to group these assertions under one get request

app/views/profile/index.html.erb
```diff
@@ -1,8 +1,6 @@
 <% page_title(@user.name) %>
 
-<div class="mb-3">
-  <h1><%= @user.name %></h1>
-</div>
+<h1 class="mb-3"><%= @user.name %></h1>
```
murny: Oops one of my unrelated changes snuck in here, but pretty simple. Just cleanup of some markup

test/integration/sitemap_test.rb
pgwillia: Good catch!  I started out using it, but moved away.

test/integration/sitemap_test.rb
pgwillia: It started as an integration test but I was running into problems with capybara interfering with what was returned.

test/integration/sitemap_test.rb
pgwillia: I was trying to organize what/how was being tested.  I'm okay with these being combined. Faster!

PR #419, opened 2018-01-31
Bug fixes for Deposit Item

murny:
First round of bug fixes:
- Increase upload file size to 1GB
- Reorder year dropdown, most recent first(2019) to least(1918)
- Fix weird edge case where on error submission where next step in progress bar is a clickable link
- Bandaid over a few weird tests, need revisiting
- Do not link related item in item view (broken anyways) (requirements will come later for this)

Will open another PR to tackle whats left:
TODO https://github.com/ualbertalib/jupiter/issues/269


Gemfile.lock
```diff
@@ -244,7 +244,7 @@ GEM
     puma (3.11.2)
     pundit (1.1.0)
       activesupport (>= 3.0.0)
-    rack (2.0.3)
+    rack (2.0.4)
```
murny: Need to bump this as this destroys your CPU on large file uploads. 

Original issue here: https://github.com/rack/rack/issues/1075

Now fixed in 2.0.4

PR #420, opened 2018-01-31
Multiple file download using jquery plugin

### LINKED ISSUE #90: Multi-file zipping for multi-file downloads

pgwillia:
Closes #90


app/assets/javascripts/jquery-multidownload/jquery-multidownload.js
```diff
@@ -0,0 +1,29 @@
+;(function($, window, document, undefined) {
```
murny: Jquery-multidownload should be put into `vendor/assets/javascripts` instead of our `app/assets/javascripts`

More info here: 
http://guides.rubyonrails.org/asset_pipeline.html#asset-organization

> vendor/assets is for assets that are owned by outside entities, such as code for JavaScript plugins and CSS frameworks. Keep in mind that third party code with references to other files also processed by the asset Pipeline (images, stylesheets, etc.), will need to be rewritten to use helpers like asset_path.

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -21,10 +21,13 @@
...
           </div>
           <div class="p-1">
-            <%= link_to(t('.download'), fileset_uri(file_set, :download), class: 'btn btn-outline-primary') %>
+            <%= link_to(t('.download'), fileset_uri(file_set, :download), class: 'download btn btn-outline-primary') %>
           </div>
         </div>
       </div>
     <% end %>
+    <div class="p-1">
+      <%= link_to(t('.download_all'), '#', id: 'download_all', class: 'btn btn-outline-primary float-right') %>
```
murny: A few comments here:

- Prefer an actual button (instead of a link with a href of `#` here, as links are for navigation and here there is no navigation). 
- ID/Classes should be seperated by dashes not underscore
- Should use ID's very sparely. Better approach is to use a class here instead and namespace it to indicate javascript usage instead of styling aka `.js-download-all`

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -21,10 +21,13 @@
             <%= link_to(t('.view'), fileset_uri(file_set, :show), class: 'btn btn-outline-primary') %>
           </div>
           <div class="p-1">
-            <%= link_to(t('.download'), fileset_uri(file_set, :download), class: 'btn btn-outline-primary') %>
+            <%= link_to(t('.download'), fileset_uri(file_set, :download), class: 'download btn btn-outline-primary') %>
```
murny: namespace `.download` to be `.js-download`

pgwillia:
Thanks @murny. These conventions are all new to me.  Anything I should read to bring me up to speed?

murny:
For conventions? Hmm here are a few I try to follow:

For CSS/HTML: 

- http://codeguide.co/ 
- https://github.com/airbnb/css (they have a good ruby/javascript guide too if curious)
- general bootstrap best practices (or look at what people are doing on popular websites/projects around github)
- Sure theres more here (and im in no means an expert for css)

Accessbility? Not sure but I found these pretty helpful though:

- https://a11yproject.com/
- https://webaim.org
- http://brunopulis.com/awesome-a11y/

PR #421, opened 2018-01-31
 Working multi-community/collection selection

mbarnett:
Also removes a bunch of old Item controller cruft from the original demo. I had to mark one authorization test as pending, as there was no obvious place to send it now that `item#new` is gone, as it won't redirect back to a POST to `item_draft#create`


test/system/authentication_test.rb
```diff
@@ -53,7 +53,9 @@ class AuthenticationTest < ApplicationSystemTestCase
 
   context 'when visiting a protected page' do
     should 'get redirected to login then back to page, if user is authorized' do
-      visit new_item_url
+      skip "there's no obvious place to send this now, as it won't redirect to the POST action for a draft item..."
+      visit root_url
+      click_link 'Deposit'
```
murny: Could be switched to profile_url? http://localhost:3000/profile?

non logged in user will get promted to logged in, then gets redirected back so good fit for this test

Otherwise just leave it now. I can come in and redo this one and the test below it that we left commented out

mbarnett: Ah yeah, that should work.

app/controllers/items/draft_controller.rb
```diff
@@ -2,6 +2,8 @@ class Items::DraftController < ApplicationController
 
   include Wicked::Wizard
 
+  before_action :initialize_communities
```
murny: Maybe an except/only on this? :create and :destroy shouldn't need communities I think. Show and Update will though

```ruby
before_action :initialize_communities, only: [:show, :update]
```

app/assets/javascripts/jupiter/item_draft.js
```diff
@@ -23,10 +25,18 @@ $(document).on('turbolinks:load', function() {
     e.preventDefault();
   });
 
-  // bring over community/collection select from items (tweaked a bit)
-  // (we only need to handle one pairing for time being)
-  $('form.js-deposit-item .js-community-select').change(function() {
-    var $collectionSelect = $('.js-collection-select');
+  $('form.js-deposit-item .add-community-collection').click(function(e) {
```
murny: nitpick: class names should be name spaced with `.js-*` 
For example `.add-community-collection`, `.remove-community-collection`, `.community-collection`, etc.

mbarnett: Is this some new styleguide thing I didn't get the memo on?

murny: pretty common practice used in industry (look at github/twitter/bootstrap/etc source code as an example)

Also defined here:
http://codeguide.co/#css-classes (bottom of list):
`Use .js-* classes to denote behavior (as opposed to style), but keep these classes out of your CSS.`

Ideally, we should namespace our custom css code as `.jupiter-*`, test selectors as `.test-*` and javascript selectors as `.js-*` and anything not namespaced you can assume it's default bootstrap/your css framework.

PR #423, opened 2018-02-02
Add edit button on item view and fix some css/html

murny:
Fixes some CSS/HTML and adds a Edit button on the item detail view

Item detail view before this merge:
image
Some weird padding/margin in the left column with the thumbnail. It's not equal in size as the comm/coll pairs. Left col is also very close to the breadcrumb while right col isn't. Etc.

image
The second `col` is actually to the right of this content and a giant whitespace is left below it. Not very mobile friendly.

Item detail view after this merge:
image
image

### DIFF
--- a/app/assets/stylesheets/logo.scss
+++ b/app/assets/stylesheets/logo.scss
@@ -1,7 +1,5 @@
 .logo-large {
   height: 400px;
-  max-width: inherit;
-  width: 300px;
 }
 
 .logo-small {
--- a/app/views/application/_thumbnail.html.erb
+++ b/app/views/application/_thumbnail.html.erb
@@ -1,13 +1,14 @@
 <%# TODO: might want to add a size parameter (e.g. list view vs item view) %>
+<%# TODO: Need to consolidate this with logo and other thumbnail/logo code in ERA as they are pretty much identical %>
 <% size_class = if local_assigns[:size].present? && size == :large
                   'logo-large'
                 else
                   'logo-small'
                 end %>
 <% if object.thumbnail&.attached? %>
-  <%= image_tag(object.thumbnail.url, alt: "#{t('search.thumbnail.thumbnail')} - #{object.title}", class: "d-flex mr-3 img-thumbnail #{size_class}") %>
+  <%= image_tag(object.thumbnail.url, alt: "#{t('search.thumbnail.thumbnail')} - #{object.title}", class: "d-flex img-thumbnail #{size_class}") %>
 <% else %>
-  <div class="d-flex justify-content-center align-items-center mr-3 img-thumbnail <%= size_class %>">
+  <div class="d-flex justify-content-center align-items-center img-thumbnail <%= size_class %>">
     <small class="text-muted">
       <i class="fa fa-5x fa-picture-o" aria-hidden="true"></i>
       <small class="sr-only"><%= "#{t('search.thumbnail.thumbnail')} - #{object.title}" %></small>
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/logo.scss
```diff
@@ -1,7 +1,5 @@
 .logo-large {
   height: 400px;
-  max-width: inherit;
```
murny: This has some weird behaviour. Max-width inherit means it gets it's width height from its parent element, which is smaller then the 300px below. Thus it will automatically change the width of the element (making it overstick a bit). We were fixing this by wrapping a flexbox div around it, but lets just fix the root issue instead.

app/views/application/_thumbnail.html.erb
```diff
@@ -1,13 +1,14 @@
...
 <% size_class = if local_assigns[:size].present? && size == :large
                   'logo-large'
                 else
                   'logo-small'
                 end %>
 <% if object.thumbnail&.attached? %>
-  <%= image_tag(object.thumbnail.url, alt: "#{t('search.thumbnail.thumbnail')} - #{object.title}", class: "d-flex mr-3 img-thumbnail #{size_class}") %>
+  <%= image_tag(object.thumbnail.url, alt: "#{t('search.thumbnail.thumbnail')} - #{object.title}", class: "d-flex img-thumbnail #{size_class}") %>
 <% else %>
-  <div class="d-flex justify-content-center align-items-center mr-3 img-thumbnail <%= size_class %>">
```
murny: `mr-3` is great on the item list pages (as the item is represented as a media object with the thumbnail on the left and content on the right. So putting some margin between thumbnail and content is a great idea.

Not a great idea when we are using the thumbnail in an item view (where it doesnt need this margin on the right)

Solution? Delete the `mr-3` so it works for all. And put `ml-3` in the media-body for the item list content instead

app/views/application/_thumbnail.html.erb
```diff
@@ -1,13 +1,14 @@
 <%# TODO: might want to add a size parameter (e.g. list view vs item view) %>
+<%# TODO: Need to consolidate this with logo and other thumbnail/logo code in ERA as they are pretty much identical %>
```
murny: TODO: we probably got at least 4 different representation of logo/thumbnail in the codebase now. Should consolidate this in a nice reusable and flexible component

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -1,33 +1,30 @@
-<div class="col">
```
murny: Useless div/col here. As it's parent class is a div/col

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -1,33 +1,30 @@
...
-  <div class="d-flex justify-content-center">
```
murny: Useless d-flex/flexbox div here. As pointed out in the css comment above

app/views/items/show.html.erb
```diff
@@ -21,7 +21,7 @@
 <div class="container">
   <div class="row">
     <%# LEFT SIDE %>
-    <div class="col-4">
+    <div class="col-md-4 mt-5">
```
murny: Need to add breakpoint (sm/md/lg/etc) classes for this to work properly on smaller screensizes

app/views/items/_thesis.html.erb
```diff
@@ -1,110 +1,105 @@
-<div>
```
murny: useless div. Less code less problems

app/views/items/_item.html.erb
```diff
@@ -1,111 +1,106 @@
-<div>
-  <h1 class="mt-3" title="<%= t('items.show.title') %>"><%= @item.title %></h1>
```
murny: Moved title into the show, since thesis/item both have the exact same markup. Can probably move more of the same stuff into here and reduce some of the duplication

murny:
Cool tip with regards to `?w=1`!  :+1:

Issue #362, opened 2018-01-04
File Ordering


weiweishi:
This is to track the file ordering issue that is currently blocked by the version of rails and active_storage we are on. The hope is that when we move to the new version (5.2), we will be able to refactor our current approach and display the files in the same order as when they are uploaded, and that is sufficient for the IR phase of the implementation.

murny:
Might be some confusion here. Currently the order the files are uploaded are the exact same way they come out for ActiveStorage/Deposit Item/Draft Items. (However this means if I drag and drop 10 files from my desktop into the upload files form, it depends on which files finish uploading first not the order they were uploaded with). So if this is the issue then there is nothing to fix here. 

With that said, there is currently no way of re-ordering files after upload (aside from...removing the files and re-uploading them in the desired order). If this is the problem, this will still not be fixed with upgrading ActiveStorage that comes with Rails 5.2 as ActiveStorage has no concept of ordering files aside from who got uploaded first (based on their primary keys in the database). So this requires work on our part to figure this out if this is needed.

Now ActiveStorage is just one piece of the puzzle. The bigger concern is Fedora. We can upload and preserve the same order (or heck you can choose whatever order you want, maybe you want to sort them in alphabetical order etc) that user's uploaded the files in ActiveStorage into Fedora (with the addition of adding the chosen thumbnail as the first file? or whatever way we will want to handle the thumbnail for this). But is there a guaranteed way of preserving this order in Fedora? I assume this is where Proxy Objects come in and help with this. But needs investigation.

mbarnett:
> But is there a guaranteed way of preserving this order in Fedora? I assume this is where Proxy Objects come in and help with this. But needs investigation.

Yes. Via Hydra-works and related layers, ordering proxies are created that maintain the file order that the files are added to Fedora in (and you can reorder with it updating the proxies).

weiweishi:
Confirmed with Leah this is not required for launch, but need to be developed in the near future. Put it in post-launch milestone.

weiweishi:
will verify with @leahvanderjagt if we can close this for Jupiter I and revisit file ordering in the refactoring work and jupiter phase 2

PR #425, opened 2018-02-02
 Thumbnails, Characterization, and Virus-Checking (Oh my!)

### LINKED ISSUE #19: Derivatives Generation Setup

### LINKED ISSUE #402: maintain file name when downloading
The file deposited through UI doesn't seem to use its' original filename when downloading? it uses a uuid instead:
http://uat.library.ualberta.ca/items/6c6397b9-f292-4542-b418-6f1db50bd4ef/download/2r36tz89220180126-21-n5cso6

mbarnett:
Closes #19 

Also closes #402 and #362.

I've made virus-checking and FITS optional for development purposes. For UAT, Staging, and Production, clamav should be running as a daemon on the machine, and fits needs to be installed and in the Rails' process' user's path.

@weiweishi For your migration script, for each item or thesis, you'll need to call `thumbnail_fileset` on the item after saving, and pass it a fileset on that item that contains the file that should be used to derive the thumbnail. Presumably this will be the first (only) file on most items, since we didn't support multi-file items in Hydranorth, so you probably just need to do something like:

```ruby
item.thumbnail_fileset(item.file_sets.first) 
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,6 @@ group :test do
   gem 'shoulda', require: false
 end
 
-#group :uat, :staging, :production do
+group :uat, :staging, :production do
   gem 'clamby'
-#end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -89,3 +89,7 @@ group :test do
   gem 'minitest-hooks'
   gem 'shoulda', require: false
 end
+
+#group :uat, :staging, :production do
+  gem 'clamby'
+#end
--- a/app/models/concerns/item_properties.rb
+++ b/app/models/concerns/item_properties.rb
@@ -1,3 +1,4 @@
+
 module ItemProperties
   extend ActiveSupport::Concern
 
@@ -131,9 +132,11 @@ length=\"#{unlocked_fileset.original_file.size}\" \
 type=\"#{unlocked_fileset.original_file.mime_type}\"\
 />"
             unlocked_fileset.save!
-            self.members += [unlocked_fileset]
-            # pull in hydra derivatives, set temp file base
-            # Hydra::Works::CharacterizationService.run(fileset.characterization_proxy, filename)
+
+            self.ordered_members += [unlocked_fileset]
+            if Rails.configuration.run_fits_characterization
+              Hydra::Works::CharacterizationService.run(unlocked_fileset.original_file)
+            end
           end
         end
       end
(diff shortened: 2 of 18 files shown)


Gemfile
```diff
@@ -89,3 +89,7 @@ group :test do
   gem 'minitest-hooks'
   gem 'shoulda', require: false
 end
+
+group :uat, :staging, :production do
+  gem 'clamby'
```
murny: Is there any instructions to install this? Would that be worth adding to readme or something? Or maybe just link to its readme: https://github.com/kobaltz/clamby#dependencies

app/models/concerns/item_properties.rb
```diff
@@ -169,8 +172,22 @@ def each_community_collection
...
-    nil
+    @thumbnail ||= ActiveStorage::Attached::One.new(:thumbnail, self)
+  end
+
+  def thumbnail_fileset(fileset)
+    raise ArgumentError, 'Thumbnail must belong to the item it is set for' unless id == fileset.owning_item.id
+    thumbnail.purge if thumbnail.present?
+    fileset.unlock_and_fetch_ldp_object do |unlocked_fileset|
+      unlocked_fileset.create_derivatives
+      # don't ask. RDF::URIs aren't real Ruby URIs for reasons that presumably made sense to someone, somewhere
```
murny: :upside_down_face:

app/models/draft_item.rb
```diff
@@ -185,30 +197,34 @@ def validate_if_visibility_is_embargo?
...
+    files.each do |file|
+      path = file_path_for(file)
+      errors.add(:files, :infected, filename: file.filename.to_s) unless Clamby.safe?(path)
+    end
+  end
+
   # Fedora file handling
   # Convert ActiveStorage objects into File objects so we can deposit them into fedora
   #
   # TODO: How to handle thumbnail? Should it be the first file uploaded to fedora?
```
murny: Can probably delete this TODO now since its now fixed!

app/models/concerns/item_properties.rb
mbarnett: I don't know man

Gemfile
mbarnett: I'll add a link to the Readme for this and FITS

mbarnett:
Added some more comments, so you'll need to re-approve

PR #429, opened 2018-02-05
Css theme cleanup

murny:
PR Fixes some tiny CSS/HTML for themeing/homepage

- Fix footer container (should have padding on either side like the header aka have a container)
- Allow flash and announcement messages to be just under navbar on homepage (instead of below banner image)
- Move main container div outside of application layout into every single view which gives more freedom for tackling things like the full-width banner/etc (lots of whitespace changes as a result)
- Fix consistency between upper and lower case in footer/navbar/on page headings/etc.
- Add a masthead with temporary tagline/lead with ERA A+V button underneath like mockups 
- Moved `welcome.scss` into seperate stylesheet instead of `jupiter.scss`
- Updated selenium webdriver 
- Add view more link on community/collections homepage section that links to browse/communities index
- Improve button links using more standard bootstrap components instead (media/list group)
- Add box shadow on fixed navbar (gives nice layering effect)
- fix broken ERA A+V link

# DESKTOP
## Previous above fold with flash/announcement:
image

## Below fold:
image

## After PR above fold with flash/announcement:
image

## below fold:
image

# MOBILE
## Before
image
## After
image
image


app/assets/stylesheets/jupiter.scss
```diff
@@ -75,19 +80,12 @@ body {
...
 
-.collection-links {
+.j-collection-links {
   a {
     text-align: left;
     white-space: normal;
   }
 }
 
-.nav-item {
```
murny: Leaving it unscoped to the header means it's bolded everywhere (admin nav menu, search nav menu, etc.)

app/assets/stylesheets/jupiter.scss
```diff
@@ -75,19 +80,12 @@ body {
     text-align: center;
     width: 100%;
   }
-
-  .era-av {
```
murny: Never used? Removed.

app/views/application/_flash.html.erb
```diff
@@ -1,15 +1,8 @@
-<% if flash[:notice] %>
-  <div class="alert alert-success" role="alert">
+<% if flash.any? %>
```
murny: Cleanup the duplication between the two flash messages into one. And move the container class in here too.

app/assets/stylesheets/jupiter.scss
```diff
@@ -33,19 +33,24 @@ body {
...
+  }
+}
+
+.j-footer {
+  // needed for sticky footer
   bottom: 0px;
   position: absolute;
   width: 100%;
-
-  .container {
```
murny: We want the container to have padding on left/right side which is the default. This is basically asking for a `container-fluid` if we want a 100% wide container.

app/views/welcome/index.html.erb
```diff
@@ -1,66 +1,76 @@
...
-    </div>
-    <div class="align-self-center px-0 px-md-3 col-md-7">
-      <%= link_to "https://era-av.libary.ualberta.ca", class: 'btn btn-jupiter-primary w-100' do %>
-        <p class="mb-0 lead">
+<%= image_tag 'campus-banner.jpg', class: 'img-fluid w-100', alt: 'banner image' %>
+
+<section class="jumbotron text-center">
+  <div class="container">
+    <h1 class="jumbotron-heading"><%= t('.welcome_header') %></h1>
+    <p class="lead">
```
murny: Copied the tagline from ERA (which is two paragraphs). We will eventually get a new one from Leah/Sonya. (I prefer this to be text-muted as well but seems to be failing contrast accessibility guidelines)

app/assets/stylesheets/jupiter.scss
```diff
@@ -1,26 +1,26 @@
-$header-height-large-screen: 76px;
-$header-height-small-screen: 136px;
-$footer-height: 76px;
+$header-height-large-screen: 81px;
+$header-height-small-screen: 139px;
+$footer-height: 72px;
```
murny: These numbers all seem a bit off with some extra padding (not sure if this was on purpose). For example, The banner/etc tucks under the header for a few pixels. So this makes header/banner, etc touch much better now.

app/views/welcome/index.html.erb
```diff
@@ -69,20 +81,17 @@
...
-          class: 'btn btn-jupiter-primary w-100 mt-3' %>
-    </div>
-    <div>
-      <%= link_to t('.theses_collection'), '#TODO', class: 'btn btn-jupiter-primary w-100 mt-3' %>
-    </div>
-    <div>
-      <%= link_to t('.afns_collection'), '#TODO', class: 'btn btn-jupiter-primary w-100 mt-3' %>
-    </div>
-  </div>
+  <div class="d-md-none list-group">
```
murny: Easier to just make them list groups instead of having to worry about the weirdness with `.btn` with regards to very long text (white-space: normal), plus easier to customize down the road. (alot of the markup wasn't needed (empty divs especially))

app/views/application/_navbar.html.erb
```diff
@@ -1,10 +1,10 @@
-<header class="navbar navbar-light fixed-top navbar-expand-md bg-jupiter-primary">
+<header class=" j-header navbar navbar-light fixed-top navbar-expand-md bg-jupiter-primary">
   <nav class="container">
 
     <%= link_to root_path, class: 'navbar-brand' do %>
-      <%= image_tag 'era-logo.png', width: '100px', class: 'align-top d-none d-md-block',
+      <%= image_tag 'era-logo.png', width: '100', class: 'align-top d-none d-md-block',
```
murny: `px` is redundant here, width expects a number which is the pixels in html5

app/views/layouts/application.html.erb
```diff
@@ -1,21 +1,20 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <meta charset="utf-8">
```
murny: charset (and probably the viewport too now that i think of it (guess we will tackle this later in go live)) should be the first tag in the head

app/views/welcome/index.html.erb
```diff
@@ -1,66 +1,79 @@
...
-  <%= image_tag 'campus-banner.jpg', class: 'img-fluid w-100', alt: 'banner image' %>
-<% end %>
-<div class="pt-md-3">
-  <%# Two columns on large screens, wrap on smaller screens %>
-  <div class="d-flex flex-row flex-wrap">
-    <div class="col-12 col-md-5 px-0 px-md-3">
-      <h4><%= t('.welcome_header') %></h4>
-    </div>
-    <div class="align-self-center px-0 px-md-3 col-md-7">
-      <%= link_to "https://era-av.libary.ualberta.ca", class: 'btn btn-jupiter-primary w-100' do %>
```
murny: library was spelled wrong resulting in bad url

config/locales/en.yml
```diff
@@ -157,20 +158,20 @@ en:
...
 
   sessions:
     new:
-      header: 'Login to ERA'
+      header: 'Log into ERA'
       description_html: 'ERA is now a
         <a href="https://ist.ualberta.ca/services/ccid-passwords/campus-computing-id" target="_blank">CCID</a>
         (Campus Computing ID) enabled service.
         If you have a CCID please use it to log in to ERA.'
-      saml_link: 'Sign in with University of Alberta CCID (Campus Computing ID)'
```
murny: Lets be consistent (Sign in vs Log in. We using Log In everywhere else so lets change it)

app/assets/stylesheets/jupiter.scss
```diff
@@ -33,61 +32,17 @@ body {
   text-decoration: inherit;
 }
 
-.footer {
-  bottom: 0px;
-  position: absolute;
-  width: 100%;
+.j-header {
```
cwant: I take it that `j-*` is to mean 'jupiter specific'?

murny: Yep, it indicates any custom css we created and is namespaced as jupiter or just "j" for short

PR #430, opened 2018-02-05
Active storage dir setup

weiweishi:
use Rails.application.secrets.active_storage_directory to set up path to the thumbnail storage server.

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -19,6 +19,7 @@
 
 shared:
   redis_url: <%= ENV['REDIS_URL'] %>
+  active_storage_directory: <%= ENV['ACTIVE_STORAGE_DIRECTORY'] || Rails.root.join("storage") %>
 
 development: 
   secret_key_base: c0a4bf2c5890d0fa86e1459dd189bc4c5a02f412067b610490885c8c312bf0cea5d988e075761ba7277a8291041c1b2e7cf6c373d4d6f43d4522bc48db76cc1a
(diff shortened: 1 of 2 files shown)


config/secrets.yml
```diff
@@ -35,6 +35,8 @@ development:
   fcrepo_password: fedoraAdmin
   fcrepo_base_path: /dev
 
+  active_storage_directory: <%= ENV['ACTIVE_STORAGE_DIRECTORY'] || Rails.root.join("storage") %>
```
murny: Since its the same for every environment maybe a good candidate for `shared` at:
https://github.com/ualbertalib/jupiter/blob/8afeb4e56a204c67d14453fb81d6cf6997917830/config/secrets.yml#L20

config/secrets.yml
```diff
@@ -50,14 +52,16 @@ test:
   fcrepo_password: fedoraAdmin
   fcrepo_base_path: /test
 
+  active_storage_directory: <%= ENV['ACTIVE_STORAGE_DIRECTORY'] %> || Rails.root.join("storage") %>
```
murny: syntax error here i think:
```
<%= ENV['ACTIVE_STORAGE_DIRECTORY'] %> || Rails.root.join("storage") %>
```
vs
```
<%= ENV['ACTIVE_STORAGE_DIRECTORY'] || Rails.root.join("storage") %>
```

Probably why travisCI is breaking on yaml parsing

config/secrets.yml
```diff
@@ -19,6 +19,7 @@
 
 shared:
   redis_url: <%= ENV['REDIS_URL'] %>
+  active_storage_directory: <%= ENV['ACTIVE_STORAGE_DIRECTORY'] || Rails.root.join("storage") %>
```
murny: Sorry I noticed they WERE different so I deleted my original comment :sweat_smile: . 

This is fine, if you want a nice fallback on uat/production. If you want this to blowup instead (when ENV['ACTIVE_STORAGE_DIRECTORY']  is not set in prod/uat) then what you had before would be a better choice.

weiweishi: I'll leave it like this - there likely won't be situations where the var is not set in prod/staging since they should all handled by ansible playbook.  The application should still be able to function if we didn't set up the env var properly and thumbnailing/uploading is happening in the wrong dir.  thanks!

PR #431, opened 2018-02-05
Item editing

### LINKED ISSUE #322: Item edit after deposit
- [ ] An edit form for item that has already been saved into Fedora
- [ ] connect to the deposit workflow so that user can still modify the item after it's published. 
@murny did I miss anything?

mbarnett:
Adds item editing, reworks file thumbnail-ActiveStorage interaction to (hopefully) work on the prod stack, fixes a bug with files that are unthumbnailable, turns off editing for theses, slices, dices, makes julienne fries.

Closes #322, #431


app/controllers/items/draft_controller.rb
```diff
@@ -6,10 +6,15 @@ class Items::DraftController < ApplicationController
 
   steps(*DraftItem.wizard_steps.keys.map(&:to_sym))
 
+  FIRST_STEP = DraftItem.wizard_steps.keys.first.to_sym
```
murny: I believe wicked has something for this called: `Wicked::FIRST_STEP`
https://github.com/schneems/wicked#keywords

(But i think it does an extra 301 redirect, so this solution of yours might be better)

mbarnett: `steps.first` returns a proc, which works with the magic `wizard_path` helpers but doesn't work in ItemsController, which doesn't have access to `wizard_path` and has to use `item_draft_path` with an id:.

In ItemsController I get this, too:
```ruby
>> Wicked::FIRST_STEP
=> "wicked_first"
>> Items::DraftController::FIRST_STEP
=> :describe_item
```

so it doesn't look like that will work with `item_draft_path` either ¯\_(ツ)_/¯

mbarnett: Huh, nevermind. Weirdly that does work even though the values are different

murny: Yeah it's different values and wicked handles this internally and redirects it to the actual step mapping. But if you look at the network it does this via an additional 301 redirect. So not sure if big deal or not /shrug

app/controllers/profile_controller.rb
```diff
@@ -5,7 +5,7 @@ class ProfileController < ApplicationController
   def index
     authorize :user, :logged_in?
     @user = current_user
-    @draft_items = @user.draft_items.where(status: :active)
+    @draft_items = @user.draft_items.unpublished
```
murny: Theres another one of these in admin users: https://github.com/mbarnett/jupiter/blob/7fcfa6481fe1f52c7ad11f3b8c36a34f6d9d8db3/app/controllers/admin/users_controller.rb#L21

mbarnett: Fixed

PR #437, opened 2018-02-06
Issue #428, file download and routing improvements

### LINKED ISSUE #428: File download improvements
From @mbarnett:

> It might be helpful if you could look at switching the file download away from what we’re doing now with  HTTP::Get to do something like https://github.com/samvera/hydra-head/blob/9dcdc3b226c56a27024b6fb78b0d67a618ba5d2a/hydra-core/app/controllers/concerns/hydra/controller/download_behavior.rb#L103
> 
> playing with it in the console, it looks like uo.original_file.stream where uo is an unlocked FileSet gives you something you can stream directly from the existing Fedora connection, which means we don’t have to worry about authorization

cwant:
Fixes #428 

* Stream files directly from Fedora
* Allow partial file download
* Tests for file sets controller
* Routing changes for file sets:
  * `items/:id/view/:file_set_id/:filename => file_sets#show`
  * `items/:id/download/:file_set_id => file_sets#download`
* Updates to site map generation to reflect above changes


app/controllers/file_sets_controller.rb
```diff
@@ -4,36 +4,84 @@ class FileSetsController < ApplicationController
 
   # TODO: expand to handle derivatives
   def show
+    @file_name = params[:file_name]
+    return head :not_found unless @file_name == @file_set.contained_filename
```
mbarnett: Is this how we want to be doing this? Shouldn't we raise a NotFound and serve a proper 404 page instead of a blank page?

app/controllers/file_sets_controller.rb
```diff
@@ -4,36 +4,84 @@ class FileSetsController < ApplicationController
...
   private
 
   def load_and_authorize_fileset
-    @file_name = params[:file_name]
-    @file_set = FileSet.find(params[:id])
-    return render status: :not_found if @file_set.blank?
-    return render status: :not_found unless @file_name == @file_set.contained_filename
+    @file_set = FileSet.find_by(params[:file_set_id])
+    return head :not_found if @file_set.blank?
+    return head :not_found unless @file_set.item == params[:id]
```
mbarnett: Same comment as above. If you use `FileSet.find` instead of `.find_by`, it will raise `JupiterCore::ObjectNotFound` and we can serve a proper 404 page in the ApplicationController

app/controllers/file_sets_controller.rb
cwant: Sorry about that. The test suite was whinging about missing templates (which may have been due to a different issue), and this was a quick fix.

mbarnett: Dang test suite is so needy.

We're actually missing a 404 page right now, which may or may not be related?

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,7 @@
 class ApplicationController < ActionController::Base
 
+  # TODO: this require is to locate JupiterCore::ObjectNotFound, needs restructuring
+  require 'jupiter_core/locked_ldp_object'
```
mbarnett: It can't find it because `config.eager_load` is set to false in environments/development.rb. Any reason we wouldn't want to set it to true (@murny maybe you have thoughts too)? The difference in boot time for me locally seems pretty minimal, and it would actually fix a bit of other weirdness in dev mode where it's possible to load ActiveFedora data before its model is defined, which leads to things getting confused.

(everything works fine in Prod, where eager-loading is the default).

murny: Include it in application.rb? I mean I forsee this jupiter_core folder living in `jupiter/lib` eventually and maybe one day even being in a `jupiter_core` gem down the road. So yeah should be included in application.rb just like how rails includes itself

mbarnett: That doesn't really solve the other issues with lazy loading, though. I'd rather just eager load the whole set of models than write a bunch of one-offs

murny: Sure lets eager/autoload it. Also should make it a `require_dependency` so changes get picked up instead of `require`

PR #441, opened 2018-02-07
Rollbar setup

### LINKED ISSUE #403: Rollbar Setup
Set up Rollbar for Jupiter UAT so that we can track errors happening in the beta launch

pgwillia:
* add gem
* use ENV var
* enable user tracking but mask user id
* closes #403

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -20,8 +20,9 @@
 shared:
   redis_url: <%= ENV['REDIS_URL'] %>
   active_storage_directory: <%= ENV['ACTIVE_STORAGE_DIRECTORY'] || Rails.root.join("storage") %>
+  rollbar_token: <%= ENV['ROLLBAR_TOKEN'] %>
 
-development: 
+development:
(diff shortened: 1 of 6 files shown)


config/secrets.yml
```diff
@@ -30,12 +31,14 @@ development:
...
   saml_idp_cert: ''
   saml_idp_sso_target_url: ''
- 
+
   fcrepo_url: <%= ENV['FCREPO_URL'] || 'http://localhost:8080/fcrepo/rest' %>
   fcrepo_user: fedoraAdmin
   fcrepo_password: fedoraAdmin
   fcrepo_base_path: /dev
 
+  rollbar_token: ''
```
murny: Can probably just exclude this, no?

pgwillia: I think it would prevent rollbar being used in those environments, yes? But I guess it doesn't really matter, maybe we want to be able to.

murny: Well everything depends on the 
`config.enabled = Rails.env.production? && config.access_token.present?` 

So unless you edit that line and remove the `Rails.env.production?` and accidentally have a `ENV['ROLLBAR_TOKEN']` set somewhere, then you may have to worry about Dev/Test running rollbar but otherwise I think your pretty safe.

PR #442, opened 2018-02-07
Fix issues with draft editing antique licenses and files on new drafts

mbarnett:
Also resolves an issue with autoloading conflicting with the cache reloader in Development mode after editing a file in some cases.


app/models/draft_item.rb
```diff
@@ -15,7 +15,8 @@ class DraftItem < ApplicationRecord
                   attribution_share_alike: 5,
                   cco_universal: 6,
                   public_domain_mark: 7,
-                  license_text: 8 }
+                  license_text: 8,
+                  unselected: 9}
```
murny: What's this look like from UI? 

Creation nothing changes?

But Edit if you try to save an item (only on review screen?) with an old license, you will just get prompted an error message, which you will then need to navigate to choose license/visibility page and choose something? Something like that?

mbarnett: Yeah. If you edit something with like CC 3.0, it gets mapped to unselected, which when you get to the license screen, means nothing is checked, and it won't let you progress past it, and you're forced to select something current.

PR #446, opened 2018-02-08
Allow unauthenticated users to view CCID-protected metadata

mbarnett:
but not download files or view thumbnails. Also adds a regression test for the issue fixed yesterday where click a facet link on an item page was broken.

### DIFF
--- a/app/models/jupiter_core/search.rb
+++ b/app/models/jupiter_core/search.rb
@@ -27,8 +27,11 @@ class JupiterCore::Search
 
     # Our query permissions are white-list based. You only get public results unless the results of +calculate_ownership_query+
     # assign you additional permissions based on the user passed to it.
+
+    # Why can't I split %Q() strings over multiple lines? Seems incorrect
     # rubocop:disable Metrics/LineLength
     fq << %Q((visibility_ssim:"#{JupiterCore::VISIBILITY_PUBLIC}" OR visibility_ssim:"#{JupiterCore::VISIBILITY_AUTHENTICATED}"#{ownership_query}))
+    # rubocop:enable Metrics/LineLength
 
     base_query << q if q.present?
     facets.each do |key, values|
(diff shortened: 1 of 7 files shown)


app/models/jupiter_core/search.rb
```diff
@@ -22,14 +22,15 @@ def self.faceted_search(q: '', facets: [], models: [], as: nil)
...
 
     base_query = []
+    fq = []
     ownership_query = calculate_ownership_query(as)
 
     # Our query permissions are white-list based. You only get public results unless the results of +calculate_ownership_query+
     # assign you additional permissions based on the user passed to it.
-    base_query << %Q((_query_:"{!raw f=visibility_ssim}#{JupiterCore::VISIBILITY_PUBLIC}"#{ownership_query}))
-    base_query << q if q.present?
+    # rubocop:disable Metrics/LineLength
```
murny: Don't you need to re enable the cop after the offending line?

app/views/application/_thumbnail.html.erb
```diff
@@ -1,11 +1,12 @@
 <%# TODO: might want to add a size parameter (e.g. list view vs item view) %>
 <%# TODO: Need to consolidate this with logo and other thumbnail/logo code in ERA as they are pretty much identical %>
+
 <% size_class = if local_assigns[:size].present? && size == :large
                   'logo-large'
                 else
                   'logo-small'
                 end %>
-<% if object.thumbnail&.attached? %>
+<% if object.thumbnail&.attached? && policy(object).thumbnail? %>
```
murny: Guess this is fine for now?


But down the road this should probably be put outside of the thumbnail render all the way up until the `file_section_sidebar` partial?
Should be like:
```ruby
if  policy(object).download_file?
  <%= render partial: 'files_section_sidebar' %>
else
  <%= render partial: 'ccid_files_section_sidebar' %>
end
```
Which would render a partial that looks like the very last mockup here:
https://jupiter.mybalsamiq.com/projects/jupiter/Item%20Detail

mbarnett: No, this needs to be here, so that the thumbnails for CCID-protected items also don't render in search results when you're logged out

app/models/jupiter_core/search.rb
mbarnett: Yeah I can re-enable it. I wouldn't even need it, but I can't figure out how to split %Q() strings over multiple lines. 

```ruby
%Q() \
%Q()
```
doesn't work

app/views/application/_thumbnail.html.erb
murny: Ah yeah your right, makes perfect sense :+1:

PR #463, opened 2018-02-09
minor updates to predicates to match metadata's data dictionary

weiweishi:
Minor changes to bring predicates inline with metadata's data dictionary so that the migration can locate the triples with these predicates.


weiweishi:
replaced by PR #475

Issue #465, opened 2018-02-09
Switch docker deployments to Postgresql


mbarnett:
- Dockerfile update: Need to switch out `mysql-client` package for postgres equivalent (`postgresql-dev` i think?):
https://github.com/ualbertalib/jupiter/blob/master/Dockerfile#L9

- Docker-compose.yml (and other docker-compose.*.yml files) update: 
  - Update mysql section of file:
  https://github.com/ualbertalib/jupiter/blob/master/docker-compose.yml#L22-L27
  to basically something like:
  ```
  postgres:
    image: postgres:alpine (alpine if you want smaller container size)
    environment:
      PGDATA: /var/lib/postgresql/data/pgdata
      POSTGRES_PASSWORD: mysecretpassword
      POSTGRES_USER: jupiter
    volumes:
      - postgres:/var/lib/postgresql/data
  ```
  - Update DATABASE_URL to point to postgres now
  - Update image name from mysql to postgres everywhere (eg: depends_on)

cwant:
Can this issue be closed?

PR #469, opened 2018-02-09
Switch to Postgres and fix a bug it caught

### LINKED ISSUE #464: Postgres driver
Switch the database driver over to pg

mbarnett:
Includes #465 

Closes #464


weiweishi:
@murny would you mind spending a little time today to fast-track the changes to docker files so that we'd be able to merge this into master?  Much appreciated!

murny:
@weiweishi Yep can do!

.travis.yml
```diff
@@ -1,28 +1,42 @@
+language: ruby
```
murny: Moved our travisCI builds to rails/ruby instead of running everything in docker containers,  we were spending close to 3-4 minutes just setting up the docker containers. 

Also this way we can utilize caching on bundler to further improve build times. And this gives us the ability to parallel our builds running different ruby versions, gemfiles(aka different rails versions), etc.

Dockerfile
```diff
@@ -1,39 +1,28 @@
...
-#     && ln -s /usr/local/fits/fits-1.0.6/fits.sh /usr/bin/fits
-
-RUN mkdir -p /app
-WORKDIR /app
-
-# Preinstall gems in an earlier layer so we don't reinstall every time any file changes.
-COPY Gemfile /app
-COPY Gemfile.lock /app
-RUN bundle install --jobs=3 --retry=3
+FROM ruby:2.5.0-alpine
```
murny: Bumped to ruby 2.5 and using alpine (smaller image size)

Dockerfile.deployment
```diff
@@ -6,7 +6,7 @@ LABEL maintainer="University of Alberta Libraries"
 
 RUN apt-get update -qq \
     && apt-get install -y build-essential \
-                          mysql-client \
+                          postgresql-dev \
```
murny: Left deployment dockerfile as a Debian image, as if we want to add fits/ffmpeg/etc, probably much easier to do then in alpine

README.md
```diff
@@ -157,11 +151,20 @@ For characterization you need to install [FITS](https://github.com/harvard-lts/f
   ```shell
   docker-compose logs web
   ```
-- If your switching between docker-compose and local development on your machine, you may encounter in weird permissions on files that docker has created (logs/tmp/etc.). Simply just `sudo rm` them.
+- If your switching between docker-compose and local development on your machine, you may encounter in weird permissions on files that docker has created (/tmp/pids/, etc.). Simply just `sudo rm` them.
+
+- If you would like to run Postgres in a container, but docker-compose reports that port 3306 is already in use, you likely have a Postgres instance already running on the host. You will need to shutdown Postgres before you can start the container. On Ubuntu, `sudo service postgresql stop` on the host will do the trick. Another option is to configure docker and the rails app to look for Postgres using a different port.
+
+# Want to setup Virus-Checking and FITS characterization?
```
murny: We were nesting everything below the docker documentation, in the docker documentation. Lets keep it seperate

README.md
```diff
@@ -79,71 +79,65 @@ To build, create, start and setup your docker containers simply run:
...
   docker-compose -f docker-compose.lightweight.yml up -d
   ```
 And everything else is how you would normally develop in a rails project.
 
 (See other sections of this README for more information about developing in a rails project environment)
 
-## Want to setup Virus-Checking and FITS characterization?
-
-The Clamby repository has [instructions](https://github.com/kobaltz/clamby#dependencies) on setting up clamav on various operating systems
+## For deployment (on UAT environment)
```
murny: Move deployment with docker into its own place, made little sense where it was previous

bin/docker-start
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -e
+
+while ! nc -z postgres 5432; do
+  echo "Waiting for database to be available..."
+  sleep 1
+done
+
+bundle exec rails db:create
```
murny: Why not just simplify this and use `db:setup`?

 Our seeds take to long. And this gives no feedback to the user, so if you spam localhost:3000 for 5 minutes, you will probably think the docker is broken. Even with running db:create/db:migrate it slows it  down a few seconds and after `docker-compose up -d` finishes, its not instant that `localhost:3000` works. But a few seconds isn't bad

bin/docker-start
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
murny: What is this? Simple script the rails web server runs to provision the database (if haven't already) and run the server. 

Previously you had to manually run db create/migration and we were already stuffing alot into the docker-compose command field with removing pids and running the server. So lets move this to a script instead (which can be added to in the future if needed)

PR #473, opened 2018-02-12
Issue #411, handle restricted collections.

### LINKED ISSUE #411: Restricted Community/Collection on deposit
Certain communities/collections will not be allowed for deposit by public. These communities/collections should not be displayed in the dropdown on the deposit form:

We will use a property with the predicate ual:restrictedCollection (boolean, single value) on community/collection. Data dictionary will be updated to reflect such change.

cwant:
Also, validate existence of community/collections for `DraftItem`.

Fixes #411


app/controllers/concerns/communities_collections_typeahead.rb
```diff
@@ -11,10 +11,12 @@ def typeahead_results(term)
       if communities.any?
         results.append(text: 'Communities', children: communities)
       end
+      # Note: Non-admin users should not see restricted collections
```
mbarnett: Is that the intention? I thought the primary use here was the Thesis collection, and that we didn't want users able to deposit into it? (but obviously still see it).

app/controllers/concerns/communities_collections_typeahead.rb
```diff
@@ -11,10 +11,12 @@ def typeahead_results(term)
...
       collections = JupiterCore::Search.faceted_search(q: "title_tesim:#{term}*",
                                                        models: [Collection], as: current_user)
                                        .sort(:community_title, :asc)
                                        .sort(:title, :asc)
+                                       .select { |c| c.restricted.blank? || current_user.admin? }
```
mbarnett: It looks like we're doing this here because the deposit form calls into type-ahead to get a list of collections in a specific community, and we want to filter out the restricted collection from deposit?

Two things:

1) I don't think we want to actually mask the restricted collection out of the real type-ahead results on /communities. Users should still be able to find the Thesis collection, no? So we should probably split up the type-ahead and deposit logic rather than tie their behaviour together.

2) Either way, even if we only want the Thesis collection hidden from the type-ahead, it would be better to put this into the query if the user isn't an admin, rather than filter through the result objects?

app/models/draft_item.rb
```diff
@@ -331,15 +331,23 @@ def visibility_after_embargo_for_uri(uri)
...
     member_of_paths['community_id'].each_with_index do |community_id, idx|
-      errors.add(:member_of_paths, :community_not_found) if community_id.blank?
-      errors.add(:member_of_paths, :collection_not_found) if member_of_paths['collection_id'][idx].blank?
+      collection_id = member_of_paths['collection_id'][idx]
+      community = Community.find_by(community_id)
+      errors.add(:member_of_paths, :community_not_found) if community.blank?
+      collection = Collection.find_by(collection_id)
+      if collection.blank?
+        errors.add(:member_of_paths, :collection_not_found)
+      elsif collection&.restricted && !user.admin?
```
mbarnett: Two notes:

1) Collection can't be blank if we've gotten here, don't use `&.` (I guess it would be concealing the fact that the universe is broken, in this case).

2) `user` here is the object's owner, which I don't think is entirely what we want -- if an item were mis-deposited, and then an admin edits the item to put it into a restricted collection, the admin doesn't take over ownership of the object (that seems undesirable in general), so the admin would actually be denied permission to move the object into the restricted collection, because the check is only looking at the owner.

Since deposit permissions are a current_user/model context thing, I think it might make more sense to hoist this permissions check up to the controller and do it via the policy object?

app/models/draft_item.rb
```diff
@@ -331,15 +331,23 @@ def visibility_after_embargo_for_uri(uri)
 
   # Validations
 
-  # TODO: validate if community/collection ID's are actually in Fedora?
   def communities_and_collections_validations
```
mbarnett: This comment sort of depends on whether the desposit permissions check stays in here or not, but if it does it's probably worth splitting this up into a couple of separate validation methods instead of having one uber-method

app/controllers/concerns/communities_collections_typeahead.rb
cwant: Thanks for catching that -- I forgot that this typeahead is used in two places. Will re-think.

app/models/draft_item.rb
cwant: Yeah, sounds good -- thanks for catching that one too.

app/controllers/concerns/communities_collections_typeahead.rb
cwant: As mentioned on slack, I mistakenly thought that the typeahead code was involved in the draft item user interface (it's not). New code pushed.

app/models/draft_item.rb
cwant: I left the collection/community existence checks in, took the other stuff out and moved to policy.

app/policies/draft_item_policy.rb
```diff
@@ -36,6 +36,13 @@ def owned?
     record && user && record.user.id == user.id
   end
 
+  def unrestricted_collections?
```
murny: Hmm this is fine but what about the UX experience of being on the receiving end of this error message?

- First, for whatever reason this returns false instead of getting a proper validation and being redirected to the create/edit page with a "Collection chosen is not valid" error message, you will get redirected back to the root_url with a "You are not authorised to do this action" error message, which is horrible user experience as first this is confusing and second they probably lost all their changes they just worked so hard on.

- Second, not sure the requirements for making a collection restricted? But let's assume an admin/developer turns a normal collection with items into a restricted collection, now a user wants to edit their item. They get redirected to the root url again with a you are not authorized message which is also confusing. (as Edit button should be shown cause this is in the ItemPolicy, but fails when it gets converted to Item Draft policy)

Couldn't a inclusion validation or something work here instead? http://guides.rubyonrails.org/active_record_validations.html#inclusion

mbarnett: Re: #2, the requirement is first and fore-most that Users not be able to deposit into Thesis Collection.

The problem with doing the validation in the model instead of via policy is outlined here: https://github.com/ualbertalib/jupiter/pull/473#discussion_r167961634. The nut of it is that we need to know who the current user making the request is, not the object's owner.

mbarnett: I get what you're saying about the bad UI on error, buuut because we're filtering the restricted collections out of the selection interface, they'd pretty much have to be cheating to hit this, in which case bouncing to the root URL isn't the worst possible scenario?

cwant: @mbarnett this is not in the model, it is in the policy. In this case, the value of the thing called `user` is the `current_user` from the controller.

cwant: @murny I think you are concerned about the UX of something that is likely never to happen. Even if it does happen, I think the UX is the least of their concerns -- they need to talk to the helpdesk.

mbarnett: Yes, I understand that. I was explaining to Shane about the problem that led to this moving into this, the policy object?

murny: You can still do this via validations (only time the user is not the object owner is when their admin.... unless we doing sharing and multiple owners which is new to me), so therefore can do something like `validate unrestricted_collections_present unless user.admin?`

I totally agree for them to be hitting this they are nefarious or we broke community index (which is very possible).

I'm okay with this, I'd be happier if this wasn't in every single draft_item policy though (this shouldnt be in create/file policies/etc just more things to go wrong)

mbarnett: That doesn't work, though. `user` in `validate unrestricted_collections_present unless user.admin?` *is* owner.

mbarnett: > Do they become the new owner of this item after editing it? (

No, they don't (and that would suck). That's *why* this can't happen in the model -- because in the model, user == owner *because you named the column user, we should probably rename it), so doing `validate unrestricted_collections_present unless user.admin?` would result in the admin getting permission denied because you're checking the owner, not the current user.

mbarnett: Now we could, theoretically, maintain a second column for a "current draft session belongs to" record, separate from ownership of the record, but I haven't thought deeply about that, so there might be a downside I haven't come up with yet.

murny: Alright cool, sounds fine to me then.

mbarnett: Wait, no, I did this differently than I thought I did.

When an item gets turned into a draft for editing, the draft belongs to current user, not the item owner. When the item gets updated from the final draft, the draft's owner doesn't replace the item's owner.

So @murny  is right, we *can* check the user field in the draft and implement the permission there. Sorry @cwant, I had you move the code based on my own misunderstanding of my code.

We might want to rename this column though

cwant: Okay, added the model tests back. Did not remove the policy -- does it matter if the policy stays in place now?

cwant: Policy changes removed now.

cwant:
Any road blocks left for this one?

app/models/draft_item.rb
```diff
@@ -330,15 +330,23 @@ def visibility_after_embargo_for_uri(uri)
...
     return if member_of_paths.blank? # caught by presence check
-    errors.add(:member_of_paths, :community_not_found) if member_of_paths['community_id'].blank?
-    errors.add(:member_of_paths, :collection_not_found) if member_of_paths['collection_id'].blank?
-    return unless member_of_paths['community_id'].present? && member_of_paths['collection_id'].present?
+    if member_of_paths['community_id'].blank? || member_of_paths['collection_id'].blank?
+      errors.add(:member_of_paths, :community_blank) if member_of_paths['community_id'].blank?
+      errors.add(:member_of_paths, :collection_blank) if member_of_paths['collection_id'].blank?
+      return
+    end
     member_of_paths['community_id'].each_with_index do |community_id, idx|
```
mbarnett: Can we split this into a few validations? Community and Collection presence, and Community and Collection existence/correctness (there's some subtlety here where you could get away with crafting the params to put the item in a collection that isn't in the community that's passed in), and user having permission to add to this collection?

Issue #478, opened 2018-02-13
Document requirements for "lower-casing" all ual predicates


mbarnett:
This also seems to involve removing underscores? Are there any other major requirements here? It would be good to specify this up front so that we can plan for mapping Ruby-friendly names into their requirements properly.

mbarnett:
>  It doesn't seem there's an easy way to get around with this constraint without some significant development at our end.

I disagree? I'm pretty sure we could get this working today, but I'd need to know what requirements I'm trying to meet before I take a shot at it. This might just be because I have more of how this stuff works in my head than anyone else, but that's all the more reason to make sure these requirements pass through proper Github issues.

So all lower case, and no underscores. Anything else?

mbarnett:
It would also be good if someone could provide a rationale for why we're violating RDF naming conventions (http://www-sop.inria.fr/acacia/personnel/phmartin/RDF/conventions.html#lexicalConventions), so that I can document our reasoning in the codebase

weiweishi:
I've booked a meeting for tomorrow morning to have a chat in-person for
this and another issue that has recently raised during the implementation
phase.




*Weiwei Shi*Digital Initiative Applications Librarian

2-10L Cameron Library, University of Alberta
780-492-7802  |  weiwei.shi@ualberta.ca
*"The University of Alberta respectfully acknowledges that we are situated
on Treaty 6 territory, traditional lands of First Nations and Métis
people."*

danydvd:
Based on our meeting today:

- All the UAL properties in Data Dictionary are now camel-cased according to RDF conventions.
- The newly added "dataDictionaryLabel" is removed as it is no longer necessary.
- Jupiter [ontology](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/jupiter_ontology.md) is now up-to-date and includes all UAL properties and terms.
- A list of all [UAL properties](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/UAL_properties.md) and terms is available in the Data Dictionary.
- Everything else remained unchanged.

mbarnett:
Thanks Danoosh. I'll tweak my PR accordingly

sfarnel:
Thanks @danydvd and @mbarnett

mbarnett:
@danydvd it looks like a couple of terms are missing from https://github.com/ualbertalib/metadata/blob/master/data_dictionary/UAL_properties.md.

[...]

PR #480, opened 2018-02-13
Separate concerns between RDF term URI naming and Ruby method names

### LINKED ISSUE #479: Change terms.yml to a map so that we can isolate RDF naming practices from Ruby
working off requirements in #478

### LINKED ISSUE #485: add department in ual terms
It seems we missed department in UAL terms definition. 

image

mbarnett:
Create some separation between how RDF term URIs manifest versus how we access them via Ruby. Cleans up names to meet both metadata URI requirements as I currently understand them from #478 and meet idiomatic Ruby method naming standards.

Also makes vocabularies strict and declares a previous undeclared term (`department`). Supercedes #475. Closes #479.

Also closes #485


cwant:
I think that the underlying code changes to the terms looks okay and provides us the flexibility to get where we want to, but the implementation is actually going in reverse of what was requested. e.g. you changed

```has_attribute :visibility_after_embargo, ::TERMS[:acl].visibility_after_embargo, solrize_for: :exact_match```

But what was requested was a match of the label with what is in the data dictionary, so the original line might be more appropriate (with corresponding changes to `config/terms.rb`:

```has_multival_attribute :embargo_history, ::TERMS[:acl].embargoHistory, solrize_for: :exact_match```

This is just a declaration of a property, so `embargo_history` (the nice ruby method name) is already what we see everywhere else in the code base. If we can't bear to see a camel-case method name, perhaps the vocabularies in the `TERMS` object should be treated more like hashes, e.g.:

```has_multival_attribute :embargo_history, ::TERMS[:acl]['embargoHistory'], solrize_for: :exact_match```

The label in the data dictionary could be found here:

https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_access_and_permissions.md#aclembargohistory

**Comment by mbarnett** [comment]
So for your example, there's a distinction between the property name, term name, and the URI the term manifests as.

Previously, with
```ruby
has_attribute :visibility_after_embargo, ::TERMS[:acl].visibility_after_embargo, solrize_for: :exact_match
```
the property name was visibility_after_embargo, the term name was visibility_after_embargo, and the manifested URI was http://terms.library.ualberta.ca/visibility_after_embargo

and with 

```has_multival_attribute :embargo_history, ::TERMS[:acl].embargoHistory, solrize_for: :exact_match
```

property name embargo_history, term name embargoHistory, manifested URI  http://terms.library.ualberta.ca/embargoHistory

My understanding from #478 is that the request is that the URIs that are emitted from the RDF gem be without underscore or camel case -- that we create separation between the term name and the manifested URI. Property names remain as before

 Weiwei wrote

> To be clear: the requirement is the uris in UAL namespace need to be all lower case without any token separator.

Thus:

```ruby
2.4.1 :001 > TERMS[:ual].record_created_in_jupiter
 => #<RDF::Vocabulary::Term:0x3ffe5e2ed668 URI:http://terms.library.ualberta.ca/recordcreatedinjupiter> 
```
etc

There's now a distinction between how it appears in the URI, and what the term reference looks like, which seems to be what they want. (Although, again, actually getting a written description of the exact requirements from them would be nice)


> This is just a declaration of a property, so embargo_history (the nice ruby method name) is already what we see everywhere else in the code base.

The property, yes. But the *term* was not referenceable that way, previously -- and now that we have some separation between how they look in code and how they look in URIs, we might as well go for being clean in referencing them in code.

weiweishi:
We can discuss this in more details at our meeting tomorrow but here are some more details:

1. only the uri structures have an impact on our metadata team's work, how term references, methods, properties are named in the application has zero impact on their work, and it is up to the dev team to come up with a convention that makes the most sense.  

We are okay with the inconsistency between the term references used in the application vs. the labels used in the data dictionary (which will be camel cased) as long as the information is documented either in the code, or on the data dictionary, to indicate how the labels should be used.

So this fulfils such requirement: 
```
TERMS[:ual].record_created_in_jupiter
 => #<RDF::Vocabulary::Term:0x3ffe5e2ed668 URI:http://terms.library.ualberta.ca/recordcreatedinjupiter>  
```

And now it's decoupled, it is up to us to decide whether we want to use `record_created_in_jupiter` as with ruby's convention, or `recordCreatedInJupiter` as in line with the data dictionary is our decision. 

2. The requirement of all lower case without underscore is a practice our metadata team has adopted based on our research and needs. Other vocabularies may take similar or different approaches, and may use camel case, or other ways to indicate tokens, and we will not change those. For example: embargoHistory and visibilityAfterEmbargo are project hydra vocabs, and we will use them as is. 

Let's have more discussion on this tomorrow morning.

mbarnett:
> And now it's decoupled, it is up to us to decide whether we want to use record_created_in_jupiter as with ruby's convention, or recordCreatedInJupiter as in line with the data dictionary is our decision.

I don't really think that's even up to us. There's a very clear industry standard out there for Ruby, and we should use it in Ruby code. The entire foundation of this project was that we should walk away from Library-specific standards and embrace doing things in an industry-normal way.

The yaml file documents the mapping between their "labels" and Ruby. If they want more documentation than that, or it to appear somewhere else, we can talk about autogenerating that documentation as part of the broader discussion of getting away from the anti-pattern of copy-and-pasting information out of the codebase that needs to happen.

mbarnett:
@weiweishi this may or may not be ready to go for the migration, depending on anything else that comes out of #478

mbarnett:
Should I be adding properties to something to use those terms anywhere?

weiweishi:
no. currently, they are not required to be searchable, facetable, or
presented anywhere on the UI. We will just need to carry the information on
the object so they will be in the triplestore.


people."*

mbarnett:
Right, but adding them to terms.yml just means gives us a way to reference an instance of RDF::Term referencing that URI. 

Without creating a corresponding property on some Item, Thesis, Collection etc that references that term, no objects will have or be able to have any information related to that term and it will not appear in any triples.

weiweishi:
Sorry, of course. It will need to be in
app/models/concerns/item_properties.rb
<https://github.com/ualbertalib/jupiter/pull/480/files#diff-543d82cd91cd9cbd7aef79c2424ea481>,
as single value properties. (find it's really hard to operate with 3 hours
of sleep, and getting up at 5 is not my thing)


people."*

mbarnett:
Aha, gotcha

mbarnett:
Northern North America filenames and ids are added

PR #440, opened 2018-02-07
Thumbnail Refactor

murny:
- Move all thumbnail code into two shareable partials (thumbnail (meant as a small 100x100 image for lists/etc) and feature_image (meant as a big 350x350 or so image on item detail page and other places)
- Fix alt attributes on images
- Make image helpers/code from deposit item more generic so we can reuse in other places
- Fix edit button icons: trash bin => pencil and paper
- Make community collection buttons the same as they are in admin area
- Restyle thumbnail section on file uploads screen
- Fix weird search test

Issue #416, opened 2018-01-31
Crawler Friendliness & Hardening


mbarnett:
All of the `<a>` tags in the facet sidebar on search, communities, collection (and anywhere else they appear) and pagination links on search should be marked `rel=nofollow` to keep bots from needlessly slamming the app by crawling variations in search results. 

We probably also want to mark the entire search page as `noindex` (not collection, community, or other pages, though)

pgwillia:
Should these and `/items/*/download` be in robots.txt as well?

murny:
Stackoverflow with exact same question:
https://webmasters.stackexchange.com/questions/19133/should-i-add-a-nofollow-attribute-to-download-links-or-disallow-the-urls-in-r

pgwillia:
- [ ] unique title and description for each page
- [ ] noindex [meta tag](https://developers.google.com/search/reference/robots_meta_tag) for search page 

> For the noindex meta tag to be effective, the page must not be blocked by a robots.txt file. If the page is blocked by a robots.txt file, the crawler will never see the noindex tag, and the page can still appear in search results, for example if other pages link to it. [[here](https://support.google.com/webmasters/topic/4598466?hl=en&ref_topic=4617736)]

- [ ] nofollow for download links 
- [ ] nofollow for facets
the discussion on the [Google SEO starter guide](https://support.google.com/webmasters/answer/7451184?hl=en) uses nofollow in situations where you don't endorse the content behind the link (untrusted content, paid links or crawl prioritization) i.e. user comments, 3rd party widgets or links you want to mention but not boost -- is this really what we're going for?
- [ ] [structured data testing tool](https://search.google.com/structured-data/testing-tool/u/0/)
- [ ] [Blocking 404 pages from being crawled through the robots.txt file.](https://support.google.com/webmasters/answer/7451184)
- [ ] alt attributes on images
- [ ] Google's Mobile-friendly test
- [ ] use meta name="viewport" tag to tell the browser how to adjust the content

mbarnett:
Re:
> the discussion on the Google SEO starter guide uses nofollow in situations where you don't endorse the content behind the link (untrusted content, paid links or crawl prioritization) i.e. user comments, 3rd party widgets or links you want to mention but not boost -- is this really what we're going for?

It's not a guarantee that bots won't follow it, but it's a hint that there's no reason to, and it can't hurt.

mbarnett:
[...]

PR #483, opened 2018-02-13
Crawler Friendliness & Hardening

pgwillia:
#416 

- [x] unique title and description for each page
- [x] noindex [meta tag](https://developers.google.com/search/reference/robots_meta_tag) for search page 

> For the noindex meta tag to be effective, the page must not be blocked by a robots.txt file. If the page is blocked by a robots.txt file, the crawler will never see the noindex tag, and the page can still appear in search results, for example if other pages link to it. [[here](https://support.google.com/webmasters/topic/4598466?hl=en&ref_topic=4617736)]

- [x] nofollow for download links 
- [x] nofollow for facets
the discussion on the [Google SEO starter guide](https://support.google.com/webmasters/answer/7451184?hl=en) uses nofollow in situations where you don't endorse the content behind the link (untrusted content, paid links or crawl prioritization) i.e. user comments, 3rd party widgets or links you want to mention but not boost -- is this really what we're going for?
- [x] ~~[structured data testing tool](https://search.google.com/structured-data/testing-tool/u/0/)~~
- [x] alt attributes on images -- With good descriptive context for our images in proximity most will remain empty strings see [here](https://github.com/ualbertalib/jupiter/pull/440#discussion_r166754581) and [here](https://github.com/ualbertalib/jupiter/pull/440#discussion_r166755342)
- [x] ~~Google's Mobile-friendly test~~
- [x] use meta name="viewport" tag to tell the browser how to adjust the content
- [x] canonical tags into the right places, ie) faceted versions of the Community page pointing back at the unfaceted canonical version.


app/views/items/_item_page_description.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= @item.sort_year %>
+<%= authors(@item)&.map { |creator| creator }
```
mbarnett: I don't think this map is doing anything?

```ruby
["foo", "bar", "baz"].map {|input| input}
 => ["foo", "bar", "baz"] 
```

app/views/items/_item_page_description.html.erb
```diff
@@ -0,0 +1,4 @@
...
+                 &.join(', ')&.html_safe %>
```
mbarnett: There's a risk here that by marking the string `html_safe`, someone could upload an item and feed malicious <script> tag into the author field, which would then render through here.

https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/safe_join is a way around this.

```ruby
safe_join(authors(@item), ', ')

**Comment by mbarnett** [comment]
Looks good other than the style issues hound is complaining about and the two notes I added.

File: app/views/collections/show.html.erb [hunk 3]
```diff
@@ -1,4 +1,7 @@
 <% page_title(@collection.title) %>
+<% content_for :head do %>
+  <meta name="description" content="<%= jupiter_truncate @collection.description %>">
```
murny: We may want to truncate these ourselves? Sounds like google truncates themselves at 250-300 characters

app/views/items/_item_page_description.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= @item.sort_year %>
```
murny: Just a thought, but wouldn't this be the place schema.org/open graph/etc comes into play? Instead of fitting all this content into the meta description, you would use the appropriate meta tags to mark up this page for the year/authors/description etc.?

app/views/items/show.html.erb
```diff
@@ -1,5 +1,6 @@
 <% content_for :head do %>
   <%= render partial: 'google_scholar_metadata' %>
+  <meta name="description" content="<%= render partial: 'item_page_description' %>">
```
murny: Just a thought, but wouldn't this be the place schema.org/open graph/etc comes into play? Instead of fitting all this content into the meta description, you would use the appropriate meta tags to mark up this page for the year/authors/description etc.?

app/views/welcome/index.html.erb
```diff
@@ -1,5 +1,9 @@
 <%= image_tag 'campus-banner.jpg', class: 'img-fluid w-100', alt: '' %>
 
+<% content_for :head do %>
```
murny: This will need to get updated now, since welcome_lead_two no longer exists/etc and banner image above has been changed

app/views/items/show.html.erb
pgwillia: That sounds like a good idea.  I fell down the rabbit hole with this a little bit already. I decided a good target to meet was the [Google SEO starter guide](https://support.google.com/webmasters/answer/7451184?hl=en) which I found looking for discussion of nofollow. The good enough I settled with was the same info that was presented in our search results for the item.  I'll open a new ticket for your suggestion.

app/views/items/_item_page_description.html.erb
pgwillia: I used https://github.com/ualbertalib/jupiter/blob/master/app/views/application/_item.html.erb#L38 as a template.  I'll make this change there too.

pgwillia: On second thought, that's the correct use. Leaving it.

pgwillia: If I had thought about this originally I should have just done 
`authors(@item).join(', ')`

mbarnett: I _think_ that's safe because it's only marking a string of concatenate links we control together, BUT I think your first instinct is right and we should change that to `safe_join` too.

It would be very easy for us to make the mistake of changing part of that without re-evaluating the safety aspect. If we use safe_join we don't leave ourselves any room to shoot ourselves in the foot.

pgwillia:
@mbarnett Is there anything else you'd like to see for this PR?

PR #487, opened 2018-02-14
Postgres uat deployment

murny:
Fix broken dockerhub build


Dockerfile.deployment
```diff
@@ -38,10 +38,8 @@ RUN bundle install --without development test --jobs=3 --retry=3
 # *NOW* we copy the codebase in
 COPY . $APP_ROOT
 
-# Precompile Rails assets. We set a dummy secret key
-ENV DATABASE_URL mysql2:assets_throwaway.db
```
murny: Lets not set them as true ENV vars, just inline them in rake command

PR #488, opened 2018-02-14
Let JUPITER_DATABASE_URL override DATABASE_URL when set.

cwant:


config/database.yml
```diff
@@ -17,6 +17,9 @@ default: &default
 development:
   <<: *default
   database: jupiter_development
+  # e.g. for connecting with docker image from docker-compose.lightweight.yml:
```
murny: Maybe a note about this in the docker-compose.lightweight.yml or in the readme for this section: 
https://github.com/ualbertalib/jupiter#docker-compose-lightweight-edition

cwant: Added it to the `docker-compose.lightweight.yml`, since they have to go there anyways to uncomment the postgres section (but may not visit the README)

PR #489, opened 2018-02-14
Add note about needing to set psql user

pgwillia:


config/database.yml
```diff
@@ -1,14 +1,5 @@
...
-#   gem install mysql2
-#
-# Ensure the MySQL gem is defined in your Gemfile
-#   gem 'mysql2'
-#
-# And be sure to use new-style password hashing:
-#   http://dev.mysql.com/doc/refman/5.7/en/old-client.html
-#
+# Rails will assume you have a psql user matching your system user
+# configure with `sudo -u postgres createuser ${USER}`
```
cwant: Might want to mention that this is an issue with locally installed postgres: using docker doesn't have this problem.

pgwillia: I'm not convinced this was the right fix anyways. Permission problems while running tests.

PR #495, opened 2018-02-15
Issue #5, custom 404 and 500 pages.

### LINKED ISSUE #5: 400/500 pages, other table-stakes production stuff

cwant:
This is what it looks like:

Fixes #5


app/views/application/4xx.html.erb
```diff
@@ -0,0 +1,8 @@
+<% page_title( t('application.4xx.title', status: response.status)) %>
+<%= image_tag 'campus-banner.jpg', class: 'img-fluid w-100', alt: '' %>
```
murny: Whats thoughts behind putting the banner on these pages?

cwant: Nothing too deep, just looked empty with only the message.

PR #500, opened 2018-02-16
Fix Thesis crasher, add regression test, catch some translation issues

mbarnett:
Wow


test/system/item_show_test.rb
```diff
@@ -47,6 +47,28 @@ def before_all
       uo.add_to_path(@community.id, @collection.id)
       uo.save!
     end
+
+    @thesis = Thesis.new_locked_ldp_object(
+      title: 'Thesis about the effects of missing regression tests',
```
cwant: :+1:

PR #506, opened 2018-02-20
Issue #504, Auto-populate community/collection on depositing form

### LINKED ISSUE #504: Deposit Item/Thesis from collection page prefills collection/community for draft item
This requirement is included as a checkbox for #211

cwant:
Auto-populate the community/collection on deposit page when coming from link on collections page.

Fixes #504


app/views/collections/show.html.erb
```diff
@@ -37,11 +37,18 @@
       <%= t('.items_list_header') %>
     </h4>
     <% if policy(:item).new? %>
-        <%= link_to create_draft_items_path, method: :post,
+      <div>
```
murny: empty div here?

cwant: Not empty, the div closes on line 51. This is to ensure that the buttons are grouped together in the flex box starting on line 35.

murny: Right, but whenever you have just a `<div>` with no class or anything it's almost always a bad sign. As a div by itself is pretty much meaningless.

How can we do this better? So you want the title to be left align and the two buttons right aligned?

How about this: https://getbootstrap.com/docs/4.0/utilities/flex/#auto-margins (second example is exactly what you want, no extra markup/div needed)

Another option is pretty much everywhere else on the website we use a button group (although we should probably use less of them as they don't look nice on mobile/small screens), this would also solve this issue

app/controllers/items/draft_controller.rb
```diff
@@ -78,7 +78,17 @@ def update
 
   # Deposit link
   def create
-    @draft_item = DraftItem.create(user: current_user)
+    create_params = { user: current_user }
+    if params[:collection].present?
+      collection = Collection.where(id: params[:collection]).first
```
mbarnett: Since we know :collection is present in the params, would it be better to do

```ruby
collection = Collection.find(params[:collection])
```
remove the defensive `if collection.present?` and let it 404 if the collection doesn't actually exist?

PR #514, opened 2018-02-22
Analytics

### LINKED ISSUE #470: SPIKE - Investigate how we use bot filters on logs
As part of the Analytics epic, we decide to use some kind of bot filtering gems to parse our log and provide us reliable analytics. Based on @pbinkley's investigation on https://github.com/Hentioe/isbot, this is an acceptable solution. 

This is a spike to determine if we can live parsing the log with the gem.

mbarnett:
Closes #470


mbarnett:
@murny needs one last approval

PR #525, opened 2018-02-28
Convert 400/500 pages to static once again

murny:
500: 
image

404:
image


app/controllers/application_controller.rb
```diff
@@ -77,12 +78,14 @@ def user_not_authorized
...
-  end
 
-  def render_500(exception = nil)
-    raise exception if exception && Rails.env.development?
-    render '5xx.html.erb', status: :internal_server_error
+    respond_to do |format|
+      format.html do
+        render file: Rails.root.join('public', '404'), layout: false, status: :not_found
+      end
+      format.js { render json: '', status: :not_found, content_type: 'application/json' }
```
murny: We need to handle 404s for non html requests

public/404.html
```diff
@@ -0,0 +1,104 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Page Not Found (404) | ERA</title>
+  <style type="text/css" media="screen">
```
murny: Inlined the styles and images. 

(Especially important for 500 page, Apache/Nginx could be serving this and if rails is breaking, then 500 page could be served with no css or other funny things happening)

Other option? We could have used a CDN, which you got some guarantees it will be up.

For example if we wanted to use bootstrap styles, could use this CDN: https://www.bootstrapcdn.com/

public/500.html
```diff
@@ -0,0 +1,102 @@
+<!DOCTYPE html>
+<html lang="en">
```
murny: Alot of 500/error pages I see in the wild, seem to ignore the Language attribute `lang="en"` here. Not sure if this is an oversight? Or maybe I am missing a certain reason why this is? Can seem to find a valid reason why you would leave this off. Plus Adding it makes AXE happy  ¯\\_(ツ)_/¯

app/controllers/application_controller.rb
```diff
@@ -77,12 +78,14 @@ def user_not_authorized
 
   def render_404(exception = nil)
     raise exception if exception && Rails.env.development?
```
murny: We might want to log this or capture this in rollbar in production? If something keeps showing RecordNotFound might be an issue worth noting?

public/404.html
pgwillia: Should we get an opinion from a stakeholder or ops about this?

app/controllers/application_controller.rb
pgwillia: Yeah, I think this is a good use of Rollbar.

app/controllers/application_controller.rb
pgwillia: Where are we expecting to serve json? I see several places where we are serving xml, is there a best practice there?

murny: The `format.js` is actually for javascript/xhr/ajax/etc. Which is used in various places (any time we do `remote: true` in a rails form, do an ajax call, or heck the entire principal of turbolinks uses something called `pjax` (pushstate + ajax = pjax)

So if any of these 404s for some reason, instead of giving the ajax request a giant html 404 page (which was happening on the image uploading for deposit item a few weeks ago? Which would render the 404 page inline where the image should of went), we can instead return an actual json 404 error with proper status code

public/404.html
murny: Not really needed. I can't imagine stakeholders/ops having a strong opinion about this. If we want to do more fancier things on these error pages, then I would lean towards bringing bootstrap/fontawesome in via a CDN for these pages. Currently static/inline CSS is fine.

Issue #524, opened 2018-02-28
Login button should redirect back to same page you were previously on


murny:
Now that we don't have a login page, it makes little sense to redirect a user back to the homepage every time  (should only happen as a fallback). If I am viewing an item and I choose to login, I shouldn't randomly appear on the homepage, I should still be viewing the previous item I was on.

(this will also resolves the CCID protected login button from redirecting back to homepage as well)

PR #532, opened 2018-03-05
Login redirect

murny:
#524


Gemfile
```diff
@@ -93,6 +93,10 @@ group :test do
   gem 'haikunator'
   gem 'minitest-hooks'
   gem 'shoulda', require: false
+
+  gem 'launchy'
```
murny: This is for `save_and_open_page` method for debugging system tests. So up to team if we want to include this or not?

Gemfile
```diff
@@ -93,6 +93,10 @@ group :test do
...
+  # Easy installation and use of chromedriver to run system tests with Chrome
+  gem 'chromedriver-helper'
```
murny: As comment says this will install chromedriver for you. This is a new default rails gem in rails 5.2 (when you go `rails new` etc. Might be easier to just use this gem install of having people install chromedriver and might make travisCI work better as well?

app/controllers/application_controller.rb
```diff
@@ -2,6 +2,7 @@ class ApplicationController < ActionController::Base
 
   include Pundit
 
+  before_action :store_user_location!, if: :storable_location?
```
murny: Essentially copied how devise was doing this: https://github.com/plataformatec/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update

Just adapted to our code.

app/controllers/application_controller.rb
```diff
@@ -86,8 +100,7 @@ def render_500(exception = nil)
   end
 
   def redirect_back_to
-    redirect_to session[:forwarding_url] || root_path
-    session.delete(:forwarding_url)
+    redirect_to session.delete(:forwarding_url) || session.delete(:previous_user_location) || root_path
```
murny: Simplifying this, by just deleting/returning the sessions at the same time they are accessed.

app/controllers/sessions_controller.rb
```diff
@@ -38,7 +38,7 @@ def create
 
     if current_user.present?
       # Was signed in successfully, redirect them back to where they came from or to the homepage
-      flash[:notice] = t('login.success', kind: auth_hash.provider)
```
murny: We removed using the `kind` of provider in the login success message, so this isn't being used anymore. Remove it.

config/routes.rb
```diff
@@ -2,6 +2,8 @@
 require_dependency 'admin_constraint'
 
 Rails.application.routes.draw do
+  root to: 'welcome#index'
```
murny: Heard mix things regarding where you should put your `root to:`. Previously you suppose to put it at the end (as this loads all your routes before you hit the root) but apparently in official docs it recommends putting it at the beginning.

http://guides.rubyonrails.org/routing.html#using-root

> You should put the root route at the top of the file, because it is the most popular route and should be matched first.

test/system/admin_items_index_test.rb
```diff
@@ -1,10 +1,10 @@
 require 'application_system_test_case'
 
-class AdminItemsShowTest < ApplicationSystemTestCase
+class AdminItemsIndexTest < ApplicationSystemTestCase
```
murny: Mistake on the class name. Should be Index instead of Show.

test/system/authentication_test.rb
```diff
@@ -64,38 +64,59 @@ class AuthenticationTest < ApplicationSystemTestCase
...
       click_link I18n.t('application.navbar.links.login')
 
-      assert_text I18n.t('login.success', kind: 'saml')
+      assert_text I18n.t('login.success')
 
       # TODO: fix this view and i18n this
       assert_text I18n.t('admin.users.created')
     end
 
-    # TODO: Nothing currently exist to nicely test this behaviour, Comment this out for time being
```
murny: Uncommented this test, and enable it. Using another person's draft item as the example URL here. Anon user should be forced to login, then once logged in, taken to root url with a unauthorised message.

test/system/authentication_test.rb
```diff
@@ -64,38 +64,59 @@ class AuthenticationTest < ApplicationSystemTestCase
...
+          )
+
+      click_link I18n.t('application.navbar.links.login')
+
+      assert_text I18n.t('authorization.user_not_authorized')
+      assert_selector 'h2', text: I18n.t('welcome.index.welcome_lead')
+    end
+  end
+
+  should 'after login should be redirected back to previous page user was on' do
```
murny: Add test for testing the redirect. If I am on a certain page (browse in this case) and I log in, I should still be on the certain page (browse)

app/controllers/application_controller.rb
```diff
@@ -16,6 +17,19 @@ class ApplicationController < ActionController::Base
...
 
+  def storable_location?
+    request.get? && !request.xhr? && request.fullpath !~ /auth/
+  end
+
+  def store_user_location!
+    uri = URI.parse(request.fullpath)
+    return unless uri
+
+    path = [uri.path.sub(/\A\/+/, '/'), uri.query].compact.join('?')
```
cwant: Might not hurt to have a comment here about the regex -- something is getting replaced by a slash, but it's not obvious to me as to what or why. (One or more slashes at the beginning of the string?)

murny: Yeah forsure can add. 

Basically what its doing, is it adds the query param (?search=hey) and next line adds the anchor tag ('#cool-stuff-here') back onto the URI so when you redirect it redirects where you came from exactly and not just the current url

test/system/authentication_test.rb
```diff
@@ -64,38 +64,59 @@ class AuthenticationTest < ApplicationSystemTestCase
...
-    #   click_link I18n.t('sessions.new.saml_link')
+    click_link I18n.t('application.navbar.links.login')
 
-    #   assert_text I18n.t('authorization.user_not_authorized')
+    assert_text I18n.t('login.success')
 
-    #   assert_selector 'h1', text: I18n.t('sessions.new.header')
-    # end
+    # Still on browse page
+    assert_selector 'h1', text: I18n.t('communities.index.header')
```
cwant: Is it worth testing being on the browse page directly, rather than inferring it? (With something like `assert_equal URI.parse(current_url).request_uri, browse_path`.)

murny: Apparently theres a `assert_current_path` helper that will verify if its on that page, so added:

`assert_current_path(communites_path)` 

here as well

app/controllers/application_controller.rb
murny: Added some comments

test/system/authentication_test.rb
```diff
@@ -46,9 +46,10 @@ class AuthenticationTest < ApplicationSystemTestCase
   end
 
   context 'when visiting a protected page' do
-    should 'get redirected to login then back to page, if user is authorized' do
+    should 'get redirected to homepage then back to page, if user is authorized' do
```
murny: No longer got a login page, update description

PR #536, opened 2018-03-06
Change `sort_year` from string to integer.

cwant:
Also, resolve issues between Jupiter's use of `int` and Solrizer's use of `integer`.

### DIFF
--- a/app/models/concerns/item_properties.rb
+++ b/app/models/concerns/item_properties.rb
@@ -19,7 +19,7 @@ module ItemProperties
     has_attribute :northern_north_america_item_id,
                   ::TERMS[:ual].northern_north_america_item_id, solrize_for: :exact_match
     has_attribute :rights, ::RDF::Vocab::DC11.rights, solrize_for: :exact_match
-    has_attribute :sort_year, ::TERMS[:ual].sort_year, solrize_for: [:search, :sort, :facet]
+    has_attribute :sort_year, ::TERMS[:ual].sort_year, type: :integer, solrize_for: [:search, :sort, :facet]
     has_attribute :visibility_after_embargo, ::TERMS[:acl].visibility_after_embargo, solrize_for: :exact_match
 
     has_multival_attribute :embargo_history, ::TERMS[:acl].embargo_history, solrize_for: :exact_match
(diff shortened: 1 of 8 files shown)


cwant:
@mbarnett I think you had better take a look at this one -- seems to be some issues between the integer type between Jupiter (`:int`) and Solrizer (`:integer`).

app/models/concerns/item_properties.rb
```diff
@@ -19,7 +19,7 @@ module ItemProperties
     has_attribute :northern_north_america_item_id,
                   ::TERMS[:ual].northern_north_america_item_id, solrize_for: :exact_match
     has_attribute :rights, ::RDF::Vocab::DC11.rights, solrize_for: :exact_match
-    has_attribute :sort_year, ::TERMS[:ual].sort_year, solrize_for: [:search, :sort, :facet]
+    has_attribute :sort_year, ::TERMS[:ual].sort_year, type: :integer, solrize_for: [:search, :sort, :facet]
```
cwant: A particular issue that is driving a lot of the core changes is that if I set `type: :int` here (instead of `:integer`), I get the error:

```
/home/cwant/gitwork/jupiter/vendor/bundle/ruby/2.4.0/bundler/gems/solrizer-4c9712f9309b/lib/solrizer/suffix.rb:73:in `block in config': Invalid datatype `:int'. Must be one of: :date, :time,
 :text, :text_en, :string, :symbol, :integer, :boolean (Solrizer::InvalidIndexDescriptor)
```

This was puzzling to me, since `LockedLdpObject` uses `:int` type for the `owner` attribute.
As a result, I've changed `:int` to `:integer` everywhere in this PR to match what Solrizer wants.

pgwillia: If this is the right direction then theres a `:bool` `:boolean` mismatch too.

mbarnett: Yeah, Tricia is correct, `bool` is wrong too. Nothing had used integer or boolean types prior to this, so I'd never checked to see if the mappings actually matched what Solrizer called the types for those.

cwant: Sounds good, will update `bool` too. There was one thing using it: the restricted collection flag (surprised it didn't complain about that)

cwant:
Thanks, looks like merging master broke the test suite. Fixed now, but the PR will need re-approval.

PR #538, opened 2018-03-07
rake task to fetch-and-unlock-every-object-and-save (reindex)

pgwillia:


lib/tasks/cleanup.rake
```diff
@@ -15,4 +15,11 @@ namespace :cleanup do
 
     puts 'Cleanup of Inactive Draft Items now completed!'
   end
+
+  desc 'fetch-and-unlock-every-object-and-save'
```
murny: It's okay to make the description human readable. Can probably use real spaces here/etc.

lib/tasks/cleanup.rake
```diff
@@ -15,4 +15,11 @@ namespace :cleanup do
...
+  task reindex: :environment do
```
murny: Throwing this in cleanup will namespace it as such, `rake cleanup:reindex`

Is this intended? Can always create a new rake task file with a better namespace if required, not an issue?

pgwillia: I thought about putting it in a different namespace but I thought `rake cleanup:reindex` made sense to me.  What would you suggest is a good namespace?

lib/tasks/cleanup.rake
pgwillia: Fair.  I was being cheeky because that was the description that came back from the standup yesterday in slack.

pgwillia:
This may cause problems with `reverse_solr_name_cache`. I'm seeing duplicates for filename_ssim and three different sort_years
```
facet_map=>
    {...
     "northern_north_america_filename_ssim"=>:northern_north_america_filename,
     "northern_north_america_item_id_ssim"=>:northern_north_america_item_id,
     ...
     "sort_year_isim"=>:sort_year,
     "sort_year_isi"=>:sort_year,
     "sort_year_sim"=>:sort_year,
     "publisher_tesim"=>:publisher,
     "publisher_sim"=>:publisher,
     "languages_tesim"=>:languages,
     "languages_sim"=>:languages,
     ...}
```

lib/tasks/cleanup.rake
murny: Nope no suggestion. Fine to leave it as is then :+1:

pgwillia:
The duplication should be fine.  Need to change https://github.com/ualbertalib/jupiter/blob/master/app/models/jupiter_core/locked_ldp_object.rb#L217 to take `:type` into account. I'll do that in the PR where sort_date is being used.

PR #542, opened 2018-03-08
Install chromedriver via gem instead

murny:
https://github.com/ualbertalib/jupiter/pull/532#discussion_r172315937

> As comment says this will install chromedriver for you. This is a new default rails gem in rails 5.2 (when you rails new this gem is in the gemfile etc.) Might be easier to just use this gem install of having people install chromedriver on their own and might make travisCI work better as well (currently travisCI is being flaky when it comes with the chromedriver/system tests)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,14 +17,6 @@ rvm:
 cache: bundler
 bundler_args: --without development production --jobs=3 --retry=3
 
-before_install: # Install ChromeDriver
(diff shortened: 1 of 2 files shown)


cwant:
Looks like the Gemfile.lock was missed in this PR, will add it in a separate PR.

murny:
Ooh thanks! Sorry!

Issue #224, opened 2017-10-30
Go Live Checklist


murny:
# Epic for capturing go-live tasks for all teams:

### Pre-launch blockers:

- [x] Multi-valued fields (#568)
- [x] page titles (Is this already done?)
- [ ] Completed user and object migration and audit (Metadata also reviewing this)
- [x] Google Analytics/Piwik set up 
- [x] make sure [all links on the site](https://gist.github.com/pgwillia/97fb2931254c02057418) is working, contact info is correct
- [x] make sure the collection image links on the homepage are pointing at the correct collections (#584)
- [x] Google Analytics set up - connect to the production instance. 
- [ ] Pushmi-Pullyu deployment
- [x] Robots.txt changed before go live (this is just a [Rails env variable change](https://github.com/ualbertalib/jupiter/blob/master/app/views/robots/robots.text.erb#L4) on Prod)
- [x] Pre-populate Redis download/view counts (Peter working on this)
- [x] DOI merged
- [x] #557
- [x] #519 (static text)
- [x] #556
- [x]  #581
- [x] Only admins can see or use visibility facet
- [x] Created displaying created not sort_year
- [x] #596
- [x] #597 
- [x] Henry to confirm w/ Tyler re: return from IST SAML server pointing at era.library.ualberta.ca instead of era-beta.library.ualberta.ca

### Can happen post-launch 

- [x] Completed accessibility scans and fixes with automated tools (Wave/AxE) #220 
- [x] Keyboard tests #223 
- [x] Fix thumb-nailing issue on production
- [x] meta tags
- [x] set up the 404 page
- [x] responsive testing on various devices
- [x] Make sure we have a favicon in place
- [x] Cross-Browser check - check site in all mainstream browsers: Chrome, Firefox, IE, Safari. 
- [x] Graceful degradation check, try functions without javascript
- [x] Google Search Console set up
- [ ] Other SEO configuration checks
- [x] check against: https://github.com/thedaviddias/Front-End-Checklist (issue found here: #562)
- [x] check against: http://a11yproject.com/checklist.html
- [x] #543 
- [x] #580 
- [x] #468
- [x] OAI data checking complete
- [ ] Finish load of related objects 

## Tests
- [ ] Completed user acceptance test and usability test. 
- [x] Performance testing
  - [x] automated Jmeter testing (Ops Team)
  - [x] stress testing with concurrent users

## Ops

- [ ] Operationalize the application - on-call process
- [x] Monitoring
[...]

PR #543, opened 2018-03-08
#562 Frontend checklist (part 1)

murny:
Addresses a few issues from this ticket: 
https://github.com/ualbertalib/jupiter/issues/562
(this is a sub ticket of the epic ticket here: https://github.com/ualbertalib/jupiter/issues/224)

Following been fixed in this PR (will open another PR for other stuff)
### HEAD
- [x] Put important meta tags at top of page (charset, viewport, etc)
- [x] Truncate long title tags, google wants them to be around 55 characters
- [x] Generate Favicons for different devices/browsers/etc
- [x] rel="noopener"
- [x] Canonical url 

### HTML
- [x] Linting of HTML, W3C compliant, broken links etc.

### CSS
- [x] Unique Ids are actually unique on the page
- [x] no embedded or inline css


app/models/presenters/facet_values/default_presenter.rb
```diff
@@ -11,15 +11,17 @@ def initialize(view, active_facets, facet_value)
   end
 
   def facet_search_link
+    # TODO: can we move this to the view?
     if @active_facets[@solr_index].present? && @active_facets[@solr_index].include?(@value)
       @view.link_to @view.query_params_without_facet_value(@solr_index, @value), rel: 'nofollow' do
-        @view.concat(@view.content_tag(:input, '', type: 'checkbox', 'aria-label' => @value, checked: true,
```
murny: Not valid html to have a input tag inside of a anchor tag like this. So alternative option is just use fontawesome check-square/square to give a similar visualization.

Also moved the count number within the link and float it right and style it as a badge, looks better this way and people can click on it as well.

(Side note, we can probably move this html concatenation out of here and into the view itself maybe?)

app/views/admin/announcements/index.html.erb
```diff
@@ -11,7 +11,7 @@
   </fieldset>
 <% end %>
 
-<h2 class="mt-5"><%= t('.current_announcements') %></h5>
+<h2 class="mt-5"><%= t('.current_announcements') %></h2>
```
murny: Lots of these with wrong tag endings or missing closing divs. Ooops

app/views/application/_navbar.html.erb
```diff
@@ -18,7 +18,7 @@
       <span class="navbar-toggler-icon"></span>
     </button>
 
-    <div class="collapse navbar-collapse" id="jupiter-main-nav" aria-expanded="false">
+    <div class="collapse navbar-collapse" id="jupiter-main-nav">
```
murny: Don't need this. If you add it, it want's a `role=` and other attributes

app/views/application/_navbar.html.erb
```diff
@@ -64,7 +64,7 @@
               <%= current_user.name %>
             </a>
 
-            <div role="menu" class="dropdown-menu dropdown-menu-right" aria-labelledby="#jupiter-user-nav-downdown">
+            <div role="menu" class="dropdown-menu dropdown-menu-right" aria-labelledby="jupiter-user-nav-downdown">
```
murny: should be the name of the id, not the ID with hash

app/views/items/draft/_files_list.html.erb
```diff
@@ -14,9 +14,11 @@
       </div>
     </div>
     <div class="col-md-8">
-      <% @draft_item.files.each do |file| %>
-        <%= render partial: 'items/draft/file', locals: { file: file } %>
-      <% end %>
+      <ul class="list-unstyled">
```
murny: rending partial which first element is a `li` within a `div`. Which isn't valid, needs to be nested within `ul`

app/views/items/draft/_progress_bar.html.erb
```diff
@@ -1,7 +1,7 @@
 <% page_title( header) %>
-<h1 class="mb-3"><%= header %></h1>
-
 <section>
+  <h1 class="mb-3"><%= header %></h1>
```
murny: Sections need a header. So could either remove the section or move the header down? Choose latter.

app/views/items/draft/choose_license_and_visibility.html.erb
```diff
@@ -29,7 +29,7 @@
 
         <% license_accordions.map do |accordion_name, licenses| %>
           <div class="card">
-            <div id="js-<%= "#{accordion_name}" %>" class="card-header" role="tab">
+            <div id="js-heading-<%= "#{accordion_name}" %>" class="card-header" role="tab">
```
murny: aria-labelledby below was referencing non existent id. Fix the id to match

app/views/items/draft/describe_item.html.erb
```diff
@@ -32,11 +32,8 @@
         label: t('.type_id'),
         label_method: :translated_name,
         value_method: :id,
-        # TODO: Bug in simple form: https://github.com/plataformatec/simple_form/issues/1427
```
murny: So maybe theres a reason for this? It's invalid html to have a empty prompt? So lets just create/move the placeholder as the prompt. Better UX anyways.

app/views/items/draft/describe_item.html.erb
```diff
@@ -79,20 +70,17 @@
...
+        prompt: t('.subjects_placeholder'),
         hint: t('.subjects_help') %>
 
     <%= f.input :date_created,
                 as: :date,
                 label: t('.date_created'),
                 order: [:year, :month, :day],
                 prompt: {
-                  day: false,
-                  month: false,
```
murny: Again if you have a `required: true` select, you need to have a non empty prompt. This requires some work around in the controller as I was exploiting this so year is required but month/day is optional.

app/views/items/draft/upload_files.html.erb
```diff
@@ -21,6 +21,7 @@
                       url: item_files_path,
                       method: :post,
                       html: {
+                        id: :upload_files_form,
```
murny: Simple_form derives the default form id from the object. However we have two simple_form forms here, which means we duplicating the ID twice. Fix this by explicitly setting the ID here.

app/views/items/draft/upload_files.html.erb
```diff
@@ -50,9 +51,9 @@
   <% end %>
 
   <%# Code below is used as the dropzone previewTemplate, will not be displayed to user %>
-  <div id="js-dropzone-preview-template" class="d-none">
+  <ul id="js-dropzone-preview-template" class="d-none">
     <li class="media my-3 p-3 border">
```
murny: Because this is a `li` needs to be nested in a `ul` instead of a `div`. Easy fix

app/views/items/draft/upload_files.html.erb
```diff
@@ -50,9 +51,9 @@
...
-      <img data-dz-thumbnail alt='' class='mr-3'>
+      <img data-dz-thumbnail src='#' alt='' class='mr-3'>
```
murny: This is a tricky one... `img`'s `src` attribute is a required attribute for valid html. 

However we don't have one, as this is a dummy template. Setting this as `src='#'` actually can be bad? As most browsers will do a another additional request to this page so it can grab the image. There's some hacks to get around this like setting `src='//:0'` which seems to be the best option? 

Another option is when we grab the template via javascript in dropzone, we remove this html from the dom entirely, so it doesn't get flagged as invalid html.

app/views/items/draft/upload_files.html.erb
```diff
@@ -77,9 +78,9 @@
                       html: { class:'js-deposit-item' } do |f| %>
 
     <h4><%= t('.saved_files') %></h4>
-    <ul id="js-files-list" class="list-unstyled">
+    <div id="js-files-list">
```
murny: file_list grown to include much more then just a list of images/files. As a result it's no longer has first element as a `li`. Fix this as you can't nest a `div` under a `ul`

app/views/kaminari/_page.html.erb
```diff
@@ -1,9 +1,3 @@
-<% if page.current? %>
-  <li class="page-item active">
-    <%= content_tag :a, page, remote: remote, rel: (page.next? ? 'next' : (page.prev? ? 'prev' : nil)), class: 'page-link' %>
```
murny: By creating an anchor tag with `remote: remote` here actually makes it as a attribute like 

`<a remote="false">1</a>`

Which is invalid html.

Instead of how `link_to` does this which sets it as a data-attr which is valid html:

`<a data-remote="false">1</a>`

Fix this locally, simply just use a link_to to current_url and set the active on the li when its current page or not.

app/views/search/_facet_results.html.erb
```diff
@@ -5,30 +5,32 @@
     <% facet.each_facet_value(0..@max_facets-1) do |facet_value| %>
       <%= render partial: 'search/facet_value', locals: { value: present(facet_value)} %>
     <% end %>
-    <% if facet.count > @max_facets %>
-      <%# Other results collapsed by default %>
-      <div class='js-hideshow'>
-        <div class="collapse" id='<%= "#{facet.solr_index}_hidden" %>'>
+  </ul>
```
murny: We got divs inside of `ul` here. This fixes this by separating the lists into two separate lists.

app/views/search/_model_results.html.erb
```diff
@@ -14,7 +14,7 @@
           <% end %>
         </div>
       </div>
-      <div class="card-body p-2 pr-3 pr-md-0 filter-list">
```
murny: Notice we were removing the padding on the facets? So they look like this:
image

Notice on the right side the facet touches the outside container.

Now they look like this:
image

app/views/search/_model_results.html.erb
```diff
@@ -80,4 +80,5 @@
 </div>
 
 <%# Overlay for when mobile filter sidebar is open %>
-<div class="filters-overlay"><div>
+<div class="filters-overlay">
+</div>
```
murny: Used an opening tag instead of closing. Fixed

config/locales/en.yml
```diff
@@ -488,12 +489,10 @@ en:
           province, or country the item's contents are about."
         time_periods: 'Is this item about a time?'
         time_periods_placeholder: 'Enter one or multiple times'
-        time_periods_help_html: 'Examples:
-          <ul>
```
murny: Having this ul/li markup inside of a `small` tag isn't valid. Just use BRs instead?

app/views/items/draft/upload_files.html.erb
murny: So apparently this was breaking system tests? Couldn't get pass the upload form for some reason. Changing this to `src='//:0'` seemed to fix it

murny: Just removing img entirely now (as html validator didnt like src='//:0' either), as we don't even show thumbnails on upload anyways

app/views/items/_item.html.erb
```diff
@@ -99,7 +99,7 @@
       href="#more-information-hidden"
       aria-expanded="false"
       aria-controls="more-information-hidden">
-    <span class='js-hideshow-control font-weight-bold font-italic' style='display:none;'>
+    <span class='js-hideshow-control font-weight-bold font-italic d-none'>
```
murny: Opt for css class (bootstrap) instead of inline style

app/views/search/index.html.erb
```diff
@@ -1,9 +1,6 @@
 <% page_title( t('.header')) %>
 <% content_for :head do %>
   <meta name="robots" content="noindex" />
-  <% if @active_tab == :community %>
-    <link rel="canonical" href='<%= communities_path %>' />
```
murny: Talked to matt about this. We believe this to be incorrect? Communities index (browse) is different content then the search with communities tab. So shouldn't be canonically it to browse like this.

config/initializers/canonical_rails.rb
```diff
@@ -0,0 +1,28 @@
+# Do yourself a favor and set these up right when you install the engine.
+
+CanonicalRails.setup do |config|
+  # Force the protocol. If you do not specify, the protocol will be based on the incoming request's protocol.
```
murny: Leaving this as all defaults (could force the port/host/etc, but lets just leave this so it derives this information from the requesting url

public/404.html
```diff
@@ -4,7 +4,7 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Page Not Found (404) | ERA</title>
-  <style type="text/css" media="screen">
```
murny: This is redundant, just remove

app/views/search/index.html.erb
pgwillia: Looking back at this I agree that this is incorrect. https://github.com/ualbertalib/jupiter/blob/master/app/views/communities/_community.html.erb#L4 should be where the canonical was placed.

murny: If you need a refresher or anything let me know. More info can be found here: https://support.google.com/webmasters/answer/139066?hl=en

Basically a way used to reduce duplicate content/urls and telling google that they are the same thing and which one is "canonical".

So for example on the search page, you may have `/search?query=random`, `/search?page=2`, `/search`, `search?tab=communities` and thousands of different permutations here and google will treat them all as _different_ pages! So your SEO juice/weighting for the `/search` page will gets split for all these different permutations (and potentially google may even penalize you for having duplicate content on your site!). A canonical URL explicitly tells google which URL is canonical and that all SEO juice/weighting from the duplicates should go to this canonical url instead. For example everything with query params on `/search` like  `/search?query=random` is actually the same as  `/search` so the canonical URL should be `/search`

I am using a gem called `'canonical-rails'` which does all this hard work for me: https://github.com/jumph4x/canonical-rails

PR #549, opened 2018-03-09
Working around bugs in the list_source stuff

### LINKED ISSUE #539: Listsource content just keeps pointing to the same thing
Below is the Listsource datastream for an object with multiple files. All of the listsource attributes point to the same thing (which is the URI for the listsource datastream with an anchor tag). It is likely that either Hydra::Works is not working correctly, or we are not using it correctly.

listsource_issue

cwant:
Also: adding seed items with three files for testing.

Fixes #539


test/models/item_test.rb
```diff
@@ -33,7 +34,6 @@ class ItemTest < ActiveSupport::TestCase
     item = Item.new_locked_ldp_object
 
     assert_nil item.visibility
-    assert_equal 0, Item.public.count
```
cwant: I was having some issues with these kinds of assertions failing when I added my test below. In all honesty, I'm not sure what they are testing, since the tests don't actually save anything (they could possibly be reworked to use `assert_changed` or `assert_difference` to make them less order-dependent).

pgwillia: I think what I was trying to show was that the Item.public method I added returned the right number of Items under those conditions.

cwant: I've reworked some of the `Item.public` tests as part of the first test -- let me know if they work for you.

pgwillia: Nice! :+1:

Issue #466, opened 2018-02-09
File titles with spaces make sitemap not well formed


pgwillia:
![screenshot from 2018-02-09 14-29-07](https://user-images.githubusercontent.com/1220762/36050981-c781dd6e-0da5-11e8-9ef0-4c40e8f794b1.png)
https://era-beta.library.ualberta.ca/sitemap-items.xml

mbarnett:
Fixed with Tricia's merge on March 9

PR #552, opened 2018-03-09
Ensure data values (including URLs) in sitemap.xml must use entity escape codes

pgwillia:
#466 requires reindex to update sitemap_link with urls encoded:  `bundle exec rake cleanup:reindex`


test/integration/sitemap_test.rb
```diff
@@ -22,9 +22,13 @@ def before_all
                   uo.add_to_path(@community.id, @collection.id)
                   uo.save!
                   # Attach a file to the item so that it has attributes to check for
-                  file = File.open(file_fixture('image-sample.jpeg'), 'r')
-                  uo.add_files([file])
-                  file.close
+                  File.open(file_fixture('image-sample.jpeg'), 'r') do |file|
+                    # Bit of a hack to fake a file name with characters that require escaping ...
+                    def file.original_filename
+                      "&'\"><ü.jpeg"
```
pgwillia: https://www.sitemaps.org/protocol.html#escaping

Issue #562, opened 2018-03-14
Review and Address issues from Front End Checklist


murny:
Full list found here: https://github.com/thedaviddias/Front-End-Checklist 

Broken down list (for what I think is relevant?) is below:

### HEAD
- [x] Put important meta tags at top of page (charset, viewport, etc)
- [x] Truncate long title tags, google wants them to be around 55 characters
- [x] Generate Favicons for different devices/browsers/etc
- [x] rel="noopener"
- [x] Canonical url 
- [x] Facebook OG and Twitter Cards
- [x] Need to figure out meta description (with regards to work already being done here https://github.com/ualbertalib/jupiter/pull/483 and we need the copy for the main description)

### HTML
- [ ] Semantic elements being used
- [x] Linting of HTML, W3C compliant, broken links etc.

### CSS
- [x] Website is responsive
- [ ] Website tested on various devices: various combinations of desktop, mobile, browsers, operating systems
- [x] Unique Ids are actually unique on the page
- [x] js prefix on css classes
- [x] no embedded or inline css

### Images
- [x] Images have been optimized for web (thanks imgbot?)
- [x] width/height set on `<img>` tag
- [x] alt text being used properly
- Should use a better logo, that has better dimensions 
### Javascript
- [x] No inline javascript
- [ ] Noscript tag?
- [ ] Do we need jquery everywhere?
- [ ] ESlint?

### Security
- [x] Scan and check website using various tools
  - https://securityheaders.io/?q=%09https%3A%2F%2Fera.library.ualberta.ca%2F&followRedirects=on (missing a few headers:  Strict-Transport-Security, Content-Security-Policy, Referrer-Policy)
  - https://observatory.mozilla.org/analyze/era.library.ualberta.ca
  - Essentially we should configuring the application to not return unnecessary headers as this keeps information silent and makes it significantly more difficult to identify the underlying frameworks. Turn on secure cookies. Turn on HSTS
- Our SSL certificate has errors: This server's certificate chain is incomplete (Invalid certificate chain)
- [x] Https used everywhere
- [x] checks for CSRF/XSS 
- [ ] ... more stuff maybe we should be checking?

### Performance
- [ ] First meaningful paint under 1 sec, time to interactive under 5 secs for average configuration (crappy android phone on slow 3g network) and 2 secs for repeat visits
  - Big suggestion is to use a CDN for javascript/css assets (cookieless domain/cachable/etc)
[...]

PR #565, opened 2018-03-15
Frontend checklist (part 2)

murny:
Continuation from #543 

Addresses a few issues from this ticket:
#562

### CSS
- [x] Center links on footer, looks much better on mobile
- [x] Remove pull-right for float-right instead (deprecation in bootstrap 4)
- [x]  Website is responsive (verified and fixed problems with regards to styling on an emulated iphone 8 plus, may be some more issues with smaller screens)

### Accessibility
- [x] Color contrast should pass WCAG AAA
- [x] Labels are used (aria-label if label can't be displayed)
- [x] Wave testing done

Found Issues:
- Communtiy/Collection Adding/Removing on Describe Item reuses the same element ID's
- Select2 is pretty bad for accessibility:
  - On Deposit item, it seems the underlying input field it uses doesn't refer to its label so you get an "Ensures every form element has a label" error
  - On Browse typeahead, its failing because of its usage of `role='combobox'` with the following warning: "Ensures elements with an ARIA role that require child roles contain them". Also placeholder text on typeahead is failing color contrast ratio

AXE false positives:
- AXE fails the color contrast on the gap in the pagination which renders a ellipsis (...). No matter what I do this fails.
- AXE Color contrast fails automatically when text is on a background image, so 6 failures on homepage (2 for banner and 4 for the featured collections). Should be a false positive for most of them.
- AXE wants radio buttons within a group which I have them in a fieldset/legend, must be a false positive on their end? (https://www.w3.org/WAI/tutorials/forms/grouping/)


app/assets/javascripts/jupiter/item_show.js
```diff
@@ -3,4 +3,13 @@ $(document).on('turbolinks:load', function() {
     event.preventDefault();
     $('.js-download').multiDownload();
   });
+
+  $('#more-information-hidden').on('hidden.bs.collapse', function() {
```
murny: Use a real accordion for expanding/hidding the extra fields on the show page. Before we were using two links and the hidden link is still in the tab order which fails accessibility. Here we just got one link which changes text depending on opened/closed

app/views/application/_item.html.erb
```diff
@@ -1,18 +1,21 @@
...
       <h5 class="mt-0"><%= link_to item.title, path_for_result(item) %></h5>
-      <div class="btn-group">
-        <% primary_file = item.file_sets.first %>
-        <% if primary_file.present? %>
-          <%= link_to fileset_uri(primary_file, :download), class: 'btn btn-secondary', rel: 'nofollow' do %>
-            <%= fa_icon 'cloud-download' %>
-            <%= t('download') %>
-          <% end %>
+
+      <% primary_file = item.file_sets.first %>
```
murny: Move the download button out of the btn group which has edit/delete actions. Putting all three of them in a btn group is very big and prevents proper responsiveness on small screens

app/views/application/_navbar.html.erb
```diff
@@ -23,9 +23,9 @@
 
         <li role="presentation" class="nav-item mr-3">
           <form class="d-inline" action="<%= search_path %>">
-            <div class="input-group" aria-label="navbar-search">
+            <div class="input-group">
               <%= search_field_tag 'search', nil, placeholder: t('.search.placeholder'),
-                  value: params[:search], id: 'navbar-search-lg', 'aria-label': 'search-box', class: "form-control" %>
+                  value: params[:search], id: 'navbar-search-lg', 'aria-label': t('.search.placeholder'), class: "form-control" %>
```
murny: Only the search field needs an aria-label. Reused the placeholder for the text so label becomes: 'Search ERA...'

app/views/application/_thumbnail.html.erb
```diff
@@ -1,15 +1,16 @@
 <% if policy(object).thumbnail? %>
   <% if thumbnail.present? && image?(thumbnail) %>
-    <%= image_tag(thumbnail.url, class: 'j-thumbnail img-thumbnail', alt: '', size: '100x100') %>
+    <%= image_tag(thumbnail.url, class: 'j-thumbnail img-thumbnail', alt: '', title: object.title, size: '100x100') %>
```
murny: Because the image is surrounded by a link, the link needs discernible text for accessibility. Adding title attr (tooltips) to the image fixes this. For non images (font awesome placeholder) we can add a sr-only span to achieve the same effect.

murny: Everything using thumbnail currently has an object with a title attribute. If this ends up changing we may have to pass in the title into the thumbnail partial instead.

app/views/collections/show.html.erb
```diff
@@ -46,7 +46,7 @@
         <%= t('.deposit_item') %>
       <% end %>
 
-      <%= link_to '#TODO', class: 'btn btn-primary' do %>
+      <%= link_to '#TODO', class: 'btn btn-primary  mt-2 mt-md-0' do %>
```
murny: Want some top padding on mobile, otherwise buttons are touching each other

app/views/communities/_community.html.erb
```diff
@@ -1,30 +1,30 @@
...
+    </div>
+  <% end %>
 
-    <div class="btn-group" role="group" aria-label="Community Actions">
-      <% if policy(community).update? %>
-        <%= link_to edit_admin_community_path(community), class:'btn btn-outline-secondary' do %>
-          <%= fa_icon('pencil-square-o') %>
-          <%= t('edit') %>
-        <% end %>
+  <div class="btn-group mb-3" role="group" aria-label="Community Actions">
```
murny: Basically whats happening here, is I am moving the buttons (edit/delete/etc) below the media content of the community. This was because on smaller screen sizes, the buttons would keep getting pushed and eventually go off the screen thus preventing responsiveness

app/views/items/_files_section_sidebar.html.erb
```diff
@@ -27,7 +27,7 @@
           </div>
         </div>
       <% end %>
-      <div class="p-1">
+      <div class="pt-3">
```
murny: Download All button was basically touching the top of the download buttons grouping. Space it out a bit

app/views/items/_item.html.erb
```diff
@@ -79,28 +80,23 @@
...
 
 <%# ADDITIONAL INFORMATION %>
-<div class='js-hideshow'>
-  <a data-toggle="collapse"
-      data-parent=".js-hideshow"
-      href="#more-information-hidden"
-      aria-expanded="false"
-      aria-controls="more-information-hidden">
-    <span class='js-hideshow-control font-weight-bold font-italic'>
+<button type="button"
```
murny: If its not a navigation link typically should opt for a `button` instead of an anchor `a`. Also essentially use a bootstrap collapsible accordion here instead of two links as stated above.

app/views/items/draft/_community_collection_input.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="js-community-collection">
   <div class="form-group">
-    <label>
+    <%= label_tag 'draft_item[community_id][]' do %>
```
murny: Link the label to the input item for community/collection

(Warning! We are sharing the ids here when a user adds multiple community/collection pairs).

app/views/items/draft/choose_license_and_visibility.html.erb
```diff
@@ -103,7 +103,7 @@
             year: t('select_a_year')
           },
           start_year: Date.today.year,
-          input_html: { class: 'mr-2 mb-2' } %>
+          input_html: { class: 'mr-2 mb-2', 'aria-label': t('.visibility.embargo') } %>
```
murny: date_select only references the label to the first select box (in this case year), need labels for the other 2, so lets just set all of them to have `'aria-label': t('.visibility.embargo')`

app/views/items/draft/choose_license_and_visibility.html.erb
```diff
@@ -116,11 +116,10 @@
           </label>
         </div>
       <% end %>
-
     </fieldset>
 
-    <div class="mt-3 pull-right">
```
murny: Buttons here were getting overlapped by footer. Pull right does some funny things when they get pushed onto new lines, wasn't acting like a `display: block`

Also apparently `pull-right|left` don't exist anymore? Maybe ill update all references to this to the new `float-left|right`

config/initializers/kaminari_config.rb
```diff
@@ -1,7 +1,7 @@
 Kaminari.configure do |config|
   config.default_per_page = 10
   # config.max_per_page = nil
-  # config.window = 4
+  config.window = 2
```
murny: Make the window smaller as pagination to way to long on mobile with 4. What this means is if active page is `?page=5`, the pagination will show links for:

First, Prev, 1, 2, 3, 4, active 5, 6, 7, 8, 9, gap, Next, Last

Now window is 2 so its just:

First, Prev, gap, 3, 4, active 5, 6, 7, gap, Next, Last

config/locales/en.yml
```diff
@@ -656,3 +656,11 @@ en:
     item_type_with_status: 'Item type'
     all_contributors: 'Author / Creator / Contributor'
     all_subjects: 'Subject / Keyword'
+
+  views:
+    pagination:
+      first: "&laquo; <span class='d-none d-md-inline'>First</span>"
```
murny: Basically what this is doing is we want to hide the text on mobile and only show the icon. So instead of `<< First`, it just gets `<<` on mobile

test/integration/collection_show_test.rb
```diff
@@ -121,8 +121,8 @@ def before_all
 
     # Only 'Fancy' items are shown
     assert_select '.jupiter-results ul.list-group .list-group-item', count: 1
-    assert_select '.jupiter-results ul.list-group .list-group-item a', text: 'Fancy Item', count: 1
-    assert_select '.jupiter-results ul.list-group .list-group-item a', text: 'Nice Item', count: 0
+    assert_select '.jupiter-results ul.list-group .list-group-item h5 a', text: 'Fancy Item', count: 1
```
murny: Since there is actually to links with the text 'Fancy Item' on the page (one for the image with title attr and one for the header (h5)), we just want to target the header one

app/views/items/draft/_community_collection_input.html.erb
mbarnett: Note that what we want to be doing is sharing the _name_ between input elements, not the _id_. Unlike IDs, it's perfectly fine (indeed, necessary) that names aren't unique when there are multiple inputs for values targeting the same field.

There's no reason we couldn't generate unique _IDs_ for every input field while they share the same name, and then the label could properly target the right input field. But if this isn't breaking anything I'd vote for doing that post-launch.

app/views/collections/show.html.erb
mbarnett: Let's just get rid of this link. I don't think manual Thesis deposit is even a glimmer in anyone's eye right now

murny: Remove this one and the one in admin items

app/views/search/_model_results.html.erb
```diff
@@ -39,7 +39,7 @@
       </div>
 
       <div class="d-md-none mt-2">
-        <%= link_to '#TODO', class: 'btn btn-jupiter-primary mb-2 js-filters-open' do %>
+        <%= button_tag type: :button, class: 'btn btn-jupiter-primary mb-2 js-filters-open' do %>
```
murny: Having this as a link_to pointing to #TODO is not ideal. This isnt used for navigation, so convert to a button instead

PR #572, opened 2018-03-20
Range facet for Sort Year

pgwillia:
#339


app/models/jupiter_core/facet_result.rb
```diff
@@ -1,10 +1,8 @@
 class JupiterCore::FacetResult
```
pgwillia: Refactored to be parent of FieldValueFacetResult and RangeFacetResult

app/models/jupiter_core/search.rb
```diff
@@ -47,6 +51,7 @@ def self.faceted_search(q: '', facets: [], models: [], as: nil)
                                               qf: calculate_queried_fields(models),
                                               facet_map: construct_facet_map(models),
                                               facet_fields: models.map(&:facets).flatten.uniq,
+                                              ranges: ranges,
```
pgwillia: Passes forward the range parameters to hang populate the form in the range facet.

app/views/search/_model_results.html.erb
```diff
@@ -54,6 +54,10 @@
                          raw_value: facet_value.value,
                          solr_index: facet_value.solr_index } %>
             <% end %>
+            <% if params[:ranges]&.fetch(facet_value.solr_index, false) %>
+              <%= render partial: 'search/range_facet_badge', locals: { value: facet_value.range.to_s.gsub('..', ' to '),
```
pgwillia: Shows the facet badge as '1880 to 2018'

app/views/search/_range_facet_badge.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= link_to query_params_without_range_value(solr_index) do %>
```
pgwillia: Otherwise the same as the _facet_badge.html.erb

app/views/search/_range_facet_results.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <% else %>
+        <% h.each do |k, v| %>
+          <% next if k == facet.solr_index %>
+          <% v.each do |i| %>
+            <%= hidden_field_tag("#{type}[#{k}][]", i) %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+    <div class='form-group'>
```
pgwillia: Still working through bootstrap styles to find a better look for the begin to end part of the form.

app/models/jupiter_core/deferred_faceted_solr_query.rb
```diff
@@ -114,7 +115,14 @@ def reify_result_set
     )
 
     @facets = facet_data['facet_fields'].map do |k, v|
-      JupiterCore::FacetResult.new(criteria[:facet_map], k, v) if v.present?
+      if k == 'sort_year_isi'
+        JupiterCore::RangeFacetResult.new(
+          criteria[:facet_map], k,
+          criteria[:ranges].fetch(k, begin: 1880, end: Time.current.year).to_h
```
pgwillia: This is where the defaults are initialized.  Is there a better place?

app/controllers/search_controller.rb
```diff
@@ -46,4 +50,20 @@ def index
     end
   end
 
+  def validate_ranges
+    params[:ranges]&.each do |facet|
+      start = params.dig(:ranges, facet, :begin)
+      finish = params.dig(:ranges, facet, :end)
+      message = ''
```
murny: Can maybe simplify this? Message will always be empty if it doesn't hit the first `unless`, so you could just combine the two `unless` into one. 

```ruby
    start = params.dig(:ranges, facet, :begin)
    finish = params.dig(:ranges, facet, :end)
    
    unless start.match?(/\d{4}/) && finish.match?(/\d{4}/) && (start.to_i <= finish.to_i)
      flash[:alert] = "#{start} to #{finish} is not a valid range"
      params[:ranges].delete(facet)
      params.delete(:ranges) if params[:ranges].empty?
    end
```
I personally prefer to use an `if` whenever possible instead of an `unless`. So you can instead do something like this instead:

```ruby

    start = params.dig(:ranges, facet, :begin)
    finish = params.dig(:ranges, facet, :end)

    next if start.match?(/\d{4}/) && finish.match?(/\d{4}/) && (start.to_i <= finish.to_i)

    flash[:alert] = "#{start} to #{finish} is not a valid range"
    params[:ranges].delete(facet)
    params.delete(:ranges) if params[:ranges].empty?
```

app/views/search/_range_facet_badge.html.erb
murny: Could you maybe reuse the same partial if it is the same? Have an option to pass in an argument for "url" into `_facet_badge`? Maybe defaults to `query_params_without_facet_value(solr_index, raw_value)` etc?

app/views/search/_model_results.html.erb
murny: Could add a `locals` for url, then you can reuse the `_facet_badge` here partial instead of creating a new one for example

app/views/search/_range_facet_results.html.erb
```diff
@@ -0,0 +1,33 @@
...
+          <% next if k == facet.solr_index %>
+          <% v.each do |i| %>
+            <%= hidden_field_tag("#{type}[#{k}][]", i) %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+    <div class='d-flex flex-wrap'>
+      <div class='p-2'>
+        <%= text_field_tag("ranges[#{facet.solr_index}][begin]", facet.range.first, class: 'form-control', size: 4, maxlength: 4, 'aria-label' => 'range_begin') %>
```
murny: Can opt in to use a `number_field_tag` here instead? Uses html5 `<input type="number">` which gives you a bit more control on the client side with what the user inputs (also does things like, when on mobile phone, gives you the number keyboard instead of alphabet keyboard etc)

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number

app/views/search/_range_facet_results.html.erb
```diff
@@ -0,0 +1,33 @@
...
+      </div>
+      <div class='p-2 align-self-center'>
+        to
+      </div>
+      <div class='p-2'>
+        <%= text_field_tag("ranges[#{facet.solr_index}][end]", facet.range.last, class: 'form-control', size: 4, maxlength: 4, 'aria-label' => 'range_end') %>
+      </div>
+    </div>
+    <div class='d-flex justify-content-begin'>
```
murny: `justify-content-begin` isnt a valid option? Probably meant `justify-content-start` instead?

Also this probably shows this isn't needed? Can just render the submit tag without any flexbox stuff?

app/views/search/_range_facet_results.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="card mt-3">
+  <div class="card-header"><%= facet.category_name %></div>
+  <%= form_tag("/search", method: "get", class: 'form-inline') do %>
+    <% search_params_hash.each do |type, h| %>
```
murny: Can't you just use the link for this stuff instead? Instead of destructing all the params and making invisible hidden inputs?

`form_tag search_path(search_params_hash), method: :get, class: 'form-inline" do`

app/models/jupiter_core/deferred_faceted_solr_query.rb
murny: Probably fine? Can make the beginning to be dynamic as well though? Like:

`begin: Time.current.year - 100.years`

app/models/jupiter_core/facet_result.rb
```diff
@@ -15,31 +13,9 @@ def initialize(facet_map, solr_index, values)
...
-  def each_facet_value(range = 0..-1)
-    keys = @values.keys.slice(range)
-    keys.each do |raw_value|
-      count = @values[raw_value]
-      if raw_value.present?
-        yield FacetValue.new(attribute_name, solr_index, raw_value, count)
-      end
-    end
-  end
+  def each_facet_value(_range = 0..-1) end
```
murny: Should this raise an exception or something (requiring child class to fill this out)? or just return an empty array (so doesn't break UI)?

app/models/jupiter_core/deferred_faceted_solr_query.rb
pgwillia: @sfbetz do you have a preference for static (1880) vs dynamic (100 years before the current year)?
screenshot from 2018-03-28 13-07-13

app/controllers/search_controller.rb
pgwillia: The reason to be looping on the params[:range] was hoping to treat range facets generically. Right now there is only range facet (Sort Year) but there could be more, maybe. I think the last occurrence of the error should be okay. How I think this would work is that the user would enter one range at a time, so there would only be one range in error at a time. Ranges in error are deleted from parameters so it would be a bit of work on the users part to have multiple ranges in error.

app/models/jupiter_core/deferred_faceted_solr_query.rb
sfbetz: Static probably makes more sense as the dates of items in ERA aren't going to change dynamically.

app/views/search/_range_facet_results.html.erb
pgwillia: I was running into trouble styling this input when it was number type. I'll give it another go.

app/views/search/_range_facet_results.html.erb
pgwillia: That's way better. Sometime Rails is pretty sweet -- just wish I knew it better.

pgwillia: In ruby when do you use braces () vs when do you not?

murny: Mostly personal preference. Depends who you ask you will get a different answer.

For example rubocop prefer no parentheses: https://github.com/bbatsov/rubocop/blob/master/manual/cops_style.md#styledefwithparentheses

Airbnb styleguide prefers opposite when passing in arguments: https://github.com/airbnb/ruby#method-definitions

app/models/jupiter_core/facet_result.rb
pgwillia: This is probably my Java background showing.  FacetResult would be an [abstract class in Java](https://docs.oracle.com/javase/tutorial/java/IandI/abstract.html).  There's probably a ruby way to do this better.  I'll opt for empty array with a comment.

PR #576, opened 2018-03-23
DOI work

mbarnett:
- [x] Porting over the rest of the DOI test suite is incoming shortly


app/models/concerns/item_properties.rb
```diff
@@ -32,7 +32,25 @@ module ItemProperties
...
+
+    # allow for a uniform way of accessing this information across regular items and theses
+    def authors
+      respond_to?(:creators) ? creators : [dissertant]
+    end
+
+    def doi_state
+      @state ||= begin
+        ItemDoiState.find_by(item_id: id) ||
+          ItemDoiState.create(item_id: id, aasm_state: (doi.present? ? :available : :not_available))
```
murny: Could add some sugar here?

```ruby
ItemDoiState.find_or_create_by(item_id: id) do |i| 
  i.aasm_state = doi.present? ? :available : :not_available
end
```

mbarnett: Mmmm, sugar

test/models/item_doi_state_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class ItemDoiStateTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
```
murny: Throw some tests here? Or just remove this test file and the fixture file? ¯\\_(ツ)_/¯

mbarnett: Yeah, good call. I'll rip out the autogen stuff.

test/services/doi_service_test.rb
```diff
@@ -0,0 +1,105 @@
+require 'test_helper'
+
+class DoiServiceTest < ActiveSupport::TestCase
+
+  EXAMPLE_DOI = 'doi:10.5072/FK2JQ1005X'.freeze
+
+  test 'DOI state transitions' do
+    assert_equal 0, Sidekiq::Worker.jobs.size
```
murny: Alternatively could have used the activejob api: http://edgeapi.rubyonrails.org/classes/ActiveJob/TestHelper.html

But this is fine

test/models/item_doi_state_test.rb
murny: K cool, rest looks good. Can approve after you rip this stuff out

test/services/doi_service_test.rb
mbarnett: Oh nice. TIL.

mbarnett:
Alright, it's switched over to the ActiveJob test interface, which is handy, because now it's easier to assert that each job is what it's expected to be.

test/test_helper.rb
```diff
@@ -2,11 +2,28 @@
...
+VCR.configure do |config|
+  config.cassette_library_dir = 'test/vcr'
+  config.hook_into :webmock
+
+  # Only want VCR to intercept requests to external URLs.
+  config.ignore_localhost = true
+end
+
+# just push all jobs to an array for verification
+Sidekiq::Testing.fake!
```
murny: Might not need all this now?

require 'sidekiq/testing'
Sidekiq::Testing.fake!
Sidekiq::Worker.clear_all

If using activejob stuff?

mbarnett: `Sidekiq::Testing.fake!` would take precedence for any test class that doesn't specifically include `ActiveJob::TestHelper`. With the `Sidekiq::Worker.clear_all` in after_all I'm guessing we want to keep jobs from piling up between runs of different test classes?

mbarnett: It's that or I hoist the   `include ActiveJob::TestHelper` up to ActiveSupport::TestCase, which I could also do if we want to mandate it across the board

murny: I'm okay with moving the Activejob::Testhelper up into ActiveSupport::TestCase? (it does the clearing of job array under the hood: https://github.com/rails/rails/blob/80dc309821732c68a6fb347230c99f2f6abf2f6f/activejob/lib/active_job/test_helper.rb#L35-L51)

Your call, whatever is easier to bite off

mbarnett: No go on hoisting it up to TestCase and using `clear_enqueued_jobs` in `after_all`. It blows up inside of `ActiveJob::QueueAdapters::SidekiqAdapter` for some reason

murny: `clear_enqueued_jobs` is a private method looks like? They do it before each test setup.

But yeah lets leave as is, we can revisit later, etc.

mbarnett: I saw your message after. Retrying it with it handling the cleaning still blows up for some reason. It looks like it doesn't like the include being inherited. Jobs becomes undefined in the DOIServiceTest if it doesn't specifically do the include itself.

Probably not worth sinking more time into right now.

app/models/item_doi_state.rb
```diff
@@ -0,0 +1,58 @@
...
+      transitions from: [:unminted, :awaiting_update], to: :available
+    end
+
+    event :altered, after: :queue_update_job do
+      transitions from: [:available, :not_available], to: :awaiting_update
+    end
+  end
+
+  def item
+    JupiterCore::LockedLdpObject.find(item_id, types: [Item, Thesis])
```
cwant: I know this PR is merged already, but I was snooping around so that I could understand this process better.
I wonder if we should have `@item ||= ...` in this method here? Without it, the `withdraw_doi` method below does two fetches from solr for the same item.

mbarnett: Good call. I'll tweak that in a bit, since I need to touch some related things anyways.

PR #586, opened 2018-03-27
Possible solution for links on homepage?

### LINKED ISSUE #584: Welcome page links
- [ ] Make sure the community and collection links on the front page are pointing to the correct communities and collections

mbarnett:
Open to suggestions on better ways to approach this other than hardcoding IDs from Prod.

Closes #584


mbarnett:
I ran with your idea of wrapping it up in a faux model, so that when the time comes we can swap in a real one fairly seamlessly.

PR #589, opened 2018-03-28
Use created/graduation_date in places we were incorrectly using sort_year

mbarnett:
Also catches a small performance issue where we were generating multiple Solr queries relating to filesets when rendering the Google Scholar metadata

### DIFF
--- a/app/views/items/_item.html.erb
+++ b/app/views/items/_item.html.erb
@@ -26,11 +26,16 @@
     </li>
   <% end %>
 
-  <% if @item.sort_year.present? %>
+  <% if @item.creation_date.present? %>
     <li class="list-unstyled list-group-item-action">
       <dl>
-        <dt><%= t('.sort_year') %></dt>
-        <dd><%= search_link_for(@item, :sort_year) %></dd>
+        <dt><%= t('.created') %></dt>
+        <% if @item.sort_year.present? %>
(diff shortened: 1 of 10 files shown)


app/views/items/_item.html.erb
```diff
@@ -26,11 +26,16 @@
...
     <li class="list-unstyled list-group-item-action">
       <dl>
-        <dt><%= t('.sort_year') %></dt>
-        <dd><%= search_link_for(@item, :sort_year) %></dd>
+        <dt><%= t('.created') %></dt>
+        <% if @item.sort_year.present? %>
+          <%# Display the recorded date created, but link to searching on the facetable sort year if it is present %>
+          <dd><%= search_link_for(@item, :sort_year, display: humanize_date(@item.creation_date)) %></dd>
+        <% else %>
+          <dd><%= humanize_date(@item.creation_date) %></dd>
```
cwant: Probably want to humanize the date in `app/views/items/_thesis.html.erb` too,

mbarnett: Are you referring to https://github.com/ualbertalib/jupiter/pull/589/files#diff-e6ecff2afd01d11cd621bf23c904635bR40 or some other date I'm not seeing there?

cwant: Shoot, yeah -- think I'm going blind.

Issue #568, opened 2018-03-16
Multi-value fields very problematic for users in deposit and edit


sfbetz:
Our elegant single-line multi-value field solution is too hard for users to figure out. I'll be doing more in-depth user testing next week, but this one is already really obvious and definitely needs a fix. Users are trying all of these things:

- types to complete the field and then presses tab to get to the next field (inadvertently clearing the current field)
- types and then carefully clicks into the next field (inadvertently clearing the current field)
- types and realize they made a mistake in their response so try to click into the text to correct (inadvertently clearing the current field) - especially frustrating in long fields like citation when the chances of a typo are high.

The only way to complete the field is to click enter, which makes users nervous because they are concerned that this may accidentally submit their deposit before they are done.

murny:
Just some comments:

First bullet with regards to tabbing, we can enable this so that tab also selects the option that was entered/hovered on. Just requires some configuration on our end (Not enabled by default, which is kinda weird to be honest).

Fixing a mistake? Yeah best option would be to remove and re-add, probably no way around that. (But there is at least indication it will be removed as the item becomes a cancel/remove button (X))

The other issues outlined above are inherited to a normal select box (which this library (and others) tries to emulate). 

- With a normal select box, if you hover/click/type a selection then click out, it doesn't select this option
- The only way to select an item from a select box would be to hit the enter key on the selection or click on the selection


Now with that said, there's definitely issues (bug with placeholders, errors dont get highlighted properly, etc.) with the library we are using (https://select2.org) and it's probably not the best one that could have been chosen (it used to be the best one a few years ago, but seems to not have been maintained very well in the recent years).

So instead of throwing this out entirely we can try a different library like this: http://selectize.github.io/selectize.js/

Which seems more intuitive and more catered to using keyboard only (tab works, keyboard navigation works much better, etc).

Otherwise if we really want to get away from this solution, we can go back to a multiple select/input fields like we have in the mockups.

mbarnett:
It looks like that Selectize tagging demo is pretty much the behaviour we want if it plays well with accessibility tools

cwant:
Only drawback is that selectize hasn't been updated since last July (100 open pull requests, 500 open issues).

sfbetz:
Selectize definitely looks better - I can click in to something I'm writing to correct a mistake without losing everything I've just typed, so fixes concerns one and three. 

If the lack of recent updates makes everyone too nervous, then I'm also fine with the multiple select / input fields as per mockups.

[...]

PR #590, opened 2018-03-28
Remove Select2 for Selectize/Easy-Autocomplete

murny:
https://github.com/ualbertalib/jupiter/issues/568

- Remove usage of select2 for typeahead on browse (/communities), use easy-autocomplete instead (why this library? we needed something that allows returning results into two different groups (communities and collections), as well be able to click on the results which can take you to a different url. Most libraries don't meet these two requirements. For example `awesomplete` doesn't meet either.)

- Remove usage of select2 for a multiple select box and multiple item creation(tagging), use selectize instead


app/assets/javascripts/jupiter/item_draft.js
```diff
@@ -52,18 +52,20 @@ $(document).on('turbolinks:load', function() {
...
-    // a larger width then normal bootstrap inputs
-    // See deposit_item.scss for similar comments.
-    // Potential fix here: https://github.com/select2/select2/pull/4898/
-    width: '100%'
+  // global selectize initailization could be moved elsewhere
+  $('.js-selectize').selectize({
+  });
+
+  // This one is for tagging/ability to create items on input
+  $('.js-selectize-create').selectize({
```
murny: Have to create a different initialisation for the tagging/creation of items as it needs a few more options.

app/assets/stylesheets/item_draft.scss
```diff
@@ -1,14 +0,0 @@
-// TODO: Hacks for select 2 widths being weird when initialized inside a hidden element
```
murny: Placeholders and width issues just work in selectize, no need to do any fancy stuff here

PR #605, opened 2018-04-04
Fix alignment of download buttons

murny:
Was being styled as middle button since we render an empty div

#604

### DIFF
--- a/app/views/application/_item.html.erb
+++ b/app/views/application/_item.html.erb
@@ -15,21 +15,23 @@
         <% end %>
       <% end %>
 
-      <div class="btn-group">
-        <% if policy(item).update? %>
-          <%= link_to edit_item_path(item), class:'btn btn-outline-secondary'  do %>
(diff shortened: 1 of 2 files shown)


app/views/application/_item.html.erb
```diff
@@ -15,21 +15,23 @@
         <% end %>
       <% end %>
 
-      <div class="btn-group">
-        <% if policy(item).update? %>
-          <%= link_to edit_item_path(item), class:'btn btn-outline-secondary'  do %>
-            <%= fa_icon('pencil-square-o') %>
-            <%= t('edit') %>
+      <% if policy(item).update? || policy(item).destroy? %>
```
murny: Surround this with an check for policies, if no buttons are being displayed, don't render an empty `<div class="btn-group></div>", This way, `justify-content-between` will push the download button all the way to the right edge, instead of somewhere in the middle which resulted in the weird styling

Issue #580, opened 2018-03-26
Last step of deposit needs some progress/action indicator


sfbetz:
During deposit, there can be some wait time one the last step while file is processing. It's long enough to look to a user like the page is frozen / broken, especially if they don't see the spinning wheel on the browser tab. We need some way to indicate that there's something happening behind the scenes (message, animation, other?).

pgwillia:
I think there are a few approaches to the problem:
1. Redirect the user to another page quickly with a message that their request is being processed.
 * make the review_and_deposit_item an [ActiveJob ](http://guides.rubyonrails.org/active_job_basics.html) to run in the background
 * no javascript involved
 * currently the wait is for the item that the user is creating to be created and stored to Fedora so that it's page can be displayed
2. Wait for the page but [display something](http://www.codemag.com/article/1507051/Display-a-Progress-Message-on-an-MVC-Page)
  * Disable all the buttons on the screen so they can’t click anything else.
  * Hide all of the buttons on the screen so they can’t click any of them.
  * Change the text on the button they just clicked on.
  * Disable the button they just clicked on.
  * Pop-up a message over the screen.
  * Gray out the page they’re on.
  * Any combination of the above items.
  * is javascript so needs to fail intelligently and all user's won't benefit
3. Use [Turbolinks suggestion](https://github.com/turbolinks/turbolinks#redirecting-after-a-form-submission) to use [XHR submission](http://guides.rubyonrails.org/working_with_javascript_in_rails.html#remote-elements)
 * see also [here](https://github.com/turbolinks/turbolinks/issues/85#issuecomment-338784510)
 * currently uses [simple_form_for which seems to be incompatible with form_with](https://github.com/plataformatec/simple_form/issues/1494) that Turbolinks likes => might be significant rewrite of draft-submission
 * is javascript so needs to fail intelligently and all user's won't benefit

In related news, I learned that [turbolinks ](https://github.com/turbolinks/turbolinks#displaying-progress) provides [a progress bar](http://www.usabilitypost.com/2013/08/19/new-ui-pattern-website-loading-bars/) by default for any page that takes longer than 500ms (.5 seconds) to load. We can style it and also choose the minimum interval before it's shown. Unfortunately Turbolinks doesn't apply to this case (HTTP POST).

murny:
I say you can do something like #2. 

Basically when you submit the form, the button text changes to say 
"<Loading Symbol> Saving Draft..." and the button is disabled.

[...]

PR #617, opened 2018-04-05
Last step of deposit needs some progress/action indicator

pgwillia:
#580

### DIFF
--- a/app/views/items/draft/review_and_deposit_item.html.erb
+++ b/app/views/items/draft/review_and_deposit_item.html.erb
@@ -221,7 +221,7 @@
             <%= fa_icon 'chevron-left' %>
             <%= t('items.draft.previous') %>
           <% end %>
-          <%= button_tag type: :submit, class: 'btn btn-primary' do %>
+          <%= button_tag type: :submit, class: 'btn btn-primary', data: { disable_with: t('items.draft.submit_draft_btn_loading_html') } do %>
             <%= fa_icon 'save' %>
             <% if @is_edit %>
               <%= t('items.draft.save_and_deposit_edits') %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -441,6 +441,7 @@ en:
       please_fix_errors: 'Please correct all errors below and try again'
       please_follow_the_steps: 'Looks like you tried to skip ahead. Please follow the steps accordingly'
       files_are_required_to_continue: 'You are required at least one file attached before continuing'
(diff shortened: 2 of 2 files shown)


app/views/items/draft/review_and_deposit_item.html.erb
```diff
@@ -221,7 +221,7 @@
             <%= fa_icon 'chevron-left' %>
             <%= t('items.draft.previous') %>
           <% end %>
-          <%= button_tag type: :submit, class: 'btn btn-primary' do %>
+          <%= button_tag type: :submit, class: 'btn btn-primary', data: { disable_with: "<i class='fa fa-spinner fa-pulse fa-1x fa-fw'></i>Saving Draft..." } do %>
```
murny: As a bonus you can i18n this if you want.

en.yml
`submit_draft_btn_loading_html:  "<i class='fa fa-spinner fa-pulse fa-1x fa-fw'></i>Saving Draft..."`

Then here you can just do

`data: { disable_with: t('.submit_draft_btn_loading_html') }`

pgwillia: I have mixed feelings about putting html in en.yml. The <i> tag doesn't need translation or localization. On the other hand it does make the html.erb easier to look at. ¯\_(ツ)_/¯

murny: Agreed, if you can help it, its better to keep html out of en.yml. 

But here your hands are kinda tied since your rendering html as a string to `disable_with` and the "Saving Draft..." part should be i18n'd since if we are localising in french this needs to change. Alternatively I suppose you can concat/string interpolate the i18n like so:

`data: { disable_with: "<i class='fa fa-spinner fa-pulse fa-1x fa-fw'></i>#{t('.submit_draft_btn_loading')}" }`

But not sure if that's worth the extra effort. I18n does support safe rendering of html using the `_html` which is what your currently doing. Suppose up to personal preference.

config/locales/en.yml
```diff
@@ -441,6 +441,7 @@ en:
       please_fix_errors: 'Please correct all errors below and try again'
       please_follow_the_steps: 'Looks like you tried to skip ahead. Please follow the steps accordingly'
       files_are_required_to_continue: 'You are required at least one file attached before continuing'
+      submit_draft_btn_loading_html: "<i class='fa fa-spinner fa-pulse fa-lg fa-fw'></i>Saving..."
```
pgwillia: Sonya wants this to say 'Saving...' for the general case.

PR #618, opened 2018-04-05
Try instrumenting Solr queries for better profiling

mbarnett:
If I've squinted at the Skylight documentation and an implementation I found in the netflix/fast_jsonapi repo correctly, this should give us some insight into what queries are being run against solr, in Skylight.


config/initializers/solr_normalizer.rb
```diff
@@ -0,0 +1,14 @@
+require 'skylight'
+
+SOLR_NORMALIZER = 'solr.query'.freeze
+JUPITER_SOLR_NOTIFIFCATION = "jupiter.#{SOLR_NORMALIZER}".freeze
```
pbinkley: typo: notifi*f*cation

config/initializers/solr_normalizer.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Skylight::Normalizers::SolrNormalizer < Skylight::Normalizers::Normalizer
+
+  register JUPITER_SOLR_NOTIFIFCATION
```
pbinkley: or was that intentional?

test/models/jupiter_core/search_test.rb
```diff
@@ -76,7 +76,20 @@ def unlocked_method_dont_let_locked_methods_mutate(attempted_title)
       end
     end
 
-    search_results = JupiterCore::Search.faceted_search(models: @@klass, q: creator2)
+    # ensure searches are sending notifications
+    events = []
+    ActiveSupport::Notifications.subscribe(JUPITER_SOLR_NOTIFIFCATION) do |*args|
```
pbinkley: last one

config/initializers/solr_normalizer.rb
mbarnett: No no that's a totally real word I DEFINITELY didn't typo. It means a notification that MIGHT happen IF something happens.

PR #619, opened 2018-04-05
Frontend checklist (part 3)

murny:
https://github.com/ualbertalib/jupiter/issues/562

### HEAD
- [x] Facebook OG and Twitter Cards
- [x] meta description

### ACCESSIBILITY
- [x] All pages have an H1 and right order for sub headings (H1 to H6)


app/helpers/application_helper.rb
```diff
@@ -3,16 +3,59 @@ module ApplicationHelper
...
+    image = @community&.logo || @item&.thumbnail
+
+    # TODO: ActiveStorage 5.2 upgrade, fix this hack
+    # needs to be a full url so can use rails_blob_url or url_for in 5.2 instead
+    image_url = request.base_url + image.url if image.present?
+    image_url || default_url
+  end
+
+  def page_type(type = nil)
+    type || 'website'
```
murny: This mostly for future use for toggling Facebook OG type field, may want to use article/etc here which allows us to mark up our items (author array, tags, published time, etc)

app/views/items/_item.html.erb
```diff
@@ -1,4 +1,12 @@
 <%# TODO: Reduce duplication between this partial and thesis, pretty much exactly the same %>
+
+<% content_for :extra_social_meta do %>
```
murny: Pull over the old social metadata we had on Hydranorth. We now have the same as old hydranorth

app/controllers/search_controller.rb
```diff
@@ -6,6 +6,8 @@ class SearchController < ApplicationController
   skip_after_action :verify_authorized
 
   def index
+    # TODO: Mass Assignment: Parameters should be whitelisted for mass assignment (Brakeman scan)
```
murny: Brakeman scan flags this as a medium issue. We need to use strong parameters better here. Not sure best way to refactor this, so just left a TODO for time being.

Also we got lots of nil checks on these as well? Why not just assign them to an empty hash like `params[:ranges] ||= {}` and handle accordingly instead of constantly checking if they are empty/nil/or using `.&`?

mbarnett: Ah ok, I think I can fix this on my end

app/assets/stylesheets/communities.scss
```diff
@@ -1,10 +1,4 @@
 // remove the underlines from everything in the media object since its a link
 .media.jupiter-community-media {
   text-decoration: none;
-
-  // hover darkens the h4, lets not do this
-  // since theres a tiny bit of padding within the li which gives a weird effect
```
murny: Doesn't seem to be an issue now (maybe because i moved the buttons out of the media div)? Color being darkened isn't a bad thing either. Leaving this out.

app/views/application/_item.html.erb
```diff
@@ -5,7 +5,7 @@
   <% end %>
   <div class="media-body ml-3">
     <div class="d-flex flex-wrap flex-lg-nowrap align-items-start">
-      <h5 class="mt-0 mr-auto"><%= link_to item.title, path_for_result(item) %></h5>
+      <h3 class="mt-0 h5 mr-auto"><%= link_to item.title, path_for_result(item) %></h3>
```
murny: This actually doesn't look that bad as an H3, but to keep consistent, ill leave this styled as an h5

app/helpers/search_helper.rb
```diff
@@ -53,7 +53,7 @@ def results_model_tab_link(model)
     # Create bootstrap nav-item, make it a link if there are results for the model
     classes = 'nav-link'
     count = @results[model].total_count
-    text = t("search.tab_header_#{model.to_s.pluralize}_with_count", count: count)
+    text = content_tag(:h2, t("search.tab_header_#{model.to_s.pluralize}_with_count", count: count), class: 'h5')
```
murny: Since we are using item/thesis partials else where and the titles are an h3, we need an h2 on this page and this makes the most logical sense to make the tabs an h2. But since this will be styled as huge text, lets style as h5

app/views/collections/show.html.erb
```diff
@@ -36,9 +34,9 @@
   <% end %>
 
   <div class="d-flex flex-wrap align-items-center justify-content-between mt-3">
-    <h4 class="mr-auto">
+    <h2 class="mr-auto">
```
murny: Whats all these header changes? Screenreaders/accessibility wants the headers to flow properly, from an H1 to an H6. We have been skipping headers going from an H1 to an H4 for example here. Instead of doing this, use the proper header and style the header to get the desired effect (bootstrap makes this easy with the header utility classes, for example making an `h2` look like an `h4` is as simple as `<h2 class="h4">`

app/views/search/index.html.erb
```diff
@@ -2,7 +2,10 @@
 <% content_for :head do %>
   <meta name="robots" content="noindex" />
 <% end %>
+
 <div class="container mt-3">
+  <h1><%= t('.header') %></h1>
```
murny: Let's add a header/h1 to search page? Just says "Search"

config/locales/en.yml
```diff
@@ -590,7 +590,7 @@ en:
     range_limit: 'Limit'
 
     index:
-      header: 'Search Items'
```
murny: Not exactly just items? As we are also showing community/collections. Lets label correctly?

app/views/items/_item.html.erb
pgwillia: Could make this a partial and add rights to https://github.com/ualbertalib/jupiter/blob/master/app/models/concerns/item_properties.rb#L37-L50

pgwillia: Should we treat the Google Scholar metadata as extra_social_meta?

app/views/welcome/index.html.erb
```diff
@@ -1,7 +1,3 @@
-<% content_for :head do %>
-  <meta name="description" content="<%= t('.welcome_lead') %>">
```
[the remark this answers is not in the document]
pgwillia: This is [Google's advice](https://support.google.com/webmasters/answer/7451184?hl=en) 
> Use unique descriptions for each page
> 
> Having a different description meta tag for each page helps both users and Google, especially in searches where users may bring up multiple pages on your domain (for example, searches using the site: operator). If your site has thousands or even millions of pages, hand-crafting description meta tags probably isn't feasible. In this case, you could automatically generate description meta tags based on each page's content.
> 
> Avoid:
> 
>     Using a single description meta tag across all of your site's pages or a large group of pages.
> 

[My conclusion](https://github.com/ualbertalib/jupiter/blob/master/test/integration/site_for_bots_test.rb#L66) was that search_path is noindex so description shouldn't matter

I guess using this description for /search doesn't really matter to me, but what are the etc?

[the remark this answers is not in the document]
murny: We can put more thought in this for sure. 

Pages that  currently get generated descriptions:
- Home page (default)
- Communtiy index and show pages (index gets a crafted one, show page uses community description)
- Collection show pages (uses collection description)
- Item show pages (uses item description)

Pages that don't:
- Admin pages (but seo cant reach here, so using the site wide default stuff is fine)
- Item deposit (but seo cant reach here, so using the site wide  default stuff is fine)
- Profile pages (but seo cant reach here, so using the site wide default stuff is fine)
- Search (seo doesn't index this, but people can share this and we need a description for social cards?)
- All static pages (contact us/about/polices) (probably doesnt hurt to have one?)

So I guess really just search and static pages? We can probably create something simple for this

¯\\_(ツ)_/¯

app/views/items/_item.html.erb
murny: Yeah good suggestion to remove duplication. Can move to that stuff to item properties. 

Mmm we could? I suppose we could generalise the content_for and call it metadata or something and can stick both in the same place

app/views/welcome/index.html.erb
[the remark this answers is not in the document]
pgwillia: I agree, your PR gives a good default. The static pages should have different descriptions. I'll add a ticket to get copy for a meta description for the static pages if this rates on anyones priority list.

PR #632, opened 2018-04-06
Specify a host and protocol manually for target URL generation 

mbarnett:
Necessary so that DOI background jobs can run properly under Resque.

### DIFF
--- a/app/services/doi_service.rb
+++ b/app/services/doi_service.rb
@@ -80,13 +80,17 @@ class DOIService
   private
 
   def ezid_metadata
+    # deal with the fact that there's no request object when running under resque
+    url_opts = { id: @item.id,
+                 host: Rails.env.production? ? 'era.library.ualberta.ca' : 'localhost',
+                 protocol: 'https' }
     {
       datacite_creator:  @item.authors.join('; '),
       datacite_publisher: PUBLISHER,
       datacite_publicationyear: @item.sort_year.present? ? @item.sort_year : '(:unav)',
       datacite_resourcetype: DATACITE_METADATA_SCHEME[@item.item_type_with_status_code],
       datacite_title:  @item.title,
-      target: Rails.application.routes.url_helpers.item_url(id: @item.id),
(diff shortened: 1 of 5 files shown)


app/services/doi_service.rb
```diff
@@ -80,13 +80,17 @@ def self.remove(doi)
   private
 
   def ezid_metadata
+    # deal with the fact that there's no request object when running under resque
```
murny: should say sidekiq not resque :trollface:

app/services/doi_service.rb
```diff
@@ -80,13 +80,17 @@ def self.remove(doi)
...
+    url_opts = { id: @item.id,
```
murny: Alternatively you can set this jupiter wide from config files (production.rb etc.). Something like:
```
# needed when using Rails.application.routes.url_helpers from a model or worker (see http://stackoverflow.com/a/5456103/545137)
Rails.application.routes.default_url_options = { host: 'era.library.ualberta.ca', port: 443, protocol: 'https' }```

Then you can use url_helpers in models/workers/etc

Comment (mbarnett) [reply app/services/doi_service.rb]: Yeah, I guess there's no reason not to do that, as long as it's loaded by Sidekiq when the job runs

Comment (murny) [reply app/services/doi_service.rb]: Should be loaded by sidekiq as they run the same code, done it before on a past project by just plopping it in production.rb etc. 

Initializer also works for example this is what gitlabhq does: https://github.com/gitlabhq/gitlabhq/blob/master/config/initializers/default_url_options.rb

File: config/environments/production.rb [hunk 3]
```diff
@@ -1,3 +1,5 @@
+Rails.application.routes.default_url_options = { host: 'era.library.ualberta.ca', port: 443, protocol: 'https' }
```
murny: Should add to the other environments too?

test/dev can be like:
Rails.application.routes.default_url_options = { host: 'localhost', port: 3000 }

mbarnett: Specifying a port actually breaks system tests, but I can specify the host in the other environments at least.

Issue #468, opened 2018-02-09
Admin needs to be able to delete items


murny:
We currently show a "Delete" button on each item on the list views (search/users profile/collection index/etc). However currently this isn't wired up.

- Need to add a delete action to Admin::ItemsController and handle the deleting of items appropriately 

Duplicate issue here #579 from sfbetz:

> Trying to delete test items leads to a 404 error.
> 
> Logged in to my sfbetz user account - trying to delete this item:
> https://era-beta.library.ualberta.ca/items/06ec0895-075d-4098-80c9-ebb1795db937
> 
> Delete fails whether I try it from delete button on a search results page like this: https://era-beta.library.ualberta.ca/search?search=lorem+ipsum
> 
> Or from user account page like this: https://era-beta.library.ualberta.ca/admin/users/877
> 
> Error message:
> 
> This era-beta.library.ualberta.ca page can’t be found
> No webpage was found for the web address: https://era-beta.library.ualberta.ca/items/06ec0895-075d-4098-80c9-ebb1795db937
> HTTP ERROR 404
> 
> I am able to successfully delete "draft" files with no issues.

mbarnett:
@sfbetz should this include Theses, or strictly Items?

mbarnett:
When you call destroy on the item, it should kick off a job to tombstone the DOI automatically, so that part is done.

You'll want to delete the activestorage blobs and and doi_state records _after_ the object is deleted. Normally there would be a dependent delete on the join table, but because the object lives in Fedora and the dependent records live in Postgres, you'll need to do it manually in an after_destroy hook.

weiweishi:
@weiweishi will verify if DOI is actually tombstoned. and if that's case this ticket can be closed

PR #635, opened 2018-04-06
Admin needs to be able to delete items

pgwillia:
#468
- [x] tests!
- [x] ~~resource sync~~
- [x] error handling


config/routes.rb
```diff
@@ -24,6 +24,8 @@
     end
   end
 
+  resources :items, module: :admin, only: :destroy
```
pgwillia: Redirects `DELETE item/:id` to Admin::ItemsController

app/models/concerns/item_properties.rb
```diff
@@ -53,6 +53,7 @@ def doi_state
       before_save :handle_doi_states
       after_create :handle_doi_states
       before_destroy :remove_doi
+      after_destroy :delete_blobs, :delete_doi_state
```
pgwillia: You'll want to delete the activestorage blobs and and doi_state records after the object is deleted. Normally there would be a dependent delete on the join table, but because the object lives in Fedora and the dependent records live in Postgres, you'll need to do it manually in an after_destroy hook.

app/controllers/admin/items_controller.rb
```diff
@@ -6,4 +6,12 @@ def index
     item_search_setup
   end
 
+  def destroy
+    @item = JupiterCore::LockedLdpObject.find(params[:id], types: [Item, Thesis])
+    authorize @item
+    @item.unlock_and_fetch_ldp_object(&:destroy!)
```
pgwillia: Buh-bye LDP Object (Fedora and Solr)

config/routes.rb
mbarnett: Small note, but I think you just need to add delete here https://github.com/pgwillia/jupiter/blob/a53213c5f2e26a039f477da7d4baa66fc5841bd4/config/routes.rb#L53

So that it becomes `resources :items, only: [:index, :delete]`

pgwillia: Small note, but do we want it to be? 
`DELETE item/:id` to Admin::ItemsController or
`DELETE admin/item/:id` to Admin::ItemsController

mbarnett: The former is strictly speaking more RESTful, but the latter has the advantage of passing through the AdminContstraint

I don't feel so strongly about the routes that I mind if admin/ appears in them, honestly.

test/integration/collection_show_test.rb
```diff
@@ -60,14 +60,16 @@ def before_all
...
+      assert_equal item_links.count, 2
       # Thumbnail link to item
       assert_includes item_links.first.inner_html, 'img-thumbnail'
       # Text link to item
       assert_equal item_links[1].text, item.title
+
       # Link to delete item
-      assert_match 'Delete', item_links.last.text
-      assert_equal item_links.last.attributes['data-method'].to_s, 'delete'
+      delete_link = css_select ".jupiter-results ul.list-group .list-group-item a[href='#{admin_item_path(item)}']"
```
pgwillia: The path to delete changed from what was expected to facilitate route passing through the AdminContstraint

test/controllers/admin/items_controller_test.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+      uo.save!
+    end
+  end
+
+  context '#destroy' do
+    should 'should destroy item if admin user' do
+      sign_in_as(users(:admin))
+
+      assert_difference(['Item.count', 'ItemDoiState.count'], -1) do
```
pgwillia: I couldn't figure out what difference to measure for ActiveStorage::Attached::One (the thumnail) being purged.

murny: ActiveStorage has two tables, ActiveStorage::Blob (table that holds the actual blob and metadata of the attachment) and ActiveStorage::Attachment (join table between model and activestorage blob). 

So either one you can check to see if they decreased (probably blob is what you want though).

app/controllers/admin/items_controller.rb
murny: Considering this is activefedora and weird things happen...

Should we handle the fail case here? What if delete fails? 

(Assuming activefedora works like activerecord...which is a big assumption) `destroy!` (bang version) will raise an exception (in activerecord it raises ActiveRecord::RecordNotDestroyed). Should we be using the non bang version? `destroy` and just let it fail quietly?

Or should we handle this gracefully and return a decent error to the user?
```ruby
if  @item.unlock_and_fetch_ldp_object(&:destroy)
  redirect_back(fallback_location: root_path, notice: t('.successful_deletion'))
else
 redirect_back(fallback_location: root_path, notice: t('.failed_deletion'))
end
```

app/controllers/admin/items_controller.rb
```diff
@@ -6,4 +6,12 @@ def index
     item_search_setup
   end
 
+  def destroy
+    @item = JupiterCore::LockedLdpObject.find(params[:id], types: [Item, Thesis])
+    authorize @item
```
pgwillia: Maybe this isn't required in the Admin::ItemsController?

murny: Not required no, as everything is already authorized from the inheritance (admincontroller) which authorizes if your an admin.

app/controllers/admin/items_controller.rb
pgwillia: If we can't trust ActiveFedora to work like ActiveRecord then maybe this would be better?
```
begin
      @item.unlock_and_fetch_ldp_object(&:destroy!)
      flash[:notice] = t('.deleted')
    rescue StandardError => e
      flash[:alert] = t('.failed')
      Rollbar.error("Error deleting #{@item.id}", e)
    end
```
That would catch any errors from the unlock_and_fetch_ldp_object step (like LDP:Gone) and anything else that Fedora/ActiveFedora might come up with. I was looking for some documentation about destroy vs destroy! and found that even the non bang version [doesn't always work as expected](https://github.com/rails/rails/issues/18056).

pgwillia:
travis-ci?!  Those tests with that seed Work on My Machine™

PR #660, opened 2018-04-10
Fix showing all results instead of limiting to 4 and fix responsiveness

murny:
From slack, issue was found where we were returning 5 json results but only 4 of them were showing. Was result of a configuration on categories in the jquery library we were using. This fixes this.


app/assets/javascripts/jupiter/communities_collections_searchbox.js
```diff
@@ -1,30 +1,36 @@
...
-      // we use a relative url "communities.json" here as this will be used for both /admin and non admin (/) paths
-      return "communities.json?query=" + phrase;
-    },
-    categories: [
-      {
-        listLocation: "communities",
-        header: "<strong>Communities</strong>",
+  if($input.length > 0) {
+    var options = {
+      adjustWidth: false,
```
murny: https://github.com/pawelczak/EasyAutocomplete/issues/154

app/assets/javascripts/jupiter/communities_collections_searchbox.js
```diff
@@ -1,30 +1,36 @@
...
+      url: function(phrase) {
+        // we use a relative url "communities.json" here as this will be used for both /admin and non admin (/) paths
+        return "communities.json?query=" + phrase;
       },
-      {
-        listLocation: "collections",
-        header: "<strong>Collections</strong>",
+      categories: [
+        {
+          maxNumberOfElements:5,
```
murny: https://github.com/pawelczak/EasyAutocomplete/blob/5ea2e794bb7d3766ae4fcc039fa590f16badbe76/src/configuration.js#L113

app/assets/javascripts/jupiter/communities_collections_searchbox.js
```diff
@@ -1,30 +1,36 @@
...
-    getValue: "name",
-    url: function(phrase) {
-      // we use a relative url "communities.json" here as this will be used for both /admin and non admin (/) paths
-      return "communities.json?query=" + phrase;
-    },
-    categories: [
-      {
-        listLocation: "communities",
-        header: "<strong>Communities</strong>",
+  if($input.length > 0) {
```
murny: Check to see if input is even on the page before we start configuring stuff

app/assets/javascripts/jupiter/communities_collections_searchbox.js
cwant: I suspect this might be too much of a bother to be worth it, but we now have two hardcoded fives in the code. If it possible to configure both in the same place? (e.g., some sort of `data-max-categories` attribute on the input, who's value is set in the controller where the `limit(5)` is.)

murny: Good suggestion, can add :+1:

PR #665, opened 2018-04-11
Harden search controller against bots playing with params

### LINKED ISSUE #650: Remove counts from unselected tabs on /search
The counts are confusing in that selected facets only apply to the selected tab, and the counts on the unselected tabs never vary (they are just counts of how many items in total exist).

Also remove useless single checkbox facet ("Public") on the Community tab, since selecting it does not reduce results at all.

### LINKED ISSUE #656: Bot Hardening
Malicious bots are running around stuffing various form and query params with garbage to try to probe for vulnerabilities. 

None of this is harmful, but it's generating 500s on the bot requests that are cluttering up Rollbar, so it _is_ annoying, and it's highlighting various places where we could be doing more to sanitize inputs so that things behave gracefully even when infinite internet monkeys are playing with the site.

mbarnett:
Also resolves issues brakeman found by better handling parameter permissions, fixes several bugs with range faceting that caused a loss of existing search queries and applied facets, improves validation of range years by restricting them the regex further, and removes counts from inactive search tabs for better UX and performance.

Closes #650. Closes #656.


app/controllers/search_controller.rb
```diff
@@ -1,72 +1,79 @@
...
+    tab ||= 'item'
+    @models = case tab
+              when 'item'
+                [Item, Thesis]
+              when 'collection'
+                [Collection]
+              when 'community'
+                [Community]
+              else
+                return redirect_to search_path(tab: :item)
```
murny: Just curious if there is a specific reason to correct this via a redirect? Instead of doing the following:

Why not just:
```ruby
tab = params[:tab]
@models = if tab == 'collection'
            [Collection]
          elsif tab == 'community'
            [Community]
          else # its assumed to be item
            [Item, Thesis]
          end
```

mbarnett: It just seemed better to canonicalize it, instead of someone somewhere linking to /search?tab=foobar or w/e and then having to support that going forward ¯\_(ツ)_/¯

app/helpers/search_helper.rb
```diff
@@ -52,16 +52,15 @@ def query_params_with_sort(sort, direction = 'asc')
   def results_model_tab_link(model)
```
murny: Guess question for the future. Did you see anything preventing this from being a view partial? Instead of a helper spitting out html? We got a few of these that should be refactored down the road

mbarnett: Yeah, I might just pull this into the view layer when I fix the tests. It's overly complicated for what it does

app/views/search/_range_facet_results.html.erb
```diff
@@ -1,11 +1,21 @@
 <div class="card mt-3">
   <div class="card-header"><%= facet.category_name %></div>
-  <%= form_tag search_path(search_params_hash), method: "get", class: 'form-group' do %>
```
pgwillia: What does using the hidden_field_tag gain us? (PR review suggested I change to use the path helper rather than invisible hidden input fields. Wondering why the flip-flop?)

mbarnett: Mostly, the PR review was wrong.

It wasn't working at all. The reason is that this form uses GET, and browsers don't _merge_ query params in the form's target URL with query params that come from the form – the form params just clobber whatever was there. Hence, right now, if you go to https://era.library.ualberta.ca/search?facets%5Ball_subjects_sim%5D%5B%5D=Burge&search=John and hit limit in the Sort Year facet, you completely lose your existing search term and facets. 

So you had it right the first time. Everything that we want to preserve has to be a part of the form.

PR #672, opened 2018-04-12
Further search cleanup

mbarnett:
Simplify partials, eliminate distinction between controllers that used :search for the search param and those that used :query, unify areas where we were doing the same thing in different ways, eliminate unnecessary locals, hide facet sidebar on Community search (where no facetable values are present unless you're an Admin)


app/helpers/search_helper.rb
```diff
@@ -1,7 +1,14 @@
 module SearchHelper
   def search_params_hash
-    # Search controller uses param[:search], all others use params[:query]
```
mbarnett: Rather than taking the time to document this difference, we probably should have eliminated it from the get-go...

app/views/application/_facet_result.html.erb
```diff
@@ -3,15 +3,20 @@
 
   <ul class="list-group list-group-flush">
     <% facet_result.each_facet_value(0..ItemSearch::MAX_FACETS - 1) do |facet_value| %>
-      <%= render partial: 'search/facet_value', locals: { value: present(facet_value)} %>
+      <li class="list-group-item">
+        <%= present(facet_value).facet_search_link %>
```
murny: Similar to another comment: https://github.com/ualbertalib/jupiter/pull/665#discussion_r180916043

is this presenter actually do alot? It's just doing an if check if facet is active then rendering the corresponding html (with checkbox or without checkbox, etc)

Maybe we can move this to the view as well (or partial since its being called twice here)

mbarnett: Yeah, I don't think this belongs in the presenter at all. I'll move it in another PR

PR #686, opened 2018-04-18
Allow for deletion of items in a DOI-safe manner

mbarnett:


app/models/concerns/item_properties.rb
```diff
@@ -135,6 +133,13 @@ def delete_doi_state
         doi_state.destroy!
       end
 
+      # for use when deleting items for later re-migration, to avoid tombstoning
+      # manually updates the underlying aasm_state to preclude running the Withdrawl job
+      def doi_safe_destroy!
+        doi_state.update_attribute(:aasm_state, 'excluded')
```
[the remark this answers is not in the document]
mbarnett: ... that's exactly why I'm using `update_attribute` 😛

pgwillia:
So the idea is that the migration scripts will use `item.doi_safe_destroy!`?

mbarnett:
Correct

app/models/concerns/item_properties.rb
```diff
@@ -135,6 +133,15 @@ def delete_doi_state
         doi_state.destroy!
       end
 
+      # for use when deleting items for later re-migration, to avoid tombstoning
+      # manually updates the underlying aasm_state to preclude running the Withdrawl job
+      # rubocop:disable Rails/SkipsModelValidations
+      def doi_safe_destroy!
+        doi_state.update_attribute(:aasm_state, 'excluded')
```
pgwillia: And this puts the doi_state directly to 'excluded' rather than going through the events?

mbarnett: Yeah, this writes directly to the column in Postgres, without going through any other mechanisms. `excluded` would normally be the state it ends up in _after_ withdrawing the DOI via EZID, but by moving it directly to that state without tripping the [declared transition method](https://github.com/mbarnett/jupiter/blob/1c3aedda85b4549d1644f4a71cc2bb74aa3b1098/app/models/item_doi_state.rb#L19), we ensure that that job is never run per [this method](https://github.com/mbarnett/jupiter/blob/1c3aedda85b4549d1644f4a71cc2bb74aa3b1098/app/models/concerns/item_properties.rb#L125)

PR #689, opened 2018-04-19
Refactor tests

pgwillia:
- [x] remove `context`
- [ ] change all `before_all` to just ruby (or `setup` or make part of the test, as appropriate)
- [x] change all `should` to `test`
- [ ] create fixtures for reused setup
- [x] use [shoulda-matchers](https://robots.thoughtbot.com/shoulda-matchers-3-0)


test/test_helper.rb
```diff
@@ -19,6 +19,14 @@
   config.ignore_localhost = true
 end
 
+Shoulda::Matchers.configure do |config|
+  config.integrate do |with|
+    with.test_framework :minitest
+    with.library :active_record
+    with.library :active_model
```
pgwillia: These were the options for configuration.  Could use `with.library :rails` instead.
```
# Choose one or more libraries:
with.library :active_record
with.library :active_model
with.library :action_controller
# Or, choose the following (which implies all of the above):
with.library :rails
```

mbarnett:
just a note re: 

> change all before_all to just ruby (or setup or make part of the test, as appropriate)

`setup` can't/shouldn't replace `before_all` because setup runs before every test block, whereas `before_all` only runs once, before all the tests in a class.

The minitest docs suggest `SETUP = begin ... end` to get the equivalent functionality of `before_all`

PR #690, opened 2018-04-20
Ensure file sets get purged when an item is deleted

mbarnett:
Stemming from zombie file cleanup earlier today

### DIFF
--- a/app/models/concerns/item_properties.rb
+++ b/app/models/concerns/item_properties.rb
@@ -163,7 +163,7 @@ module ItemProperties
           Rollbar.warn("Couldn't purge fileset #{fs.id} with item id #{id}") unless success
         end
 
-        self.ordered_members = [] unless self.destroyed?
+        self.ordered_members = [] unless destroyed?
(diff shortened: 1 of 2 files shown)


app/models/concerns/item_properties.rb
```diff
@@ -53,7 +53,7 @@ def doi_state
       before_save :handle_doi_states
       after_create :handle_doi_states
       before_destroy :remove_doi
-      after_destroy :purge_thumbnail, :delete_doi_state
+      after_destroy :purge_thumbnail, :delete_doi_state, :purge_files
```
pgwillia: Could add to this test https://github.com/ualbertalib/jupiter/blob/master/test/controllers/admin/items_controller_test.rb#L39

murny: Looks good. Agreed, try and add some test for purge_files/purge_drafts and ill approve

weiweishi:
Does anything block this PR from being merged?

PR #695, opened 2018-04-23
Add task to recover Solr index from Fedora

mbarnett:
Also fixes a bug in Collection's community title indexing.


test/models/collection_test.rb
```diff
@@ -30,8 +30,8 @@ class CollectionTest < ActiveSupport::TestCase
 
   test 'community must exist' do
     community_id = generate_random_string
-    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular).id,
-                                                  community_id: community_id)
+    collection = Collection.new_locked_ldp_object(title: 'foo', owner: users(:regular).id)
+    collection.unlock_and_fetch_ldp_object { |uo| uo.community_id = community_id }
```
pgwillia: What's the difference here?  If `community_id` can only be set on an unlocked item there are a bunch of tests that need this change.

mbarnett: The difference is that, if you deliberately set community_id to a value that will break this: https://github.com/ualbertalib/jupiter/blob/master/app/models/collection.rb#L19, then the initializer can never complete because it needs to be able to generate a solr document to base the wrapping object on. 

This test is kind of tricky in that it's trying to test the validation of something that also makes the object unconstructable because of that `additonal_solr_index` call. You can paper over it by using `&.`, but then you're back in the case where it's never actually valid to not have a community there, it's just that you're warping your application code to make the test's life easier (and hiding bugs in the process).

It's a corner case that only impacts this test – and the _real_ solution is to get the community name into Fedora on the collection record. These cross-record solr dependencies are a minefield and there's no really good way to avoid the fragility they introduce (and if we can move away from Fedora soon enough, this stops being an issue anyways).

pgwillia: Ah.  I see now.  Thanks for the explanation.

lib/tasks/recover.rake
```diff
@@ -0,0 +1,69 @@
...
+    # the AF IR{Classes} are instantiated when these classes load, so they need to be loaded prior to any call to
+    # ActiveFedora::Base#find
+    RECOVERED_CLASSES = [Thesis, Item, Collection, Community, FileSet].freeze
+
+    # Unfortunately, an inter-object dependency crept into the way we're using +additional_search_index+ in Collection.
+    # Collections attempt to retrieve a Community to index its title, without saving that information to Fedora
+    # – this creates an ordering dependency between Collections and Communities such that the Community needs to be
+    # indexed prior to indexing the Collection.
+    #
+    # We got lucky in that there are not expected to be a large number of Collections in this phase of Jupiter
```
pgwillia: I'm not sure if this part of the comment adds anything to the code. Any future abuse of `additional_search_index` probably won't see this warning. Take it out or move it closer to the problem?

lib/tasks/recover.rake
```diff
@@ -0,0 +1,69 @@
...
+      puts
+      print "#{idx}/#{count}"
+    else
+      print '.'
+    end
+  end
+
+  def add_to_solr(ir_object)
+    klass = ir_object.class.owning_class
+    object = klass.send(:new, ldp_obj: ir_object)
```
pgwillia: What happens to a JupiterCore::LockedLdpObject that is created with new but not saved?

lib/tasks/recover.rake
mbarnett: I think the justification should live with the weird code it justifies. "Organizational knowledge" about future consequences we have watch out for, and in my experience people tend to remember it more if it they read it as part of an explanation of why some weird code is doing what it does than if you bury it with the API docs (nobody, to a first approximation, ever bothers to read the API docs).

lib/tasks/recover.rake
mbarnett: I'm not sure I understand the question?

pgwillia: I'm not sure I understand the lifecyle of this object.  Will it live on somewhere or get garbage collected? What I think is happening here is that a new object is created just so that it's to_solr can be collected and sent to Solr.  The object already exists in Fedora, and this copy is never saved so I'm just wondering what happens to it. Not any critique, just for my information.

lib/tasks/recover.rake
pgwillia: I agree that the first part of the explanation is useful and good explanation of the 'weird code'. The second part of the comment reads like a complaint. If it's meant as a warning about abusing `additional_search_index` then I'd argue that it should be put closer to the implementation of the problem: around here https://github.com/ualbertalib/jupiter/blob/f71cbbde276ebe1e9e1fcf1c81ca7693dc21b02d/app/models/collection.rb#L18

weiweishi: Suggestion: @mbarnett Could we add the comment here (in additional to recover.rake) as Tricia suggested, https://github.com/ualbertalib/jupiter/blob/f71cbbde276ebe1e9e1fcf1c81ca7693dc21b02d/app/models/collection.rb#L18 just so that we cover all the bases. 

Since inline comments are part of our documentation, maybe rephrase the second paragraph to a more clean documentation of the limitation, instead of potentially be interpreted as an unreasonable user requirement? e.g: 
* additional_search_index works in Jupiter phase 1 because there is not expected to be a large number of Collections
* however Solr-only data that isn't recreatable by inspecting only the given object's Fedora record, and no others, vastly complicate scenarios like this.
* this needs to be revisit if we are dealing with large number of filesets/items ordering 
* this method should be used with caution etc etc.

mbarnett: Yes, I'm planning on circling back to this to update it once other work is off of my place.

However, I do think we should explicitly capture the fact that this current situation is the result of, if not an "unreasonable user requirement" precisely, an aggressively vocal stakeholder pushing his idea for how we should solve this problem, which I warned would lead to this exact situation. In the future, rather than rely on me running around catching every possible violation of this, it would be nice if stakeholders took developer objections more seriously.

To be frank: We got lucky here. We could _easily_ have ended up in a situation where recovery was much more difficult. We need to be disseminating knowledge of this near-miss throughout the team, because it reinforces several important facts:

1) Stakeholders are not developers, do not have (and should not be expected to have) a grasp of the full context of development problems.

2) Developers _need_ to be thinking about the bigger picture, and not just "doing what a stakeholder told them to do". Check with your teammates before doing something just because so-and-so from upstairs "told you" to only put that data in Solr.

3) The primary datastore cannot purely represent one team's idea of "how the data should be stored", because as this case illustrates, there are complicated consequences to seemingly-simple inclusions or exclusions of data from a record.

weiweishi: Thanks. @mbarnett. I see the pain point here, and those conversations need to happen, and we'd want to address the issues illustrated here in a more collaborative and mutually respectful way, and with that note, I want to reassure you that the momentum on the retrospective, where some of these issues were raised - hasn't been lost yet, and we will initiate some of those conversations soon. 

But since we are considering inline comments as part of our documentation practice, it would be more helpful for others in the team and potential future adopters if the technical issues are phrased in a more objective way, while we deal with other issues on teamwork/collaboration/requirement gathering etc, and the associated frustration offline from github, with more meaningful face-to-face conversations.

mbarnett: I hear what you're saying, and like I said I will be circling back to address the feedback on this Pull Request. I don't think updating it is currently the highest priority task on my plate.

But, as a note: face-to-face dissemination of knowledge about why the codebase is the way it is, what led it to be that way, the consequences of it being that way, and related knowledge is among the worst possible ways to do this. It robs potential new teammates and our future selves forgetting these conversations of essential knowledge and context.

If we claim to value openness and open source participation, we should not fear speaking in the open about the challenges that shape our codebase, because this helps other people to understand it best.

weiweishi:
Yes, totally, this is not a high priority task comparing to some other
things on your plate, and can wait until you have time to circle back.

And to clarify, I am in support of your point of having the knowledge on "why
the codebase is the way it is, what led it to be that way, the consequences
of it being that way, and related knowledge" live within the codebase. My
point is that these should be expressed in a more neutral objective way,
without things that can be interpreted as blaming, to prevent a
healthy conversation from happening in the future when solving the issue.

*Issues *beyond this or any particular technical issue regarding our
design/decision making/team collaboration processes can be recorded if it
helps people to understand the context but again in a
neutral/objective way. And the *discussions* around these topics are the
ones I think should be taken offline, not because of trying to hide the
issue, but because it would be better facilitated and more efficient in
reaching an agreement.


people."*

murny:
@mbarnett whats left on this? anything? PR looks good to me, I'll approve if nothing left here

mbarnett:
Closing this to quickly re-apply the changes and feedback on top of latest master rather than mess around trying to resolve out of date conflicts. Will reopen shortly.

PR #697, opened 2018-04-24
remove context/should blocks and only use shoulda-matchers

pgwillia:
#417

### DIFF
--- a/test/controllers/items/files_controller_test.rb
+++ b/test/controllers/items/files_controller_test.rb
@@ -2,7 +2,9 @@ require 'test_helper'
 
 class Items::FilesControllerTest < ActionDispatch::IntegrationTest
 
+  # TODO: candidate for fixture
   def before_all
+    super
     @community = Community.new_locked_ldp_object(title: 'Books', owner: 1).unlock_and_fetch_ldp_object(&:save!)
     @collection = Collection.new_locked_ldp_object(title: 'Fantasy Books',
                                                    owner: 1,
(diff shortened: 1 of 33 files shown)


test/controllers/items/files_controller_test.rb
```diff
@@ -2,7 +2,9 @@
 
 class Items::FilesControllerTest < ActionDispatch::IntegrationTest
 
+  # TODO: candidate for fixture
```
murny: Is this actually a candidate? I would have assume ActiveFedora stuff won't work with fixtures? 

Or if it does, it will work much like how factorygirl/factorybot was working with activefedora. Which was pretty poorly as it had no idea of how to clean up after itself, leaking tons of records everywhere, requiring us to just clean the entire fedora instance after every test.

test/test_helper.rb
```diff
@@ -6,10 +6,10 @@
 require 'minitest/hooks/test'
 require 'minitest/mock'
 require 'rails/test_help'
-require 'shoulda'
 require 'sidekiq/testing'
 require 'vcr'
 require 'webmock/minitest'
+require 'shoulda/matchers'
```
murny: The `require: false` was removed from the gemfile, so you technically don't need to require this right here as it's already required.

test/test_helper.rb
```diff
@@ -19,6 +19,14 @@
   config.ignore_localhost = true
 end
 
+Shoulda::Matchers.configure do |config|
```
murny: I think you had a question about this and why we don't just use `with.library :rails`

https://github.com/ualbertalib/jupiter/pull/689#discussion_r182847323

The answer is we don't want `with.library :action_controller` as this is kinda deprecated and not "Rails" way anymore. You also need to install another gem called 'rails-controller-testing' which brings back these removed methods from rails

test/controllers/items/files_controller_test.rb
pgwillia: You're exactly right, ActiveFedora stuff will not work with fixtures. The Testing with Rails guide describes fixtures as a 
>way of organizing test data

I was interpreting these TODOs as capturing the setup for the objects in some kind of file (perhaps YML like fixtures so that it would be transferable if we move away from AF) and building from that maybe with a couple helper methods.
I was thinking that there would be something to gain by having some stock test data for these objects rather than re-inventing the wheel every time (or copy/pasting as the case may be) even if we still have to clean up fedora after every test. Maybe it's not worth it?

test/test_helper.rb
pgwillia: Hmmm.  I thought I was getting errors without it.  I'll take another look.

pgwillia: Nope, whatever was causing the error wasn't this.  Pushing that change now.

PR #700, opened 2018-04-25
Add Content Security Policy

murny:
Uses new Rails 5.2 CSP api, to add CSP to jupiter


app/views/items/draft/_progress_bar.html.erb
```diff
@@ -19,8 +19,7 @@
   <div class="progress">
     <%= content_tag :div,
                     progress_bar_text,
-                    class: "progress-bar",
-                    style: "width: #{progress_bar_percentage}%",
```
murny: Can't use inline styles, as this is against CSP, refactor to use utility class instead (w-25/w-50/w-75/etc)

config/application.rb
```diff
@@ -38,5 +38,29 @@ class Application < Rails::Application
...
+    # but we need a way to calculate nounces/hashes/etc
+    # another improvement is for a better way to inject matomo_url into here instead of hardcoding
+    config.action_dispatch.default_headers = {
+      'X-Frame-Options' => 'SAMEORIGIN',
+      'X-XSS-Protection' => '1; mode=block',
+      'X-Content-Type-Options' => 'nosniff',
+      'X-Download-Options' => 'noopen',
+      'X-Permitted-Cross-Domain-Policies' => 'none',
+      'Referrer-Policy' => 'strict-origin-when-cross-origin',
+      'Content-Security-Policy' =>
```
murny: We can hold off on this until Rails 5.2. But the below does work just fine. However 2 caveats:

Because we are using inline scripts for analytics, this goes against CSP. So there's ways around this such as calculating a hash for the script or calculation nuances. But this might be out of scope and hopefully Rails 5.2 CSP DSL has a nice way around this. So with that said, I am just allowing unsafe-inline scripts... which sucks but better then what we had before.

Second one, is because our matomo_url from Rails.applications.secrets is a full url `analytics.library.ualberta.ca/piwik` , this doesn't play very well with CSP. So just hardcoded our Piwik analytics url for now

config/application.rb
```diff
@@ -38,5 +38,30 @@ class Application < Rails::Application
...
+      'X-Download-Options' => 'noopen',
+      'X-Permitted-Cross-Domain-Policies' => 'none',
+      'Referrer-Policy' => 'strict-origin-when-cross-origin',
+      # TODO: Move to Rails 5.2 new DSL for Content Security Policy
+      # http://guides.rubyonrails.org/security.html#content-security-policy
+
+      # TODO: Also, I am allowing unsafe-inline for script-src which kinda defats the purpose of CSP,
+      # but we need a way to calculate nounces/hashes/etc
+      # another improvement is for a better way to inject matomo_url into here instead of hardcoding
+      'Content-Security-Policy' =>
```
murny: We can hold off on this until Rails 5.2. But the below does work just fine. However 2 caveats:

Because we are using inline scripts for analytics, this goes against CSP. So there's ways around this such as calculating a hash for the script or calculation of nounces. But this might be out of scope and hopefully Rails 5.2 CSP DSL has a nice way around this. So with that said, I am just allowing unsafe-inline scripts... which sucks but better then what we had before.

Second one, is because our matomo_url from Rails.applications.secrets is a full url analytics.library.ualberta.ca/piwik , this doesn't play very well with CSP. So just hardcoded our Piwik analytics url for now

murny:
Need to pivot this to use CSP API now that we on Rails 5.2:

http://edgeguides.rubyonrails.org/security.html#content-security-policy

Especially enabling nonces on the inline script tags

config/application.rb
mbarnett: Should we move this CSP stuff into the new initializer? https://github.com/ualbertalib/jupiter/blob/master/config/initializers/content_security_policy.rb

murny: Yep for sure. This can all get removed now as headers are now enabled by default now and we can transition to the new CSP api using intializer and use nounces

config/initializers/content_security_policy.rb
```diff
@@ -4,20 +4,36 @@
...
+  policy.img_src     :self, :https, :data,
+                     'analytics.library.ualberta.ca',
+                     'www.google-analytics.com'
+  policy.object_src  :none
+  policy.script_src  :self, :https, :unsafe_inline,
+                     'analytics.library.ualberta.ca',
+                     'www.google-analytics.com',
+                     'www.googletagmanager.com'
+  policy.style_src   :self, :https,
+                     # hash for whitelisting turbolinks progress bar
```
murny: Turbolinks seems to doing some inline styles for the progress bar. Issue can be found here: https://github.com/turbolinks/turbolinks/issues/330

Calculate hash to whitelist this

config/initializers/content_security_policy.rb
```diff
@@ -4,20 +4,36 @@
...
+                     "'sha256-voXja0NHK+kj/CO6kVFGewEz+qyDFbxR+WW6e9vfN3o='",
+                     # hash for sidekiq dashboard
+                     "'sha256-Y/9Ml03MSwcSrqR+veSpLzwCRJlw4Sk1S/X68fa51ec='",
```
murny: Same for sidekiq and jquery multidownload library (couldn't find any others?)

config/initializers/content_security_policy.rb
```diff
@@ -4,20 +4,36 @@
...
-#   policy.font_src    :self, :https, :data
-#   policy.img_src     :self, :https, :data
-#   policy.object_src  :none
-#   policy.script_src  :self, :https
-#   policy.style_src   :self, :https
 
-#   # Specify URI for violation reports
-#   # policy.report_uri "/csp-violation-report-endpoint"
-# end
+# TODO: Should be able to use nounces with UJS? But can't seem to get this to work. Needs more investigation.
```
murny: As comment says, couldn't seem to get this to work with UJS. Everything works fine except for our javascript views (`js.erb`). Couldn't get these to work, keep failing CSP.

Seems they are still hashing out some issues with this (for example bug fix in certain browsers which will be available in next release: https://github.com/rails/rails/pull/32600). 

For now just using `:unsafe_inline` for script src

config/initializers/content_security_policy.rb
```diff
@@ -4,20 +4,34 @@
...
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
-# Rails.application.config.content_security_policy do |policy|
-#   policy.default_src :self, :https
-#   policy.font_src    :self, :https, :data
-#   policy.img_src     :self, :https, :data
-#   policy.object_src  :none
-#   policy.script_src  :self, :https
-#   policy.style_src   :self, :https
+# TODO: Should be able to use nounces with UJS? But can't seem to get this to work. Needs more investigation.
```
murny: As comment says, couldn't seem to get this to work with UJS. Everything works fine except for our javascript views (js.erb). Couldn't get these to pass CSP with nonces (tried http://edgeguides.rubyonrails.org/security.html#content-security-policy and read source code....should have been working?). Seems to be some bugs still being ironed out in rails like https://github.com/rails/rails/pull/32600 which will be fixed with next release. 

For now, will just allow `:unsafe_inline` for `script_src` :cry: . Will need to revisit this in the future.

PR #712, opened 2018-04-30
refactor duplicates in tests

### LINKED ISSUE #417: Rewrite "RSpec"y Context/Should tests into before_all/test

pgwillia:
Closes #417 for now.

From `gem install flay && flay test > duplicates.txt`

- [x] JOHNDOE_AUTH_HASH
- [x] removed second @item1
- [x] Random Item Fixture
  test/models/file_set_test.rb:35
  test/models/item_test.rb:331
- [x] Fancy Item Fixture
  test/controllers/file_sets_controller_test.rb:13
  test/integration/sitemap_test.rb:12
  test/integration/sitemap_test.rb:43
- [x] Nice Community and Collection fixture
  test/controllers/admin/collections_controller_test.rb:5
  test/controllers/collections_controller_test.rb:5
- [x] Books Community and Collection fixture
  test/controllers/items/files_controller_test.rb:6
  test/system/deposit_item_test.rb:5
- [x] Books Community and Collection fixture
  test/controllers/items/draft_controller_test.rb:5
  test/models/draft_item_test.rb:5
- [x] Fancy Community & Collection fixture, maybe Fancy Item fixture
  test/system/item_show_test.rb:15
  test/system/item_show_test.rb:36
- [x] Community and foo Collection fixture, maybe Random Item fixture?
  test/models/file_set_test.rb:32
  test/models/item_test.rb:328

### DIFF
--- a/test/ldp_fixtures/collection.yml
+++ b/test/ldp_fixtures/collection.yml
@@ -1,0 +1,12 @@
+nice:
+  Community: nice
+  title: 'Nice collection'
+  owner: 1
+fancy:
+  Community: fancy
+  title: 'Fancy collection'
+  owner: 1
+books:
+  Community: books
+  title: 'Fantasy Books'
+  owner: 1
--- a/test/system/authentication_test.rb
+++ b/test/system/authentication_test.rb
@@ -2,19 +2,20 @@ require 'application_system_test_case'
 
 class AuthenticationTest < ApplicationSystemTestCase
 
+  JOHNDOE_AUTH_HASH = OmniAuth::AuthHash.new(
+    provider: 'saml',
+    uid: 'johndoe',
+    info: {
+      email: 'johndoe@ualberta.ca',
+      name: 'John Doe'
+    }
+  )
+
   test 'should log in and log out work with correct credentials' do
     visit root_url
 
     Rails.application.env_config['omniauth.auth'] =
-      OmniAuth.config.mock_auth[:saml] =
-        OmniAuth::AuthHash.new(
-          provider: 'saml',
-          uid: 'johndoe',
-          info: {
-            email: 'johndoe@ualberta.ca',
-            name: 'John Doe'
-          }
-        )
+      OmniAuth.config.mock_auth[:saml] = JOHNDOE_AUTH_HASH
 
     click_on I18n.t('application.navbar.links.login')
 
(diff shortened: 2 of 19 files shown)


test/system/authentication_test.rb
```diff
@@ -2,19 +2,20 @@
 
 class AuthenticationTest < ApplicationSystemTestCase
 
+  JOHNDOE_AUTH_HASH = OmniAuth::AuthHash.new(
```
murny: Nice refactor :+1:

test/ldp_fixtures/collection.yml
```diff
@@ -0,0 +1,12 @@
+nice:
```
murny: Did you see any reason why we need so many versions of different collections/communities? Or can we be reusing the same ones over and over again (remove a few of these and reuse one or two instead)?

pgwillia: Some have subtle differences in [visibility](https://github.com/ualbertalib/jupiter/blob/c17b8976ba6c88d11861a11fb1c5c4e767e01303/test/models/thesis_test.rb#L10) [fedora3_uuid](https://github.com/ualbertalib/jupiter/blob/master/test/controllers/redirect_controller_test.rb#L14), or [owner](https://github.com/ualbertalib/jupiter/blob/b0e18d3c6cd99cc684d1e931d623624c25854f51/test/models/collection_test.rb#L7), etc. For the ones that are there, for sure there only really needs to be one or two. I guess just laziness on my part.

murny:
That thought also crossed my mind, but looking at travisCI runs, it doesn't seem all that much slower? Did you do any bench marking to see the difference of performance? I

t's probably fine how it is. Going off TravisCI build times doesn't seem that big of a difference. (even if its a bit slower, this helps future proof us with the migration path off fedora etc).

Issue #550, opened 2018-03-09
Jupiter: CCID protected thumbnail of UofA logo should have some text on it indicating its CCID protected


murny:
UofA Thumbnail should have the `Item restricted to University of Alberta users` text on the thumbnail itself, so it shows up in the search list etc.

- Need to update the image
- Can probably remove the `Item restricted to University of Alberta users` text on the show page now since it's in the image

murny:
This is now in master. Closing issue

PR #714, opened 2018-05-01
Add text to ccid protected logo

murny:
#550 

Kinda doesn't work as well in practice, since our thumbnails are 100x100 in size so text becomes hard to read on search result pages...

Previous:
image
image
Changes:
image
image

Using lock icon instead:

image
image


app/views/application/_feature_image.html.erb
```diff
@@ -7,5 +7,5 @@
     </div>
   <% end %>
 <% else %>
-  <%= image_tag('uofa-logo.jpg', class: 'j-feature-image img-thumbnail',  alt: '', size: '350x350') %>
+  <%= image_tag('uofa-logo.png', class: 'j-feature-image img-thumbnail',  alt: '', size: '350x350') %>
```
murny: PNG apparently better for text images, half the size too.

config/locales/en.yml
```diff
@@ -420,7 +420,7 @@ en:
       download: 'Download'
       download_all: 'Download All'
       header: 'Files'
-      ccid_restricted_item: Item restricted to University of Alberta users
+      ccid_restricted_item: Item Restricted to University of Alberta Users
```
murny: Make it the same as the image, title case this

sfbetz:
I think this is okay. But, we could also use a nice grey closed padlock icon under the logo instead of the text? That's pretty visual and easy to understand (I think).

weiweishi:
Do we need to get the image from someone? Or @murny would you be able to make this change?

PR #718, opened 2018-05-03
Rails 5.2

mbarnett:
- Replaces GID monkeypatches/table changes in ActiveStorage with the original polymorphic setup, and uses shim objects for compatibility with Fedora objects. This involves migrating all of the current data into the more standard format.
- ActiveStorage URLs have changed, so this adds a redirect for the form currently in production
- Overrides signing in ActiveStorage blobs
- Overrides Blob Controller to use permanent URLs and eliminate the new mandatory redirect
- CSP and other features still need to be enabled.


config/routes.rb
```diff
@@ -1,6 +1,8 @@
 require 'sidekiq/web'
 require_dependency 'admin_constraint'
 
+# rubocop is bad and recommends insane things
+# rubocop:disable Style/FormatStringToken
```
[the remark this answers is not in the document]
mbarnett: Not useless in the version of Rubocop we're on. Once we're merged we should upgrade Rubocop finally.

Gemfile
```diff
@@ -7,7 +7,7 @@ end
 
 # Core Rails stuff
 gem 'puma', '~> 3.11'
-gem 'rails', '~> 5.1.6'
+gem 'rails', '~> 5.2.0'
```
murny: Did we want to be using ActiveStorage variant stuff? If so we should add `gem 'mini_magick'`?

app/helpers/page_layout_helper.rb
```diff
@@ -40,11 +40,8 @@ def page_description(description = nil)
   def page_image
     default_url = image_url('era-logo.png')
     # We only have images on community and item/thesis show pages
-    image = @community&.logo&.attachment || @item&.thumbnail&.attachment
-
-    # TODO: ActiveStorage 5.2 upgrade, fix this hack
-    # needs to be a full url so can use rails_blob_url or url_for in 5.2 instead
-    image_url = root_url + image.url if image.present?
+    image = @community&.logo_attachment || @item&.thumbnail_attachment
+    image_url = root_url + url_for(image) if image.present?
```
murny: What about blob url instead of the root_url concatenation? Like so:
`image_url = rails_blob_url(image) if image.present?`

Gemfile
mbarnett: I'm going to look at that as part of the backgrounding of all file ingestion into fedora, since it touches on thumbnail generation etc.

bin/rails
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
```
murny: This stuff still valid I think? We still using spring in rails 5.2?

mbarnett: The upgrade script ripped it out. I've had it entirely turned off locally forever, so someone who actually uses it might need to re-enable it.

That said, it's very fast with bootspring now. To the point I'm not sure Spring adds anything.

db/schema.rb
```diff
@@ -10,20 +10,18 @@
...
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name"
-    t.string "record_gid"
-    t.integer "blob_id"
+    t.bigint "blob_id", null: false
     t.datetime "created_at"
+    t.bigint "record_id"
+    t.string "record_type"
     t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
-    t.index ["record_gid", "blob_id"], name: "index_active_storage_attachments_on_record_gid_and_blob_id", unique: true
```
murny: should we add a index back? Seems like default one is:
`    t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
`

test/helpers/page_layout_helper_test.rb
```diff
@@ -82,7 +82,7 @@ class PageLayoutHelperTest < ActionView::TestCase
     @community.logo.attach io: File.open(file_fixture('image-sample.jpeg')),
                            filename: 'image-sample.jpeg', content_type: 'image/jpeg'
 
-    assert_equal root_url + @community.logo.attachment.url, page_image
+    assert_equal root_url + url_for(@community.logo_attachment), page_image
```
murny: blob_url thing here too maybe instead of root_url concatenation

app/controllers/active_storage/blobs_controller.rb
```diff
@@ -0,0 +1,15 @@
+class ActiveStorage::BlobsController < ActiveStorage::BaseController
+
+  include ActiveStorage::SetBlob
+
+  def show
```
murny: Just wanted to understand what was going on here. I am assuming we overriding this just for the signed/expiration of blob ids stuff correct?

mbarnett: Yeah, there's a complicated workflow here. By default:

1) Blobs sign their URLs, so when you do `url_for(blob)`, you get `/rails/active_storage/blobs/signed_id/filename`. The signature expires in 5 minutes.

2) when the browser asks for the file from that URL, Rails verifies the signed ID , and then generates _another_ signed expiring URL of the form `/rails/active_storage/disk/signed_key/filename` and issues a redirect to that.

3) When the browser asks for _that_ file, Rails again verifies the signature and expiration, and then sends the file.

This is apparently done to mimic the way the other providers work (verify the id, redirect to an S3 URL, which also verifies the expiring URL, etc)

Issue #702, opened 2018-04-26
PMPY implementation in Jupiter


weiweishi:
I believe all the work has been down to bring PMPY inline with Jupiter's data model, we need to implement this in production ASAP. 
- [x] Add code to Jupiter on save to push IDs into preservation queue #720 
- [x] Update logging to capture additional information https://github.com/ualbertalib/pushmi_pullyu/issues/144 (PR here: https://github.com/ualbertalib/pushmi_pullyu/pull/146)
- [x] Test of PMPY in test env
- [x] coordinate with Ops to implement
  - [x] set up a new container in ERA (called jupiter)

murny:
Work for the first two bullets are completed in PRs (#720  and https://github.com/ualbertalib/pushmi_pullyu/pull/146). 

So once these are reviewed and merged we can get this on staging for testing. 

In mean time I'm going to go work on a different issue (update my CSP PR to work with 5.2).

weiweishi:
All tasks are completed. PMPY is in place.

PR #720, opened 2018-05-07
Add preservation callback for items to push item ids into redis

murny:
Code to resolve first bullet in #702


app/models/concerns/item_properties.rb
```diff
@@ -169,6 +170,34 @@ def purge_files
...
+        # TODO: <removed temporarily> log to external service iff result != true
+        true
+      rescue StandardError
+        # we trap errors in writing to the Redis queue in order to avoid crashing the save process for the user.
+        Rollbar.error("Error occured in push_item_id_for_preservation, Could not preserve #{id}", e)
+        true
+      end
+
+      # rubocop:disable Style/GlobalVars
+      def preserve
```
murny: Hydranorth has moved this out to a lib file: https://github.com/ualbertalib/HydraNorth/blob/master/lib/hydranorth/preservation_queue.rb

Which don't think we need to do here. Makes it much simiplier this way (no need to pass in items id etc)

app/models/concerns/item_properties.rb
```diff
@@ -169,6 +170,34 @@ def purge_files
...
+        queue_name = Rails.application.secrets.preservation_queue_name
+
+        $queue ||= ConnectionPool.new(size: 1, timeout: 5) { Redis.current }
```
murny: I assume connection pool wants this to be a global variable for it to do the connection pooling stuff. So I disabled rubocop for this

app/models/concerns/item_properties.rb
```diff
@@ -169,6 +170,34 @@ def purge_files
...
 
+      def push_item_id_for_preservation
+        result = preserve
+
+        Rails.logger.warn("Could not preserve #{id}") if result != true
+        # TODO: <removed temporarily> log to external service iff result != true
+        true
+      rescue StandardError
+        # we trap errors in writing to the Redis queue in order to avoid crashing the save process for the user.
+        Rollbar.error("Error occured in push_item_id_for_preservation, Could not preserve #{id}", e)
```
murny: There was a TODO in the old code for adding a rollbar notification? https://github.com/ualbertalib/HydraNorth/blob/master/app/models/concerns/hydranorth/generic_file/preservation.rb#L18

So added one here

app/models/concerns/item_properties.rb
```diff
@@ -169,6 +170,34 @@ def purge_files
         self.ordered_members = []
       end
 
+      def push_item_id_for_preservation
+        result = preserve
+
+        Rails.logger.warn("Could not preserve #{id}") if result != true
+        # TODO: <removed temporarily> log to external service iff result != true
```
murny: Should we be logging this to rollbar as well?  Or logging this is fine?

lib/tasks/jupiter.rake
```diff
@@ -1,10 +1,10 @@
-namespace :cleanup do
+namespace :jupiter do
```
murny: This rake task was out growing it's "cleanup" name. Rename this file and namespace to be jupiter.

test/models/item_test.rb
```diff
@@ -418,4 +418,139 @@ class ItemTest < ActiveSupport::TestCase
     end
   end
 
+  # Preservation queue handling
```
murny: Brought over existing tests from hydranorth: https://github.com/ualbertalib/HydraNorth/blob/master/spec/models/generic_file_spec.rb#L217-L312

Converted them to minitest and using activesupport time helpers

test/test_helper.rb
```diff
@@ -31,6 +31,7 @@
 
 class ActiveSupport::TestCase
 
+  # TODO: Remove after 5.2 PR gets merged
```
murny: This helper gets added in 5.2, so we can just remove ours once we finish upgrading

app/models/concerns/item_properties.rb
mbarnett: Rollbar, probably? There's like zero monitoring of the Rails logs in production

murny: Okay adding this to the PR right now. Will need reapproval after.

Issue #526, opened 2018-03-01
Embargo expiry job


mbarnett:
We need to set up a Sidekiq job for scanning items with VISIBILITY_EMBARGO and whose embargo_end_date is < current date,and transition them to the post-embargo visibility. I believe there's also a requirement to push information regarding this change into the embargo_history attribute?

murny:
Need to investigate how to run this periodically/reoccurring? Job probably runs once a day and checks for any expired embargo_end_date. Need to use a cron job or use a cron-like gem to give this behaviour for us with sidekiq.

https://github.com/ondrejbartas/sidekiq-cron
https://github.com/javan/whenever
???

murny:
Is there a requirement for how often we want to be checking embargo dates? Every day? At 6 AM or something?

mbarnett:
Hydranorth did it at midnight IIRC

murny:
Looks like it was 10:01 PM? From digging into the ansible playbook here's the relevant section:
```
     - name: cronjob for remove embargoes
       cron: name="embargoes" hour=22 minute=1 job="cd /var/www/sites/hydranorth && RAILS_ENV={{rails_env}} bundle exec rake hydranorth:remove_lapsed_embargoes"
       tags: embargoes
       when: ansible_hostname == "bournemouth"
```

I guess we can agree it be daily. I'll make it 10 PM (no 1 minute cause that's silly) unless someone says otherwise

murny:
This is now in production. Closing issue

PR #731, opened 2018-05-24
End embargo job

murny:
#526 

- Creates ActiveJob for querying solr for Items/Theses that have expired embargo dates and transitions these to the visibility  defined in `visibility_after_embargo`, and clears embargo fields
- Uses Sidekiq-cron for running job on a cron schedule (10 PM daily)
- Adds tests for job logic of removing expiry of expired embargoes

### DIFF
--- a/app/jobs/embargo_expiry_job.rb
+++ b/app/jobs/embargo_expiry_job.rb
@@ -1,10 +1,11 @@
-class CheckEmbargoExpiryJob < ApplicationJob
+class EmbargoExpiryJob < ApplicationJob
 
   queue_as :default
 
-  def perform(*args)
+  def perform(*_args)
     # TODO: Should be a better way to query solr?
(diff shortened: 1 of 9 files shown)


app/jobs/embargo_expiry_job.rb
```diff
@@ -0,0 +1,46 @@
+class EmbargoExpiryJob < ApplicationJob
+
+  queue_as :default
+
+  def perform(*_args)
+    # TODO: Should be a better way to query solr?
+    # Can't do date logic like this:
+    # Item.where(visibility: ItemProperties::VISIBILITY_EMBARGO, embargo_end_date: [* TO NOW])
+    # And faceted_search has visibility based on current_user
+    # So dropped down to perform_solr_query to get what I need
```
mbarnett: I think this works for the time being. I'm going to throw something in the backlog to add a less low-level interface that will work for this use-case

PR #735, opened 2018-05-28
Update Rubocop and fix offenses

murny:
- Upgrade Rubocop from 0.51 to 0.56


app/models/jupiter_core/deferred_faceted_solr_query.rb
```diff
@@ -159,9 +159,9 @@ def reify_result_set
 
   def sort_clause
     model = criteria[:restrict_to_model].first.owning_class
-    indexes = criteria[:sort].present? ? criteria[:sort] : model.default_sort_indexes
+    indexes = criteria[:sort].presence || model.default_sort_indexes
```
mbarnett: Am I the only one that finds `presence` less readable?

I'm not married to removing it but I'm not loving it either

bin/yarn
```diff
@@ -1,11 +1,9 @@
...
 Dir.chdir(APP_ROOT) do
-  begin
-    exec 'yarnpkg', *ARGV
-  rescue Errno::ENOENT
-    warn 'Yarn executable was not detected in the system.'
-    warn 'Download Yarn at https://yarnpkg.com/en/docs/install'
-    exit 1
-  end
+  exec 'yarnpkg', *ARGV
+rescue Errno::ENOENT
```
mbarnett: I think rubocop screwed this up. Now if APP_ROOT doesn't exist (possible if you're cd'd into the dir in one terminal after `rm`-ing it in another, that error will get rescued here and the wrong error message displayed.

It's a real corner-case, admittedly.

murny: Can revert, rubocop autocorrected this. So can disable this file for whatever it's doing.

app/models/jupiter_core/deferred_faceted_solr_query.rb
murny: I'm indifferent? So if anyone feels strongly about this, we can disable

mbarnett: @cjwant || @pgwillia ? Any strong feelings? I can learn to live with this, I just don't find it terribly obvious.

cwant: Sorry, no strong opinion.
In honesty, I had to look `presence` up -- would be nice if the method had a more illustrative name (oh well).

mbarnett: 99% of my feeling on this is "I wish it had a slightly different name".

I guess we live with it. My eyes aren't bleeding or anything.

cwant: In this very particular case, maybe `fetch` would be more readable?

murny: gif

mbarnett: Between the two, I guess I like `present` more. It's kind of a "which do I like finding more, a  used bandaid or a toe-nail" situation though

PR #759, opened 2018-06-28
Remove fileset usage from web request processing in favour of ActiveStorage

mbarnett:
Deprecates Item#file_sets in favour of using ActiveStorage to service all web requests, and ingest files into Fedora in the background to maintain compatibility with the existing preservation pipeline.

TODO:

- [x] ordered_members scrambled/jobs crash if attempt to update the old way
- [x] weird bug on drafts where every file has same thumbnail
- [ ] more testing around migrating existing data
- [x] convert gc task to job & have it queue future versions of itself?


app/helpers/items_helper.rb
```diff
@@ -38,4 +45,13 @@ def humanize_date(dateish)
   rescue ArgumentError
     dateish
   end
+
+  private
+
+  def file_url(file, action)
+    # The slash in the controller name below allows this to work in admin namespace
```
murny: Is this code comment still relevant? Don't see a slash in `controller: 'downloads'`?

config/initializers/clamby.rb
```diff
@@ -1,5 +1,5 @@
 if defined?(Clamby)
   Clamby.configure(check: true,
-                   daemonize: true,
+                   daemonize: false,
```
pgwillia: Seems to be an unrelated change?  Are there any implications to this?

app/helpers/items_helper.rb
mbarnett: Will remove 👍

test/models/community_test.rb
```diff
@@ -44,45 +44,11 @@ class CommunityTest < ActiveSupport::TestCase
     c.logo.attach io: File.open(file_fixture('image-sample.jpeg')),
                   filename: 'sample1.jpeg', content_type: 'image/jpeg'
 
-    # Assert database records exist
```
pgwillia: I'm assuming that you're removing these tests because we don't need to test ActiveStorage (rails should do that for us)?

test/models/file_set_test.rb
```diff
@@ -36,19 +36,23 @@ class FileSetTest < ActiveSupport::TestCase
...
       unlocked_item.add_to_path(community.id, collection.id)
       unlocked_item.save!
-      unlocked_item.add_files([file])
-      unlocked_item.save!
+      Sidekiq::Testing.inline! do
+        item.add_and_ingest_files([file])
+      end
+    end
+    file_set = nil
+    # we need to call this deprecated method to verify the model is working
```
pgwillia: I don't think I know why we need to call a deprecated method to verify the model.  What do we do when the method is really gone?

app/helpers/items_helper.rb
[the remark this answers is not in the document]
pgwillia: @murny jinx.  This is an old comment and can be removed.

app/controllers/downloads_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  include ActiveStorage::SetBlob
+
+  # Needed for Activestorage compatibility, normally happens in ActiveStorage::BaseController for controllers that subclass is
+  # see https://github.com/rails/rails/blob/master/activestorage/app/controllers/active_storage/base_controller.rb
+  before_action do
+    ActiveStorage::Current.host = request.base_url
+  end
+
+  before_action :load_and_authorize_file, only: [:view, :download]
+  after_action :update_download_count, only: [:view, :download], unless: -> { request.bot? }
```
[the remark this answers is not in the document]
mbarnett: It doesn't run.

Conceptually when an action runs in a controller it looks something like:

```ruby
def run_action
  run_before_actions
  yield # run the actual action
  run_after_actions
end```

If the action raises an exception during the yield it unwinds the stack upwards until caught, so `run_after_actions` never executes.

File: app/models/concerns/item_properties.rb [hunk 6]
```diff
@@ -53,11 +54,43 @@ def doi_state
...
+      files_attachment_shim.logo_file
+    end
+
+    def add_and_ingest_files(file_handles = [])
+      return if file_handles.blank?
+      raise 'Item not yet saved!' if id.nil?
+
+      file_handles.each do |fileio|
+        attachment = files.attach(io: fileio, filename: File.basename(fileio.path)).first
+        # binding.pry
```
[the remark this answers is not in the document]
mbarnett: d'oh

app/views/items/draft/_thumbnail.html.erb
```diff
@@ -0,0 +1,14 @@
+<% begin %>
```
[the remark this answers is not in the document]
mbarnett: Ugh, yes, although it's not immediately obvious why:

On the Deposit page it renders a thumbnail for _every_ file attached to an item, not just the "thumbnail" file, and it does this via JS that goes through update_files_list.js.erb, so this file ends up being called on everything wheres `thumbnail_url` only deals with the particular file that's been designated to represent the Item as a thumbnail.

I'll add a comment trying to clarify this.

db/migrate/20180607201513_add_logo_id_to_attachment_shims.rb
```diff
@@ -0,0 +1,5 @@
+class AddLogoIdToAttachmentShims < ActiveRecord::Migration[5.2]
```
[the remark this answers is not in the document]
mbarnett: Not in this case. Rails can automatically infer how to rollback an add_column call in a Migration `change` method

test/integration/site_for_bots_test.rb
```diff
@@ -122,11 +126,11 @@ def before_all
...
     assert_select "meta[name='citation_author'][content='#{@thesis.dissertant}']"
     assert_select "meta[name='citation_publication_date'][content='#{@thesis.creation_date}']"
-    assert_select format("meta[name='citation_pdf_url'][content='%<citation_url>s']",
-                         citation_url: Rails.application.routes.url_helpers.url_for(
-                           controller: :file_sets, action: :show, id: @thesis.id,
-                           file_set_id: @thesis.file_sets.first.id,
-                           file_name: @thesis.file_sets.first.contained_filename,
+    assert_select format("meta[name='citation_pdf_url'][content='%s']",
+                         Rails.application.routes.url_helpers.url_for(
+                           controller: :downloads, action: :view, id: @thesis.id,
```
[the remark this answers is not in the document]
mbarnett: The 'view' action in the downloads controller in this case corresponds to the button marked view in the UI, which is intended to load the file for viewing in the browser instead of downloading it.

There's a lot of terminological overlap/conflict with REST that led to confusion I tried to avoid by specifically naming the actions after the corresponding UI labels (previously, I believe, the download button triggered the REST 'show' action which led to the file downloading, not "showing" in the "view" sense).

config/initializers/clamby.rb
mbarnett: Ah, yeah, that's a bad commit -- a local change to work around something on my machine. Will remove.

app/jobs/garbage_collect_blobs_job.rb
```diff
@@ -0,0 +1,22 @@
+class GarbageCollectBlobsJob
```
murny: Instead of using sidekiq unique, an initializer to kick off the job, and every job then proceeding to enqueue the job again, etc, you could have set this up as a cron job using ['sidekiq-cron'](https://github.com/ondrejbartas/sidekiq-cron)?

Take a peak at `config/schedule.yml` and the EmbargoExpiryJob, etc. If you want some ideas of how to do this

test/models/community_test.rb
mbarnett: Yeah, ActiveStorage has its own tests to ensure it's managing its tables how it sees fit. Any tests we write to depend on its inner workings will be fragile and break whenever it changes, so it's better to stop testing at the boundary where our code ends.

(these tests made a lot more sense when we were using a hacked-up version of an unofficial prototype)

test/models/file_set_test.rb
mbarnett: We need something to verify that when we add files to an Item, and the background job runs, the files got ingested into file sets. 

But this is just about the only place we should still be calling this method (hence the deprecation warning, so that it's obvious if someone writes a new PR or something an inadvertently uses file_sets instead of files -- the warning will make it nice and obvious).

When this method is really gone it will presumably because we've stopped dealing with filesets entirely, at least in the generating them for Fedora sense.

app/jobs/garbage_collect_blobs_job.rb
mbarnett: sidekiq-cron wasn't a thing when I branched, mostly. I'll switch it around.

pgwillia:
Test coverage is probably okay 87.92% (vs 90.13% in master).  Here's where I think we might be able to improve.
* https://github.com/mbarnett/jupiter/blob/d835744c3ded07e583f0bf66d78fe36301e4674f/app/models/jupiter_core/active_storage_macros.rb#L76
* https://github.com/mbarnett/jupiter/blob/d835744c3ded07e583f0bf66d78fe36301e4674f/app/models/concerns/item_properties.rb#L233
* https://github.com/mbarnett/jupiter/blob/d835744c3ded07e583f0bf66d78fe36301e4674f/app/models/concerns/item_properties.rb#L66-L72
* https://github.com/mbarnett/jupiter/blob/d835744c3ded07e583f0bf66d78fe36301e4674f/app/controllers/downloads_controller.rb

app/jobs/file_attachment_ingestion_job.rb
```diff
@@ -0,0 +1,83 @@
...
+          # Store file properties in the format required by the sitemap
+          # for quick and easy retrieval -- nobody wants to wait 36hrs for this!
+          escaped_path = CGI.escape_html(Rails.application.routes.url_helpers.url_for(
+                                           controller: :downloads,
+                                           action: :view,
+                                           id: unlocked_obj.id,
+                                           file_set_id: unlocked_fileset.id,
+                                           file_name: unlocked_fileset.contained_filename,
+                                           only_path: true
+                                         ))
```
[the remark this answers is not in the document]
mbarnett: We need to update Rubocop because right now Hound and Rubocop want those brackets in different places

mbarnett:
@pgwillia you actually caught a couple of instances of dead code that crept in when I resolved a bunch of merge conflicts with master

app/jobs/garbage_collect_blobs_job.rb
murny: Is uniqueness still a concern here now that you switched to cron job?

mbarnett: I think so? I'm being a bit paranoid, but if we tweak how often it runs, and it runs for long enough, weird things are going to happen if a second job starts before the first job ends since nothing locks the tables). I don't think sidekiq-cron has any notion of not double-pumping long-running jobs.

It's not a huge deal but it is kind of defensive.

murny: Sounds good :+1:

test/controllers/items_controller_test.rb
```diff
@@ -1,7 +0,0 @@
-require 'test_helper'
-
-class ItemsControllerTest < ActionDispatch::IntegrationTest
```
murny: Probably should keep this and add tests for this?

I guess show action is at least being tested in system/integration tests (which most of our integration tests should either be moved to controller tests or system tests) but yeah not a big deal either

¯\_(ツ)_/¯

Issue #709, opened 2018-04-30
Jupiter: Thesis Deposit


weiweishi:
Requirements:
- [ ] Allow admin users to deposit/edit theses with associated specialized metadata as outlined here:
https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_thesis.md
- [ ] No need to show item type on deposit/edit
- [ ] Can only deposit into Graduate Studies and Research, Faculty of / Theses and Dissertations
and St. Stephen's community/collections
- [ ] Graduation date should also be treated as a controlled vocabulary. Proceed with two dropdowns: optional: Spring or Fall + required year dropdown. 
- [ ] It uses a default license. Language as below:
"This thesis is made available by the University of Alberta Libraries with permission of the copyright owner solely for non-commercial purposes. This thesis, or any portion thereof, may not otherwise be copied or reproduced without the written consent of the copyright owner, except to the extent permitted by Canadian copyright law."
- [ ] Doesn't need to have a CCID protected option for visibility

Wireframes:
![Image of Thesis Deposit Mockup](https://jupiter.mybalsamiq.com/mockups/6178693.png)
https://jupiter.mybalsamiq.com/projects/jupiter/Deposit%20-%20Thesis

sfarnel:
Thanks @weiweishi 
Theses can be deposited into any of the three St. Stephen's College collections.
Graduation date is tricky. Any new UA theses will have it in the form noted in the data dictionary, but legacy UA and St. Stephen's new theses may or may not. If there is a way to handle this that would be great. Copying in @danydvd and @anayram so we can think through how best to express this in the data dictionary

weiweishi:
thanks @sfarnel 
this seems to be in contradiction with the comments on the wireframe:
" All theses deposited by this form should go into FGSR community --> Theses and dissertations collection. There may be these that don't belong to that community but will only be deposited via batch. "

So I just wanted to clarify.

sfarnel:
Thanks @weiweishi @leahvanderjagt can confirm whether or not a St. Stephen's would ever come not in a batch. Perhaps not!

weiweishi:
Confirmed with @leahvanderjagt that admin should be able to deposit into both FGSR and St. Stephen's College community and their collections.

murny:
Is there a defined list of required fields for thesis on the form?

According to here the properties that are required in the data dictionary:
https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_thesis.md#required

[...]

PR #761, opened 2018-06-29
Thesis deposit

murny:
#709 

TODO:
- Refactor views to use partials -- might be tough, lots of differences like urls, i18n, some are hugely different etc. Leaving this off the table for now


app/models/concerns/draft_properties.rb
```diff
@@ -0,0 +1,133 @@
+module DraftProperties
```
murny: Moved as much as possible between draft items and theses models into a shared concern

app/controllers/concerns/draft_actions.rb
```diff
@@ -0,0 +1,146 @@
+module DraftActions
```
murny: Moved as much as possible for the draft actions for draft controller for both theses and items into a shared concern. Also did same for the file actions controller

app/views/items/draft/describe_item.html.erb
```diff
@@ -1,4 +1,4 @@
-<% page_title(header) %>
+<% page_title(t('.header')) %>
```
murny: This fixes a bug in production right now. Currently the general title is repeated twice...for example if im on deposit item, describe item step, the page title becomes:

`Deposit Item | Deposit Item | ERA`

When it should be

`Describe Item | Deposit Item | ERA`

app/views/collections/show.html.erb
```diff
@@ -37,11 +37,22 @@
     <h2 class="mr-auto">
       <%= t('.items_list_header') %>
     </h2>
-    <% if policy(:item).new? %>
-      <%= link_to create_draft_items_path(collection: @collection.id), method: :post,
-          class: 'btn btn-primary' do %>
-        <%= fa_icon('plus-circle') %>
-        <%= t('.deposit_item') %>
+
+    <% if @collection.restricted.blank? %>
```
murny: Only show deposit button on collection show page if you can actually deposit inside this community/collection pair. 

Previously, you could go to the FGSR thesis collection and press the deposit item button, which once you fill out the describe form, you will get an error saying this collection is restricted. Which doesn't make much sense.

db/seeds.rb
```diff
@@ -21,7 +21,7 @@
 
   # start fresh
   [Announcement, ActiveStorage::Blob, ActiveStorage::Attachment, JupiterCore::AttachmentShim,
-   Identity, User, Type, Language].each(&:destroy_all)
+   Identity, User, Type, Language, Institution].each(&:destroy_all)
```
pgwillia: Ah, this is why it wasn't working for me without re-seeding data.

app/controllers/admin/theses/draft_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  include Wicked::Wizard
+  include DraftActions
+
+  steps(*DraftThesis.wizard_steps.keys.map(&:to_sym))
+
+  private
+
+  def authorize?
+    false
```
cwant: What does `false` here mean -- the item is open to everybody and doesn't go through Pundit?

murny: Because thesis deposit is nested under /admin, it inherits the admin authorization:
https://github.com/ualbertalib/jupiter/blob/master/app/controllers/admin/admin_controller.rb#L10

(its protected also on the route level via a constraint)

So we don't need to do any further authorization checking within each action here. Hence we just turn it off for theses. 

If that makes sense?

app/models/thesis.rb
```diff
@@ -55,6 +55,76 @@ def item_type_with_status_code
...
+        new_attachment = ActiveStorage::Attachment.create(record: thesis.files_attachment_shim,
+                                                          blob: attachment.blob, name: :shimmed_files)
+        FileAttachmentIngestionJob.perform_later(new_attachment.id)
+      end
+
+      thesis.set_thumbnail(thesis.files.find_by(blob_id: draft_thesis.thumbnail.blob.id))
+    end
+
+    draft_thesis.uuid = thesis.id
+    draft_thesis.save!
```
cwant: Is there a reason to keep the draft around? (In case they want to edit again?) Looks like this is called when the draft is converted into a Fedora object and the user is redirected to the show page.

murny: Yeah its for editing again. We add the Thesis's ID to the Draft Thesis uuid column and they are kept in sync like this. Then when a user edits a Item/Thesis, we look up to see if a draft version exists? If so use that otherwise create a new draft object.

(Also future proofs us a bit, maybe one day we get rid of Fedora's Thesis's and Draft Thesis's become the new Thesis object)

db/seeds.rb
murny: More likely was the Line on 413-416. Where you need Institutions seeded otherwise nothing shows up in the form

pgwillia: I also meant to say nice refactor for the item/thesis drafts.

app/controllers/admin/theses/draft_controller.rb
cwant: Does make sense. As somebody who hasn't looked at the stuff in a while, that wasn't obvious. Just a suggestion, but it might not hurt to add a comment, or rename to `already_authorized_via_admin?` or such (flipping the true/false).

murny: Good suggestion, can rename to make it better

app/controllers/concerns/draft_actions.rb
```diff
@@ -0,0 +1,146 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :initialize_communities, only: [:show]
+    before_action :set_draft, only: [:show, :update, :destroy]
+  end
+
+  def show
```
cwant: Seems that the `show` action is really an `edit` action. This is fine, but I wonder if it should have been called `edit` rather than `show`.

app/controllers/concerns/draft_actions.rb
```diff
@@ -0,0 +1,146 @@
...
+  def authorize?
+    true
+  end
+
+  def draft_class
+    DraftItem
+  end
+
+  def item_class
+    Item
```
cwant: It's pretty cool how you're able to do this, thanks to the abstract code in the concern. One thought: potentially you could move more code to the concern by deriving these private methods from `item_class` (e.g., construct the needed symbols using `item_class.name.underscore` or `item_class.model_name.singular`)

app/controllers/concerns/draft_actions.rb
murny: I'd agree but this comes from the wicked gem that we are using for the wizard steps and how they like it: https://github.com/schneems/wicked

No documentation on changing it to edit and probably wouldn't work. Would make urls a bit weird too (`localhost:3000/admin/theses/1/draft/describe_item/edit`, with `edit` at the end of everything)

murny:
Replying to https://github.com/ualbertalib/jupiter/pull/761#pullrequestreview-138861597

If you edit a thesis which belongs to a non restricted collection (AKA not FGSR or St. Stephen's), you will be forced to update the community/collection pair and choose either FGSR/St Stephens. This is why you got an error and couldn't proceed past the describe thesis step. 

In other environments we will need to seed these values. No different then what we do for other dropdowns like Types, etc. Currently they are static and probably won't change. But who knows?

Testing can be improved in general. Some of these I hardly touched at all and show general gaps in our testing (item controller has no controller test at all and stub file was recently deleted: https://github.com/ualbertalib/jupiter/pull/759/files/e289a2f3355f0506ba6d99b66c2f898e8ce2d0d7#r202116700) . Model tests could be improved, but its a trade off...these things that are missing are things touching fedora directly which are finicky to break and very slow. Plus most of these gaps are covered by system tests. Were still at 88% + coverage for `rails test` and 82% + for `rails test:system`. If these drop below 80% then absolutely we should revisit testing and add some more tests. Can also bring back coveralls (https://coveralls.io/) and fail builds to enforce this like we do on pushmi_pullyu

app/models/draft_thesis.rb
```diff
@@ -0,0 +1,167 @@
+class DraftThesis < ApplicationRecord
+
+  include DraftProperties
+
+  DEFAULT_RIGHTS = 'This thesis is made available by the University of Alberta Libraries'\
```
cwant: Should this go through `I18n`?

murny: Ooh good find. Absolutely should go through i18n

app/views/admin/theses/draft/_progress_bar.html.erb
```diff
@@ -0,0 +1,28 @@
+<% page_title(header) %>
```
cwant: Should setting the page title happen in this partial? Don't know if this is related to the bug you mentioned elsewhere in the conversation about this PR.

test/controllers/admin/theses/draft_controller_test.rb
```diff
@@ -0,0 +1,266 @@
...
+  test 'should be able to get to show page for a draft thesis' do
+    sign_in_as @admin
+
+    draft_thesis = draft_theses(:inactive)
+
+    get admin_thesis_draft_url(id: draft_thesis.wizard_step, thesis_id: draft_thesis.id)
+    assert_response :success
+  end
+
+  test 'should prevent prevent user from skipping ahead with uncompleted steps' do
```
cwant: Sounds like you really want to 'prevent' this :)

test/controllers/items/files_controller_test.rb
```diff
@@ -27,17 +27,14 @@ def before_all
   end
 
   test 'should be able to attach files to a draft item' do
-    assert_difference('ActiveStorage::Attachment.count', 1) do
-      post item_files_url(@draft_item), params: { file: @file_attachment }, xhr: true
-    end
+    # TODO: Why this failing?
```
cwant: Thought: related to file ingest now getting background-ed by the recent ActiveStorage work?

test/models/draft_thesis_test.rb
```diff
@@ -0,0 +1,214 @@
...
+      creator: 'Jane Doe',
+      description: 'Really random description about this random thesis',
+      graduation_term: 'Spring',
+      graduation_year: 2018,
+      member_of_paths: { community_id: [@community.id], collection_id: [@collection.id] },
+      visibility: DraftThesis.visibilities[:embargo],
+      rights: 'License text goes here'
+    )
+
+    assert_not draft_thesis.valid?
```
cwant: Other tests test the error message, but not this one -- don't know if that was intentional

test/controllers/items/files_controller_test.rb
murny: You might be onto something :thinking:

app/views/admin/theses/draft/_progress_bar.html.erb
murny: Basically how the title helper works is its a stack, where the titles get push at the end of the stack, aka (first in, first out). So you push titles to it, it creates an array [ 'decribe item', 'deposit item' , 'era'] and then it creates the `describe item | deposit item | era` for the page title.

So the step partials will add it's page title piece (aka the "Describe Thesis"/"Review and deposit thesis"/"Choose licence and visibility", ETC), and the H2 title with the step title.

This progress bar partial will add the generic page title, add the H1 with the generic page title and the progress bar.

(bug was the step partials were adding the generic title, instead of it's step title)

app/models/draft_thesis.rb
murny: This is done now. Plus the graduation term dropdown has also been `i18n`'d

test/models/draft_thesis_test.rb
murny: Fix these, adding the extra assertion for specific error message or a number of errors coming back

murny:
Hey @cwant or anyone, can you please give this another review when you get the chance?

- Fixed feedback from review:
  - Add better testing of error messages
  - fix double "prevent"
  - I18n a few missing things
  - renamed needs authorization method and add comment
  - derived a bunch of methods within concerns using `draft_class` and `item_class` instead
- Added Draft Theses to cleanup rake task
- Added Draft Theses to show under user profile or users show page

PR #781, opened 2018-07-20
Some fixes for UAT environment

murny:
- Add tmp/pids for worker container
- fix bug with UAT environment and default_url location
- log rails logs to stdout
- sync solr config into uat on predeploy
- upgrade ruby version of web image


.dockerignore
```diff
@@ -3,6 +3,7 @@
 *.md
 
 tmp
+!tmp/pids/.keep
```
murny: Need to keep this directory for sidekiq worker, else sidekiq worker doesn't start up as directory doesn't exist

.env_deployment_sample
```diff
@@ -11,6 +11,7 @@ ROLLBAR_TOKEN=
 GOOGLE_ANALYTICS_TOKEN=
 MATOMO_URL=
 MATOMO_SITE_ID=
+RAILS_LOG_TO_STDOUT=true
```
murny: When you do `docker-compose logs web` we are currently only showing the skylight logs, should be showing everything? Should be logging to standard out

config/environments/uat.rb
```diff
@@ -1,3 +1,5 @@
+Rails.application.routes.default_url_options = { host: 'uat.library.ualberta.ca', port: 80, protocol: 'http' }
```
murny: This was breaking all pages that show a thumbnail.

murny: Also be nice to get lets-encrypt in here and serve UAT with HTTPS instead... maybe one day

Dockerfile.deployment
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.3.4
+FROM ruby:2.5.0
```
murny: Using 2.3.4 actually breaks a few things in Jupiter. 

For example on search results, when filtering by year (start and end year), we use String's `match?` which is only available in ruby 2.4+.

bin/predeploy
```diff
@@ -14,11 +14,16 @@ readonly branch_name="master"
...
 
 mkdir -p "${install_directory}"
-mkdir -p "${install_directory}/config"
 curl -o "${install_directory}/deploy" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/deploy"
 curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.deployment.yml"
+
+mkdir -p "${install_directory}/config"
 curl -o "${install_directory}/config/nginx.conf" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/config/nginx.conf"
 
+mkdir -p "${install_directory}/solr/config"
```
murny: If starting from scratch, we need a way for the solr container to pick up the config files? How do these get synced over on the server? Assuming this was done manually before, lets add back to `predeploy` script

.env_deployment_sample
weiweishi: Are the logs kept in /var/logs on the UAT server? I think we had some issues with logs filling up the directory and never getting cleaned out - wondering if we could do some setup in the docker containers to rotate the log?

bin/predeploy
weiweishi: Makes sense

config/environments/uat.rb
weiweishi: Mind opening a ticket in the backlog for this? I'd like to see this happen in UAT. I think we have a ticket for UAT improvement. Could we add this there?

.env_deployment_sample
murny: Sounds like this would best be a change to docker-compose yml file? 

Apparently you can configure the logging driver and options from here.

Decent blog post about it: https://medium.freecodecamp.org/how-to-setup-log-rotation-for-a-docker-container-a508093912b2

murny: Created ticket here: https://github.com/ualbertalib/jupiter/issues/794 

To explore a solution for this

weiweishi:
That sounds good! thanks!


people."*

PR #782, opened 2018-07-20
Bump reported version for release 1.1.2

pgwillia:


app/models/jupiter_core.rb
```diff
@@ -14,5 +14,5 @@ class LockedInstanceError < StandardError; end
   VISIBILITIES = [VISIBILITY_PUBLIC, VISIBILITY_PRIVATE, VISIBILITY_AUTHENTICATED].freeze
 
   # This'll do for the time being I suppose ¯\_(ツ)_/¯
```
murny: Probably be better to move the version stuff out of here.

Move this out into a `lib/jupiter/version.rb` file or something better

murny:
Probably better to do these version bumbs when we actually cut the tag release. So you can reference the two in the git history

¯\_(ツ)_/¯

pgwillia:
Call this a fire-drill. Think this process needs more discussion in general.

murny:
100% agree :+1:

Issue #796, opened 2018-07-26
UAT Bug: Investigate how we can store attachments better across containers?


murny:
Currently in our docker setup for UAT, we run the rails app in the `web` container and sidekiq in a separate container called `worker`. Which is fine.

However the issue is how activestorage stores the images. It currently creates and stores them from the rails app within the `web` container under a folder called `storage`. The problem is when Sidekiq runs to process images into fedora and various other tasks, sidekiq is running in a separate container with no context of the `web` container or where the images are stored. This leads to pretty much every sidekiq job failing as a result with an error like this:

```
Error performing FileAttachmentIngestionJob from Sidekiq in 16.79ms: Errno::ENOENT (No such file or directory @ rb_sysopen - /app/storage/5R/Hz/5RHzJLfcZ4zh3zVxnraVa5VY)
```

Solution? I assuming the best way to fix this, would be to store the images in a volume on the server and map this volume between both the `web` and `worker` container. Which we are kinda already doing! We are doing this for the `web` container, but need to do this also for the `worker` container.

Very similar to how we do the `assets` for `nginx` and `web` container.

murny:
PR now in master. Closing out issue

PR #797, opened 2018-07-26
Map file-storage volume into worker container, so sidekiq can process images

murny:
#796 

Applied fix on UAT already and seems to be working now.

### DIFF
--- a/docker-compose.deployment.yml
+++ b/docker-compose.deployment.yml
@@ -52,6 +52,8 @@ services:
   worker: &rails
     restart: always
     image: ualbertalib/jupiter:deployment
+    volumes:
+      - file-storage:/app/storage/
     command: bundle exec sidekiq
     env_file: .env_deployment
     depends_on:


docker-compose.deployment.yml
```diff
@@ -52,6 +52,8 @@ services:
   worker: &rails
     restart: always
     image: ualbertalib/jupiter:deployment
+    volumes:
+      - file-storage:/app/storage/
```
weiweishi: So this will be the file storage for both thumbnails and actual files?

weiweishi: or assets is still where the thumbnail is stored?

murny: Yeah anything ActiveStorage consumes is stored in `/app/storage/`. So that be any thumbnails and potentially files if they are consumed by ActiveStorage.

weiweishi: :+1:

murny: `assets` volume is for any precompiled assets. So we precompile our assets and throw them in a shared volume with nginx. So Nginx can serve these assets directly (as this is what nginx excels at!) instead of the rails server

weiweishi:
Thanks for the clarification!
[image: Andy Samberg Thanks GIF]



people."*

PR #806, opened 2018-08-02
Relevance sorting

### LINKED ISSUE #693: Jupiter: The main search results should sort by relevance by default
It's unintuitive and doesn't make much sense that when you search for a specific title, it can be pages deep, because we currently sort by title.

Leah writes: 

> The user story is: "When I search ERA,  I expect my results to be sorted according to the returned item's relevance to the text I have submitted to the search form." or something like that. I agree we could look at how Sam has dealt with this stuff.

Collection pages and other pages displaying a faceted interface should continue to sort by title in the absence of a specific search parameter.

Needs consultation w/ Metadata re: if they want to track how specific fields are weighted.

mbarnett:
Adds relevance sorting as an optional sort field for faceted Solr queries. 

Relevance is now the *default* sort for /search if-and-only-if the user has included a query -- searching for "everything" (aka a blank query) results in an alphabetical sort by default (perhaps this should be record_created_at, but it isn't in Production, currently?). This was done because relevance sorting without a query term results in fairly non-sensical orderings.

`:relevance` is special-cased to always default to descending order when order is not explicitly over-riden, and only descending relevance order is added as an option in the user interface (there doesn't seem to be much of a use-case for "show me the least relevant things first").

Closes #693


Gemfile
```diff
@@ -112,6 +112,7 @@ group :test do
   gem 'vcr', require: false
   # Easy installation and use of chromedriver to run system tests with Chrome
   gem 'chromedriver-helper'
+  gem 'rails-controller-testing'
```
murny: Wondering if theres a better way to test this, than to bring in `rails-controller-testing` gem? Assert the html code coming back instead of using `assigns`? or you can ask for a json response and check that instead?  Do this via a system test, etc?

murny:
Rest of PR looks good 👍. Just be nice if we can test this without bringing in `rails-controller-testing` gem

Gemfile
mbarnett: We've got a bunch of tests in other places to verify that various results are on the page, but they all seem to have worse problems than just verifying the actual data?

eg) we do this in the system search test: https://github.com/ualbertalib/jupiter/blob/master/test/system/search_test.rb#L285

But it's actually kind of a bad test, because it makes sure that those things appear in that order, it doesn't establish that there isn't also a bunch of other results mixed in between those (and that's really hard to disprove with a regex-based test).

The other way we've tried doing it is by doing tests that assert nth-child presence for each result (eg. https://github.com/ualbertalib/jupiter/blob/master/test/system/communities_pagination_and_sort_test.rb#L63) but these are the tests that keep breaking randomly and I'm reluctant to double-down on a pattern that's been flapping on CI.

Long story short, I went this route because I'm starting to think that we're better off just looking directly at the actual data via `rails-controller-testing` rather than trying to sniff the HTML in all of these imprecise ways just to verify that we got the data we expected?

I'm open to other ideas but this seemed like a quick way forward.

murny: I think we can agree those system tests for search_test and communities pagination and sorting needs a lot of love and refactoring. Completely agree we can do this better and less brittle. 

All I am saying is I think you can achieve something like you are striving for without using `rails-controller-testing`. Usually when something is removed from rails core there is typically a good reason for it (usually... :neutral_face: ).  

I'd personally would like to keep a controller test for all of our controllers. Search currently doesn't have one, neither does items etc. So I like your approach and what your trying to do is good work.

Asserting the `@response.body` is pretty easy to do. If you just want to check if things are on the page (e.g. `assert_match item_id, @response.body`). However I'd agree this gets tricky when you want things to come back in a special order, such as this case. Would need fancy regex/xpath stuff which isn't ideal.

What if we just expose a json response though? This allows you do far better checking here without the magically regex. Plus makes sure your testing more things than just an assignment of a variable (`@posts = Post.all;  assert_equal Post.all, assigns(:posts);) For example:

```
  test 'should get results in alphabetical order when no query present' do
    get search_url
    assert_response :success
    json_response = JSON.parse(@response.body)
    assert json_response['data'].is_a?(Array)
     assert_equal json_response['data'].count,  Item.count
     assert_equal [@item1.id, @item4.id, @item2.id, @item5.id, @item3.id]], json_response['data'].map(&:id)
  end
```

But up to you.  If you still feel `assigns` is the best route, than i'll approve.

mbarnett: Yeah, that's a really good insight. Extending our support to JSON both makes this more testable and gives us a lot more flexibility on the front-end to move towards streaming in results. 👍 

Will make those changes shortly.

mbarnett:
Ok, I've removed `rails-controller-testing` and changed things around by adding JSON support to /search and using that for testing the ordering, as we discussed.

PR #810, opened 2018-08-03
Temporarily disable flapping tests

mbarnett:
These 3 tests continue to flap heavily on CI, which is getting in the way of cleaning up our pull-request backlog. 

I'm proposing that we temporarily skip the broken parts of these tests until we can finally get them working reliably, as the noise is detracting from the usability and reliability of the rest of our test suite.


mbarnett:
I've also added a ticket to track investigating the issue and restoring these tests: https://github.com/ualbertalib/jupiter/issues/811

mbarnett:
I think that one's already covered? Maybe I'm blind. There's a skip on line 73 of that file in this PR.

I think you're right about the wait interval, but we inserted a 5 second wait to "fix" it and it's slowly been getting worse again, so a pure timing fix doesn't seem reliable in the long term.

pgwillia:
That test failed at line # 25

mbarnett:
Whoops, good catch. I've bumped the skip forwards in this and the corresponding Thesis test.

PR #812, opened 2018-08-03
Missed one flapping test in previous PR

mbarnett:


mbarnett:
Agreed. I'm just going to bypass these two tests entirely for now, as it seems random at what point the "<ERROR><ERROR><ERROR>" issue crops up on Travis.

PR #841, opened 2018-08-31
Add migration task

mbarnett:
Restores pre-merge migration task, and adds changelog notes for this release.

### DIFF
--- a/lib/tasks/jupiter.rake
+++ b/lib/tasks/jupiter.rake
@@ -35,4 +35,39 @@ namespace :jupiter do
     (Item.all + Thesis.all).each { |item| item.unlock_and_fetch_ldp_object(&:preserve) }
     puts 'All Items and Theses have been added to preservation queue!'
   end
+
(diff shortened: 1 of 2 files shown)


lib/tasks/jupiter.rake
```diff
@@ -35,4 +35,39 @@ namespace :jupiter do
...
+    puts "Found #{orphan_blobs.count} orphans. Purging..."
+    orphan_blobs.each do |blob|
+      blob.purge
+      print '.'
+    end
+    puts
+    puts 'done!'
+  end
+
+  desc 'turn existing filesets into attachments'
```
murny: Just to clarify, this will be a one time thing right? Once we migrate everything from filesets to attachments (via running this task), chances are we will never use this again?

mbarnett: It should be, yeah. I suppose that theoretically starting from only a backup of Fedora you could recreate both Solr and the data on the gluster (assuming it had all successfully been ingested in the background) using this for the latter and `rake jupiter:recover` for the former, though.

PR #860, opened 2018-09-14
Improve fileset migration task

mbarnett:


mbarnett:
Assigning the first item in the fileset as the thumbnail is already in there.

PR #866, opened 2018-09-18
Update downloads_controller to serve file view properly. 

weiweishi:
Update the downloads_controller to:
- use disposition: 'inline' for file viewing
- use type instead of content-type when sending data
- use find_by instead of where.first to fetch file in the fileset. 

Updated how urls and filename are handled with regular Rails/ActiveStorage way. And added more tests for the download controllers.


app/helpers/items_helper.rb
```diff
@@ -1,12 +1,4 @@
 module ItemsHelper
-  def file_view_url(file)
```
murny: We don't need all these custom helpers? Just use rails routes properly

app/jobs/file_attachment_ingestion_job.rb
```diff
@@ -11,7 +11,7 @@ def perform(attachment_id)
     item = attachment.record.owner
 
     path = ActiveStorage::Blob.service.send(:path_for, attachment.blob.key)
-    original_filename = attachment.blob.filename.to_s
+    original_filename = attachment.filename.to_s
```
murny: Don't need to walk the chain of the blob object to get filename, can get it right from the activestorage object

test/integration/site_for_bots_test.rb
```diff
@@ -143,11 +142,10 @@ def before_all
     end
     assert_select "meta[name='citation_publication_date'][content='#{@item.creation_date}']"
     assert_select format("meta[name='citation_pdf_url'][content='%s']",
-                         Rails.application.routes.url_helpers.url_for(
-                           controller: :downloads, action: :view, id: @item.id,
+                         file_view_item_url(
+                           id: @item.id,
                            file_set_id: @item.files.first.fileset_uuid,
-                           file_name: @item.files.first.filename,
-                           only_path: true
```
murny: This was a bug here. We need the absolute url not just the path (relative url)!

From here: https://scholar.google.ca/intl/en/scholar/inclusion.html#indexing

> The content of the tag is the absolute URL of the PDF file; for security reasons, it must refer to a file in the same subdirectory as the HTML abstract.

weiweishi:
Assume this is no longer WIP. I am looking into the manual testing right now.

weiweishi:
Validated all the file download/view links, filename saved in fedora etc, all looks good to me. Now sure if there's an easy way for me to manually test the redirect urls.

murny:
Yeah had hard time testing the redirects as well since they expect some weird formats and kept failing on validation errors  (`/public/view/item/uuid:{{uuid}}/{{datastream? random numbers}}/:filename`)

So not sure best way to verify this either /shrug

test/controllers/downloads_controller_test.rb
```diff
@@ -0,0 +1,44 @@
+require 'test_helper'
+
+class DownloadsControllerTest < ActionDispatch::IntegrationTest
```
murny: New tests for testing the downloads controller, view and download actions. Making sure they file is being streamed back with the proper headers and content

Issue #811, opened 2018-08-03
Jupiter: Fix flapping tests


mbarnett:
Three tests are continually flapping on Travis CI (although they seem reliable locally). The flapping portions were disabled in these PRs https://github.com/ualbertalib/jupiter/pull/810/files and https://github.com/ualbertalib/jupiter/pull/812/files, but we should fix this ASAP to restore the coverage for the portions of the tests currently being skipped.

pgwillia:
1987 / 2739 LOC (72.54%) covered now vs 2183 / 2739 LOC (79.7%) covered recently.
Specifically the deposit wizard has lost out on coverage.

weiweishi:
We will timebox it for 2 days for Sprint 4.

murny:
Another flapping test that's been cropping up a few times currently in travisCI is this one:
https://travis-ci.org/ualbertalib/jupiter/builds/428341082
```

Failure:
SearchControllerTest#test_should_get_results_in_relevance_order_when_a_query_is_present [/home/travis/build/ualbertalib/jupiter/test/controllers/search_controller_test.rb:90]:
--- expected
+++ actual
@@ -1 +1 @@
-["8aac2982-8a18-416c-b4b7-6b7a60d238b2", "d26e9b4e-26ee-4994-bfcc-ad34b2ca33a6", "bc044ae5-3a1c-4d39-a459-52ae1cda38da"]
+["8aac2982-8a18-416c-b4b7-6b7a60d238b2", "bc044ae5-3a1c-4d39-a459-52ae1cda38da", "d26e9b4e-26ee-4994-bfcc-ad34b2ca33a6"]
bin/rails test test/controllers/search_controller_test.rb:86
```
Essentially the array isn't coming back in the correct order :thinking:

mbarnett:
But only sometimes? Maybe relevance sort isn't stable in solr...

murny:
Yeah only sometimes, its flapped on last couple dependency upgrades...rerunning the build fixes it.

pgwillia:
Replying to https://github.com/ualbertalib/jupiter/issues/811#issuecomment-421146853

The problem is that the scores of those two documents are identical.
```
"response":{"numFound":3,"start":0,"maxScore":0.81081206,"docs":[
      {
        "id":"4b9ef927-99db-403b-a751-5b3e335346b1",
        "score":0.81081206},
      {
        "id":"462facb0-edf6-4bac-92f4-ca84ebb335b4",
[...]

PR #875, opened 2018-09-25
Small improvements for flapping tests

pgwillia:
* fixed one flapping test by changing the score
* only skip flapping tests on travis so that we still run the tests locally

#811

### DIFF
--- a/test/system/collections_pagination_and_sort_test.rb
+++ b/test/system/collections_pagination_and_sort_test.rb
@@ -14,7 +14,7 @@ class CollectionsPaginationAndSortTest < ApplicationSystemTestCase
   end
 
   test 'anybody should be able to sort and paginate collections' do
-    skip 'The rest of this test continues to flap on CI for unknown reasons that should be investigated ASAP'
+    skip 'This test continues to flap on CI for unknown reasons that should be investigated ASAP' if ENV['TRAVIS']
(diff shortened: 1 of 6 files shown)


test/system/collections_pagination_and_sort_test.rb
```diff
@@ -14,7 +14,7 @@ def before_all
   end
 
   test 'anybody should be able to sort and paginate collections' do
-    skip 'The rest of this test continues to flap on CI for unknown reasons that should be investigated ASAP'
+    skip 'This test continues to flap on CI for unknown reasons that should be investigated ASAP' if ENV['TRAVIS']
```
weiweishi: Like how we only skip this test on Travis - local run seems to be okay.

test/system/search_test.rb
```diff
@@ -245,7 +245,8 @@ def before_all
     click_button 'Search'
     assert_equal URI.parse(current_url).request_uri, search_path(search: 'Fancy')
 
-    # Default sort is by title
+    # Default sort is by relevance
+    # TODO this test has flapped in the past because the score of each document is equal
```
weiweishi: Would you be changing this test this round? Or this is a to-do for future note?

pgwillia: It's a todo for future note.

Issue #888, opened 2018-10-11
[Jupiter] Debug Garbage Collect Job


weiweishi:
**Describe the bug**
We have seen error messages https://rollbar.com/ualbertalib/jupiter/items/369/ regarding the permissions to remove files from era-assets. 

We have also experienced very high CPU usage on postgreSQL
https://helpdesk.library.ualberta.ca/otrs/index.pl?Action=AgentTicketZoom;TicketID=47896

that may or may not be related. 

Additional things to consider:
- maybe that job needs to watch the clock when it runs?
- limit the number of entries processed?
- self-throttle so as not to eat all the I/O on the DB server?
- better use of config/scheduler.yml, as it runs on all three application servers, and really it only needs to be run once. (adding an env variable as the condition to load the schedule.yml in config/initializer/sidekiq.rb, and let the env variable to be controlled by ansible, to determine which app server is responsible for these scheduled jobs, some ideas: https://github.com/moove-it/sidekiq-scheduler/issues/88)
- performance test/tune the MySQL query

weiweishi:
The scheduled cron job will be temporarily stopped before this issue can be fixed - and we can run the manual rake task to remove the orphaned assets.

PR #890, opened 2018-10-15
888 tune sql for garbage collect active storage

weiweishi:
To tune the SQL query to fetch the records that are orphaned in activestorage, and allow env var to control if scheduled jobs need to be run (so that ansible playbook can set the one app server where the job needs to be run, instead of running concurrently on 3 app servers. )

### DIFF
--- a/app/jobs/garbage_collect_blobs_job.rb
+++ b/app/jobs/garbage_collect_blobs_job.rb
@@ -10,8 +10,9 @@ class GarbageCollectBlobsJob
   # See the note on: https://github.com/rails/rails/blob/master/activestorage/app/models/active_storage/attachment.rb#L5-L8
 
   def perform
-    orphan_query = 'SELECT * FROM active_storage_blobs asb WHERE asb.id NOT IN '\
-                   '(SELECT distinct blob_id FROM active_storage_attachments)'
+    orphan_query = 'SELECT * FROM active_storage_blobs WHERE active_storage_blobs.id IN '\
+                   '( SELECT id FROM active_storage_blobs asb EXCEPT SELECT DISTINCT blob_id '\
+                   'FROM active_storage_attachments asa INNER JOIN active_storage_blobs asb ON asb.id=asa.blob_id)'
     orphan_blobs = ActiveStorage::Blob.find_by_sql(orphan_query)
     orphan_blobs.each(&:purge)
   end
--- a/app/jobs/garbage_collect_blobs_job.rb
+++ b/app/jobs/garbage_collect_blobs_job.rb
@@ -9,10 +9,17 @@ class GarbageCollectBlobsJob
   # Therefore we need a process for removing blobs that become orphaned over time as all of their associated attachments cease to exist.
   # See the note on: https://github.com/rails/rails/blob/master/activestorage/app/models/active_storage/attachment.rb#L5-L8
 
+  # This query is to get any ActiveStorage::Blob ids that are not in the ActiveStorage::Attachment table - meaning that the blob was orphaned as there is no more association found within the attachment table.
+  # This is equivalent to:
+  # ActiveStorage::Blob.where([
+  # "id NOT IN (?)",
+  # ActiveStorage::Attachment.pluck("blob_id")
+  # )]
+  # but faster.
+
   def perform
-    orphan_query = 'SELECT * FROM active_storage_blobs WHERE active_storage_blobs.id IN '\
-                   '( SELECT id FROM active_storage_blobs asb EXCEPT SELECT DISTINCT blob_id '\
-                   'FROM active_storage_attachments asa INNER JOIN active_storage_blobs asb ON asb.id=asa.blob_id)'
+    orphan_query = 'SELECT id FROM active_storage_blobs asb EXCEPT SELECT DISTINCT blob_id '\
+                   'FROM active_storage_attachments asa INNER JOIN active_storage_blobs asb ON asb.id=asa.blob_id'
     orphan_blobs = ActiveStorage::Blob.find_by_sql(orphan_query)
     orphan_blobs.each(&:purge)
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and releases in Jupiter project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Fixed
+- fixes typo for conference paper item type [#879](https://github.com/ualbertalib/jupiter/issues/879)
 
 ## [1.2.4] - 2018-10-09
 ### Changed
--- a/app/models/draft_item.rb
+++ b/app/models/draft_item.rb
@@ -54,7 +54,7 @@ class DraftItem < ApplicationRecord
   ITEM_TYPE_TO_URI_CODE = { book: :book,
                             book_chapter: :chapter,
                             conference_workshop_poster: :conference_workshop_poster,
-                            conference_workshop_presenation: :conference_workshop_presenation,
+                            conference_workshop_presentation: :conference_workshop_presentation,
                             dataset: :dataset,
                             image: :image,
                             journal_article_draft: :article,
--- a/app/services/doi_service.rb
+++ b/app/services/doi_service.rb
@@ -6,7 +6,7 @@ class DOIService
     article_submitted: 'Text/Submitted Journal Article',
     book: 'Text/Book',
     chapter: 'Text/Chapter',
-    conference_workshop_presenation: 'Other/Presentation',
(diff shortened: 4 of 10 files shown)


app/jobs/garbage_collect_blobs_job.rb
```diff
@@ -10,8 +10,9 @@ class GarbageCollectBlobsJob
   # See the note on: https://github.com/rails/rails/blob/master/activestorage/app/models/active_storage/attachment.rb#L5-L8
 
   def perform
-    orphan_query = 'SELECT * FROM active_storage_blobs asb WHERE asb.id NOT IN '\
-                   '(SELECT distinct blob_id FROM active_storage_attachments)'
+    orphan_query = 'SELECT * FROM active_storage_blobs WHERE active_storage_blobs.id IN '\
```
murny: Is it possible to get an explanation of the difference here? Maybe even a code comment?

Previously we wanted to get any blobs ids not in the attachment table (meaning the blob was orphaned as there is no more association found within the attachment table).

The new query I am not quite sure what's different with it or what has changed?





Also an aside it be nice if we can keep this within ActiveRecord and resist the temptation of having to resort to raw SQL. For example the previous query could have been written in AR like so:

```ruby
orphan_query = 'SELECT * FROM active_storage_blobs asb WHERE asb.id NOT IN '\
                   '(SELECT distinct blob_id FROM active_storage_attachments)'
orphan_blobs = ActiveStorage::Blob.find_by_sql(orphan_query)
```

```ruby
orphan_blobs = ActiveStorage::Blob.includes(:active_storage_attachments).
      references(:active_storage_attachments).
      where.not(active_storage_attachments: { blob_id: active_storage_blobs.id })
```

(code not tested, but the idea is there, can google how to do joins/etc using activerecord)

weiweishi: The original query takes 25+mins to complete on the production server, with over 130,000 enties in the db. And it has become a significant performance issue especially it runs concurrently on all three app servers. 

So Neil uses the sql's explain query to analyze the original query. And the new query is supposed to do the exactly same, but instead of selecting all fields (when doing a huge venn diagram with itself), only asking for the 'id' field, and query for all the fields after the ids are returned with this:
```
select * from active_storage_blobs where active_storage_blobs.id in ( SELECT id FROM active_storage_blobs asb except SELECT distinct blob_id FROM active_storage_attachments asa  inner join active_storage_blobs asb  on asb.id=asa.blob_id)
```
The analyze from sql explain is:

```
jupitertest=# explain analyze  SELECT id FROM active_storage_blobs asb except SELECT distinct blob_id FROM active_storage_attachments asa  inner join active_storage_blobs asb  on asb.id=asa.blob_id ;
                                                                                                           QUERY PLAN                                                                                              
              
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------
 SetOp Except  (cost=34182.86..35110.91 rows=92805 width=8) (actual time=221.043..221.043 rows=0 loops=1)
   ->  Sort  (cost=34182.86..34646.88 rows=185610 width=8) (actual time=188.598..207.806 rows=185758 loops=1)
         Sort Key: "*SELECT* 1".id
         Sort Method: external merge  Disk: 3984kB
         ->  Append  (cost=0.00..15399.21 rows=185610 width=8) (actual time=0.002..122.287 rows=185758 loops=1)
               ->  Subquery Scan on "*SELECT* 1"  (cost=0.00..4036.10 rows=92805 width=8) (actual time=0.002..19.307 rows=92879 loops=1)
                     ->  Seq Scan on active_storage_blobs asb  (cost=0.00..3108.05 rows=92805 width=8) (actual time=0.002..12.038 rows=92879 loops=1)
               ->  Subquery Scan on "*SELECT* 2"  (cost=0.05..11363.11 rows=92805 width=8) (actual time=9.654..92.354 rows=92879 loops=1)
                     ->  Unique  (cost=0.05..10435.06 rows=92805 width=8) (actual time=9.653..84.525 rows=92879 loops=1)
                           ->  Merge Join  (cost=0.05..10203.04 rows=92805 width=8) (actual time=9.651..74.407 rows=92887 loops=1)
                                 Merge Cond: (asa.blob_id = asb.id)
                                 ->  Index Only Scan using index_active_storage_attachments_on_blob_id on active_storage_attachments asa  (cost=0.00..4350.91 rows=92807 width=8) (actual time=0.013..10.462 rows=9
2888 loops=1)
                                       Heap Fetches: 34775
                                 ->  Index Only Scan using active_storage_blobs_pkey on active_storage_blobs asb  (cost=0.00..4495.15 rows=92805 width=8) (actual time=9.631..42.392 rows=92886 loops=1)
                                       Heap Fetches: 30478
 Total runtime: 221.815 ms
(16 rows)
```

This new query takes 221 ms to complete instead of 5 mins on staging, and takes 800+ms to complete instead of 21 mins on production. 

There is probably better ways to handle this without the rough mysql and we can certainly test it with the full production records to make sure it is as effective. The main issue right now is to prevent this monsterous query to bring down our production db server.

murny: Some random comments for this (https://github.com/ualbertalib/jupiter/issues/888):

-  `And it has become a significant performance issue especially it runs concurrently on all three app servers.`

Does this mean the job runs 3 times every 12 hours when it gets called? Once on each app server? The entire "scheduling" part of sidekiq is from a sidekiq addon called `sidekiq-cron`, it should work with multiple processes. And should only run once. More info here:
https://github.com/ondrejbartas/sidekiq-cron#under-the-hood
Specifically this part:

> Sidekiq-Cron is safe to use with multiple sidekiq processes or nodes. It uses a Redis sorted set to determine that only the first process who asks can enqueue scheduled jobs into the queue.

So don't think this is the issue? Maybe something else is causing it to run multiple times? Like it failing and retrying and retrying etc etc. So maybe the config `Rails.application.secrets.run_scheduled_jobs` isn't needed?

- We should maybe scale back this job? Instead of running it every 12 hours, run it once a day at  a late hour? Like after midnight or something? Should be sufficient enough to run this once a day or even longer? (once a week etc?) As I assume the amount of orphans would be small as this only happens when a DraftItem/Item/Thesis/etc gets nuked?

-  From looking at the Sidekiq UI, there is some concerns. This job is failing a lot and resulting in a ton of retries: 
image
Looks like mostly permission errors? As a result this job never actually succeeds and instead is being reran almost every hour or more.

- We need to be careful with the scheduled time. These cron times are in UTC time zone! So 10 PM is actually 4 AM, and 12 hours on each day is actually 6 AM and 6 PM (nothing serious as all these hours are outside of our normal business hours but something to think about).

- Tuning this query is good work! But needs a code comment or two so the next person can know what this is doing? as its not exactly clear what this is doing and probably could be simplified still. For example probably can just take the raw array of ID's back from the second query (one in the brackets) and purge using this in activerecord, instead of running another `select *` to get this as objects again. Makes this easier to read. (Also Again this could be written in ActiveRecord. There is something called `.pluck` that does exactly this (instead of a `select *` it can "pluck" the field you want (like `select id`) which you can then loop through the array of Id's and purge accordingly).

weiweishi: Great points @murny. And helped me learn about more about Sidekiq.(not aware of the admin dashboard, that's very helpful)

I think you are absolutely right, that the issue is with the retries. We've seen the error message of "Permission Denied" in Rollbar and will be our next step to troubleshoot (was thinking that we need to stop the bleeding asap). That explains why we didn't see the performance issue right after the release launch, and just started to see it last week. The jobs accumulate, and keep retrying every hour.  

I agree that we should scale down the scheduling of this job. This doesn't seem to occur often. I will change it to run every week. and given the number of objects we get from production (1 since the release), I think it's probably good enough. 

Will also add some notes around the mysql query. Will test the ActiveRecord approach and monitor the performance.

weiweishi: This works with native ActiveRecord method
```
ActiveStorage::Blob.where([
 "id NOT IN (?)",
ActiveStorage::Attachment.pluck("blob_id")
)]
```
Gets the correct object back in the array:
```
#<ActiveRecord::Relation [#<ActiveStorage::Blob id: 43924, key: "CHVtk88TTyrLQBssg5gPpHaC", filename: "Radler_Karola_2011_MTSHon.pdf.jpg", content_type: "image/jpeg", metadata: {}, byte_size: 10915, checksum: "Rj6U1J6d4ZWaUSZ9ublQ5A==", created_at: "2018-04-11 15:46:38">]>
```
However, the operation takes 1 mins 26 secs to complete on production data. 

While the operation with the SQL query takes less than 1 second to complete. 
```
SELECT id FROM active_storage_blobs asb except SELECT distinct blob_id FROM active_storage_attachments asa  inner join active_storage_blobs asb  on asb.id=asa.blob_id
```
I think the SQL route seems more efficient when dealing with big amount of data?

PR #907, opened 2018-10-31
can delete additional contributors and similar fields

### LINKED ISSUE #830: [Jupiter] Additional Contributors field edits do not save
My first bug! Be kind.

User-reported bug.

In ERA, edits made to the Additional Contributors field on the item deposit form do not persist.

I.E. - user entered names in the Additional Contributors field and decided afterwards to delete them. The names would reappear when the changes were saved, however, and could not be permanently deleted. 

E.G.
https://doi.org/10.7939/R3FX74C3W

You can also search by author "Jonas Valloton" for more examples.

This field should not be a mandatory text field.

Anna

pgwillia:
closes #830

### DIFF
--- a/app/models/draft_item.rb
+++ b/app/models/draft_item.rb
@@ -71,7 +71,7 @@ class DraftItem < ApplicationRecord
   has_many :draft_items_languages, dependent: :destroy
   has_many :languages, through: :draft_items_languages
 
-  before_validation :strip_input_fields, on: :update
+  before_validation :strip_input_fields
 
   validates :title, :type, :languages,
             :creators, :subjects, :date_created,
--- a/app/models/draft_item.rb
+++ b/app/models/draft_item.rb
@@ -296,7 +296,7 @@ class DraftItem < ApplicationRecord
   def strip_input_fields
     attributes.each do |key, value|
       self[key] = value.reject(&:blank?) if value.is_a?(Array)
-      self[key].present? || self[key] = nil
+      self[key] = nil if self[key].blank?
     end
   end
 
--- a/app/models/draft_item.rb
+++ b/app/models/draft_item.rb
@@ -71,6 +71,8 @@ class DraftItem < ApplicationRecord
   has_many :draft_items_languages, dependent: :destroy
   has_many :languages, through: :draft_items_languages
 
+  before_validation :strip_input_fields, on: :update
+
   validates :title, :type, :languages,
             :creators, :subjects, :date_created,
             :member_of_paths,
@@ -289,6 +291,13 @@ class DraftItem < ApplicationRecord
 
   def license_not_unselected
     errors.add(:license, :missing) if license == 'unselected'
+  end
+
+  def strip_input_fields
+    attributes.each do |key, value|
+      self[key] = value.reject(&:blank?) if value.is_a?(Array)
+      self[key].present? || self[key] = nil
+    end
   end
 
 end
--- a/app/views/items/draft/describe_item.html.erb
+++ b/app/views/items/draft/describe_item.html.erb
@@ -52,7 +52,6 @@
     <%= f.input :creators,
         label: t('.creators'),
         collection: @draft.creators || [],
-        include_hidden: false, # TODO: figure out this, need to parse empty strings out in model
         input_html: {
           multiple: true,
           data: { placeholder: t('.creators_placeholder') },
@@ -65,7 +64,6 @@
     <%= f.input :subjects,
         label: t('.subjects'),
         collection: @draft.subjects || [],
-        include_hidden: false,
         input_html: {
           multiple: true,
           data: { placeholder: t('.subjects_placeholder') },
@@ -155,7 +153,6 @@
             <%= f.input :contributors,
                 label: t('.contributors'),
                 collection: @draft.contributors || [],
-                include_hidden: false,
                 input_html: {
                   multiple: true,
                   data: { placeholder: t('.contributors_placeholder') },
@@ -167,7 +164,6 @@
             <%= f.input :places,
                 label: t('.places'),
                 collection: @draft.places || [],
-                include_hidden: false,
                 input_html: {
                   multiple: true,
                   data: { placeholder: t('.places_placeholder') },
@@ -179,7 +175,6 @@
             <%= f.input :time_periods,
                 label: t('.time_periods'),
                 collection: @draft.time_periods || [],
-                include_hidden: false,
                 input_html: {
                   multiple: true,
                   data: { placeholder: t('.time_periods_placeholder') },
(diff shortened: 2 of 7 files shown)


app/models/draft_item.rb
```diff
@@ -291,4 +293,11 @@ def license_not_unselected
     errors.add(:license, :missing) if license == 'unselected'
   end
 
+  def strip_input_fields
+    attributes.each do |key, value|
+      self[key] = value.reject(&:blank?) if value.is_a?(Array)
+      self[key].present? || self[key] = nil
```
pgwillia: Is there a preference between this style and
`self[key] = nil unless self[key].present?`

app/views/items/draft/describe_item.html.erb
```diff
@@ -52,7 +52,6 @@
     <%= f.input :creators,
         label: t('.creators'),
         collection: @draft.creators || [],
-        include_hidden: false, # TODO: figure out this, need to parse empty strings out in model
```
pgwillia: Figured this out! Use a `before_validation` to parse empty strings out of model

test/models/draft_item_test.rb
```diff
@@ -221,4 +221,31 @@ def before_all
     assert draft_item.valid?
   end
 
+  # https://github.com/ualbertalib/jupiter/issues/830
+  test 'should be able to delete additional contributors' do
```
pgwillia: I tested for additional contributors but I also made the same change to subjects, places, time_periods and citations.  Do these need tests also?

test/system/item_edit_test.rb
```diff
@@ -0,0 +1,64 @@
...
+      File.open(Rails.root + 'app/assets/images/era-logo.png', 'r') do |file|
+        item.add_and_ingest_files([file])
+      end
+    end
+
+    login_user user
+
+    visit item_path item
+
+    click_on I18n.t('application.navbar.links.login') # TODO: not sure why the login didn't persist
```
pgwillia: Couldn't figure this out. Have to login again, but not sure why line 32 didn't do that for me.

test/system/item_edit_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    click_on I18n.t('edit')
+
+    assert_selector 'h1', text: I18n.t('items.draft.header_edit')
+    click_on I18n.t('items.draft.describe_item.additional_fields')
+
+    # clear ALL additional contributors
```
pgwillia: https://github.com/ualbertalib/jupiter/blob/master/test/system/deposit_item_test.rb#L124-L129 informed this.  Was trying to make this a private method but couldn't figure out how to remove one at a time with the selectize code.

test/system/item_edit_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    Sidekiq::Testing.inline! do
+      File.open(Rails.root + 'app/assets/images/era-logo.png', 'r') do |file|
+        item.add_and_ingest_files([file])
+      end
+    end
+
+    login_user user
+
+    visit item_path item
```
pgwillia: Maybe I should `assert_selector 'h6', text: 'Additional contributors'`

app/models/draft_item.rb
murny: Although they are similar in this case they are not actually the same?

`self[key].present? || self[key] = nil` 

Will either return `true` or `nil` depending on weather `self[key].present?` is true or false (which can have some weird side effects)

Where as, 
`self[key] = nil unless self[key].present?`

Always returns nil (if `self[key].present` is true, then nothing happens (aka nil),  otherwise its returning the value of the assignment which happens to be nil)

So probably the second style is more true to what your actually trying to achieve here.

pgwillia: Thanks.  I'll make this correction.

test/models/draft_item_test.rb
murny: Probably could make this test a bit more generic and test a couple more fields (don't need to test them all)? test "#strip_input_fields should strip empty strings from array fields" or something along those lines

test/models/draft_item_test.rb
```diff
@@ -221,4 +221,31 @@ def before_all
...
+      date_created: Date.current,
+      description: 'Really random description about this random book',
+      member_of_paths: { community_id: [@community.id], collection_id: [@collection.id] },
+      contributors: ['Carly Brown']
+    )
+
+    draft_item.assign_attributes(
+      contributors: ['']
+    )
+    draft_item.save
```
murny: Shouldn't need to save this item? Since this is on the `before_validation` method, this should work without having to go all the way to the database (slows down tests)

pgwillia: The validations weren't running unless I did a save and the test failed.  Was there something I missed?

test/system/item_edit_test.rb
murny: Hmm that's very weird... :thinking:

test/models/draft_item_test.rb
murny: `draft_item.valid?` should run the validations.

Take a peak at some of the test near the top of this file, like this one:

```
 test 'should run validations when on choose_license_and_visibility wizard step' do
    user = users(:regular)

    draft_item = DraftItem.new(
      user: user,
      status: DraftItem.statuses[:active],
      wizard_step: DraftItem.wizard_steps[:choose_license_and_visibility],
      title: 'Book of Random',
      type: types(:book),
      languages: [languages(:english)],
      creators: ['Jane Doe', 'Bob Smith'],
      subjects: ['Best Seller', 'Adventure'],
      date_created: Date.current,
      description: 'Really random description about this random book',
      member_of_paths: { community_id: [@community.id], collection_id: [@collection.id] },
      # technically a draft_item is valid in this wizard step already, since license/visibility are given defaults
      # but this case let's force the validation to fail
      license: nil,
      visibility: nil
    )

    assert_not draft_item.valid?

    draft_item.assign_attributes(
      license: DraftItem.licenses[:attribution_non_commercial],
      visibility: DraftItem.visibilities[:open_access]
    )

    assert draft_item.valid?
  end
```

Can mess around with the data and keep calling `valid?` to see if proper validations are being run/etc.

murny: Ah it's probably cause we have the `before_validation` being run only on `:update`? So the object needs to exist in the database.

So if that's the case, maybe grab on of the fixtures draft items as a starting base? etc?

test/system/item_edit_test.rb
pgwillia: The login_user brings you to the root_path, navigating to that item page still has the same issue of not being logged in.

murny: Alright no worries, we can debug another day. At least its got a `TODO` comment on it :+1:

PR #923, opened 2018-11-13
Use ez.test.datacite.org for testing of doi minting code

### LINKED ISSUE #911: [Jupiter] Test Datacite DOI API
We now have a test account with Datacite, so we should test Jupiter's EZID-based DOI management against Datacite's emulation of the EZID API. When the tests are successful we can unblock #711.

It looks like [doi_service_test.rb](https://github.com/ualbertalib/jupiter/blob/c4901e866d9753f4a0f51aa45efd20f254bb0bb5/test/services/doi_service_test.rb) does all the tests we need: minting, updating, update unavailable, removal.

pgwillia:
closes #911 
closes #861
- upgrades ezid-client gem which includes Datacite compatibility
- `Ezid::Client.configure.host` points at 'ez.test.datacite.org'
- added `datacite_resourcetypegeneral` to accomodate required field for minting
- removed `datacite_resourcetype*` from updates (causes error)
- removed `_export` from identifiers being made unavailable. Datacite doesn't really have this concept.
- add comments with TODOs when fixes available https://github.com/datacite/cheetoh/pull/21 and ez.test.datacite.org

TODO:
- [x] why does update complain about 'Ezid::Error: :resource_type_general' as an unpermitted param?
- [x] why is `_status` 'unavailable' duplicated
- [x] why is `_export` change not respected

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -106,6 +106,7 @@ GEM
       sass (>= 3.5.2)
     brakeman (4.3.1)
     builder (3.2.3)
+    byebug (10.0.2)
     canonical-rails (0.2.4)
       rails (>= 4.1, < 5.3)
     capybara (3.11.1)
@@ -142,8 +143,9 @@ GEM
     et-orbi (1.1.4)
       tzinfo
     execjs (2.7.0)
-    ezid-client (1.7.1)
+    ezid-client (1.8.0)
       hashie (~> 3.4, >= 3.4.3)
+      nokogiri
     faker (1.9.1)
       i18n (>= 0.7)
     faraday (0.12.2)
@@ -274,6 +276,9 @@ GEM
     pry (0.12.2)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
+    pry-byebug (3.6.0)
+      byebug (~> 10.0)
+      pry (~> 0.10)
     pry-rails (0.3.7)
       pry (>= 0.10.4)
     public_suffix (3.0.3)
@@ -467,7 +472,7 @@ DEPENDENCIES
   clamby
   connection_pool
   dropzonejs-rails
-  ezid-client
+  ezid-client (~> 1.8.0)
   faker
   font-awesome-rails
   haikunator
@@ -485,6 +490,7 @@ DEPENDENCIES
   omniauth-saml
   pg (~> 1.1.3)
   pry
+  pry-byebug
   pry-rails
   puma (~> 3.12)
   pundit
--- a/app/services/doi_service.rb
+++ b/app/services/doi_service.rb
@@ -51,7 +51,12 @@ class DOIService
   def update
     return unless @item.doi_state.awaiting_update?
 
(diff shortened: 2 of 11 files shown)


config/initializers/ezid_client.rb
```diff
@@ -1,4 +1,5 @@
 Ezid::Client.configure do |config|
+  config.host = 'ez.test.datacite.org'
```
pgwillia: Should be ez.datacite.org for production!

app/services/doi_service.rb
```diff
@@ -87,11 +94,11 @@ def ezid_metadata
       datacite_publisher: PUBLISHER,
       datacite_publicationyear: @item.sort_year.presence || '(:unav)',
       datacite_resourcetype: DATACITE_METADATA_SCHEME[@item.item_type_with_status_code],
+      datacite_resourcetypegeneral: DATACITE_METADATA_SCHEME[@item.item_type_with_status_code].split('/').first,
       datacite_title:  @item.title,
       target: Rails.application.routes.url_helpers.item_url(id: @item.id),
       # Can only set status if been minted previously, else its public
-      status: @item.private? && @item.doi.present? ? UNAVAILABLE_MESSAGE : Ezid::Status::PUBLIC,
-      export: @item.private? ? 'no' : 'yes'
```
pgwillia: Removing `_export`  because there is no real analog in Datacite.

config/initializers/ezid_client.rb
pbinkley: Any reason why this is hard-coded here, rather than set in secrets or elsewhere in the config?

config/secrets.yml
```diff
@@ -34,8 +34,8 @@ shared:
 development:
   secret_key_base: c0a4bf2c5890d0fa86e1459dd189bc4c5a02f412067b610490885c8c312bf0cea5d988e075761ba7277a8291041c1b2e7cf6c373d4d6f43d4522bc48db76cc1a
 
-  ezid_default_shoulder: doi:10.5072/FK2
-  ezid_user: apitest
+  ezid_default_shoulder: doi:10.21967/FK2
+  ezid_user: CISTI.UAL
```
murny: Should the password be kept private? Asking because we are exposing it in this PR (i'll add message in slack for details about where/how)

Gemfile.lock
```diff
@@ -142,8 +143,9 @@ GEM
     et-orbi (1.1.4)
       tzinfo
     execjs (2.7.0)
-    ezid-client (1.7.1)
+    ezid-client (1.8.0)
```
murny: I think we need this version to be at least 1.8.0 correct? Should we lock this in our gemfile?
```
gem 'ezid-client', '~> 1.8.0'
```

config/initializers/ezid_client.rb
pgwillia: That's a good idea.  It was like that when I got here and once it changes, it shouldn't change again for a Really Long Time, but putting it in secrets would allow us to change at configuration rather than new code deployment.

PR #925, opened 2018-11-14
Thesis ingest rewrite

weiweishi:
This is to rewrite the batch ingest script to follow the same pattern for batch ingesting of regular ERA items. 

This, however, doesn't include the file download. The PDFs need to be downloaded separately and upload to the application server.


lib/tasks/batch_ingest.rake
```diff
@@ -1,5 +1,7 @@
 INGEST_REPORTS_LOCATION = Rails.root.join('tmp', 'ingest_reports')
 INDEX_OFFSET = 1
+THESIS_COMMUNITY_ID = "db9a4e71-f809-4385-a274-048f28eb6814"
```
murny: Is it possible to look these up dynamically? Search for a community/collection that is of name 'FGSR'/'Theses'  and is a "restricted" community/collection? Or could you push this to the batch ingest CSV?

You can't guarantee this UUID across different environments and what not and I'd hope we can prevent similar logic that we had in HN to handle this.

lib/tasks/batch_ingest.rake
```diff
@@ -39,6 +41,46 @@ namespace :jupiter do
       exit 1
     end
   end
+
+  #To-Do: adding ways to authenticate and download the files. Right now this is a manual step
```
murny: Why is thesis files handled differently then generic items?  Why not have the files supplied with the manifest like we do for normal items?

weiweishi: they are handled slightly differently right now as no student is involved in helping the preparation of the ingest package - maybe this is something @leahvanderjagt and @ERAHelp can help working out as a standard task for the students working on ERA helpdesk.

lib/tasks/batch_ingest.rake
```diff
@@ -4,59 +4,91 @@ INDEX_OFFSET = 1
...
+
+    log 'FINISH: Batch ingest completed!'
+  else
+    log "ERROR: Could not open file at `#{full_csv_path}`. Does the csv file exist at this location?"
+    exit 1
   end
 end
 
+def thesis_community_id
+  solr_service = ActiveFedora::SolrService.instance
```
murny: Shouldn't have to drop down to ActiveFedora for these things. This is where JupiterCore comes into play and all the abstraction work we did on top of ActiveFedora.

Can simple just use `thesis_id = Thesis.where(title: 'Graduate Studies and Research, Faculty of').first.id`

More info: https://github.com/ualbertalib/jupiter/blob/33cf117443825524ae6d0638a1f09cb81e141758/app/models/jupiter_core/locked_ldp_object.rb#L341

lib/tasks/batch_ingest.rake
```diff
@@ -4,59 +4,91 @@ INDEX_OFFSET = 1
...
+  result = solr_service.conn.get 'select', params: { q: ['title_ssi:"Graduate Studies and Research, Faculty of"'] }
+  response = result['response']['docs'][0]
+  thesis_comm_id = response['id']
+  thesis_comm_id
+end
+
+def thesis_collection_id
+  result = ActiveFedora::SolrService.instance.conn.get 'select',
```
murny: Same suggestion as above, with regards to using JupiterCore `.where`

lib/tasks/batch_ingest.rake
```diff
@@ -4,59 +4,85 @@ INDEX_OFFSET = 1
...
-
-    log 'START: Batch ingest started...'
+    csv_path = args.csv_path
+    batch_ingest_csv(csv_path)
+  end
 
+  desc 'batch ingest for theses from a csv file - used to batch ingest theses from Thesis Deposit'
+  task :batch_ingest_theses, [:csv_path] => :environment do |_t, args|
     csv_path = args.csv_path
+    batch_ingest_csv(csv_path, 'thesis')
```
murny: The flag approach works, but we are still duplicating a lot of things. 

You can still reduce this. The only difference is these two lines:

`thesis = thesis_ingest(thesis_data, index, csv_directory, checksums)`
vs 
`item = item_ingest(item_data, index, csv_directory)`

And the checksums call for theses
`checksums = generate_checksums(csv_directory)`

So if you want to stick with the flag approach, move a guard check on the checksums method:

`checksums = generate_checksums(csv_directory) if ingest_type == 'thesis'`

And then `if else` using the ingest_type for the two different ingest methods (item_ingest and thesis_ingest).

This way the entire CSV parsing code can remain the same between the two (maybe call the thesis_data/item_data in the loop what it actually is to keep it the same, `row` or `row_data`).

lib/tasks/batch_ingest.rake
```diff
@@ -4,59 +4,85 @@ INDEX_OFFSET = 1
...
+
+    log 'FINISH: Batch ingest completed!'
+  else
+    log "ERROR: Could not open file at `#{full_csv_path}`. Does the csv file exist at this location?"
+    exit 1
   end
 end
 
+def thesis_community_id
+  thesis_comm_id = Community.where(title: 'Graduate Studies and Research, Faculty of').first.id
```
murny: You don't need to return this variable. As in ruby, the last thing in the method is what's returned, so this is redundant. And since you don't need to use this variable again, you don't need to declare it. 

So could just be like this now:
```suggestion
  def thesis_community_id
    Community.where(title: 'Graduate Studies and Research, Faculty of').first.id
  end
```

mbarnett:
@weiweishi same question: is there any reason for me not to close this? (we can always refer back to this PR even if it's closed)

weiweishi:
Different than the other PR. This is something we'd need to merge into the
codebase to allow batch ingest. I never got around to address some of
Shane's comments, maybe it worth adding it to the backlog to complete the
remaining work needed so this can be merged.


people."*

mbarnett:
Sounds good. We'll need to plan on updating this to work with the Fedora removal, as well.

ConnorSheremeta:
Closing this as this has been completed here: https://github.com/ualbertalib/jupiter/pull/1670

PR #943, opened 2018-12-03
update changelog for 1.2.7

weiweishi:


murny:
Looks fine, should we update the version here too now? https://github.com/ualbertalib/jupiter/blob/master/app/models/jupiter_core.rb#L17

PR #945, opened 2018-12-04
Use ez.test.datacite.org for testing of doi minting code [update]

pgwillia:
- removed workarounds for issues that were fixed in api https://github.com/datacite/cheetoh/pull/21


pgwillia:
Added a changelog entry.  I'm trying to get better about remembering to do this!

Issue #55, opened 2017-07-28
[Jupiter] Add version.rb file for jupiter?


murny:
Nice to have but a `version.rb` file for jupiter project would be handy. Just like what most ruby gems, these are still really handy in applications as well. 

Can use this version to put the version in the html source of our website and other places like on the admins dashboard. This is really helpful for debugging, so you know what version of code is in production.

Also instead of manually cutting tags and pushing them up into github like we are currently doing for hydranorth this should all be done via rake tasks (no different then what a ruby gem has for free)

Simple approach? Copy what a ruby gem has.

Some different examples from big ruby projects:

https://github.com/gitlabhq/gitlabhq/blob/master/lib/gitlab/version_info.rb
https://github.com/discourse/discourse/blob/master/lib/version.rb
https://github.com/diaspora/diaspora/blob/master/config/defaults.yml#L7
https://github.com/tootsuite/mastodon/blob/master/lib/mastodon/version.rb

mbarnett:
Added a VERSION constant in jupiter_core.rb

weiweishi:
Adding a version.rb file seems to be a simple enough task and will put us on a better track for release tagging/management. It seems that all we need to do is to use one of the version management files from @murny's examples, and add this to application.rb file like so?https://github.com/tootsuite/mastodon/blob/master/config/application.rb#L14

murny:
Yep exactly. That's all you need to do :+1:

weiweishi:
- move versions from model to version.rb
- gemcutter gem?
- find better ways to present the version information?

weiweishi:
timebox for 1 working day

murny:
Work is now done and in master. 

Leaving this issue open until we test the release management process of cutting and tagging new versions and add documentation to README.md

mbarnett:
Pretty sure we can call this done now, yes?

PR #985, opened 2019-01-16
Add version file, put version file in meta tag, add tool for release management

murny:
https://github.com/ualbertalib/jupiter/issues/55

- [x] Move version to a separate file and update code
- [x] Add version to meta generator tag
- [x] Figure better process for bumping versions and cutting tags
  - Using a gem called `bump` (https://github.com/gregorym/bump): 

```
bump --help
Bump your gem version.

Usage:
    bump current    # show current version
    bump file       # show version file path
    bump pre        # increase prerelease version of your gem (1.0.0-X) [alpha, beta, rc, ]
    bump patch      # increase patch version of your gem (1.0.X)
    bump minor      # increase minor version of your gem (1.X.0)
    bump major      # increase major version of your gem (X.0.0)
    bump set 1.2.3  # set the version number to the given value

Options:
        --no-commit                  Do not make a commit.
    -m, --commit-message MSG         Append MSG to the commit message.
        --no-bundle                  Do not bundle.
        --tag                        Create git tag from version (only if commit is true).
        --replace-in FILE            Replace old version with the new version additionally in this file
    -h, --help                       Show this.
```

So think all we need to do is....for example if doing a minor patch update....run `bump patch --tag` and it will bump the version, commit it, and tag a release all at once.

Can test this and maybe add proper documentation to README when we cut the next release for jupiter

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -15,6 +15,7 @@
 
     <title><%= page_title(t('site_name')) %></title>
     <meta name="description" content="<%= page_description %>">
+    <meta name="generator" content="<%= "Jupiter #{Jupiter::VERSION} - https://github.com/ualbertalib/jupiter" %>">
 
     <%# Open Graph - http://ogp.me/ %>
     <meta property="og:type"         content="<%= page_type %>">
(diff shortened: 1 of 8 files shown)


app/views/layouts/application.html.erb
```diff
@@ -15,6 +15,7 @@
 
     <title><%= page_title(t('site_name')) %></title>
     <meta name="description" content="<%= page_description %>">
+    <meta name="generator" content="<%= "Jupiter #{Jupiter::VERSION} - https://github.com/ualbertalib/jupiter" %>">
```
weiweishi: :+1: Like how it's now properly embedded in the metatag.

PR #1040, opened 2019-03-12
add current and future ruby to travis

pgwillia:
This application currently runs in production with Ruby 2.4.  Some of our other application run with 2.5 and 2.6 is the newest.  We can use travis to smoke test whether or not the changes we make are future compatible.

The con of this is that we're more likely to be caught in the flapping tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,13 @@ addons:
   chrome: stable
 
 rvm:
-  - 2.4.1
+  - 2.4
+  - 2.5
 
 cache: bundler
 bundler_args: --without development production --jobs=3 --retry=3
+before_install:
+  - gem install bundler:1.17.3
 
 before_script:
   - bundle exec danger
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -576,4 +576,4 @@ DEPENDENCIES
   wicked
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
jefferya: The rpm build environment Henry was using for another project (cwrc_preservation) was having trouble with Bundler v2. Maybe check with him before merging.

pgwillia:
```suggestion
   1.17.3
```

.travis.yml
```diff
@@ -11,10 +11,14 @@ addons:
...
 rvm:
-  - 2.4.1
+  - 2.4
+  - 2.5
+  - 2.6
 
 cache: bundler
 bundler_args: --without development production --jobs=3 --retry=3
+before_install:
+  - gem install bundler
```
pgwillia:
```suggestion
  - gem install bundler:1.17.3
```

Issue #996, opened 2019-01-24
[Jupiter] Remove validation for embargo date if embargo option is not selected


weiweishi:
**Describe the bug**
When the option of visibility: embargo is not selected, if there is embargo date information left from previous edits, it gives users a 500 Error

 **Screenshots**
![image](https://user-images.githubusercontent.com/2047614/51706407-91725980-1fdb-11e9-91f8-933455f2f2e0.png)

**Expected behavior**
When the option of visibility: embargo is not selected, the data in the embargo date field should be ignored, and should not be validated when saving. 

**Additional context**
This issue was first reported in #995 (part 2)

PR #1041, opened 2019-03-13
Remove validation for embargo date if embargo option is not selected

jefferya:
Fixes: #996 

Purpose: improve validation to handle case where `visibility: embargo` is not set and the embargo date is present. 

Changes:
- Remove validation for embargo date if embargo option is not set
- Add test for failure when object has visibility open combined with an embargo date


app/models/concerns/item_properties.rb
```diff
@@ -100,7 +100,6 @@ def as_json(_options)
       # If you're looking for rights and subject validations, note that they have separate implementations
       # on the Thesis and Item classes.
       validates :embargo_end_date, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
-      validates :embargo_end_date, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
       validates :visibility_after_embargo, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
```
jefferya: The embargo date validation differs from draft item to item models. I'm not sure why?

test/models/draft_item_test.rb
```diff
@@ -167,6 +167,12 @@ def before_all
     )
 
     assert draft_item.valid?
+
+    draft_item.assign_attributes(
+      visibility: DraftItem.visibilities[:open_access]
+    )
+
+    assert draft_item.valid?
```
jefferya: ToDo: This test should also be updated in test/models/item_test.rb

weiweishi:
@jefferya see you've removed WIP. Is this ready to be reviewed?

test/models/draft_item_test.rb
```diff
@@ -167,6 +167,12 @@ def before_all
     )
 
     assert draft_item.valid?
+
+    draft_item.assign_attributes(
```
ConnorSheremeta: There should probably be a similar check like this in /test/models/draft_thesis_test.rb

ConnorSheremeta:
No commits have been made since my comments. The error is still present for theses and I believe there should be a test added in /test/models/draft_thesis_test.rb

PR #1068, opened 2019-04-05
wrap download button in policy check

pgwillia:
WIP - writing a regression test


pbinkley:
Merging as is for quick deployment, will add reversion test afterwards.

Issue #954, opened 2018-12-06
[ERA] Bring in rack attack for rate limiting


murny:
For other apps in DI (avalon, discovery/neosdiscovery, etc), we have taken the approach to deal with enthusiastic web scrappers by banning them (or in some cases, just banning all web scrappers).

A far better approach would be  to rate limit them (and any other spammy users) and giving them a proper 429 (Too Many Requests) status code.

This is all easily done by using this gem:
https://github.com/kickstarter/rack-attack

We can lead by example by adding this to jupiter.

Simply add the gem and an intializer to jupiter! Done!

The rack-attack initializer in jupiter would be something like this (https://github.com/kickstarter/rack-attack/blob/master/docs/example_configuration.md):

```ruby
  ### Throttle Spammy Clients ###

  # If any single client IP is making tons of requests, then they're
  # probably malicious or a poorly-configured scraper. Either way, they
  # don't deserve to hog all of the app server's CPU. Cut them off!
  #
  # Note: If you're serving assets through rack, those requests may be
  # counted by rack-attack and this throttle may be activated too
  # quickly. If so, enable the condition to exclude them from tracking.

  # Throttle all requests by IP (60rpm)
  #
  # Key: "rack::attack:#{Time.now.to_i/:period}:req/ip:#{req.ip}"
  throttle('req/ip', limit: 300, period: 5.minutes, &:ip)
```

(We don't need to rate limit logins/etc as we off source this to the central IST's SAML server)

weiweishi:
- part of the testing would involve checking the logs for 429 error message
- error page for 429
- EZProxy traffic? 
- this will apply to all rails apps.

PR #1077, opened 2019-04-10
Bring in rack attack for rate limiting

ConnorSheremeta:
-Added rack attack gem and rack_attack.rb initializer file
#954
Successful throttling screenshot:
954SS

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -39,5 +39,7 @@ module Jupiter
 
     config.redis_key_prefix = "jupiter.#{Rails.env}."
 
+    # Use rack-attack as middleware
+    config.middleware.use Rack::Attack
   end
 end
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -39,5 +39,8 @@ class Application < Rails::Application
 
     config.redis_key_prefix = "jupiter.#{Rails.env}."
 
+    # Use rack-attack as middleware
```
pgwillia: What information does the comment add? Could remove the comment and let the syntactic sugar of Ruby on Rails do the talking.

config/environments/test.rb
```diff
@@ -48,4 +48,7 @@
 
   # FITS characterization
   config.run_fits_characterization = false
+
+  # Remove rack-attack so tests aren't rate limited
```
pgwillia: Good idea!

PR #1079, opened 2019-04-11
use the file_download_item helper on download links

pgwillia:
previously this would show the active storage blob url and not check for 
authorization or update the download count.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 ## [Unreleased]
 
 ### Added
+- regression tests for downloading restricted items from search results [PR#1070](https://github.com/ualbertalib/jupiter/pull/1070)
 
-- regression tests for downloading restricted items from search results [PR#1070](https://github.com/ualbertalib/jupiter/pull/1069)
+### Fixed
+- use the download url helper on the search results page [PR#1079](https://github.com/ualbertalib/jupiter/pull/1079)
 
 ## [1.2.11] - 2019-04-05
 
--- a/app/views/application/_item.html.erb
+++ b/app/views/application/_item.html.erb
@@ -6,18 +6,19 @@
   <div class="media-body ml-3">
     <div class="d-flex flex-wrap flex-lg-nowrap align-items-start">
       <h3 class="mt-0 h5 mr-auto"><%= link_to item.title, item_path(item) %></h3>
-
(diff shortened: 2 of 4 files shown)


pbinkley:
@pgwillia confirmed that downloads will still be sourced to the ActiveStorage blob and not directly to Fedora, which is what we want.

With this change in place, could we block external acess to the ActiveStorage blob urls completely? We would have to be sure to let thumbnails (and other derivatives) through. Something to think about if the authentication-check is going to be difficult to implement.

pbinkley:
I'd like to hold off on merging this until @weiweishi has seen it, in case there's relevant history behind the use of blob urls on the search results pages that I'm not aware of.

weiweishi:
@pgwillia @pbinkley this PR looks good to me. I think it's an oversight here previous to not using the proper file_download_item_url helper method to constructure the download URL. This should not be treated any differently than other download URLs.

weiweishi:
I wonder, here where it uses `key`, should we also changed it to use `signed_id`, just for precaution?
https://github.com/ualbertalib/jupiter/blob/master/app/controllers/active_storage/blobs_controller.rb#L8

according to the official document, key should never be revealed to the users? 
https://api.rubyonrails.org/classes/ActiveStorage/Blob.html#method-i-key

We however monkey-patched it，and I can't figure out why...
https://github.com/ualbertalib/jupiter/blob/master/config/initializers/active_storage.rb#L6

pgwillia:
> We however monkey-patched it，and I can't figure out why...
> https://github.com/ualbertalib/jupiter/blob/master/config/initializers/active_storage.rb#L6

I wonder if it was for backward compatible thumbnails?
https://github.com/ualbertalib/jupiter/blob/e8bf60a04ce1c065ed643ba50bc2cdb1b7baa4a5/config/routes.rb#L115

pgwillia:
> We however monkey-patched it，and I can't figure out why...
> https://github.com/ualbertalib/jupiter/blob/master/config/initializers/active_storage.rb#L6

https://github.com/ualbertalib/jupiter/pull/718/files#r185930910

pgwillia:
https://github.com/ualbertalib/jupiter/pull/631#issue-180038152

pbinkley:
So ActiveStorage wanted to meet the use-case of providing short-lived download links, which did not suit our use-case because we want thumbnails etc. to be cacheable, so we by-passed it. Except now we don't want the old ones to work any more. But we've already revealed the keys in the harvested urls. Is the key changeable?

PR #1081, opened 2019-04-12
nginx configuration change to deny access to active_storage blobs

pgwillia:
Should be handled in the background and not grant access to the average user.

The equivalent for apache (jupiter.conf) is
```
 RewriteEngine On
 RewriteCond %{REQUEST_METHOD} ^TRACE
 RewriteRule .* - [F]
 RewriteCond %{REQUEST_URI} /active_storage/blobs [NC]
 RewriteRule ^.*$ - [F,G]
```

Use the docker deployment images
```
docker-compose -f docker-compose.deployment.yml up -d
docker-compose -f docker-compose.deployment.yml run web rails db:migrate
docker-compose -f docker-compose.deployment.yml run web rails db:setup
docker-compose -f docker-compose.deployment.yml run web rails c
```
Visit localhost and pick an item with download.  Make note of the id, for example '8927cb4f-9d9c-4cd1-b2cb-78aad9cc536a'

In the rails console find a blob url
```
include Rails.application.routes.url_helpers
item = JupiterCore::LockedLdpObject.find('8927cb4f-9d9c-4cd1-b2cb-78aad9cc536a', types: [Item, Thesis])
file = item.files.first
url_for(file)
```
Make note of the response.  It will be something like `http://uat.library.ualberta.ca/rails/active_storage/blobs/qaqoPUYCMvwp9GPoNBeaT8DK/840559b8-e953-4606-b417-546b0fb27f16`

From this branch you should not be able to visit the url
Screenshot from 2019-04-12 15-24-53
But all the variants for thumbnails should still work
Screenshot from 2019-04-12 15-42-37
Screenshot from 2019-04-12 15-42-24

From the master branch
```
# add --force-recreate to update the volumes for the nginx config change
docker-compose -f docker-compose.deployment.yml up -d --force-recreate
```
Screenshot from 2019-04-12 15-34-31


pgwillia:
Thumbnails etc appear to continue to work with this change.

pbinkley:
We should configure Apache to return ```510 Gone``` for these, so that any search engine that has already harvested them will know to delete them (and not to punish us for 404s).

pgwillia:
> We should configure Apache to return `510 Gone` for these, so that any search engine that has already harvested them will know to delete them (and not to punish us for 404s).

> We should configure Apache to return `510 Gone` for these, so that any search engine that has already harvested them will know to delete them (and not to punish us for 404s).

510 Not Extended
Further extensions to the request are required for the server to fulfill it.

410 Gone
This response would be sent when the requested content has been permanently deleted from server, with no forwarding address. Clients are expected to remove their caches and links to the resource. The HTTP specification intends this status code to be used for "limited-time, promotional services". APIs should not feel compelled to indicate resources that have been deleted with this status code.

403 Forbidden
The client does not have access rights to the content, i.e. they are unauthorized, so server is rejecting to give proper response. Unlike 401, the client's identity is known to the server.

weiweishi:
I wonder if 403 is a more proper response? The content is not gone, but just not premitted to be viewed through centain routes?

We'd be able to use .htaccess file:

```
RewriteEngine On
RewriteCond %{REQUEST_URI} /active_storage/blobs [NC]
RewriteRule ^.*$ - [F,L]
```
Or in apache configuration:
```
<Location />
RewriteEngine On
RewriteCond %{REQUEST_URI} ^/active_storage/blobs [NC]
RewriteRule ^.*$ - [F,L]
</Location>
```

pbinkley:
As discussed this morning: since we'll never serve to the public from this URI again, I think 410 is appropriate: there's no point in any external system remembering this URI for any kind of future access. 

(And sorry about the 510 Can't Remember Response Code error)

pbinkley:
I suppose the fully correct response would be a ```301 Moved Permanently```, redirecting to the helper-based url, which would then issue the 403. That would have to implemented in the application, though, so that it could calculate the target url. I think 410 is good enough, given the small exposure of the blob urls.

pbinkley:
I've emailed @henryzhang87  asking for the configuration change as outlined in @weiweishi 's comment above, with 410 response codes.

pgwillia:
Since the equivalent is already in production can we merge this to catch up uat?

Issue #865, opened 2018-09-18
Add tests for Controllers


murny:
Some of our controllers are missing tests (DownloadsController and ItemsController are probably the two biggest but there is a bunch of others here (profiles/robots/static pages/etc))

We should be testing these actions and verifying they are coming back with the proper responses (in DownloadsController case, are the proper disposition and content_type headers set properly?).

ConnorSheremeta:
It appears as though DownloadsController [has had tests added already](https://github.com/ualbertalib/jupiter/blob/master/test/controllers/downloads_controller_test.rb) (although simplecov is still showing 0% coverage for downloads_controller.rb for some reason.) Would it be safe to assume that this issue can be closed?

pgwillia:
Yeah, looks like the tests for downloads controller were added the day after the issue was created. https://github.com/ualbertalib/jupiter/pull/866/commits/2b814d4a8410b82e3bd203f82528b17de304735f

In coverage they're not getting credit?
![Screenshot from 2019-04-12 13-10-46](https://user-images.githubusercontent.com/1220762/56060546-6834b000-5d24-11e9-95a2-f74987aa30bd.png)
  Maybe because of the route changing the name?
https://github.com/ualbertalib/jupiter/blob/4a90d1c4108acc20ca5e6c34e6e1cb38d04dfb1f/config/routes.rb#L24-L25


There are still improvements needed for this and others
![Screenshot from 2019-04-12 13-05-50](https://user-images.githubusercontent.com/1220762/56060327-c1501400-5d23-11e9-8a17-498341b057e8.png)

PR #1114, opened 2019-05-01
Add tests for controllers

ConnorSheremeta:
- Added tests for admin/items controller
- Added tests for items controller
- Added tests for profile controller
- Added tests for robots controller
#865


test/controllers/items_controller_test.rb
```diff
@@ -0,0 +1,69 @@
...
+      graduation_date: '2017-03-31',
+      visibility: ItemProperties::VISIBILITY_EMBARGO,
+      embargo_end_date: 2.days.from_now.to_date,
+      visibility_after_embargo: CONTROLLED_VOCABULARIES[:visibility].public
+    ).unlock_and_fetch_ldp_object do |unlocked_thesis|
+      unlocked_thesis.add_to_path(@community.id, @collection.id)
+      unlocked_thesis.save!
+    end
+
+    @admin = users(:admin)
```
weiweishi: Should we add a test for non-authorized user? not being able to view or edit an item?

test/controllers/profile_controller_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+
+class ProfileControllerTest < ActionDispatch::IntegrationTest
+
+  test 'should get profile index' do
+    sign_in_as(users(:admin))
+    get profile_url
+    assert_response :success
```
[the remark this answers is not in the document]
ConnorSheremeta: done

test/controllers/items_controller_test.rb
ConnorSheremeta: Made tests for editing an item. Making a test for a non-authorised user not being able to view an authorisation required item wasn't possible as the response for get item_url(@item) isn't different for authorised/un-authorised users. It would be possible to do feature testing and checking for "Item Restricted to University of Alberta Users" text or the lack of "View"/"Download" text. I did however add some tests to the downloads controller for those actions.

PR #1166, opened 2019-06-12
Direct solr connection

mbarnett:
Changes the way Jupiter connects to Solr -- instead of letting ActiveFedora establish a connection to Solr and then piggybacking on that, this creates a dedicated Solr connection directly based on the configuration in `solr.yml`

### DIFF
--- a/app/models/jupiter_core/solr_client.rb
+++ b/app/models/jupiter_core/solr_client.rb
@@ -1,9 +1,12 @@
 class JupiterCore::SolrClient
+
   include Singleton
 
-  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read("#{Rails.root}/config/solr.yml")).result, [], [], true)[Rails.env].symbolize_keys
+  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read(Rails.root.join('config', 'solr.yml'))).result,
+                               [], [], true)[Rails.env].symbolize_keys
 
   def connection
-    @solr_connection ||= RSolr.connect :url => SOLR_CONFIG[:url]
+    @connection ||= RSolr.connect url: SOLR_CONFIG[:url]
   end
+
 end
--- a/app/models/jupiter_core/solr_client.rb
+++ b/app/models/jupiter_core/solr_client.rb
@@ -1,0 +1,9 @@
+class JupiterCore::SolrClient
+  include Singleton
+
+  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read("#{Rails.root}/config/solr.yml")).result, [], [], true)[Rails.env].symbolize_keys
+
+  def connection
+    @solr_connection ||= RSolr.connect :url => SOLR_CONFIG[:url]
+  end
+end
--- a/lib/tasks/recover.rake
+++ b/lib/tasks/recover.rake
@@ -60,6 +60,6 @@ namespace :jupiter do
 
     solr_data = {}
     object.unlock_and_fetch_ldp_object { |uo| solr_data = uo.to_solr }
-    ActiveFedora::SolrService.add(solr_data, 'softCommit' => true)
+    JupiterCore::SolrClient.instance.connection.add(solr_data, 'softCommit' => true)
   end
 end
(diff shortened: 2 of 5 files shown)


app/models/jupiter_core/solr_client.rb
```diff
@@ -0,0 +1,12 @@
+class JupiterCore::SolrClient
+
+  include Singleton
+
+  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read(Rails.root.join('config', 'solr.yml'))).result,
+                               [], [], true)[Rails.env].symbolize_keys
+
```
weiweishi: I don't quite understand what this line of code is doing? Do you mind giving me a little explanation?

app/models/jupiter_core/solr_client.rb
```diff
@@ -0,0 +1,12 @@
+class JupiterCore::SolrClient
+
+  include Singleton
+
+  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read(Rails.root.join('config', 'solr.yml'))).result,
```
weiweishi: Curious to see that you have the configuration file hardcoded in the model? wondering if there's better way handling this?

app/models/jupiter_core/solr_client.rb
mbarnett: In a nutshell: the existing [Solr config file](https://github.com/ualbertalib/jupiter/blob/master/config/solr.yml) is a Yaml file that also contains ERB statements (anything in `<% ... %>`) and makes use of Yaml aliases (The `&default` references used for the UAT, Staging, and Prod settings).

So to load it correctly you need to: read the file in, run it through ERB to replace all of the `<% ...%>` blocks with the proper content, and then parse the resulting Yaml. Because we don't need or expect arbitrary object deserialization in this config file, we use `safe_load` rather than a raw `load` to disallow object deserialization, which is being paranoid but at least a bit safer.

The downside of `safe_load` is that by default it disables aliases in Yaml files, which we use. The `[], [], true` arguments just tell it to whitelist no classes or symbols for deserialization, and enable aliases.

app/models/jupiter_core/solr_client.rb
mbarnett: Like most of jupiter_core this is less model than model infrastructure -- so it might get moved in a future PR as I'm working on the larger Solr export infrastructure now.

Hardcoding the name...I don't know that there's any better way to do that. We seem to be in reasonable company here, eg) [Sidekiq](https://github.com/mperham/sidekiq/blob/455e9d56f46f0299eaf3b761596207e15f906a39/lib/sidekiq/cli.rb#L224), [Skylight](https://github.com/skylightio/skylight-ruby/blob/ebd13decb39e2f9d32c7244c193142c985df1740/lib/skylight/railtie.rb#L18)

The alternative to this is what ActiveFedora tried to do with a bunch of indirection to be "flexible", but the results are [fairly unreadable](https://github.com/samvera/active_fedora/blob/1df7e4506811a8d50d425e9dc9becec255b18758/lib/active_fedora/file_configurator.rb)  (and even that, in the end, [falls back to just looking for "solr.yml"](https://github.com/samvera/active_fedora/blob/1df7e4506811a8d50d425e9dc9becec255b18758/lib/active_fedora/file_configurator.rb#L200))

*At some level* you have to specify what the name of the file that contains the stuff you want in it. You could make that configurable, but how much call is there to have your solr config live in `config/foobar.yml` instead of `config/solr.yml`?. And then you're just hardcoding the environment variable or config file name that contains *that* setting instead of this one.

It's turtles all the way down, at some point.

weiweishi: Okay. Thanks for the explanation. Very helpful

lib/tasks/recover.rake
```diff
@@ -60,6 +60,6 @@ namespace :jupiter do
 
     solr_data = {}
     object.unlock_and_fetch_ldp_object { |uo| solr_data = uo.to_solr }
-    ActiveFedora::SolrService.add(solr_data, 'softCommit' => true)
+    JupiterCore::SolrClient.instance.connection.add(solr_data, 'softCommit' => true)
```
pgwillia: Looks good -- no behaviour change.  For information only there was some discussion about if the `softCommit` bit is actually doing anything: https://github.com/rsolr/rsolr/issues/198#issuecomment-486806192

PR #1167, opened 2019-06-12
Solr name mangling

mbarnett:
Replaces all calls into `Solrizer.solr_name` with Jupiter-internal code to handle mapping Solr type/role to stem mappings.

### DIFF
--- a/app/models/jupiter_core/solr_name_mangler.rb
+++ b/app/models/jupiter_core/solr_name_mangler.rb
@@ -1,35 +1,36 @@
 module JupiterCore::SolrNameMangler
-
   SOLR_MANGLED_STEMS_BY_TYPE = {
-    :path => {:pathing => 'dpsim'},
-    :boolean => {:exact_match => 'ssim'},
-    :date => {
-      :sort => 'dtsi',
-      :exact_match => 'ssim'
+    path: { pathing: 'dpsim' },
+    boolean: { exact_match: 'ssim' },
+    date: {
+      sort: 'dtsi',
+      exact_match: 'ssim'
     },
-    :integer => {
-      :search => 'isim',
-      :sort => 'isi',
-      :range_facet => 'isi',
-      :exact_match => 'ssim'
+    integer: {
+      search: 'isim',
+      sort: 'isi',
+      range_facet: 'isi',
+      exact_match: 'ssim'
     },
-    :string => {
-      :pathing => 'dpsim',
-      :facet => 'sim',
-      :search => 'tesim',
-      :sort => 'ssi',
-      :exact_match => 'ssim'
+    string: {
+      pathing: 'dpsim',
+      facet: 'sim',
+      search: 'tesim',
+      sort: 'ssi',
+      exact_match: 'ssim'
     },
-    :text => {
-        :search => 'tesim'
+    text: {
+      search: 'tesim'
     }
-  }
+  }.freeze
 
   def self.mangled_name_for(name, type:, role:)
     stem = SOLR_MANGLED_STEMS_BY_TYPE.dig(type, role)
-    raise JupiterCore::SolrNameManglingError, "Unmapped type/role combination for requested mangle of #{name}: type: #{type}, role: #{role}" unless stem.present?
+    if stem.blank?
+      raise JupiterCore::SolrNameManglingError,
+            "Unmapped type/role combination for requested mangle of #{name}: type: #{type}, role: #{role}"
+    end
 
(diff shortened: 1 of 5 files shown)


app/models/jupiter_core/solr_name_mangler.rb
```diff
@@ -0,0 +1,36 @@
+module JupiterCore::SolrNameMangler
```
weiweishi: The class name is rather long? Would we be okay to use 
JupiterCore::SolrName

app/models/jupiter_core/solr_name_mangler.rb
```diff
@@ -0,0 +1,36 @@
...
+      search: 'tesim',
+      sort: 'ssi',
+      exact_match: 'ssim'
+    },
+    text: {
+      search: 'tesim'
+    }
+  }.freeze
+
+  def self.mangled_name_for(name, type:, role:)
```
weiweishi: Question: 
When solr fields have multiple roles, like date for sort and facet etc, we will just have two lines of call for this model to create two solr names, correct?

mbarnett: Yes, this is called in a loop in LockedLDPObject currently, so eg) for Item publisher, defined as `  has_attribute :publisher, ::RDF::Vocab::DC.publisher, solrize_for: [:search, :facet]`, this is called twice: `mangled_name_for(:publisher, type: :string, role: :search)` and `mangled_name_for(:publisher, type: :string, role: :facet)` resulting in `publisher_tesim` and `publisher_sim` respectively.

In the next series of pull requests using this gets pulled out of LockedLDPObject and abstracted behind some SolrExporter infrastructure.

app/models/jupiter_core/solr_name_mangler.rb
mbarnett: I think that's more misleading? `SolrName` as a noun implies that instances represent encapsulated "SolrNames". This isn't a class and doesn't get instantiated, it's a service module that encapsulates the logic of mangling a name. If anything, something like "SolrNameManglingService" would probably be more accurate and communicative.

I actually think "SolrNameMangler" is pretty short as names go, but then again I'm responsible for "LockedLDPObject", "DeferredFacetedSolrQuery", etc. I guess that puts me in the [German Naming Convention](https://chrisdone.com/posts/german-naming-convention/) school of thought.

pgwillia: What do you think of `SolrNameTailor`?  Then instead of
>verb: severely mutilate, disfigure, or damage by cutting, tearing, or crushing.
it would be
>verb: (of a tailor) make (clothes) to fit individual customers.
    "he was wearing a sports coat that had obviously been tailored in New York"

mbarnett: This is reorganized and renamed slightly in the new PR, so we might want to take this conversation over there, now?
 
With respect to NameTailoring, my thought is that "the process of adding gibberish to a field name in order to encode type information that will later be picked up and acted upon by other processes further down the line" has, historically at least, almost always been specifically referred to as [Name Mangling](https://en.wikipedia.org/wiki/Name_mangling) in common usage and the literature. 

So like in the wikipedia example, `Wikipedia::Article.format` is transformed into "_ZN9Wikipedia7article6formatE" by the C++ compiler's name mangling facilities  so that when the linker picks up this data it knows what types are associated with the method, which is more-or-less the direct equivalent of what we're doing when we turn "title" into "title_tesim" so that Solr knows that "title" is a text stored indexed multivalue (or I think that's what that stands for, anyways...).

I guess my argument here is that if the goal is to broadly follow industry and academic standards, Name Mangling is what the jargon for this process *is*, general connotations about industrial equipment catching fingers notwithstanding. 

I'm not _opposed_ to calling it something else per se, but we should definitely be conscious of the fact that we'd be moving away from something immediately communicative to anyone familiar with the terminology and towards something that doesn't really mean anything standard and which needs to be specifically looked up and learned for Jupiter, which to me trends towards the Hydra pattern of having things like "PowerConverters", which didn't really mean anything unless you looked up and memorized their bespoke metaphors.

pgwillia: I follow your argument. Thanks for the thoughtful response.  My suggestion wasn't particularly useful, I see that now.

mbarnett: It's always useful to ask questions! I mean I don't want to get away with naming things all manner of nonsense just because it makes sense to me.

Issue #1215, opened 2019-07-19
Unable to execute command "fits.sh": command not found


pgwillia:
View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/579/](https://rollbar.com/ualbertalib/jupiter/items/579/)

These errors only occur on one of our app servers because `fits.sh` doesn't exist, `fits` does.

Need to [configure](https://github.com/samvera/hydra-derivatives#dependencies) fits_path or create a symlink to fits.sh
`Hydra::Derivatives.fits_path = '/opt/local/fits/bin/fits.sh'`

pgwillia:
fits.x86_64                        1.2.0-3 may fix this
https://code.library.ualberta.ca/gitweb/?p=rpmspecs.git;a=commitdiff;h=b09e3e698847609398476027a96b5b50f3192e62

PR #1217, opened 2019-07-19
added initializer for fits characterization

pgwillia:
Allows us to configure path to fits.  Requires adding a new ENV VAR/secret upon deployment

#1215

### DIFF
--- a/config/initializers/characterization.rb
+++ b/config/initializers/characterization.rb
@@ -1,0 +1,1 @@
+Hydra::Derivatives.fits_path = Rails.application.secrets.fits_path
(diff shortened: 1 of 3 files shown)


config/initializers/characterization.rb
```diff
@@ -0,0 +1 @@
+Hydra::Derivatives.fits_path = Rails.application.secrets.fits_path
```
weiweishi: The use of Hydra::Derivatives hopefully will be short-lived. Characterization information captured through Hydra::Derivatives will be replaced by information gathered through ActiveStorage etc.

Issue #1003, opened 2019-01-28
[Jupiter] Display graduation date in "Fall 2017" format


pbinkley:
With agreement from FGSR and Leah, we'd like to display the graduation date field for theses in the "Fall 2016" format rather than the current "2016-11" format. The underlying data will still use "2016-11", so it won't be changed. The edit form for theses also does not need to change (since it is only used by admin users). The change applies only to search results and full item displays. OAI is out of scope. 

The graduation_date field is normalized to YYYY-MM, with MM = 06 or 11. This should map to "Spring YYYY" when the MM is 06, and "Fall YYYY" when the MM is 11. Fall back to just YYYY for any other value of MM or when MM is absent.

Related to #798. Weiwei has reviewed email discussion of that issue at the time to clarify some of the history.

weiweishi:
The original requirements on the deposit form will remain unchanged - consider this as a new requirement, for public interface.

pgwillia:
https://github.com/ualbertalib/jupiter/blob/e797c4f9961bff205e5482c190f884102db9a6c2/app/views/items/_thesis.html.erb#L40

https://github.com/ualbertalib/jupiter/blob/e797c4f9961bff205e5482c190f884102db9a6c2/app/helpers/items_helper.rb#L24-L40

Here's how the deposit side is mapped to 'Spring' 'Fall'
https://github.com/ualbertalib/jupiter/blob/e797c4f9961bff205e5482c190f884102db9a6c2/app/models/draft_thesis.rb#L5-L12
https://github.com/ualbertalib/jupiter/blob/e797c4f9961bff205e5482c190f884102db9a6c2/app/models/draft_thesis.rb#L48-L49
https://github.com/ualbertalib/jupiter/blob/e797c4f9961bff205e5482c190f884102db9a6c2/app/models/draft_thesis.rb#L160-L164

PR #1220, opened 2019-07-23
display graduation date in 'Fall 2017' format

pgwillia:
#1003

### DIFF
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -23,19 +23,35 @@ module ItemsHelper
 
   # We have a lot of messy "date-ish" data. Created dates coming through the Draft interface are actual date types
   # saved into Fedora/Solr as strings at the moment. Much of the legacy data is generally a freeform string containing
-  # anything from "2017/09/12" to '2013' to "2012-09-26T11:18:38Z" (on Theses) to "Fall 1978" to "Unknown"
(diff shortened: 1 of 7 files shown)


app/helpers/items_helper.rb
```diff
@@ -23,19 +23,35 @@ def description(object)
...
     return I18n.t('date_unknown') if dateish.blank?
 
-    Date.parse(dateish).iso8601
+    begin
+      Date.parse(dateish).iso8601
+    rescue ArgumentError
+      term = Date.strptime(dateish, '%Y-%m')
+
+      # Fall back to just YYYY for any other value of MM or when MM is absent.
+      return "#{SEASON[term.month]} #{term.year}".strip
```
weiweishi: So when term.month doesn't exist, it will return an empty string or nil?

pgwillia: Is the case your thinking of `assert_equal '2013', humanize_date('2013')`?

In the case of dateish = '2013': `Date.parse` returns `ArgumentError`, then `Date.strptime` returns `ArgumentErrror` so it ends up on line 47 and just prints the string '2013'.

weiweishi: and from your comment I gathered it would be "2013" if the term is for example "2013-04"?

PR #1238, opened 2019-08-07
Leverage annotations to annotate draft classes

mbarnett:
Also adds a DraftCollection and DraftCommunity model to facilitate further migration work.

Some columns on drafts have TODOs on annotations at this time, as further ActsAsRdfable gem work is required to make certain scenarios like join tables work well. This PR constitutes the basis for further work to support changes to the controllers and views to remove ActiveFedora support.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,4 +1,12 @@
 GIT
+  remote: https://github.com/mbarnett/acts_as_rdfable.git
+  revision: d1d3b79cab1f074976407e42f9e28d24de5131c8
+  tag: v0.01
+  specs:
+    acts_as_rdfable (0.1.0)
+      rails (~> 5.2.3)
+
+GIT
   remote: https://github.com/ualbertalib/active_fedora.git
   revision: 8e94346e6ed98df46c994726749542075e11ecd0
   branch: backport_rails52_fixes
@@ -504,6 +512,7 @@ DEPENDENCIES
   aasm
   active-fedora!
   active_link_to
+  acts_as_rdfable!
   addressable (~> 2.6.0)
   better_errors (>= 2.3.0)
   binding_of_caller
@@ -573,4 +582,4 @@ DEPENDENCIES
   wicked
 
 BUNDLED WITH
-   1.17.3
+   2.0.2
--- a/app/models/draft_collection.rb
+++ b/app/models/draft_collection.rb
@@ -7,7 +7,7 @@ class DraftCollection < ApplicationRecord
     config.creators has_predicate: ::RDF::Vocab::DC.creator
   end
 
-  def update_from_fedora_collection(collection, for_user)
+  def update_from_fedora_collection(collection, _for_user)
     draft_attributes = {
       collection_id: collection.id,
       visibility: collection.visibility,
--- a/app/models/draft_collection.rb
+++ b/app/models/draft_collection.rb
@@ -1,0 +1,38 @@
+class DraftCollection < ApplicationRecord
+
+  acts_as_rdfable do |config|
+    config.community_id has_predicate: ::TERMS[:ual].path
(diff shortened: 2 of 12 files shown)


Gemfile.lock
```diff
@@ -1,3 +1,11 @@
+GIT
+  remote: https://github.com/mbarnett/acts_as_rdfable.git
+  revision: d1d3b79cab1f074976407e42f9e28d24de5131c8
+  tag: v0.01
+  specs:
+    acts_as_rdfable (0.1.0)
```
weiweishi: So from the look of it, this gem will translate/map the data model from active record to the metadata profiles created by the metadata team? But the actual configuration of the mapping is all done in the jupiter codebase? I'm surprised that no one else has created such a simple yet essential utility.

db/schema.rb
```diff
@@ -157,6 +195,14 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "rdf_annotations", force: :cascade do |t|
```
weiweishi: So please help me understand this table and what the RdfConfig is doing? What is kept in this table? and what would be configured through RdfConfig? Does the config overwrite what's in here? Or it will add new rows in the table?

test/models/draft_community_test.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class DraftCommunityTest < ActiveSupport::TestCase
+
+  def before_all
+    super
+    @community = locked_ldp_fixture(Community, :books).unlock_and_fetch_ldp_object(&:save!)
+    @collection = locked_ldp_fixture(Collection, :books).unlock_and_fetch_ldp_object(&:save!)
+  end
+
+  test 'can be made into a draft' do
```
weiweishi: do we need the test for publish draft_community/draft_collection to Collection/Community objects?

mbarnett: Not really, as that's never going to happen (DraftCollection in Postgres being "published" to ActiveFedora). DraftCommunity is going to *become* Community, in another pull request, so all of the tests that currently operate on the ActiveFedora-based Community will instead run against ActiveRecord-based DraftCommunity (which will have been renamed to Community). Same deal with Collection.

db/schema.rb
mbarnett: Right now it contains the table name, column name, and RDF predicate associated with that column, for any ActiveRecord class that "acts as rdfable". It's overwritten on Application boot, which may or may not be a terrible idea, and will probably change as the API evolves.

Gemfile.lock
mbarnett: Right; it gives us a simple way to associate column in an ActiveRecord data model with predicates, that other code can use when generating RDF. I'm still working out what the best API approach for this is, and it probably won't be nailed down until the OAI stuff gets done.

> I'm surprised that no one else has created such a simple yet essential utility.

I've been wondering this forever, but it mostly seems to come down to everyone else trying to store data in RDF, with predicates specified per-object, rather than treating it as an export artifact.

app/models/draft_collection.rb
```diff
@@ -0,0 +1,38 @@
+class DraftCollection < ApplicationRecord
+
+  acts_as_rdfable do |config|
+    config.community_id has_predicate: ::TERMS[:ual].path
+    config.description has_predicate: ::RDF::Vocab::DC.description
+    config.restricted has_predicate: ::TERMS[:ual].restricted_collection
+    config.creators has_predicate: ::RDF::Vocab::DC.creator
+  end
+
+  def update_from_fedora_collection(collection, _for_user)
```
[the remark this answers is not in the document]
mbarnett: Correct. The class name is going to end up changing to Collection in another PR that removes the ActiveFedora dependency anyways, so this is all just temporary as I work on this.

PR #1242, opened 2019-08-09
Rename sort to order, and change arguments to it and find_by

mbarnett:
This allows us to better align calling code with the ActiveRecord API, easing replacement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ### Changed
 - DeferredSimpleSolrQuery#sort renamed to 'order' and its two arguments replaced with a key-value, to better align with ActiveRecord
-  API and ease removal of ActiveFedora
+  API and ease removal of ActiveFedora.
+- Change LockedLDPObject#find_by to take a named 'id:' parameter, to better align callers with ActiveRecord
 - i18n fallback to english (configuration change) [PR#1058](https://github.com/ualbertalib/jupiter/pull/1058)
 - pin rubocop version for hound [PR#1080](https://github.com/ualbertalib/jupiter/pull/1080)
 - Replaced use of ActiveFedora's Solr connection with a direct connection to Solr setup locally.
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 - DraftCollection and DraftCommunity models for further migration work
 
 ### Changed
+- DeferredSimpleSolrQuery#sort renamed to 'order' and its two arguments replaced with a key-value, to better align with ActiveRecord
```
weiweishi: thanks, TIL difference between `.sort_by` and `.order`, and how `.order` works with RDBMS and is preferred in most of the cases.

app/models/jupiter_core/deferred_simple_solr_query.rb
```diff
@@ -31,7 +31,16 @@ def offset(num)
     self
   end
 
-  def sort(attr, order = nil)
+  def order(attrs = {})
+    ord = nil
+    attr = if attrs.is_a? Hash
```
weiweishi: curious in what case would `attrs` not be a hash?

mbarnett: When I did a find and replace of `sort` to `order`, if I'd missed an calls of `sort(attr)`, it could have ended up as `order(:title)` or something, and this helped me catch that. Also if anyone writes new code in the old style down the road, etc

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -320,7 +320,7 @@ def self.find(id, types: [])
   # find with "return nil if no object with that ID is found" semantics
   # Note: This behaves differently then AR find_by.
   # As it can only take a single argument which is an ID (which is a limitation from ActiveFedora)
-  def self.find_by(id)
+  def self.find_by(id:)
```
pgwillia: TIL that parameter colon (id:) means that id is required.

mbarnett: Required, and required to be named (ie, you can't do `find_by(1234)`, it has to be `find_by(id: 1234)`

mbarnett:
Again, this is now in `integration_postmigration`, so finally getting around to closing this

Issue #1219, opened 2019-07-22
Add embargo_history to embargo removal. 


weiweishi:
**Is your feature request related to a problem? Please describe.**
For auditing purpose, we'd like to keep a history for embargoes. We have implemented `embargo_history` in the data model, and have migrated the data. We should keep the embargo history for any change to embargo through the background job, or through the UI. 
 
**Describe the solution you'd like**
`embargo_history` is an array of strings. 
We can construct the text as such:
"An embargo was deactivated on _DateTime_when_embargo_was_lifted_. Its release date was _embargo_date_. Intended visibility after embargo was _visibility_after_embargo_" 
This can apply for both elapsed embargoes lifted through the background job, as well as any modification done through the UI.

weiweishi:
Check with @leahvanderjagt to see how we should capture when the embargo is activated. 
suggested text:
"An embargo was activated on DateTime_when_embargo_was_activated. Its release date will be embargo_date. Intended visibility after embargo will be visibility_after_embargo".

PR #1252, opened 2019-08-21
Lifting of embargo now stores item in embargo_history

ConnorSheremeta:
- When embargo is lifted through either the UI or background job, an entry is addded to embargo_history
Fixes: #1219
Example text: "embargo_history_ssim":["An embargo was deactivated on 2019-08-22 12:07:53 -0600. Its release date was 2023-01-01T00:00:00.000Z. Intended visibility after embargo was http://terms.library.ualberta.ca/public"]


test/system/deposit_item_test.rb
```diff
@@ -84,6 +84,11 @@ def before_all
     assert_text I18n.t('items.draft.successful_deposit')
     assert_selector 'h1', text: Item.last.title
 
+    # Check to make sure there isn't any embargo_history
+    item_id = current_url.split('/')[-1]
```
pgwillia: I wasn't sure what the [-1] was doing here until I looked it up.  Could do 
```suggestion
    item_id = current_url.split('/').last
```

pgwillia: @ConnorSheremeta do you have any thoughts about this?

ConnorSheremeta: Using .last is definitely more clear in what it's doing, committing your suggestion

test/system/deposit_thesis_test.rb
```diff
@@ -82,6 +82,11 @@ def before_all
     assert_text I18n.t('admin.theses.draft.successful_deposit')
     assert_selector 'h1', text: Thesis.last.title
 
+    # Check to make sure there isn't any embargo_history
+    item_id = current_url.split('/')[-1]
```
pgwillia: One more occurance
```suggestion
    item_id = current_url.split('/').last
```

Issue #1247, opened 2019-08-14
Improvement on rack-attack configuration


weiweishi:
We have noticed occasionally ERA will give 429 error message for regular web access as well as OAI harvesting after we introduced rake-attack in #954. 

We would like to improve the default configuration:
- [x] Add logging to blocked access to analyze what is legitimate access, and what is malicious spamming access. This can be done as such:
```
ActiveSupport::Notifications.subscribe('rack.attack') do |name, start, finish, request_id, req|
  if req.env["rack.attack.match_type"] == :throttle
    request_headers = { "CF-RAY" => req.env["HTTP_CF_RAY"],
                        "X-Amzn-Trace-Id" => req.env["HTTP_X_AMZN_TRACE_ID"] }

    Rails.logger.info "[Rack::Attack][Blocked]" <<
                      "remote_ip: \"#{req.remote_ip}\"," <<
                      "path: \"#{req.path}\", " <<
                      "headers: #{request_headers.inspect}"
  end
end
```
- [x] We can selectively choose the routes we'd like to protect. This can include the home path "/", search "/search", authentication "/auth/saml", anything else we should consider?
- [x] We should safelist IPs include LAC's and staff IP ranges? Those should be set as something that can be managed by Ansible/env vars, so changes don't require codebase change. 

resources:
[https://blog.bigbinary.com/2018/05/15/how-to-mitigate-ddos-using-rack-attack.html]
[https://github.com/kickstarter/rack-attack]

ConnorSheremeta:
Did some bench marking regarding lookup times of whitelisted IP's using different structures to hold those ips: array, hash, and btree with a depth of 5. My hash implementation for some reason was yielding the same result regardless of the number of ips so I've removed those results. Doing 100,000 requests yeilds (in seconds):

                                1 ip          20ips          80ips         500ips      10,000ips
                  array     0.009489       0.034668       0.087295       0.479266       8.882781
                  btree     0.035576       0.062198       0.087248       0.116727       0.164749

At 80 whitelisted ips it would be more efficient to use a btree over an array

mbarnett:
re: hash table

![I am in shock](https://i.giphy.com/media/2yqyPVbzZ7xDaIgUj2/giphy.gif)

I'm not surprised it beats the array at higher IPs, but early on that's very surprising. I'm going to poke at it a bit just to see if I get similar results, because if so there's no reason not to just use a hashtable approach across the board?

PR #1265, opened 2019-08-28
Improvement on rack-attack configuration

ConnorSheremeta:
- Add logging to log and rollbar when an IP is blocked
- Protect home, search, and auth/saml routes by blocking at 60 hits/minute
- Safelist IPs based off of a comma seperated environment variable

Fixes: #1247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 - allow batch ingest to lookup older licenses [#1115](https://github.com/ualbertalib/jupiter/issues/1115)
 - Added selectize '|' delimiter to separate authors or subjects [#1211](https://github.com/ualbertalib/jupiter/issues/1211)
 
+### Added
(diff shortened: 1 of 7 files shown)


config/secrets.yml
```diff
@@ -134,3 +134,4 @@ production:
   solr_url: <%= ENV['SOLR_URL'] %>
 
   skylight_authentication: <%= ENV['SKYLIGHT_AUTHENTICATION'] %>
+  rack_attack_safelisted_ips: <%= ENV['RACK_ATTACK_SAFELISTED_IPS'] %>
```
pgwillia: Why a production only secret?  Is there a reason not to make this shared?  `rack_attack`  is initialized in every environment.

weiweishi: @ConnorSheremeta Would you be able to make this change - implement it for every env, so they are available for configuration as needed.

ConnorSheremeta: Done! As for the comment below, I believe it is expected that safelisting wouldn't work for an external IP when running rails locally

PR #1272, opened 2019-08-30
ActiveRecord Migration -- Part 1

mbarnett:
This is the first of 2 PRs for migrating completely off of ActiveFedora. This PR is designed to _straddle_ both Fedora and Postgresql just long enough to get the data out of the former and into the latter.

### This PR introduces

- New ActiveRecord-based tables & models that will serve as the basis of the application going forward: ArCommunity, ArCollection, ArItem, and ArThesis (don't sweat the names, they change back to Community, Collection, Item, and Thesis in the next PR). These aren't fully fledged in this PR, they're just enough to receive the data from Fedora. (It was orignally thought that we could just use DraftItem and DraftThesis as the new models, but there's enough differences between attribute names and data representation that doing so would have required a lot more work and rewriting of forms and views. By creating new, distinct models, almost all of the views and controllers work untouched).

- Adds a second id column (`upcoming_id`) to DraftItem & DraftThesis, a second `draft_item_id` column to `draft_item_languages` (`upcoming_draft_item_id`) and a second record_id column to `active_storage_attachments` (you guessed it: `upcoming_record_id`). The purpose of all of these columns is the same: because the new Ar* models use `uuid`s instead of integers for their IDs, the original foreign key column in ActiveStorage isn't capable of holding their IDs for attachments -- thus we need to change the `record_id` column to be UUID instead of integer. DraftItem &  DraftThesis then need to change to use UUID for _their_ IDs, so they can continue to work with ActiveStorage. And _that_, finally, necessitates changing the DraftItem-Languages join table's `draft_id` column.

- Mints new UUIDs for the DraftItem/DraftThesis tables. These will become the primary ID column in the next PR.

- A new rake task, `jupiter:get_me_off_of_fedora` to copy data out of fedora and mint UUIDs for tables that are changing to use them

### Deployment Considerations

A version of Jupiter containing everything up to this PR -- BUT NOT INCLUDING THE NEXT PR -- needs to be deployed to the environment being migrated (eg: Production), `bundle exec rake db:migrate` needs to be run to create the new tables and columns, and finally `bundle exec rake jupiter:get_me_off_of_fedora` needs to be run, to copy all of the data out of Fedora and into Postgresql. Once the rake task is run, **the web application will no longer function** (Solr stops reporting results for the ActiveFedora/LockedLdpObjects, because the IDs are taken over by the ActiveRecord versions), until the second PR (https://github.com/ualbertalib/jupiter/pull/1273) is deployed, so a second deployment  needs to be prepped to follow this one immediately after running the migration task.


app/models/concerns/item_properties.rb
```diff
@@ -8,28 +8,25 @@ module ItemProperties
...
-    has_multival_attribute :is_version_of, ::RDF::Vocab::DC.isVersionOf, solrize_for: :exact_match
-    has_multival_attribute :member_of_paths, ::TERMS[:ual].path, type: :path, solrize_for: :pathing
+    has_multival_attribute :embargo_history, ::TERMS[:acl].embargo_history
+    has_multival_attribute :is_version_of, ::RDF::Vocab::DC.isVersionOf
+    has_multival_attribute :member_of_paths, ::TERMS[:ual].path
 
     # See `all_subjects` in including class for faceting
-    has_multival_attribute :subject, ::RDF::Vocab::DC11.subject, solrize_for: :search
-
-    additional_search_index :doi_without_label, solrize_for: :exact_match,
```
murny: Did we mean to drop this? (if i recall this was for searching so they don't need to prefix `doi:` in their search)

mbarnett: It got relocated: https://github.com/ualbertalib/jupiter/pull/1272/files#diff-e1727bc7068cea61ff3939a69373ba9aR51

Every ActiveRecord model intended for Solr indexing has a corresponding "SolrExporter" that solely handles declaring how the model's attributes should be indexed into Solr.

app/models/draft_item.rb
```diff
@@ -1,5 +1,7 @@
 class DraftItem < ApplicationRecord
 
+  scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
```
murny: Do we need the `.or(where(is_published_in_era: nil)) }`? Default for this column is false

app/models/concerns/item_properties.rb
murny: Ah makes sense! Thanks 👍

app/models/draft_item.rb
mbarnett: In testing, it turns out some of the time a null gets directly written here for reasons I was never able to fully convince myself I tracked down. It basically boils down to being extra-safe with the three-valued logic, because it is weirdly possible in PostgreSQL to explicitly write a null into a column with a false default, and they aren't treated as identical in selects.

app/models/ar_item.rb
```diff
@@ -0,0 +1,188 @@
...
+  validate :publication_status_presence,
+           if: ->(item) { item.item_type == CONTROLLED_VOCABULARIES[:item_type].article }
+  validate :publication_status_absence, if: ->(item) { item.item_type != CONTROLLED_VOCABULARIES[:item_type].article }
+  validate :publication_status_compound_uri, if: lambda { |item|
+    item.item_type == CONTROLLED_VOCABULARIES[:item_type].article && item.publication_status.present?
+  }
+  validate :license_xor_rights_must_be_present
+
+  def self.from_draft(draft_item)
+    # item = ArItem.find(draft_item.uuid) if draft_item.uuid.present?
```
murny: Any reason for commenting this out? Seems like a good reason to have 🤔

app/models/draft_item.rb
mbarnett: Arguably though this column should just never have been nullable now that I think about it? I'm going to create a separate ticket to clean up the existing data and stick a non-nullable constraint on there, because it doesn't really make any sense as-is.

murny: Yeah makes sense 👍. Seperate ticket is good idea 😄 . Sure there might be some cleanup once we fully on ActiveRecord and data has been all safely ported over to get back to normal rails standards and don't have to be paranoid about stuff

app/models/ar_item.rb
mbarnett: Bug. Merge error from when I kept rebasing this damned work to stay on top of master, actually.

It's fixed in the second PR. I could pull out this branch and fix it here for posterity but the ripple effect of subsequent rebases would be nightmarish?

Any bugs or issues you find that aren't fixed on the tip of integration_postmigration, I'm going to have to fix in a separate PR...

murny: No worries. If its fixed then all good. I'll see it when I get to your next PR 😆

app/models/jupiter_core/solr_services/client.rb
```diff
@@ -0,0 +1,21 @@
+class JupiterCore::SolrServices::Client
+
+  include Singleton
+
+  SOLR_CONFIG = YAML.safe_load(ERB.new(File.read(Rails.root.join('config', 'solr.yml'))).result,
```
murny: Minor - Be nice to just get off this solr.yml file? and pull the url directly from `Rails.application.secrets.solr_url`

Less files/code, less problems

db/migrate/20190806193807_add_ar_community.rb
```diff
@@ -0,0 +1,19 @@
+class AddArCommunity < ActiveRecord::Migration[5.2]
```
murny: Probably future work once we off Fedora, but once we fully on activeRecord we should probably beef up these migrations? Probably biggest thing is adding `null: false` for things that are required. For example here, we have community title as a required field on the UI:
```
```

So this should also be reflected here:

```
t.string :title, null: false
```

db/migrate/20190806193807_add_ar_community.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    enable_extension 'uuid-ossp'
+    enable_extension 'pgcrypto'
+    create_table :ar_communities, id: :uuid,  default: 'uuid_generate_v4()' do |t|
```
murny: Interesting enough from google the collision rate on a UUID using V4 is like one in a billion so I guess we good 😆

db/migrate/20190814210221_add_is_era_published_column.rb
```diff
@@ -0,0 +1,6 @@
+class AddIsEraPublishedColumn < ActiveRecord::Migration[5.2]
+  def change
+    add_column :draft_items, :is_published_in_era, :boolean, default: false, index: true
```
murny: Minor: Is there any reason we didn't go with something generic like`published` vs `is_published_in_era`? If I recall a long time ago we wanted to keep Jupiter generic enough and ERA brand free so to speak

app/models/jupiter_core/solr_services/client.rb
mbarnett: Yeah that's a good point, I should be able to fold this into secrets entirely on the postmigration branch now that ActiveFedora is no longer relying on it.

test/models/draft_item_test.rb
```diff
@@ -24,20 +24,20 @@ def before_all
   end
 
   test 'should not be able to create a draft item without user' do
-    draft_item = DraftItem.new
+    draft_item = DraftItem.drafts.new
```
murny: Super minor: This probably doesn't matter but `.new`ing through the `.drafts` scope probably not needed 🤔 ?

`DraftItem.new` vs `DraftItem.drafts.new`?

db/migrate/20190814210221_add_is_era_published_column.rb
mbarnett: It might have been taken by Fedora? Maybe? We truly picked an awful column name.

test/models/draft_item_test.rb
mbarnett: At one point `DraftItem` was going to just become `Item` instead of creating a separate model, so I refactored everything to use the scope. It then became clear that there's so much messy legacy data that we'd have had to rip all of the validations off of Drafts, and that it was easier to just go the two model route for now. 

So yeah, currently, the scope is kind of vestigial.

mbarnett:
Closing this PR since it's been reviewed and the final work on this should be based on `integration_migration_phase1`

PR #1273, opened 2019-08-30
ActiveRecord Migration -- Part 2 [DON'T MERGE YET]

mbarnett:
DO NOT MERGE THIS PR UNTIL A RELEASE HAS BEEN CUT FROM MASTER CONTAINING https://github.com/ualbertalib/jupiter/pull/1272

This is Part 2 of the ActiveRecord Migration. It needs to be deployed immediately after running `bundle exec rake jupiter:get_me_off_of_fedora` on a deployment of Part 1. After deploying this, `bundle exec rake db:migrate` needs to be run again, to clean up various tables, drop the old integer ID columns on various tables and rename the new UUID ones, etc. Once the DB migration is complete, the web application should be fully functional again.

This PR largely just deletes everything based on ActiveFedora/LockedLdpObject, renames the Ar* models, and fleshes them out with the original functionality, and cleans up various spots in views or controllers where bits of active_fedora knowledge had leaked in. 

### Size

It's gigantic -- sorry about that -- but there were no half-steps that would still pass tests.

### System Test Flapping

This PR also makes the flapping problem in the system tests *much worse*. After digging into this for a couple of days, I've realized that the system tests are actually very badly written as they currently exist. Rails expects the system tests to operate off of fixtures that are loaded *once*, before all of the tests run, and expects that any extra data created by the system tests will be cleaned up after the test (or at least, that subsequent tests won't care).

Existing system tests create data constantly, don't clean up after themselves, and *do* break if data from another test is still around. Example: [This test file](https://github.com/ualbertalib/jupiter/blob/msb/update_views_controllers/test/system/communities_pagination_and_sort_test.rb) is a big offender. It creates 11 Communities, and then tests that those are the only 11 being given in search results -- BUT other tests also create communities, so depending on what random order tests run in there may be as many as 15 search results, which causes some assertions to fail. Conversely, if you decide to try to fix this by wiping the database between tests, [These tests](https://github.com/ualbertalib/jupiter/blob/msb/update_views_controllers/test/system/profile_index_test.rb#L6) will fail, because fixtures only load once at the start of all system tests, and will not reload again afterwards (I spent a good deal of time trying to hack around this, to no avail). With the database wiped, there is no user record for these tests to run against.

This all (accidentally)  wasn't so bad when most of the tests were based on ActiveFedora, and got wiped between tests, but with the data in ActiveRecord it becomes much more frequent that a random ordering of tests will leave stale data in a way that some other test breaks. Wiping postgresql between tests doesn't work, as it breaks the one-shot loading of fixtures that Rails does automatically, leading to different tests breaking.

The only fix here is to re-write these tests properly. As this is a pre-existing problem with the way tests are written that is just exacerbated by the different rules of ActiveRecord, that's out of scope for this PR, so I've tried to fix little bits where I could, marked a couple of the worst offenders as `skip`, and left the rest flapping. Fixing this with proper rewrites of these tests would be a good idea at some point.

### Performance

Hovers around 5-8x faster for views, tests, and seeding. The more write-intensive the workload (eg, seeding, which was pure writes to ActiveFedora and Solr previous, and Postgresql and Solr now), the higher the win (4 minutes and 20 seconds under ActiveFedora, 34 seconds under Postgresql, for the aforementioned seeds). 

### Other notes

Calls to `unlock_and_fetch_ldp_object` were largely removed, and otherwise replaced with `.tap` where necessary (mutation that needed to return the full object was a common pattern, so the replacement was clear). Some of these `.tap` calls are redundant but harmless, and can be removed at leisure.


app/models/collection.rb
```diff
@@ -1,44 +1,46 @@
...
-  # scenarios. See recover.rake for information on the particular problems this is causing and why we want to
-  # eliminate it.
-  additional_search_index :community_title, solrize_for: :sort,
-                                            as: -> { Community.find(community_id).title if community_id.present? }
-
-  def community
-    Community.find(community_id)
+class Collection < JupiterCore::Depositable
+
+  scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
```
murny: There is no column on `Collection` for `is_published_in_era` and don't think we have draft Collection/Communities?

I assume this shouldn't be here? Maybe copy paste?...or if it does we need to be scoping this from Items/Theses (member_items or member_theses) and this needs to get renamed accordingly (draft_items/draft_theses)?

Same comment applies to Community.rb as we have this scope there as well.

mbarnett: Good catch. I'll remove those.

mbarnett:
Similarly closing this, as it's been reviewed and the final results will come from merging _posmigration and master when we're ready to deploy.

Issue #1057, opened 2019-03-26
chromedriver-helper is deprecated after 2019-03-31. 


pgwillia:
Please update to use the 'webdrivers' gem instead.
See flavorjones/chromedriver-helper#83

chromedriver-helper would download the latest webdriver so that the tests just worked on travis. It doesn't look like the suggested replacement will do this for us ualbertalib/discovery#1574

ConnorSheremeta:
Webdriver now downloads the latest driver just like chromedriver-helper. Some workarounds needed for VCR though

PR #1294, opened 2019-09-10
Replace deprecated chromedriver-helper gem with webdrivers gem

ConnorSheremeta:
- Installed and configured webdrivers gem
- Updated travis distribution from trusty to xenial
Fixes: #1057


Gemfile
```diff
@@ -119,7 +119,7 @@ group :test do
   gem 'launchy'
   gem 'vcr', require: false
   # Easy installation and use of chromedriver to run system tests with Chrome
```
mbarnett: We might as well get rid of this comment, since it's no longer accurate.

test/test_helper.rb
```diff
@@ -16,6 +16,12 @@
 
   # Only want VCR to intercept requests to external URLs.
   config.ignore_localhost = true
+
+  # Allow drivers to download
```
mbarnett: Flip side, let's maybe expand this comment to link to the document that explains the need for this, mostly so that we can remember where we got this idea and/or find updated instructions for fixing the issue if things change https://github.com/titusfortner/webdrivers/wiki/Using-with-VCR-or-WebMock

Issue #1286, opened 2019-09-09
Refactor System Tests to Eliminate Order-Dependence


mbarnett:
### The Problem

The system test cases have been flapping for a while, and this is made much worse by https://github.com/ualbertalib/jupiter/pull/1273 -- this is being caused by the way we load ActiveRecord fixtures once at the beginning of a system test run, and the way that ActionDispatch::SystemTestCase *doesn't* truncate data from tables at the end of any given test file. Both of these behaviours interact poorly with the way we system tests are structured, eg) [CollectionsPaginationAndSortTest](https://github.com/ualbertalib/jupiter/blob/master/test/system/collections_pagination_and_sort_test.rb) [creates 11 collections](https://github.com/ualbertalib/jupiter/blob/master/test/system/collections_pagination_and_sort_test.rb#L9) and then [assumes those are the only 11 in the database](https://github.com/ualbertalib/jupiter/blob/master/test/system/collections_pagination_and_sort_test.rb#L19). 

This both causes the test to be unreliable (sometimes, some other test that creates a collection will have run before it, invalidating the assertion on line 19) AND creates unreliability for other tests (none of the 11 collections created here are deleted, so all hang around and can potentially invalidate assertions in tests that run *after* this).

On the flip side, if we were to manually run database cleaner in the setup phase of each file, this breaks other system tests like [DepositItemTest](https://github.com/ualbertalib/jupiter/blob/master/test/system/deposit_item_test.rb), which [relies on fixtures](https://github.com/ualbertalib/jupiter/blob/master/test/system/deposit_item_test.rb#L17). Because fixtures are loaded only once at the beginning of all tests, if the test database data has been deleted prior to this test running (in effect, if this test file is not the first one run), this call to load a fixture will error out.

This can only be resolved be fixing the inconsistencies with our approach to using data in system  tests. Likely the best approach is to let everything that can be loaded in fixtures be loaded that way, to write the tests in such a way as to minimize both the creation of new objects and to minimize assumptions made about those being the *only ones present*, and wherever appropriate, to have tests clean up after themselves.

### Short-term

If we're not going to remediate this immediately, we need to disable system tests on CI. An unreliable test is worse than a useless one.

mbarnett:
The remaining work on this is being tracked in a spike in clubhouse. Connor improved the situation with his PR, but there's still some flakiness in a couple of spots.

PR #1312, opened 2019-09-26
Modify Jupiter to use new ActsAsRDFable migration API

mbarnett:
Changes Jupiter to use the new ActsAsRDFable migration API from https://github.com/mbarnett/acts_as_rdfable/pull/1.

(Note that the Jupiter Gemfile in this PR currently points at the last commit on that PR – this should be updated prior to merging once that PR is merged).

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,4 +1,5 @@
 class Collection < JupiterCore::Depositable
+  acts_as_rdfable
 
   scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
 
@@ -16,15 +17,7 @@ class Collection < JupiterCore::Depositable
     self.visibility = JupiterCore::VISIBILITY_PUBLIC
   end
 
-  acts_as_rdfable do |config|
-    config.title has_predicate: ::RDF::Vocab::DC.title
-    config.fedora3_uuid has_predicate: ::TERMS[:ual].fedora3_uuid
-    config.depositor has_predicate: ::TERMS[:ual].depositor
-    config.community_id has_predicate: ::TERMS[:ual].path
-    config.description has_predicate: ::RDF::Vocab::DC.description
-    config.restricted has_predicate: ::TERMS[:ual].restricted_collection
-    config.creators has_predicate: ::RDF::Vocab::DC.creator
-  end
+  acts_as_rdfable
 
   def path
     "#{community_id}/#{id}"
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -18,8 +18,6 @@ class Collection < JupiterCore::Depositable
     self.visibility = JupiterCore::VISIBILITY_PUBLIC
   end
 
-  acts_as_rdfable
-
   def path
     "#{community_id}/#{id}"
   end
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -1,4 +1,5 @@
 class Collection < JupiterCore::Depositable
+
   acts_as_rdfable
 
   scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
--- a/db/migrate/20190926213733_annotate_tables_with_rdf.rb
+++ b/db/migrate/20190926213733_annotate_tables_with_rdf.rb
@@ -1,0 +1,89 @@
+class AnnotateTablesWithRdf < ActiveRecord::Migration[5.2]
+  def change
+    add_rdf_table_annotations for_table: :items do |t|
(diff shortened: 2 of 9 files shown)


app/models/collection.rb
```diff
@@ -1,5 +1,7 @@
 class Collection < JupiterCore::Depositable
 
+  acts_as_rdfable
```
pgwillia: What does calling `acts_as_rdfable` do without the configuration block?
https://github.com/mbarnett/acts_as_rdfable/blob/3f8471b93f943ed48cff644c70f8a603645ba441/lib/acts_as_rdfable/acts_as_rdfable_core.rb#L5-L15

db/migrate/20190926213733_annotate_tables_with_rdf.rb
```diff
@@ -0,0 +1,89 @@
+class AnnotateTablesWithRdf < ActiveRecord::Migration[5.2]
+  def change
+    add_rdf_table_annotations for_table: :items do |t|
```
pgwillia: This is really interesting.  So rather than the predicate being logic in the model, it's a property in a database?  All this is moving to an RdfAnnotation ActiveRecord model with table, column and predicate as attributes.

Do you expect that this will impact performance at all?

pgwillia:
How do you put the triple back together?  Before, I think it was 
`item, Item.attribute_metadata[:title], item.title`

mbarnett:
@pgwillia I'm still working on an actual API for that half of the functionality (the readback for rendering RDF-XML etc), but effectively the triple right now is `item, item. rdf_annotation_for_attr[:title], item.title`, where `rdf_annotation_for_attr` is [one of the methods injected by the ActsAsRDFable gem](https://github.com/mbarnett/acts_as_rdfable/pull/1/files#diff-576b2eaa1326d6daeb28547695e1b59cR12) when a model is marked `acts_as_rdfable`

pgwillia:
> @pgwillia I'm still working on an actual API for that half of the functionality (the readback for rendering RDF-XML etc), but effectively the triple right now is `item, item. rdf_annotation_for_attr(:title), item.title`, where `rdf_annotation_for_attr` is [one of the methods injected by the ActsAsRDFable gem](https://github.com/mbarnett/acts_as_rdfable/pull/1/files#diff-576b2eaa1326d6daeb28547695e1b59cR12) when a model is marked `acts_as_rdfable`

Thanks, that answers a couple of my questions.

db/migrate/20190926213733_annotate_tables_with_rdf.rb
mbarnett: Yeah, so the intention was always to keep the predicate data in the database rather than solely in the model, partly out of some idea that a snapshot of the database is always enough to be used to reconstruct the triples without needing reference to outside logic, and partly out of a practical desire to be able to run things like an OAI server as a separate service than the main webapp, without having to resort to tying the codebases together at the model layer.

The old codeblock prototype worked by loading the predicates into the DB on application boot, so to answer the performance question, this should perform better as predicates are only created or changed during migrations, instead of at every single application boot. That approach was easy, and gave us one central place to look for all predicate information, but it seemed to encourage a situation where every app that wanted to work with predicate data would need, say, its own `Item` model, with each app's `Item` having the exact same `acts_as_rdfable` block. If two blocks fell out of sync (say the predicate for title changed in one app but not the other), whichever app booted last would "win", and lead to a lot of inconsistency.

So the idea with the new approach is that we say "Jupiter is in charge of handling migrations for the ERA database", and the OAI application/engine is expected to only read from the DB, and ActsAsRDFable will gain an API to let it access the predicate data without needing a duplicate set of matching declarations.

pgwillia:
> You must use Bundler 2 or greater with this lockfile.
FYI I think there was some issue with bundler 2 deploying on our servers.

mbarnett:
Appears to be related to this: https://github.com/ualbertalib/jupiter/issues/1086 ? Guessing it might be related to the requirement to upgrade to RubyGems 3?

We should get some clarity on what "issues" there are, and get them resolved sooner rather than later. Bundler 1.X is effectively End-of-Lifed

mbarnett:
Holding off on merging until the Gem is merged and I can repoint the Gemfile at a proper tag

pgwillia:
I suspect your system tests are failing because the user fixtures aren't setup.

```
class ApplicationsSystemTestCase < ActionDispatch::SystemTestCase
  fixtures :users
end
```

mbarnett:
It's a known issue that we've made some progress on related to #1286. I'm currently looking into the remaining issues with the system tests, but we're not currently letting them block merges into the oai branch

PR #1335, opened 2019-10-22
Disable Rack::Attack

mbarnett:
We're seeing intermittent 429s in production that don't really make sense given the current Rack::Attack configuration, and the easiest way to rule out it mistriggering seems to be to take it out of production for the time being

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -39,7 +39,5 @@ module Jupiter
 
     config.redis_key_prefix = "jupiter.#{Rails.env}."
 
-    config.middleware.use Rack::Attack
-
   end
 end


mbarnett:
It would probably help if I commit everything before opening the pull request, in the future

mbarnett:
Aaaand I need to downgrade bundler

Gemfile.lock
```diff
@@ -579,4 +576,4 @@ DEPENDENCIES
   wicked
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
ConnorSheremeta: Weren't we getting issues with bundler 2 on staging/prod?

pgwillia: This change might delay getting this deployed.  What do you think about addressing this annoyance at another time?

mbarnett:
Bundler now downgraded

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+## [1.2.18] - 2019-09-24
```
pgwillia:
```suggestion
## [1.2.18] - 2019-10-23
```

mbarnett:
This is the smallest PR that ever needed like 18 different commits ;)

Issue #1354, opened 2019-11-15
Rollbar issue #367


weiweishi:
As part of the Epic #1297 to clean up the Rollbar error message, we'd need to investigate what has caused the issue, and whether this is a bug in the application. 

View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/367/](https://rollbar.com/ualbertalib/jupiter/items/367/)

```

NoMethodError: undefined method `owning_class' for nil:NilClass
  File "/var/www/sites/jupiter/app/models/jupiter_core/locked_ldp_object.rb", line 391, in reconnect_owning_jupiter_object!
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 74, in block (4 levels) in perform
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 74, in each
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 74, in block (3 levels) in perform
  File "/var/www/sites/jupiter/app/models/jupiter_core/locked_ldp_object.rb", line 66, in unlock_and_fetch_ldp_object
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 23, in block (2 levels) in perform
  File "/var/www/sites/jupiter/app/models/jupiter_core/locked_ldp_object.rb", line 66, in unlock_and_fetch_ldp_object
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 22, in block in perform
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 15, in open
  File "/var/www/sites/jupiter/app/jobs/file_attachment_ingestion_job.rb", line 15, in perform
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/activejob-5.2.3/lib/active_job/execution.rb", line 39, in block in perform_now
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/activesupport-5.2.3/lib/active_support/callbacks.rb", line 109, in block in run_callbacks
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/i18n-1.7.0/lib/i18n.rb", line 297, in with_locale
[...]

PR #1368, opened 2019-12-03
Delete the list node and indirect container proxy when you purge a fileset

pgwillia:
Our current application isn't cleaning up all the proxies when you purge
a fileset.  This results in an `owning_class` error when trying to
perform a FileAttachementJob. `unlocked_obj.ordered_members.to_a` is `[nil]`

Hydra-works is confusing.  I can see why this might have been missed the
first time around

#1354

### DIFF
--- a/app/models/concerns/item_properties.rb
+++ b/app/models/concerns/item_properties.rb
@@ -206,10 +206,15 @@ module ItemProperties
 
       def purge_filesets
         FileSet.where(item: id).each do |fs|
-          fs.unlock_and_fetch_ldp_object(&:delete)
+          fs.unlock_and_fetch_ldp_object do |fileset|
+            self.ordered_members.delete(fs) # delete the list node
+            self.members.delete(fs) # delete the indirect container proxy
+            fileset.delete # delete the fileset
+          end
         end
 
-        self.ordered_members = []
+
+        self.save
       end
 
       def push_item_id_for_preservation
--- a/app/models/concerns/item_properties.rb
+++ b/app/models/concerns/item_properties.rb
@@ -207,14 +207,13 @@ module ItemProperties
       def purge_filesets
         FileSet.where(item: id).each do |fs|
           fs.unlock_and_fetch_ldp_object do |fileset|
-            self.ordered_members.delete(fs) # delete the list node
-            self.members.delete(fs) # delete the indirect container proxy
+            ordered_members.delete(fs) # delete the list node
+            members.delete(fs) # delete the indirect container proxy
             fileset.delete # delete the fileset
           end
         end
 
-
-        self.save
+        save
       end
 
       def push_item_id_for_preservation
--- a/lib/tasks/jupiter.rake
+++ b/lib/tasks/jupiter.rake
@@ -35,17 +35,18 @@ namespace :jupiter do
     (Item.all + Thesis.all).each do |item|
       item.unlock_and_fetch_ldp_object do |uo|
         changed = false
+        indexes = []
         uo.ordered_member_proxies.each_with_index do |proxy, index|
           next if proxy.proxy_for
-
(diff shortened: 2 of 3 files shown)


lib/tasks/jupiter.rake
```diff
@@ -29,6 +29,25 @@ namespace :jupiter do
...
+        changed = false
+        uo.ordered_member_proxies.each_with_index do |proxy, index|
+          unless proxy.proxy_for
+            uo.ordered_member_proxies.delete_at(index)
+            puts "#{item.id} has a nil proxy"
+            changed = true
+          end
+        end
+        uo.save! if changed
+      end
```
pgwillia: This needs work.  Doesn't get all the nil proxies on first pass.

app/models/concerns/item_properties.rb
```diff
@@ -206,10 +206,15 @@ def add_communities_and_collections(communities, collections)
...
+          fs.unlock_and_fetch_ldp_object do |fileset|
+            self.ordered_members.delete(fs) # delete the list node
+            self.members.delete(fs) # delete the indirect container proxy
+            fileset.delete # delete the fileset
+          end
         end
 
-        self.ordered_members = []
+
+        self.save
```
pgwillia: This is necessary before `#ordered_members.to_a` is next called, but results in extraneous 
> Could not preserve 8e91fd6a-c0c2-4de9-80b7-4fe4fc1a228f
log messages.  See https://rollbar.com/ualbertalib/jupiter/items/270/
> This error occurs when the item is already in the preservation queue. This could be a succession of multiple saves within a short period of time

pgwillia:
I know you're busy @mbarnett but this WIP could use an early review to see if it makes any sense to you.

mbarnett:
I think the general fix makes sense (at least, to the extent that I've ever understood ActiveFedora proxies...)

I'm trying to figure out if we could dynamically clean this up on the fly rather than reprocessing Fedora. I'll mull it over, maybe we can chat about it on Friday

app/models/concerns/item_properties.rb
```diff
@@ -206,10 +206,14 @@ def add_communities_and_collections(communities, collections)
 
       def purge_filesets
         FileSet.where(item: id).each do |fs|
-          fs.unlock_and_fetch_ldp_object(&:delete)
+          fs.unlock_and_fetch_ldp_object do |fileset|
+            ordered_members.delete(fs) # delete the list node
```
pgwillia: rubocop didn't like the use of `self.ordered_members.delete(fs)`

app/models/concerns/item_properties.rb
```diff
@@ -207,14 +207,13 @@ def add_communities_and_collections(communities, collections)
...
-            self.members.delete(fs) # delete the indirect container proxy
+            ordered_members.delete(fs) # delete the list node
+            members.delete(fs) # delete the indirect container proxy
             fileset.delete # delete the fileset
           end
         end
 
-
-        self.save
+        save
```
pgwillia: rubocop didn't like `self.save` either.

lib/tasks/jupiter.rake
```diff
@@ -29,6 +29,27 @@ namespace :jupiter do
     puts 'Reindex completed!'
   end
 
+  desc 'cleanup orphaned proxies'
```
murny: Assume this is a one time thing just to catch the list nodes/proxy nodes we missed in the past? Or would we have to run this periodically?

pgwillia: Yes, exactly.  This is a one time thing just to catch the list nodes/proxy nodes we missed in the past.

weiweishi:
checked in with @mbarnett and agreed that we can merge this one in on master and deploy in production to help cleaning the data.

PR #1377, opened 2019-12-12
Create tests for list identifiers response [ch111]

ConnorSheremeta:


test/integration/oaisys_list_identifiers_test.rb
```diff
@@ -0,0 +1,359 @@
...
+                     item_type: CONTROLLED_VOCABULARIES[:item_type].article,
+                     publication_status:
+                       [CONTROLLED_VOCABULARIES[:publication_status].published],
+                     license: CONTROLLED_VOCABULARIES[:license].attribution_4_0_international,
+                     subject: ['Items']).tap do |uo|
+      uo.add_to_path(@community.id, @collection1.id)
+      uo.save!
+    end
+
+    # I hate having a sleep here but it is required for testing of both from and until arguments
```
murny: Could you explain a bit more about this? Just curious for my own understanding. Perhap we can think of a better way around the sleeps 😄

ConnorSheremeta: Because the date is based on 'record_created_at' (which is set automatically) they're too close together to actually be able to restrict the set with to and from arguments

mbarnett: Rather than using sleep here, maybe a cleaner way of doing it would be to use Rails' time-travel helpers to move one creation backwards or forwards in time, so that there's a distinct difference in the created at times? We do something similar here:

https://github.com/ualbertalib/jupiter/blob/integration_postmigration/test/models/item_test.rb#L432

There's some documentation on it here: https://guides.rubyonrails.org/testing.html#additional-testing-resources

test/integration/oaisys_list_identifiers_test.rb
```diff
@@ -0,0 +1,359 @@
...
+
+class ExpectArgsTest < ActionDispatch::IntegrationTest
+
+  include Oaisys::Engine.routes.url_helpers
+
+  setup do
+    @routes = Oaisys::Engine.routes
+  end
+
+  def before_all
```
murny: No change required just a mental note. In a post fedora world we should be able to get off using `before_all` hook for seeding/setting up tests (and hopefully remove this gem). Fixtures and before_each should be enough

test/integration/oaisys_list_identifiers_test.rb
```diff
@@ -0,0 +1,359 @@
+require 'test_helper'
+
+class ExpectArgsTest < ActionDispatch::IntegrationTest
+
+  include Oaisys::Engine.routes.url_helpers
```
murny: This is throwing errors on TravisCI?

```
/home/travis/build/ualbertalib/jupiter/test/integration/oaisys_list_identifiers_test.rb:5:in `<class:ExpectArgsTest>': uninitialized constant ExpectArgsTest::Oaisys (NameError)
```

Might have to colon colon the beginning of it? 
```
include ::Oaisys::Engine.routes.url_helpers
```
Or why else would this not be initialized? 🤔

ConnorSheremeta: This is happening because Oaisys isn't currently defined in the Gemfile. It will be defined once https://github.com/ualbertalib/jupiter/pull/1361 is merged. However, this tests will still fail until the two open PR's inside Oaisys are merged and the most recent commit is referenced in the gemfile

Issue #1376, opened 2019-12-12
post-migration failing tests


pgwillia:
Failure:
- [x] DepositItemTest#test_be_able_to_deposit_and_edit_an_item_successfully [/home/pjenkins/Code/jupiter/test/system/deposit_item_test.rb:117]:
```
Expected nil to not be nil.
```
- [ ] ItemEditTest#test_can_edit_item [/home/pjenkins/Code/jupiter/test/system/item_edit_test.rb:56]:
```
expected to find text "Item Deposit was successful!" in "Skip to Main Content\nSearch\nCollections\nDeposit\nAbout\nJohn Snow\nThis is a test message!\nEdit Item\nStep 1\nDescribe Item\nStep 2\nChoose License and Visibility\nStep 3\nUpload Files\nStep 4\nReview and Deposit Item\nStep 4 of 4\nReview and Update Item\nPlease review your item before depositing into ERA\nEdit\nUpload Files\nera-logo.png\n4.28 KB\nEdit\nChoose License and Visibility\nLicense\nAttribution 4.0 International (CC BY 4.0)\nVisibility\nOpen access (will be seen by everyone)\nEdit\nDescribe Item\nTitle\nBook of Random\nType of Item\nBook\nLanguages\nEnglish\nAuthors/Creators\nJane Doe\nBob Smith\nSubjects/Keywords\nBest Seller\nAdventure\nDate created\n2018-10-24\nDescription or Abstract\nReally random description about this random book\nCommunity and Collection\nFancy Community / Fancy collection\nBy depositing this item into ERA you agree to the ERA Deposit Policy agreements\nPrevious step\nSave\nAbout ERA · Contact Us · Policies · Powered by Jupiter"
```
- [x]  ItemShowTest#test_Search_faceting_on_item_values_is_not_broken [/home/pjenkins/Code/jupiter/test/system/item_show_test.rb:82]:
```
expected to find css "a[href='/search?tab=item']" 1 time but there were no matches
```
- [x] SearchTest#test_admin_should_be_able_to_filter_the_public_and_private_items [/home/pjenkins/Code/jupiter/test/system/search_test.rb:328]:
```
expected to find css "a.nav-link.active" but there were no matches
```
- [x] SearchTest#test_anybody_should_be_able_to_filter_the_public_items [/home/pjenkins/Code/jupiter/test/system/search_test.rb:118]:
```
expected to find css "a.nav-link.active" but there were no matches
```
[...]

PR #1378, opened 2019-12-12
WIP fix deposit item test

pgwillia:
missing some logic about creating embargo_history
move setup to fixtures
clean up after test that modifies index

#1376


pgwillia:
#1382

PR #1379, opened 2019-12-12
index truncation happens in between tests

pgwillia:
JupiterCore::SolrServices::Client.instance.truncate_index in test_helper
truncates the index created by setup and subsequent tests fail.  One workaround is to
turn these into massive tests until resolved

#1379


pgwillia:
#1382

PR #1382, opened 2019-12-13
1376 combined test fixes

pgwillia:
- remove skips 
- make some item and thesis fixtures and reuse them in similar tests
- Any test that creates an Item leaves a stale index entry which won't be found in the database which is cleaned on teardown of each test. 
-  fix deposit item test
missing some logic about creating embargo_history
move setup to fixtures
clean up after test that modifies index
-  fix admin users show test
The items created not through fixtures mess up the user in subsequent tests.
#1376

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -55,9 +55,16 @@ class Item < JupiterCore::Doiable
     item.description = draft_item.description
 
     # Handle visibility plus embargo logic
+    if draft_item.visibility_as_uri == CONTROLLED_VOCABULARIES[:visibility].embargo
+      item.visibility_after_embargo = draft_item.visibility_after_embargo_as_uri
+      item.embargo_end_date = draft_item.embargo_end_date
+    else
+      # If visibility was previously embargo but not anymore
+      item.add_to_embargo_history if item.visibility == CONTROLLED_VOCABULARIES[:visibility].embargo
+      item.visibility_after_embargo = nil
+      item.embargo_end_date = nil
+    end
     item.visibility = draft_item.visibility_as_uri
-    item.visibility_after_embargo = draft_item.visibility_after_embargo_as_uri
-    item.embargo_end_date = draft_item.embargo_end_date
 
     # Handle license vs rights
     item.license = draft_item.license_as_uri
--- a/app/models/jupiter_core/depositable.rb
+++ b/app/models/jupiter_core/depositable.rb
@@ -180,6 +180,13 @@ class JupiterCore::Depositable < ApplicationRecord
                    (changes['visibility'][1] != JupiterCore::VISIBILITY_PRIVATE)
   end
 
+  def add_to_embargo_history
+    self.embargo_history ||= []
+    embargo_history_item = "An embargo was deactivated on #{Time.now.getlocal('-06:00')}. Its release date was " \
+    "#{embargo_end_date}. Intended visibility after embargo was #{visibility_after_embargo}"
+    self.embargo_history << embargo_history_item
+  end
+
   def push_item_id_for_preservation
     if preserve == false
       Rails.logger.warn("Could not preserve #{id}")
--- a/test/fixtures/items.yml
+++ b/test/fixtures/items.yml
@@ -1,0 +1,26 @@
+fancy:
+  visibility: <%= JupiterCore::VISIBILITY_PUBLIC %>
+  owner: regular
+  title: 'Fancy Item'
+  creators: ['Joe Blow']
+  created: 'Fall 2017'
+  record_created_at: <%= 5.day.ago.to_s(:db) %>
+  languages: <%= [CONTROLLED_VOCABULARIES[:language].english] %>
+  license: <%= CONTROLLED_VOCABULARIES[:license].attribution_4_0_international %>
+  item_type: <%= CONTROLLED_VOCABULARIES[:item_type].article %>
+  publication_status: <%= [CONTROLLED_VOCABULARIES[:publication_status].published] %>
+  subject: ['Fancy things']
+  member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
+admin:
+  visibility: <%= JupiterCore::VISIBILITY_PUBLIC %>
+  owner: admin
+  title: 'Admin Item'
+  creators: ['Joe Blow']
+  created: 'Winter 2017'
+  record_created_at: <%= 5.day.ago.to_s(:db) %>
+  languages: <%= [CONTROLLED_VOCABULARIES[:language].english] %>
+  license: <%= CONTROLLED_VOCABULARIES[:license].attribution_4_0_international %>
+  item_type: <%= CONTROLLED_VOCABULARIES[:item_type].article %>
+  publication_status: <%= [CONTROLLED_VOCABULARIES[:publication_status].published] %>
+  subject: ['Ownership']
+  member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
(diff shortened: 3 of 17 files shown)


test/fixtures/items.yml
```diff
@@ -0,0 +1,26 @@
+fancy:
+  visibility: <%= JupiterCore::VISIBILITY_PUBLIC %>
+  owner: regular
+  title: 'Fancy Item'
+  creators: ['Joe Blow']
+  created: 'Fall 2017'
+  record_created_at: <%= 5.day.ago.to_s(:db) %>
```
pgwillia: This was important to get the objects to validate.

lagoan: Just out of curiosity, which part was needed for the objects to validate ? 
Was it  ```record_created_at: <%= 5.day.ago.to_s(:db) %>``` ?

test/system/admin_items_index_test.rb
```diff
@@ -4,50 +4,12 @@ class AdminItemsIndexTest < ApplicationSystemTestCase
...
-             publication_status: [CONTROLLED_VOCABULARIES[:publication_status].published],
-             subject: ['Ownership'])
-        .tap do |uo|
-      uo.add_to_path(community.id, collection.id)
-      uo.save!
-    end
+    # creating the index from the fixtures requires a save?
+    items(:fancy).save
+    items(:admin).save
+    thesis(:nice).save
```
lagoan: This is much more cleaner code!
I think fixtures just save the data on the database and do not trigger actions, factories would be following all of the actions though. Factories have come up in a couple of other conversations so we may want to take a look at them.

If anything, I would remove the question mark from the comment :D

test/system/admin_users_show_test.rb
```diff
@@ -153,47 +153,10 @@ class AdminUsersShowTest < ApplicationSystemTestCase
...
-             languages: [CONTROLLED_VOCABULARIES[:language].english],
-             license: CONTROLLED_VOCABULARIES[:license].attribution_4_0_international,
-             item_type: CONTROLLED_VOCABULARIES[:item_type].article,
-             publication_status: [CONTROLLED_VOCABULARIES[:publication_status].published],
-             subject: ['Ownership'])
-        .tap do |uo|
-      uo.add_to_path(community.id, collection.id)
-      uo.save!
-    end
+    # creating the index from the fixtures requires a save?
```
lagoan: Same as before, removing the question mark maybe mention the possibility of the future use of factories.

test/fixtures/items.yml
pgwillia: Yes.  Exactly that.  Probably others too, but this was the part I had to add when changing from Item.create to the fixtures.

test/system/deposit_thesis_test.rb
```diff
@@ -116,26 +104,30 @@ def setup
     _, item_results, _ = JupiterCore::Search.perform_solr_query(q: item_id, fq: 'id:' + item_id, rows: 1)
     assert_not_nil item_results.first['embargo_history_ssim']
     logout_user
+
+    # This method will add an item that exists in Solr with the member of paths matching the books/fantasy_books communities/collections fixtures
+    # in order for the other test to work we need to clean this out.
+    JupiterCore::SolrServices::Client.instance.truncate_index
```
lagoan: The way that I've written tests in the pasts would clean up the full system for each test, meaning the fixtures and indexes would be cleared so that each test be completely independent. What are your thoughts on using something like before and after for every test ?
The tests will take longer but we would know for sure the data is not bothered by other test.

test/system/profile_index_test.rb
```diff
@@ -55,49 +55,11 @@ class ProfileIndexTest < ApplicationSystemTestCase
...
-             languages: [CONTROLLED_VOCABULARIES[:language].english],
-             license: CONTROLLED_VOCABULARIES[:license].attribution_4_0_international,
-             item_type: CONTROLLED_VOCABULARIES[:item_type].article,
-             publication_status: [CONTROLLED_VOCABULARIES[:publication_status].published],
-             subject: ['Ownership'])
-        .tap do |uo|
-      uo.add_to_path(community.id, collection.id)
-      uo.save!
-    end
+    # creating the index from the fixtures requires a save?
```
lagoan: Quick edit on the comment as before to be consistent

test/system/search_test.rb
```diff
@@ -109,6 +109,11 @@ def setup
     end
   end
 
+  def teardown
+    # is clearing the database but not the index, for that it needs the following
+    JupiterCore::SolrServices::Client.instance.truncate_index
+  end
```
lagoan: Same as before, just want to discuss cleaning the system for each test run.

test/system/admin_users_show_test.rb
```diff
@@ -219,6 +182,7 @@ class AdminUsersShowTest < ApplicationSystemTestCase
     refute_selector 'div.jupiter-results-list li.list-group-item .media-body a', text: 'Nice Item'
 
     logout_user
+    JupiterCore::SolrServices::Client.instance.truncate_index
```
lagoan: I made some comments on other files for this PR starting a chat about cleaning all of the system for each test. Just continuing on that conversation for this line of code.

test/system/deposit_item_test.rb
```diff
@@ -2,24 +2,13 @@
```
lagoan: The test code looks so much cleaner using fixtures :D

test/system/deposit_thesis_test.rb
pgwillia: Yes, that would be ideal! @mbarnett  kind of outlines the challenges of going straight there https://github.com/ualbertalib/jupiter/pull/1273  What I tried to address with this PR was to get the tests to pass.  I acknowledge that there is much more that could/should be done with cleaning this up.

test/system/search_test.rb
pgwillia: I started to go down the road for adding this to the test helper but the tests that have setup with `before_all` would break and the problem became a bigger jumble.

test/system/admin_users_show_test.rb
pgwillia: There is system clean up between Test classes
https://github.com/ualbertalib/jupiter/blob/075e377c0a6b41665690dfc01d03833b03c5051e/test/test_helper.rb#L55-L61

app/models/item.rb
```diff
@@ -55,9 +55,16 @@ def self.from_draft(draft_item)
...
     # Handle visibility plus embargo logic
+    if draft_item.visibility_as_uri == CONTROLLED_VOCABULARIES[:visibility].embargo
+      item.visibility_after_embargo = draft_item.visibility_after_embargo_as_uri
+      item.embargo_end_date = draft_item.embargo_end_date
+    else
+      # If visibility was previously embargo but not anymore
+      item.add_to_embargo_history if item.visibility == CONTROLLED_VOCABULARIES[:visibility].embargo
+      item.visibility_after_embargo = nil
+      item.embargo_end_date = nil
+    end
```
pgwillia: @mbarnett this logic was missing.  I just wanted to check with you to make sure that this wasn't intentional. I remember reading that we were thinking of adding the papertrail gem to do some of this work, but it's not there yet, so I put this back in.

- So the test that was checking if an item goes from embargoed to not has it's embargo_history recorded would fail.  
- The validation would fail if the visibility was changed but there were still non nil values for `visibility_after_embargo` and `embargo_end_date`
`ActiveRecord::RecordInvalid: Validation failed: Embargo end date must be blank, Visibility after embargo must be blank`

app/models/jupiter_core/depositable.rb
```diff
@@ -180,6 +180,13 @@ def transitioned_from_private?
                    (changes['visibility'][1] != JupiterCore::VISIBILITY_PRIVATE)
   end
 
+  def add_to_embargo_history
+    self.embargo_history ||= []
+    embargo_history_item = "An embargo was deactivated on #{Time.now.getlocal('-06:00')}. Its release date was " \
+    "#{embargo_end_date}. Intended visibility after embargo was #{visibility_after_embargo}"
+    self.embargo_history << embargo_history_item
+  end
```
pgwillia: @mbarnett this is the balance of the missing embargo_history logic.

PR #1406, opened 2020-01-07
Looks like sprockets v4.0.0 upgrade changed to use manifest file heav…

murny:
…ily, which was not configured correctly

Should fix travisCI Builds.

More info here:
https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs
and issue with our problem with solution here:
https://github.com/rails/sprockets/issues/597


app/assets/stylesheets/items.scss
```diff
@@ -1,9 +1,6 @@
-$cc-background: $gray-100;
-
 .community-collection {
   .card {
-    // TODO: have real designer pick this
-    background-color: $cc-background;
+    background-color: $gray-100;
```
murny: Clean this up a bit. Don't need the secondary variable and TODO isn't really relevant anymore 😅

config/initializers/assets.rb
```diff
@@ -11,4 +11,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-Rails.application.config.assets.precompile += %w[admin.js]
+# Rails.application.config.assets.precompile += %w[admin.js]
```
murny: Apparently you don't do this anymore in Sprockets v4.0.0, they want you to use manifest instead.

app/assets/config/manifest.js
```diff
@@ -1,3 +1,5 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
-//= link_directory ../stylesheets .css
```
murny: So this is the problem. It was loading it in a order we didn't want. We want to load just the application.css file and the order that is specified by that file.

app/assets/config/manifest.js
```diff
@@ -1,3 +1,5 @@
...
+
+//= link application.css
+//= link admin.js
```
murny: Instead of `Rails.application.config.assets.precompile += %w[admin.js]` we put it here

Tested both quickly by running Rails s and looking at admin section. All seems good

PR #1412, opened 2020-01-09
WIP Rdfable active storage

lagoan:
This is to add rdf to ActiveStorage attachments and blobs. The AIP needs information from the Fileset and the file itself and this is an option to provide it through the API.

This work may conflict with the work done by @mbarnett so I want to double check this is the right approach.


lagoan:
The commit log is really messy. Will clean up and present only commits related to this matter.

mbarnett:
@lagoan Rather than waste your time  cleaning up and submit another PR just to clean up the commit history (small benefit to using branches on the main repo rather than the fork, you could just set the target branch to your other branch and have this PR show what's new in this branch rather than every diff between it and integration), here's what I was about to submit in review:

> Yeah, after you mentioned this in standup yesterday I took a guess at your use-case last night (would really have appreciated if we could have had this discussion in Slack yesterday, by the way, so that the rest of the team could have gotten eyes on it, and we could have headed off this duplication of effort at the outset) and am approaching this in a way that doesn't require us to reopen the class.

> On my end, I've added a method:

> ActsAsRdfable.add_annotation_bindings!(instance, table_name: nil) which adds the acts_as_rdfable bindings to the specific instance's metaclass. This way we don't pollute instances that don't need to deal with rdf annotations with methods they don't require, and we don't need to concretely create files to re-open each and every class that isn't under our control that we want to add bindings to (optional table name allowing us to bind annotations to classes that aren't even necessarily ActiveRecord classes. Still playing with this part...)

> The downside of the reopen approach is that as the internals of implementations we don't control change, class names can change, and the breakage due to the fact that we needed to rename files on our end is silent (the re-opened redefinitions silently become simple definitions of new classes) – I'm already 99% the existing monkey-patching on ActiveStorage::Blob is broken by 6.0, so that's something we're looking to get away from. By injecting into the metaclass of a targeted instance we're less dependent on implementation details outside of our control, like the class name, along with the generally tighter scoping of the patching. Hopefully this should be more resiliant to changes in Rails going forward.

> I'm switching gears back to serializing instances into the various formats this gem needs to handle (general rdf-xml, oai_dc, oai_etdms at the moment), so to be clear, just assume that that part will show up for you at some point and that you don't need to handle it for now (we can square the circle of any mismatch between what oaisys needs and what pmpy needs afterwards). If you run into any other usecases the above doesn't seem to cover, please mention them on Slack asap so we can stay coordinated. Thanks!

PR #1414, opened 2020-01-13
WIP AIP API to be used by pushmi_pullyu

lagoan:
This PR is a continuation of the initial PR #1370.

Changes include:
Fixes for the comments done for the initial PR.
Adding API entry point for items' filesets.

This is a small PR but wanted to get some feedback before moving forward with this side of the project.

### DIFF
--- a/app/controllers/aip/v1/items_controller.rb
+++ b/app/controllers/aip/v1/items_controller.rb
@@ -26,7 +26,7 @@ class Aip::V1::ItemsController < ApplicationController
     # table entries
 
     @item.files.each do |file|
-      fileset_url = request.original_url + "/filesets/#{file.fileset_uuid}"
+      fileset_url = "#{request.original_url}/filesets/#{file.fileset_uuid}"
       add_statement!(
         subject: subject,
         predicate: CONTROLLED_VOCABULARIES[:pcdm].has_member,
@@ -34,15 +34,13 @@ class Aip::V1::ItemsController < ApplicationController
       )
     end
 
-    # Add owners email seperatedly because is a relation with 2 steps in
-    # distance
-
-    owners_email = @item.owner.email
+    # Add owners email seperatedly because there is currently no predicate set
+    # to specify this relation directly
 
     add_statement!(
       subject: subject,
       predicate: RDF::Vocab::BIBO.owner,
-      object: owners_email
+      object: @item.owner.email
     )
 
     triples = @graph.dump(:n3)
@@ -51,6 +49,12 @@ class Aip::V1::ItemsController < ApplicationController
 
   def file_sets
     authorize @item
+
+    # The underscore for xml.uuid_ is intentional. Nokogiri builder makes
+    # use of method_missing to create its xml model. There could be
+    # problems with preexisting methods so we play it safe and add an underscore
+    # to avoid unintended behaviuor. You can find more info here:
+    # https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Builder
 
     builder = Nokogiri::XML::Builder.new do |xml|
       xml.file_order do
--- a/app/controllers/aip/v1/items_controller.rb
+++ b/app/controllers/aip/v1/items_controller.rb
@@ -1,0 +1,87 @@
+require 'rdf/n3'
+require 'rdf/vocab'
+import RDF
+
+class Aip::V1::ItemsController < ApplicationController
+
+  before_action :load_item, only: [:show, :file_sets]
+
+  def show
+    authorize @item
+    @graph = RDF::Graph.new
+    subject = RDF::URI(request.url)
+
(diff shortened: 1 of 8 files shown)


app/controllers/aip/v1/items_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+      fileset_url = request.original_url + "/filesets/#{file.fileset_uuid}"
+      add_statement!(
+        subject: subject,
+        predicate: CONTROLLED_VOCABULARIES[:pcdm].has_member,
+        object: fileset_url
+      )
+    end
+
+    # Add owners email seperatedly because is a relation with 2 steps in
+    # distance
```
mbarnett: Just curious, but I'm not sure I follow (or agree with) the comment here? 

```ruby
owners_email = @item.owner.email
...
object: owners_email
```
and
```ruby
...
object: @item.owner.email
```

are in a really strict sense, both equally violations of the "Law of Demeter". The really strict orthodox approach would be for `@item` to have a method call like `owner_email`  that then call's `owner.email` internally.

Whether or not you use an intermediate variable here, this controller ends up "knowing" about the internal structure of `item` and its relationship with `owner`, which is what the Law of Demeter suggests is the actual structural issue that complicates later changes to the system.

(all of that said it's not really a very egregious violation just because in practice the relationship isn't likely to change much. I just don't think the intermediate variable improves anything over just `object: @item.owner.email`)

app/controllers/aip/v1/items_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        predicate: rdf_annotation.predicate,
+        object: value
+      )
+    end
+
+    # handle files separately since they are not a part of the acts_as_rdfable
+    # table entries
+
+    @item.files.each do |file|
+      fileset_url = request.original_url + "/filesets/#{file.fileset_uuid}"
```
mbarnett: Not a big deal, but we might as well just use interpolation for both? `"#{request.original_url}/filesets/#{file.fileset_uuid}"`?

app/controllers/aip/v1/items_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    render plain: triples, status: :ok
+  end
+
+  def file_sets
+    authorize @item
+
+    builder = Nokogiri::XML::Builder.new do |xml|
+      xml.file_order do
+        @item.files.each do |file|
+          xml.uuid_ file.fileset_uuid
```
mbarnett: Is the trailing underscore on `uuid` here intentional? Might be worth adding a comment to document if so?

app/controllers/aip/v1/items_controller.rb
lagoan: I think the comment needs to be clarified. 

What I meant by the 2 steps has to do with the predicate. The Item model does not (as of now) have a predicate for ```RDF::Vocab::BIBO.owner``` that contains the value for the owners email. The email is on ```item.owner.email``` which needs to access 2 properties from item to get the value.

I agree that there is no need here for the temp variable. Will address this comment by passing the value directly.

While we are on the subject, will there be some sort of accessors from the acts_as_rdfable gem to fetch these types of values, will there need to be duplicates, or do you have something else in mind?

app/controllers/aip/v1/items_controller.rb
lagoan: You are right.

app/controllers/aip/v1/items_controller.rb
lagoan: Yes, the underscore is intentional.

From the [nokogiri builder documentation](https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Builder)

Special Tags: The builder works by taking advantage of method_missing. Unfortunately some methods are defined in ruby that are difficult or dangerous to remove. You may want to create tags with the name “type”, “class”, and “id” for example. In that case, you can use an underscore to disambiguate your tag name from the method call.

Thought to be better safe than sorry and not worry about an ```uuid``` method.

app/controllers/aip/v1/items_controller.rb
mbarnett: I'm currently leaning towards creating an intermediary presenter between the renderer and the model, so that the renderer can just make assumptions about what it needs, (eg call `presenter.email`), and the presenter acts as the thing that knows what to call on the model to get the relevant data, so that the gem doesn't make any assumptions about the model itself.

app/controllers/aip/v1/items_controller.rb
mbarnett: Ah yeah, that makes sense. Might be just to explain and point people at the nokogiri docs, because I will definitely forget about this in a week and ask the same question again ;)

app/controllers/aip/v1/items_controller.rb
lagoan: Nice!

PR #1421, opened 2020-01-15
Danger should not be installed in production. 

murny:
Update Gemfile to reflect this. Also added require false, as we don't need to load Danger.

Also changed `spec` to `test` in Dangerfile


murny:
@ConnorSheremeta. Thanks! And good reminder about Discovery. I can throw a quick PR up for Discovery as well 👍

PR #1429, opened 2020-01-23
Add rdf annotations to ActiveStorage::Blob

lagoan:
Files need to provide information through prefixes, the migration
included provides this functionality.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/mbarnett/acts_as_rdfable.git
-  revision: 1314de0a93f623b742fccb257bb7f2cf905fe675
+  revision: b3f9011b3c42c3bfd3bb2fe1d4f526dc30de401a
   tag: v0.2.1
   specs:
     acts_as_rdfable (0.2.1)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/mbarnett/acts_as_rdfable.git
-  revision: 1314de0a93f623b742fccb257bb7f2cf905fe675
+  revision: b3f9011b3c42c3bfd3bb2fe1d4f526dc30de401a
```
ConnorSheremeta: It's a little weird to see the Gemifile.lock pointing towards b3f9011b3c42c3bfd3bb2fe1d4f526dc30de401a while the Gemfile still [points towards the 0.2.1 tag](https://github.com/lagoan/jupiter/blob/add_active_storage_blob_rdf/Gemfile#L34) ([which is 29c415b](https://github.com/ualbertalib/acts_as_rdfable/releases).) Otherwise looks good!

lagoan: Good eye!

I had to change the revision so that it would use the gem's latest version using @mbarnett fix. We should move the tag to that revision on the acts_as_rdfable repo.

mbarnett: I think you should have permissions to tag it, but let me know if you don't.

(not actually certainly you can move a tag, but feel free to tag a 0.2.2)

lagoan: Just updated the acts_as_rdfable gem tag and jupiter's Gemfile file.

PR #1437, opened 2020-01-24
Set up papertrail gem

ConnorSheremeta:
- Set up papertrail gem
- Created callback
- Fixed bug with error on save due to logo_id

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,6 +15,8 @@ class ApplicationController < ActionController::Base
               ActiveRecord::RecordNotFound,
               ActionController::RoutingError, with: :render_404
 
+  before_action :set_paper_trail_whodunnit
+
   def service_unavailable
     head :service_unavailable, 'Retry-After' => 24.hours
   end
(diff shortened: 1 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -15,6 +15,8 @@ class ApplicationController < ActionController::Base
               ActiveRecord::RecordNotFound,
               ActionController::RoutingError, with: :render_404
 
+  before_action :set_paper_trail_whodunnit
```
murny: Forgive me if this is answered else where, but I am lacking context for myself. But are we not missing some other files to this PR? Don't we need to add papertrail to the gemfile, add versions table migration, `has_paper_trail` to Models, etc.? 

Looks like we just skipped to step 4: https://github.com/paper-trail-gem/paper_trail#1b-installation

ConnorSheremeta: I was intending to do it in two different PR's but I should've realised one would break without the other, it's all here now!

murny:
Looks pretty good 👍 .

Curious though if we can disable papertrail for the test suite? Probably unneeded hits on the database when running automated tests

They have a little guide here: https://github.com/paper-trail-gem/paper_trail#7a-minitest

PR #1441, opened 2020-01-27
Working version of Aip api

lagoan:
This PR includes the AIP API in working condition.

Changes will need to be made after some discussions with the metadata team regarding the predicates we use for each API file used to create the AIP.

Feedback is welcome from everyone.


app/controllers/aip/v1/aip_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+    render plain: triples, status: :ok
+  end
+
+  def download_file
+    send_data(
+      ActiveStorage::Blob.service.download(@file.blob.key),
+      disposition: 'attachment',
+      type: @file.blob.content_type,
+      filename: @file.filename.to_s
+    )
```
mbarnett: We might need to reconsider whether or not this is an efficient way of accomplishing this?

The blobs themselves live in a SAN, and some of them can get quite large (1 Gig+ files), which is only going to get bigger in the future. Rather than having the Rails application do all of the work of first handling the upload from the user and then a subsequent redownload of the file from Rails to PMPY, Rails could simply tell PMPY the location on the SAN for the files, saving a lot of slow preservation download traffic against the public-facing Rails app?

@weiweishi not sure whether the SAN is currently attached to the PMPY server right now?

app/policies/item_policy.rb
```diff
@@ -32,4 +32,11 @@ def thumbnail?
     download?
   end
 
+  # This policy are used for the AIP V1 API. Pundit does not allow use of
+  # namespaces in its policies
+
+  def file_sets?
+    admin? || user_is_authenticated_for_record?
+  end
```
mbarnett: Same concerns as above.

app/policies/thesis_policy.rb
```diff
@@ -32,4 +32,11 @@ def thumbnail?
     download?
   end
 
+  # This policy are used for the AIP V1 API. Pundit does not allow use of
+  # namespaces in its policies
+
+  def file_sets?
+    admin? || user_is_authenticated_for_record?
+  end
```
mbarnett: Same concerns as above.

app/policies/active_storage/attachment_policy.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def fixity_file?
+    admin? || user_is_authenticated_for_record?
+  end
+
+  def original_file?
+    admin? || user_is_authenticated_for_record?
+  end
+
```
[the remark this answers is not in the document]
lagoan: You are right on that we should keep records fully open to just the pmpy user. I misunderstood the policy for ```user_is_authenticated_for_record``` thinking it would only let an authenticated user with right permissions for a particular record from the line in ``depositable_policy.rb``

```
user_is_authenticated? && record_requires_authentication?
```
I need to go over the policies.

Makes sense to make it internal until we are certain that the API is in good shape/

Will make it admin only for now.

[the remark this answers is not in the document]
mbarnett: If you have an idea for a better name for `user_is_authenticated_for_record` I'm _definitely_ open to changing it, because I had to double-check the implementation myself. It's overly confusing.

test/controllers/aip/v1/aip_filesets_controller_test.rb
```diff
@@ -0,0 +1,104 @@
...
+  # Transactional tests were creating a problem where a collection defined as a
+  # fixture would only be found sometimes (a race condition?)
+  self.use_transactional_tests = false
+
+  def setup
+    @admin_user = users(:admin)
+    @regular_user = users(:regular)
+    @public_model = items(:fancy)
+    @private_model = items(:fancy_private)
+    @model = Item.name.underscore.pluralize
```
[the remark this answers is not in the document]
lagoan: I understand what you are saying. 

I kept these names from an initial idea where I would just test an item or a thesis since they behaved similar but that does not makes sense now that they are tested separately.

Will follow your suggestion.

test/controllers/aip/v1/aip_filesets_controller_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    graph = get_n3_graph(url)
+    assert_response :success
+    assert_equal true, graph.graph?
+  end
+
+  test 'should download image/jpeg file' do
+    file_path = Rails.root.join('tmp/storage/DO/OM')
+    FileUtils.mkpath file_path
+    FileUtils.cp file_fixture('blobs/DOOMik3dEPshhSpd4mkh9xpB'), file_path
```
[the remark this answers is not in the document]
lagoan: Yes, this is dependent on the behavior of ActiveStorage. Will look into ingesting a file these tests.

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -0,0 +1,63 @@
...
+  def setup
+    @admin_user = users(:admin)
+    @regular_user = users(:regular)
+    @public_model = items(:fancy)
+    @private_model = items(:fancy_private)
+    @model = Item.name.underscore.pluralize
+  end
+
+  test 'should be able to show a visible item' do
+    sign_in_as @regular_user
```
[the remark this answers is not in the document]
lagoan: I misunderstood your previous comment on policies. I was under the impressions that the API would be open and would show records based on the user's permissions. 

@weiweishi, who makes the call on this ?

test/controllers/aip/v1/aip_filesets_controller_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    url = aip_v1_model_fileset_fixity_url(
+      model: @model,
+      id: @public_model,
+      file_set_id: @public_model.files[0].fileset_uuid
+    )
+
+    graph = get_n3_graph(url)
+    assert_response :success
+    assert_equal true, graph.graph?
```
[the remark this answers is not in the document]
lagoan: Yes, they are simple tests for now. As I wrote in the comment above:
```
  # Basic tests checking if response has n3 serialization.
  # This should be changed to verify the response content is correct.
```
We will be sure of the response content after some discussions with the metadata team. I need to make this clearer.

test/fixtures/active_storage_blobs.yml
```diff
@@ -0,0 +1,23 @@
+# The attachment - blob relation does not work if the id is not
+# set manually on the blob and attachment fixtures. 
+# Need to revisit this code at a later time
+blob_1:
+  id: <%= ActiveRecord::FixtureSet.identify(:blob_1) %>
+  key: "DOOMik3dEPshhSpd4mkh9xpB"
```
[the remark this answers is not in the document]
lagoan: This makes sense. As before, will look into programmatically ingesting a file from within the test.

app/controllers/aip/v1/aip_controller.rb
lagoan: I see your concern for network load.

I assumed we would be using the permission mechanisms from the policies for this leaving the API open to the public.

If we are keeping all of the records available only to the pmpy user and we assume that the SAN will always be connected to the pmpy server we could follow your suggestion.

test/controllers/aip/v1/aip_items_contoller_test.rb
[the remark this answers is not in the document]
mbarnett: I think this would touch on a few people? 

We'd need to take a look at the contents of the AIP from a security perspective because there might be some data in there that isn't otherwise publicly exported, we'd need the Service Manager for ERA to want this to be something we were offering (and maybe some guidance on a period of time it is reasonable to support something like this), and our team for the implications of committing to certain things from a technical perspective (ie. how many/which serialization formats do we want to make publicly available and therefore have to put work into supporting for X amount of time)...

I'm just not sure any of this is in-scope at the moment, and my off-the-cuff take is that I'd rather standardize on a few forward-thinking industry-adopted formats (JSON-LD or something) rather than take on a commitment to n3, which seems pretty niche.

We could always decide to make some version of this public later, but walking it back once it's public is a different story if anyone sees it in the codebase and decides to start building some project around it.

config/routes.rb
```diff
@@ -117,4 +117,16 @@
...
+
+  # AIP API v1
+  namespace :aip, defaults: { format: :n3 } do
+    namespace :v1 do
+      get '/:model/:id', to: 'aip#show', as: 'model'
+      get '/:model/:id/filesets', to: 'aip#file_sets', as: 'model_filesets'
+      get '/:model/:id/filesets/:file_set_id', to: 'aip#file_set', as: 'model_file_set'
+      get '/:model/:id/filesets/:file_set_id/download', to: 'aip#download_file', as: 'model_fileset_download'
+      get '/:model/:id/filesets/:file_set_id/fixity', to: 'aip#fixity_file', as: 'model_fileset_fixity'
+      get '/:model/:id/filesets/:file_set_id/original_file', to: 'aip#original_file', as: 'model_fileset_original_file'
```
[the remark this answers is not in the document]
lagoan: Will make the change for this.

test/controllers/aip/v1/aip_filesets_controller_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  include AipActions
+
+  # Transactional tests were creating a problem where a collection defined as a
+  # fixture would only be found sometimes (a race condition?)
+  self.use_transactional_tests = false
+
+  def setup
+    @admin_user = users(:admin)
+    @regular_user = users(:regular)
```
murny: `@regular_user` never used again here in this file 🤔 ?

test/controllers/aip/v1/aip_filesets_controller_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    @public_model = items(:fancy)
+    @private_model = items(:fancy_private)
```
murny: `@private_model` never used either in this file.

app/controllers/aip/v1/aip_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+  # Return the url from the request to be used as the statement's subject for
+  # each rdf annotation for the requested digital object
+  def self_subject
+    RDF::URI(request.url)
+  end
+
+  def load_and_authorize_asset
+    case params[:model]
+    when Item.name.underscore.pluralize
+      @asset = Item.find(params[:id])
```
murny: super tiny nitpick: When I hear `@asset` in a Rails world, I think of css/javascript/images/etc. You name these things in the tests as `@model` instead of `@asset` as well. Maybe stick with `@model` considering the param is `model`?  or call it `object`? or `item`? etc?

app/controllers/aip/v1/aip_controller.rb
```diff
@@ -0,0 +1,260 @@
...
+  end
+
+  # Return the url from the request to be used as the statement's subject for
+  # each rdf annotation for the requested digital object
+  def self_subject
+    RDF::URI(request.url)
+  end
+
+  def load_and_authorize_asset
+    case params[:model]
```
murny: super tiny nitpick: Think Case statement here is a bit overkill for a conditional with 3 breaks. Unless you can expect 4+ more cases here (which be a good use case of a case statement)? 

```
if params[:model] == Item.name.underscore.pluarlize
  blah
elsif params[:model] == Thesis.name.underscore.pluarlize
  blah
else
  raise ActiveRecord::RecordNotFound
end
``` 
Is probably more clearer and reads easier.

lagoan: We are expecting more models down the road so we can leave this as cases.

murny: Sounds like a valid reason 👍. Ignore me then 😄

app/controllers/aip/v1/aip_controller.rb
lagoan: You are right. I will call it ``digital_asset`` to avoid confusion.

test/test_helper.rb
```diff
@@ -8,6 +8,7 @@
 require 'sidekiq/testing'
 require 'vcr'
 require 'webmock/minitest'
+Dir[Rails.root.join('test/controllers/concerns/*')].sort.each { |f| require f }
```
murny: See a couple problems with this and your concern. For two main reasons:

1. Naming. Can we call this what it is? It's a support helper/module for tests, not a controller concern. 
Your concern is just test helper methods for your AIP controller tests.  Which leads me to think this should go elsewhere. Maybe in a support folder. A support folder is a rspec best practice, which we can just steal for minitest/testunit. But it works exactly as what your doing here. e.g: `test/support/aip_actions_helpers.rb`. We can also adopt this idea and extract alot of the test helpers and configuration out of `test_helper.rb` to here too in the future? To me when I see something in `test/controllers` (or models/systems/etc) I expect them to be tests. Even if its under `/concerns`, It should be a test testing a real controller concern. But ultimately this is just a module to help test controller code. Hence why I think it should live somewhere else, such as a `support folder`

2. Here we are requiring every single module/concern from `test/controllers` into the test suite. We shouldn't require this for every single test? What if we running model tests or system tests which have nothing to do with controllers or needing these modules or even if its a controller test it might have nothing to do with AIPs.

test/fixtures/rdf_annotations.yml
```diff
@@ -0,0 +1,296 @@
+predicate_0:
```
murny: Are these even being used anywhere? Or will they? From a quick look nothings using them in our test suite

app/controllers/aip/v1/aip_controller.rb
lagoan: Will be naming the parameter and variables ``entity`` instead as it will be following a naming standard without conflicts in a Rails world.

test/fixtures/rdf_annotations.yml
lagoan: They are being used to check the output from the AIP API calls, the predicates are used to build the graphs through the ``rdf_annotations`` method.

config/routes.rb
```diff
@@ -117,4 +118,31 @@
   get '/rails/blobs/:key', to: redirect('/rails/active_storage/blobs/%{key}/thumbnail.jpg')
 
   match '/oai/(*all)', to: 'application#service_unavailable', via: [:get, :post]
+
+  # AIP API v1
+
+  namespace :aip,
+            constraints: EntityConstraint.new,
```
murny: Nice work with the constraint 👍

app/controllers/aip/v1/aip_controller.rb
```diff
@@ -0,0 +1,273 @@
...
+    :file_sets,
+    :file_paths
+  ]
+  before_action :load_and_authorize_file, only: [
+    :file_set,
+    :fixity_file,
+    :original_file
+  ]
+
+  skip_before_action :verify_authenticity_token
```
murny: Could you explain a bit about this? This is for CSRF tokens verification correct? Is there a reason we want to skip this check? 


This check shouldn't run on any of these controller actions anyways since they are all `get` requests. GETs should be safe and idempotent. You can see the underlying code here:

https://github.com/rails/rails/blob/f33d52c95217212cbacc8d5e44b5a8e3cdc6f5b3/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L233-L246

And you will notice it will never get ran, so we shouldn't need to skip this check manually here since its already skipped

lagoan: You are correct on this. I just made the change.

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -0,0 +1,156 @@
+require 'test_helper'
+require 'json-schema'
+require Rails.root.join('test/support/aip_helper')
```
murny: shouldn't need the `Rails.root.join` here? Should just be able to just use `require 'test/support/aip_helper'`?

```suggestion
require 'test/support/aip_helper'
```

lagoan: I tried that but I get the error ``LoadError: cannot load such file -- test/support/aip_helper``
Maybe I need to add it to a path somewhere ?

lagoan: This works:
```
require_relative '../../../support/aip_helper.rb'
```
Don't know if its better than the original though since it tells me where the file actually is relative to the project:
```
require Rails.root.join('test/support/aip_helper')
```

test/support/aip_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+              }
+            }
+          }
+        }
+      },
+      required: [:files]
+    }
+  end
+
+  def create_entity(
```
mbarnett: I'd suggest that we should add a TODO: mentioning that this can be removed when we settle on a consistent way of defining test data and explaining the issues you ran into when trying to replace this with fixtures, so that someone doesn't try to clean this up and just run right into the issues you've already discovered.

lagoan:
Fixtures continued to be a problem failing on some tests based on the seed used. Tests flip too much and it is something we need to address. 

For now we can instantiate entities on the setup method as needed. We can look back on this problem once we move away from minitest-hooks.

test/support/aip_helper.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    entity = entity_class.new(parameters).tap do |current_entity|
+      current_entity.add_to_path(community.id, collection.id)
+      current_entity.save!
+    end
+
+    Sidekiq::Testing.inline! do
+      entity.add_and_ingest_files(
+        files.map do |file|
+          File.open(file, 'r')
```
mbarnett: This is a bit of a non-obvious Ruby anti-pattern: you're actually leaking file handles (the OS never sees them as closed until the process ends).

The only guaranteed-safe way to use `File.open` is with a block:

```ruby
files.map do |file|
  File.open(file, 'r') do |file_handle|
    entity.add_and_ingest_files([file_handle])
  end
end
```

`file_handle` is guaranteed to be closed after the block even in the presence of exceptions or underlying OS interrupts hitting the process.

(I know this is less efficient in that you're ingesting in two separate calls rather than one, which is an issue with the ingest API, but it's easier to stick to the safe pattern)

lagoan: I did not know this. Will make the change right now.

test/support/aip_helper.rb
```diff
@@ -51,6 +51,15 @@ def file_paths_json_schema
...
   end
 
+  # TODO: We will very likely stop using the create_entity method once settle on
+  # a consistent way of defining data for our tests. Initially the entities were
+  # fetched from fixtures instead of being instantiated, however multiple tests
+  # for Item entity started flipping with this approach, likely having to do
+  # with the order the tests were run. The errors included:
+  # - No items on the database
+  # - Items not being found on database by id
+  # - Items defined to have files did not contain them
```
mbarnett: Nice!

test/controllers/aip/v1/aip_items_contoller_test.rb
murny: We can leave it as is 👍

lagoan:
Yes!

PR #1443, opened 2020-01-29
Addressing issues from PR #1441

lagoan:
From the commit message:

Add model route constraint
    
    A route constraint will bounce a request for an unknown model faster
    than catching it on the AIP API controller.
    
    Change AIP API to be private
    
    The AIP API will be closed off from the public and will only be used
    by admins for maintenance and the PMPY client.
    
    Rename model and asset references to entity
    
    There was a discrepancy in the naming conventions for parameters,
    methods, and variables throughout the AIP API that referenced objects
    in the system.
    
    References to these objects will be named entity in order to avoid
    naming issues with rails commonly used words and OOP definitions.
    
    Improve way helper module is loaded
    
    Move module from concerns directory to a support directory and handle
    it as a helper. Require it only when needed.
    
    Add AIP API file path entry point
    
    The PMPY client will not download files directly through the API,
    instead it will fetch the file path and access the file directly on the
    system.
    
    This implementation provides the path for a single fileset, this can be
    changed to provide the paths for all files to lower calls made to server
    
    Add json schema validator to entity tests
    
    Remove previous place holder test and now properly check for json
    response format for Item and Thesis entities.
    
    Add tests checking if files on entity actually exist


test/support/aip_helper.rb
```diff
@@ -0,0 +1,75 @@
...
+              }
+            }
+          }
+        }
+      },
+      required: [:files]
+    }
+  end
+
+  def create_entity(entity_class: Item, parameters: {}, files: [])
```
lagoan: This method will be removed once we move to parallel tests as we will be testing using only fixtures

test/controllers/aip/v1/aip_theses_controller_test.rb
```diff
@@ -0,0 +1,148 @@
...
+        owner_id: @admin_user.id,
+        dissertant: 'Joe Blow',
+        visibility: JupiterCore::VISIBILITY_PUBLIC,
+        graduation_date: '2017-03-31'
+      },
+      files: [
+        file_fixture('image-sample.jpeg'),
+        file_fixture('text-sample.txt')
+      ]
+    )
```
lagoan: This will be changed back to use fixtures and will follow @murny suggestions for handling files https://github.com/ualbertalib/jupiter/issues/1445#issuecomment-580441734

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -0,0 +1,156 @@
...
+        item_type: CONTROLLED_VOCABULARIES[:item_type].article,
+        publication_status: [CONTROLLED_VOCABULARIES[:publication_status].published],
+        license: CONTROLLED_VOCABULARIES[:license].attribution_4_0_international,
+        subject: ['Items']
+      },
+      files: [
+        file_fixture('image-sample.jpeg'),
+        file_fixture('text-sample.txt')
+      ]
+    )
```
lagoan: This will be changed back to use fixtures and will follow @murny suggestions for handling files https://github.com/ualbertalib/jupiter/issues/1445#issuecomment-580441734

config/routes.rb
```diff
@@ -117,4 +117,36 @@
   get '/rails/blobs/:key', to: redirect('/rails/active_storage/blobs/%{key}/thumbnail.jpg')
 
   match '/oai/(*all)', to: 'application#service_unavailable', via: [:get, :post]
+
+  # AIP API v1
+
+  entity_contraint_regex = /(?:#{Item.name.underscore.pluralize}|#{Thesis.name.underscore.pluralize})/
```
mbarnett: let's encapsulate this into a proper constraint class. The logic is only going to get more complicated over time as new models are added, and it's better to keep the routes file simple & declarative.

mbarnett:
Also going back to re-review some of the feedback on #1441, do we have a time frame on https://github.com/ualbertalib/jupiter/pull/1441#discussion_r371475778? If this is coming soon we might as well wait to make those tests useful before merging this. If we're expecting it's going to take a while before we have an idea of what graph contents to expect we need to create a ticket so that this work doesn't get lost

config/routes.rb
lagoan: Makes sense

lagoan:
> I've added a bit of feedback about the routing constraint, but this PR doesn't seem like it's ready for review yet (comments about re-instating Item fixtures) and I'm a bit confused about what's going on here?
> 

The tests currently work as intended. There is more work to be done for sure and at this point we can merge this code and move forward with another ticket to rework the tests. This will allow the new tickets to be smaller and easier to review. Hope this clarifies things.

> The cycle of review->changes specifically addressing feedback->re-review should all be happening in the same Pull Request, so that we can keep track of what was and wasn't addressed properly. Issues from PR #1441 should continue to be addressed on PR #1441.
> 
> I'm guessing the confusion here is with the discussion about closing PRs last week? That was specifically about _Work In Progress_ PRs where people were soliciting feedback, addressing it in the PR, and then leaving the PR open for an extended period of time with no apparent activity before requesting more feedback on significant new work that had been done in the meantime, turning it into a kind of endless rolling PR.
> 
> To the extent that we had any strong feelings on the subject, my understanding was that the consensus was that we all kind of vaguely felt WIP PRs should be closed after all feedback was addressed. I don't think anyone intended that all PRs (particular full, final PRs) be closed after feedback was given but prior to anyone assessing whether it was addressed or not.
> 
> Was that everybody else's understanding?

I get what you are saying. We may have understood different things from that conversation. @weiweishi is finalizing the document that came from that meeting and will be happy to adjust my workflow to the recommendations outlined there.

I have been closing the WIP PRs as the feedback was addressed and this is one of the points we discussed where we would be respectful of the person giving feedback and respond to the issued raised accordingly.

This is (hopefully) the last PR for this portion of work. The changes have been substantial and this gives a solid base for improvement down the road. For the next pieces I will make sure to stub out more sections of code to keep the PRs smaller and easier to digest.

I will reopen PR #1441 and we can work from there.

lagoan:
Closing this PR to open #1441. Will copy your comments over.

Issue #1430, opened 2020-01-23
Upgrade Rails gem to latest v6.x


murny:
- [x] Upgrade gem to Rails 6
- [x] Can run the update task: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task 
  - Can also check out http://railsdiff.org/ to see diffs of a `rails new` app throughout version changes
- [x] Run rails with the old defaults for now (in later issues we will tackle the bigger upgrades such as webpacker or zeitwerk)
- [x] Follow through official upgrade docs: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0 and fix any changes/deprecations
- [ ] Enable parallel testing (EDIT: Cannot complete until #1445 is taken on first)
- [ ] Maybe convert community/collection description textareas to use ActionText as a proof of concept? As these are admin only and be a good candidate 🤔 

Should have a working build of code on Rails 6 after the above is done (using sprockets instead of webpacker and no zeitwerk/etc)

murny:
PR has been merged and work completed (parallel testing work pushed off to another issue #1445. ActionText was a nice to have, but going to push this off until we have a real reason to use it)

PR #1450, opened 2020-01-31
#1430 Rails 6

murny:
Takes on most of issue #1430 

- Bump gem
- Update all generated files from `rails new` generator
- Cleanup deprecations
- Fix ActionStorage regression
- Tests are passing, so should give some confidence


app/models/jupiter_core/depositable.rb
```diff
@@ -126,7 +126,8 @@ def add_and_ingest_files(file_handles = [])
     raise 'Item not yet saved!' if id.nil?
 
     file_handles.each do |fileio|
-      attachment = files.attach(io: fileio, filename: File.basename(fileio.path)).first
+      files.attach(io: fileio, filename: File.basename(fileio.path))
```
murny: Heres the ActiveStorage regression.

Previous `.attach` would return all the attachments. Now it returns true or false if it was successful at attaching (much like ActiveRecord Update/Create/Delete methods): https://github.com/rails/rails/commit/28db8ba60e726d695cf35710cc43ea45966464e9#diff-60af6faf13016eb62a7885e5eecd9b34

So Added a line, to get the last attached file from `.attachments` association which should give the same functionality we had previously.

config/application.rb
```diff
@@ -39,8 +46,5 @@ class Application < Rails::Application
 
     config.redis_key_prefix = "jupiter.#{Rails.env}."
 
-    # Set an action on unpermitted parameters to raise an exception, used to validate parameters in Oaisys.
-    config.action_controller.action_on_unpermitted_parameters = :raise
```
murny: We define this in every environment, don't need to define it in application.rb as well. Deleted

test/controllers/downloads_controller_test.rb
```diff
@@ -58,7 +58,7 @@ def before_all
                            file_name: @file.filename)
 
     assert_response :success
-    assert_equal @response.content_type, 'text/plain'
```
murny: `.content_type` deprecated in the sense of how we using it here, infavor of `.media_type`

app/controllers/application_controller.rb
```diff
@@ -92,7 +90,7 @@ def render_404(exception = nil)
 
     respond_to do |format|
       format.html do
-        render file: Rails.root.join('public/404'), layout: false, status: :not_found
+        render file: Rails.root.join('public/404.html'), layout: false, status: :not_found
```
murny: Deprecation, they want the full absolute path now

app/models/jupiter_core/depositable.rb
mbarnett: The error handling definitely leaves something to be desired here. Partially because we had all those cross-datastore non-transactional issues previously. We should be able to abort things properly now?

config/environments/production.rb
```diff
@@ -85,8 +86,26 @@
...
+  #
+  # The `database_resolver_context` class is used by the middleware to set
+  # timestamps for the last write to the primary. The resolver uses the context
+  # class timestamps to determine how long to wait before reading from the
+  # replica.
+  #
+  # By default Rails will store a last write timestamp in the session. The
+  # DatabaseSelector middleware is designed as such you can define your own
+  # strategy for connection switching and pass that into the middleware through
+  # these configuration options.
```
mbarnett: Nothing to do with anything, but this is kind of cool

config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -0,0 +1,45 @@
...
+
+# Don't force requests from old versions of IE to be UTF-8 encoded.
+# Rails.application.config.action_view.default_enforce_utf8 = false
+
+# Embed purpose and expiry metadata inside signed and encrypted
+# cookies for increased security.
+#
+# This option is not backwards compatible with earlier Rails versions.
+# It's best enabled when your entire app is migrated and stable on 6.0.
+# Rails.application.config.action_dispatch.use_cookies_with_metadata = true
```
mbarnett: I'd vote in favor of turning this on. Decent explanation of why here: https://medium.com/@dylansreile/rails-6-0-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-586146f371e8

murny: Great article, learned a few things 👍. But looks like we already have it turned on 😄 

Looks like I can delete this file (will do in next PR). It's actually not doing anything and everything that is commented out here is actually enabled already. As I have already updated and running `config.load_defaults = 6.0` in our `application.rb`

More info here:
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults

app/models/jupiter_core/depositable.rb
murny: Wrapped logic in a `if attached` check, so if it wasn't attached then it just gets skipped. Obviously not ideal though! So also put a `#TODO` on it so we can come  back to it and do something smarter (raise error or something?)

config/initializers/new_framework_defaults_6_0.rb
murny: Removed this file and the 5.2 defaults file

Issue #1432, opened 2020-01-24
Transition to Zeitwerk for Autoloading


murny:
Update to Zeitwerk: https://github.com/fxn/zeitwerk
- [x] Hopefully minor to no work required
- [x] Cleanup any compatibility issues
- [x] Remove old dependency code (if we have require_dependencies or using eager_load_paths, etc)
- [x] Port over code and refactor our dependency of canonical-rails as this gem does not work with Zeitwerk. See comment (https://github.com/ualbertalib/jupiter/issues/1432#issuecomment-580604457)below for more info

murny:
Will probably need to port over the code we need for our canoncial urls as currently this [canonical-rails](https://github.com/jumph4x/canonical-rails) gem will not work with zeitwerk and they haven't fixed it yet: https://github.com/jumph4x/canonical-rails/issues/54

```
DEPRECATION WARNING: Initialization autoloaded the constant CanonicalRails::TagHelper.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload CanonicalRails::TagHelper, for example,
the expected changes won't be reflected in that stale Module object.

`config.autoloader` is set to `classic`. This autoloaded constant would have been unloaded if `config.autoloader` had been set to `:zeitwerk`.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /Users/murny/Code/jupiter/config/environment.rb:5)
```

For us it be should be very simple, we only need a few things for our canonical helper. So maybe something like this:
```
  def canonical_href
    "https://era.library.ualberta.ca/#{request_path}"
  end

  private

  def request_path
    return '' if request.path == '/'

    request.path
  end
```
Which is great, saves us from depending on another gem 😄 

Will look into this more when I tackle the issue.

murny:
This is now merged into master. Closing out issue

PR #1455, opened 2020-02-06
#1432: Rails 6 - Turn on zeitwerk

murny:
Cleanup inflections and eager loading, replace canonical-rails gem with simple helper


config/initializers/inflections.rb
```diff
@@ -16,4 +16,6 @@
 # end
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.acronym 'PMH'
+  inflect.acronym 'DOI'
```
murny: Zeitwerk is very particular about inflections. If you use `Doi` in one place, but `DOI` elsewhere then it will complain about one of them (more likely DOI as it doesn't know what that is as no inflections exist). Added inflections and updated code for DOI and URI offenders

app/views/layouts/application.html.erb
```diff
@@ -29,16 +31,13 @@
     <%#  Twitter Card - https://dev.twitter.com/cards/types/summary %>
     <meta name="twitter:card"        content="summary">
     <meta name="twitter:site"        content="@uofalibraries">
-    <meta name="twitter:url"         content="<%= request.base_url + request.fullpath %>">
+    <meta name="twitter:url"         content="<%= canonical_href %>">
```
murny: This is actually a canonical, using `request.fullpath` would give query params and other junk which we actually don't want here.

app/helpers/page_layout_helper.rb
```diff
@@ -57,4 +57,8 @@ def page_image
   def page_type(type = nil)
     type || 'website'
   end
+
+  def canonical_href(request_path = request.path)
+    "https://era.library.ualberta.ca#{request_path == '/' ? '' : request_path}"
```
murny: Simple helper we use to replace canoncial-rails gem. 

Why hardcode the domain? Well incase our staging environment goes live/public, it should continue to link that SEO back to our real domain (or if we drop our redirect to SSL, it should continue to give HTTPS, etc).

We weren't really using much else of canonical-rails, as you can see from the initializer which was commented out. We don't care about whitelisting query params, or anything else. So this simple helper will do all what we need.

config/application.rb
```diff
@@ -41,9 +38,6 @@ class Application < Rails::Application
     # Run skylight in UAT for performance metric monitoring pre-launch
     config.skylight.environments += ['uat']
 
-    # Finding jupiter_core code before the ApplicationController loads
-    config.eager_load_paths.prepend("#{config.root}/app/models/jupiter_core")
```
murny: Was causing issues, we shouldn't need this anymore 🤔

test/helpers/page_layout_helper_test.rb
```diff
@@ -83,4 +83,11 @@ class PageLayoutHelperTest < ActionView::TestCase
     assert_equal page_image, @community.thumbnail_url
   end
 
+  test 'canonical_href is returning the correct canoncial url' do
```
murny: No idea how to mock/inject into the `request` object here. Overriding `@controller.request` or `@request` values did nothing. So allowed an optional argument which I can pass in my path to make testing easier. These tests are pretty simple as a result. `site_for_bots_test.rb` is covering this much better in a real scenario though

app/helpers/page_layout_helper.rb
mbarnett: Sanity check: `root_url` gives me `http://localhost:3000/` locally and presumably `https://era.library.ualberta.ca/` in prod. Note the trailing slash. Should our canonical href match the hrefs we use for the root_url on page?

lagoan: Would it not be better to set the domain on an environment configuration file? Just in case we need to change it for whatever reason down the road.

test/helpers/page_layout_helper_test.rb
```diff
@@ -83,4 +83,11 @@ class PageLayoutHelperTest < ActionView::TestCase
...
+    assert_equal 'https://era.library.ualberta.ca', canonical_href(nil)
+    assert_equal 'https://era.library.ualberta.ca', canonical_href('/')
+    assert_equal 'https://era.library.ualberta.ca/search', canonical_href('/search')
+    assert_equal 'https://era.library.ualberta.ca/search/nested', canonical_href('/search/nested')
```
lagoan: Same as the comment for app/helpers/page_layout_helper.rb

app/helpers/page_layout_helper.rb
mbarnett: Also: completely agree on using the Prod domain as the href for staging etc, but wonder if we shouldn't pull it into the shared portion of the config? I can go either way on this -- we don't want or need it to vary by environment, and maybe someone would mistakenly change it in the config thinking it should change, but hardcoding it in the code instead of config just feels a bit weird at the same time.

config/application.rb
mbarnett: IIRC I think I had to do that because of some of the on-the-fly class generation I was doing to paper over ActiveFedora. This should (??) be safe now.

app/helpers/page_layout_helper.rb
murny: Yeah thats a good call. Can maybe create a constant for it somewhere. Question is where 🤔 (i18n, maybe application.rb?). We do technically set it in production.rb already (via Rails.application.routes.default_url_options so maybe) I can reach into that and reuse that

mbarnett: I like the idea of pulling it from the production.rb settings if we can do that, or setting it somewhere else and using it in production.rb. DRYer, less chance of accidentally changing one but not the other (if the domain ever changes, which granted isn't terribly likely)

murny: RE: trailing slash you are correct `canonical-rails` was doing this on any `:index` action by default. So if you had Posts routes, the canonical for index would be `example.com/posts/` and any other would not have a trailing slash, e.g show page be `example.com/posts/1`. Homepage is an index route hence `example.com/`, but other static content about/contact/etc are not (`example.com/about`) but search is an index `example.com/search/`). 

I prefer not having it, keeps code simpler and more consistent (all of our `link_to`/`href` link to the non-trailing slash versions, etc). canonical-rails likes it as it's more of a "directory" feeling. 

However impact on SEO from changing it, not sure 🤔 . Apparently google doesnt care about the root url:
image
But it would treat `example.com/search/` vs `example.com/search` differently 😢. So not sure a good solution for that. So we might need to redirect from trailing slash to non trailing slash to keep our SEO which might be even more work 😆

mbarnett: We should probably double-check with what Google Scholar expects (assuming that is literally ever documented anywhere...). I'm leaning stick with whatever we do currently, buuuut maybe we'd have more Scholar Juice if https://era.library.ualberta.ca/items/d8d783c1-b0bb-42cf-8266-497dfee4e1f1 looked like https://era.library.ualberta.ca/items/d8d783c1-b0bb-42cf-8266-497dfee4e1f1/, since there's a file that lives "below" that "directory" in the ancient way GS interprets URIs?

mbarnett: I'm almost positive that it's more work to consistently add slashes to everything, though, given previous experience with that

mbarnett: Actually you know what, this isn't an SEO ticket, we should probably just document this question and move on rather than try to work in improving our SEO juice here

lib/jupiter.rb
```diff
@@ -0,0 +1,3 @@
+module Jupiter
```
murny: The idea is stolen from other popular rails apps, e.g:

- Discourse: https://github.com/discourse/discourse/blob/master/lib/discourse.rb
- Gitlab: https://github.com/gitlabhq/gitlabhq/blob/master/lib/gitlab.rb

Basically a central place for any configuration/helpers/setup for your application. Using it here to store the constant for the production url instead of hardcoding it everywhere

app/helpers/page_layout_helper.rb
murny: Hardcoded URL changes have been fixed in last commit. Created an issue here https://github.com/ualbertalib/jupiter/issues/1463 for SEO questions as suggested by Matt

PR #1459, opened 2020-02-07
Set up admin view of papertrail edit history

ConnorSheremeta:
- Created partial and helper
- Added differ gem
- Added to fixtures
- Added tests

Screenshot:
Screenshot from 2020-02-07 09-49-21


app/helpers/admin/item_history_helper.rb
```diff
@@ -0,0 +1,96 @@
+module Admin::ItemHistoryHelper
+  def get_history(versions)
```
lagoan: This method is a bit large. What do you think about refactoring it to split out some of the logic ?

app/helpers/admin/item_history_helper.rb
```diff
@@ -0,0 +1,96 @@
...
+    history = []
+    versions.each do |version|
+      if version.changeset.present?
+        history_item = {}
+        history_item[:date] = version.created_at.in_time_zone('Mountain Time (US & Canada)').to_formatted_s(:long)
```
lagoan: This string is used twice in the method, maybe we can use a variable to keep it defined only once. 
A minor thing, I am used to split up method calls when the line gets long to make it more readable, but that is more of a preference thing I guess.

app/assets/javascripts/jupiter/item_show.js
```diff
@@ -5,11 +5,20 @@ $(document).on('turbolinks:load', function() {
   });
 
   $('#more-information-hidden').on('hidden.bs.collapse', function() {
-    $('.js-shown-text').toggleClass('d-none');
-    $('.js-hidden-text').toggleClass('d-none');
+    $('.js-mi-shown-text').toggleClass('d-none');
+    $('.js-mi-hidden-text').toggleClass('d-none');
```
lagoan: Since these two calls are used twice I would put them in another function as to not repeat the logic.

app/assets/javascripts/jupiter/item_show.js
```diff
@@ -5,11 +5,20 @@ $(document).on('turbolinks:load', function() {
...
   $('#more-information-hidden').on('shown.bs.collapse', function() {
-    $('.js-shown-text').toggleClass('d-none');
-    $('.js-hidden-text').toggleClass('d-none');
+    $('.js-mi-shown-text').toggleClass('d-none');
+    $('.js-mi-hidden-text').toggleClass('d-none');
+  });
+
+  $('#edit-history-hidden').on('hidden.bs.collapse', function() {
+    $('.js-eh-shown-text').toggleClass('d-none');
+    $('.js-eh-hidden-text').toggleClass('d-none');
```
lagoan: Same as previous comment to create another function.

app/assets/javascripts/jupiter/item_show.js
ConnorSheremeta: Good point, found another way to not repeat the logic

app/helpers/admin/item_history_helper.rb
ConnorSheremeta: I checked the default max method length in rubocop and it's only 10 lines! I'm way over, I've split it up.

app/assets/javascripts/jupiter/item_show.js
```diff
@@ -4,12 +4,13 @@ $(document).on('turbolinks:load', function() {
     $('.js-download').multiDownload();
   });
 
-  $('#more-information-hidden').on('hidden.bs.collapse', function() {
-    $('.js-shown-text').toggleClass('d-none');
-    $('.js-hidden-text').toggleClass('d-none');
+  $('#more-information-hidden').on('hidden.bs.collapse shown.bs.collapse', function() {
+    $('.js-mi-shown-text').toggleClass('d-none');
```
murny: Just curious what does `mi` and `eh` mean here?

ConnorSheremeta: More information and edit history. I thought the class would be too long otherwise.

murny: It's okay if its longer if it's more informative and helpful to other developers. I would have never have guessed that's what `mi` or `eh` means 😅 . Which means I can't just look at the class and know exactly what they are doing.

Expanding them out, e.g: `.js-more-information-shown-text`/`js-edit-history-shown-text` are not that terrible 🤔 . Or drop the `-text` to save some space... `js-more-information-shown` is pretty good too and still clearly indicates its meaning

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,42 @@
+<% if current_user.present? && current_user.admin? %>
+  <div>
```
murny: Empty divs are usually a sign we can just remove them. Unless you exploiting the div for a `display: block` which then indicate it as such `<div class='d-block'>` So probably can just render the button without a wrapping div here.

app/helpers/admin/item_history_helper.rb
murny: Maybe two thoughts here that could help clean up some of this. 

- Look into Draper (Matt brought this gem in recently). If this is all presentation logic around an Item/Thesis then this could just live in a presenter instead of a helper
- Is this all view logic? Some of this may be able to live in the model if it makes sense. To help you think about this, think If we exposed an API for changesets  (or if I wanted to see changesets for an item in the rails console) would you need to reuse this logic here in this helper to make this possible? If so then probably not just view logic, and should get put back into the model

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+        <dl>
+          <dt><%= 'Edited by' %></dt>
+          <dd><%= history_item[:user_info] %></dd>
+        </dl>
+
+        <% history_item[:changes].each do |attribute, change|  %>
+          <dl>
+            <dt><%= attribute %></dt>
+            <dd><%= raw Differ.diff_by_word(change[1].to_s, change[0].to_s).format_as(:html) %></dd>
```
murny: You may have to be careful here with `raw`? What would happen if a user edited a item's description? I would assume this would become part of the changeset here? What would happne then if that user inserted a javascript exploit (or other malicious code) right into the description field? Are we somehow escaping this?

Raw ignores the default escaping that Rails adds and now in the above example we would potentially putting this javascript malicious code right into the HTML without escaping it. Which means anyone viewing this HTML here (admins) could be in for a surprise.

app/views/items/_item.html.erb
```diff
@@ -105,3 +105,6 @@
 <div class="collapse" id='more-information-hidden'>
   <%= render partial: 'item_more_information' %>
 </div>
+
+<%# Edit History %>
```
murny: Kinda a useless comment as should be clear with the `<%= render partial: 'edit_history' %>` what is happening

app/helpers/admin/item_history_helper.rb
```diff
@@ -0,0 +1,119 @@
+module Admin::ItemHistoryHelper
```
murny: May be misleading as this helper isn't actually being used inside a `admin` namespaced view. It's being used within the item show page (just that its surround by an auth check)

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            aria-expanded="false"
+            aria-controls="edit-history-hidden"
+            class="btn btn-outline-secondary mb-3 js-edit-history-btn font-italic">
+    <span class="js-eh-hidden-text">
+      <%= t(:edit_history) %>
+      <%= fa_icon('chevron-down')%>
+    </span>
+
+      <span class="d-none js-eh-shown-text">
+        <%= t(:show_less) %>
```
murny: Should this be something else then `Show Less...`? Like `Hide Edit History` or something?

app/views/items/_edit_history.html.erb
ConnorSheremeta: No, Differ and Papertrail don't do any escaping. I'm now using sanitize to only allow the differ tags!

app/views/items/_edit_history.html.erb
ConnorSheremeta: I'm using a div to make it look proper, the button used to display beside the other one and the accordion contents weren't contained in the column. I've added `display: block` to the div. Is there a better way to make it look proper? Maybe.

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+      <span class="d-none js-edit-history-shown">
+        <%= t(:hide_edit_history) %>
+        <%= fa_icon('chevron-up')%>
+    </span>
+    </button>
+  </div>
+
+  <div class="collapse" id='edit-history-hidden'>
+    <% clean_history = clean_history(@item.edit_history)  %>
```
mbarnett: So this line is another, subtler indication that locating this logic in a helper violates the separation of concerns between Views and Models.

What we want is for the views to work with objects it doesn't really know anything about – we want our views to just blindly call `obj.clean_history.each ...` on the objects they're handed. But by locating the "cleanup" logic in a helper, we're forcing the view to "know" a few things that it shouldn't:

1) it knows that Items have a method `edit_history` that contains data that _isn't_ appropriate for display
2) it knows that it has to call a different method, `clean_history`, which takes the output of `edit_history` as an argument, in order to get data appropriate for display.
3) Similarly it knows that it needs to call `change_diff` to transform the values coming from the `history_item` hash.

This might sounds like a really trivial detail, and I probably sound a bit weird talking about what code "knows"; it's a pretty subtle idea, but any time you have something knowing something it shouldn't about some other part of the application, you've got a situation that's going to make future refactoring, or more generally just any future _change_ to the codebase, more difficult. The extreme case, where every part of the application knows everything about every other part of the application, is generally called The Big Ball of Mud pattern, and it is an absolute bear to make any changes to.

Even in the less extreme case, like this, you end up running into problems: one obvious one is that we might want to display this history in a different way on a new page: by centering this logic in the view, we're forced to repeat the logic in our second view – the violation of encapsulation has led us to end up writing code that isn't DRY, and to make it DRY, we end up having to fix the encapsulation issue. Again, anti-patterns cause more anti-patterns to cause up in reaction to the original one.

In a lot of ways one of the fundamental skills that we're always working on as programmers is learning how best to _minimize_ as much as possible the knowledge any part of the codebase has about the rest of the world. The better we get at minimizing that knowledge, the more flexible our code ends up being in the long run.

app/models/item.rb
```diff
@@ -180,4 +180,20 @@ def self.valid_visibilities
...
+      history_item = {}
+      history_item[:date] = version.created_at
+      history_item[:user] = version.whodunnit
+      history_item[:changes] = version.changeset
+      history_item[:version_id] = version.id
+
+      history << history_item if history_item[:changes].present?
+    end
+    history
+  end
```
mbarnett: This is the flip-side of the violation of concerns coming from the View knowing about the Model – here, the model is repackaging the version information for ease of consumption by the View. 

The Model here is forced to know too much about the View, because the View knows too much about the Model. One sign of an anti-pattern is that its presence in the code starts causing additional anti-patterns to crop up as a consequence.

test/application_system_test_case.rb
```diff
@@ -30,4 +30,18 @@ def logout_user
...
+    was_enabled_for_request = PaperTrail.request.enabled?
+    PaperTrail.enabled = true
+    PaperTrail.request.enabled = true
+    begin
+      yield
+    ensure
+      PaperTrail.enabled = was_enabled
+      PaperTrail.request.enabled = was_enabled_for_request
+    end
+  end
```
mbarnett: This is a really nice little way of building a reusable DSL for testing the versioning 👏

ConnorSheremeta:
Thank you for all the information Matt! It helped a ton!

app/models/item.rb
ConnorSheremeta: Should I even touch the models now that I'm using presenters?

app/assets/stylesheets/jupiter.scss
```diff
@@ -1,6 +1,8 @@
 $header-height-large-screen: 81px;
 $header-height-small-screen: 139px;
 $footer-height: 72px;
+$light-green: #d4fdd5;
+$light-red: #f99;
```
pgwillia: With a mind to accessibility perhaps red/green aren't the best choice despite being used a lot for showing change.
https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html

ConnorSheremeta: The diffing gem's default was a strike through for removed text and an underline for added text, I thought that it wasn't the most clear so I just used colors instead. Maybe its better of to have both the strike and underline along with the color? I've added a screenshot of what that would look like below. Do you think that would be better? That way someone could still interpret it without seeing the different colors.

Screenshot from 2020-02-18 12-48-46

pgwillia: Yes. I think the strike through and underline will give a visual cue for those that don't see the colour difference.

murny: Yeah i like strikethrough/underline. Looks good 👍. Maybe just check the color contrast on the red/green colors you choose on the background? You can use any accessibility checker for that. (Axe being the main one, but i think google lighthouse does it too now)

Should give you a report like so:
image


Bootstrap also provides red/green shades based off our red/green variable. Might recommend trying to reuse that as well. E.g: https://github.com/twbs/bootstrap/blob/master/scss/_variables.scss#L132-L140

ConnorSheremeta: I've added the strikethrough and underline. Axe didn't like the contrast on the red background, I updated the colors (based them off github's diff colors) and it now passes. For some reason I was unable to use bootstrap's color variants, hmm

app/decorators/item_decorator.rb
```diff
@@ -0,0 +1,59 @@
...
+      formatted_date = date.to_formatted_s(:long)
+      user_info = h.user_info(version.whodunnit)
+
+      { date: formatted_date, user_info: user_info, changes: changes }
+    end.compact.reverse
+  end
+
+  private
+
+  def format_changeset(changes, version)
```
mbarnett: A few things here:

1) The word 'format' is used heavily as a name for methods in this PR, but it's a super vague word that's not very descriptive of what the methods do. I think that's a big clue that this method is doing too many things – we're transforming all kinds of different values in all kinds of different ways, so we're forced to pick a very vague catchall word to try to describe it all. Ditto "clean" and "cleanup" -- above in `clean_history` we first "format changes" and then we run "cleanup changes" on the changes we just "formatted". It's not really clear what any of this _means_ when you read it, leading to more questions than answers (why does something we just "formatted" need to be "cleaned up"? etc).

2) Take it from someone who did a lot of Scala recently: nesting maps of maps is a really difficult set of functional transformations for the reader to keep track of in their head. This goes double when the inner transformations are being done _mutably_. When you're calling mutable functions on data during functional programming-style transformations, there's a very good chance you're going to accidentally do something you didn't mean to – for instance here, `changes.map` returns a _new_ hash of arrays, but because the inner `change` was _mutably mapped_ we've also simultaneously overwritten those values in the _original_ `changes` hash with the transformation, as well as putting the transformed values in the new hash. Any further code that tries to touch the `versions` attribute of this instance of the item will see your transformations instead of the original data, which is almost certainly going to be really unexpected and wasn't what you meant to do.

3) Both of the above make this just a really really complicated piece of code to read and understand the consequences of.

Here's my suggestion for one possible way to take a different approach (I haven't fully tried this out, but hopefully it's a jumping off point for different ways to approach this for you):

How about taking the concept of decorators and applying it Papertrail's changesets? You don't literally need to use Draper, but just the basic idea of creating a small class that wraps the changeset hash and has methods that access the hash and give back the appropriate values.

`ItemDecorator` could then have a method, `history`, which then returns an array of these wrapped changesets. So eg something like:

```ruby
class HumanizedChangeSet
  def initialize(helper, changeset)
    @h = helper
    @changeset = changeset
    @html_diffs = []
    generate_html_diffs
  end

  def visibility
     @changeset[:visibility].map {|visibility| @h.humanize_uri(:visibility, visibility)}
  end

  ... # etc all of the different attributes
  def html_diffs; @html_diffs; end
private

  def generate_html_diffs
    @changeset.keys.each do |attribute_name|
       # skip changes to these attributes – we don't care to present them
       next if ['logo_id', 'updated_at', 'embargo_history', 'aasm_state'].include?  attribute_name
       transformed_values = self.send(attribute_name)
       @html_diffs << Differ.diff_by_word(transformed_values[1].to_s, transformed_values[0].to_s).format_as(:html)
    end
  end
```

and then `ItemDecorator` just needs to do something like:

```ruby
class ItemDecorator < Draper::Decorator

def history
  object.versions.map do |version|
    HumanizedChangeSet.new(h, version.changeset)
  end
end
```

and then the view can just do:

```ruby
item.history.each do |changeset|
  ...render changeset.html_diff
end
```  

This code is untested, but you get the idea. What we're trying to do is, instead of running a big set of transformations over a hash and forcing the reader to imagine in their head what the end result of the hash is after several consecutive layers of transformation, we break apart and encapsulate the data and the different transformations individually, so that the overall flow of what's happening and what the end result looks like is easier to understand.

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,41 @@
...
+
+      <span class="d-none js-edit-history-shown">
+        <%= t(:hide_edit_history) %>
+        <%= fa_icon('chevron-up')%>
+    </span>
+    </button>
+  </div>
+
+  <div class="collapse" id='edit-history-hidden'>
+    <% @item.decorate.clean_history.each do |history_item|  %>
```
mbarnett: Whoops, forgot to add this comment to my original review:

We should probably follow Draper's best practices and call `.decorate` in the controller. As far as I can tell this should be safe, as we're still delegating everything to the underlying item, so the rest of the view shouldn't care (and if it does care, we might as well start fixing that now)

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+
+  def generate_html_diffs
+    @changeset.keys.each do |attribute_name|
+      # skip changes to these attributes – we don't care to present them
+      next if ['logo_id', 'updated_at', 'embargo_history', 'aasm_state'].include? attribute_name
+
+      transformed_values = begin
+        send(attribute_name)
+      rescue NoMethodError
+        @changeset[attribute_name]
```
mbarnett: This could use a comment explaining what the cases where we might expect this to happen are.

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+      end
+
+      html_diff = Differ.diff_by_word(transformed_values[1].to_s, transformed_values[0].to_s).format_as(:html)
+      sanitized_html_diff = @h.sanitize(html_diff, tags: ['del', 'ins'])
```
mbarnett: Probably a brief comment here as well, explaining where these tags come from and what you're doing here.

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+
+      attribute_name = case attribute_name
+                       when 'dissertant'
+                         'Author'
+                       else
+                         attribute_name.tr('_', ' ').titlecase
+                       end
```
mbarnett: And one more comment above line 48 explaining the 'why 'of this block

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+
+  def supervisors
+    transform_array_attribute(:supervisors)
+  end
+
+  def committee_members
+    transform_array_attribute(:committee_members)
+  end
+
+  def transform_visibility_attribute(attribute)
```
mbarnett: I'd suggest `humanize_visibility_attribute`? Kind of gets back to my thoughts about 'transform' being a very non-specific word that we should avoid if we can use a more specific description.

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+    @changeset[attribute].map { |visibility| @h.humanize_uri(:visibility, visibility) }
+  end
+
+  def transform_date_attribute(attribute)
```
mbarnett: `localize_date_to_timezone` or something?

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+    @changeset[attribute].map do |date|
+      next nil if date.nil?
+
+      date.in_time_zone(TIMEZONE).to_formatted_s(:long)
+    end
+  end
+
+  def transform_array_attribute(attribute)
```
mbarnett: Maybe `concat_array_attribute`? Again, just trying to replace `transform` with a more specific word that describes what the transformation _is_.

lib/humanized_change_set.rb
```diff
@@ -0,0 +1,226 @@
...
+  def created
+    @changeset[:created].map do |date|
+      next nil if date.nil?
+
+      @h.humanize_date(date)
+    end
+  end
+
+  def item_type
+    transformed_item_type = []
```
mbarnett: Last one, probably something like "humanized" or "localized" or "translated" rather than "transformed".

mbarnett:
It seems like the version of the target branch GitHub is comparing against is way out of date – like it thinks this PR is creating files that already exist in the target branch. We might need to close this an open a new PR to get it to un-stick or something?

ConnorSheremeta:
closing in favour of https://github.com/ualbertalib/jupiter/pull/1562 fell free to reopen if needed

PR #1491, opened 2020-02-21
Add file checksum to output from file paths

lagoan:
PMPY will be doing file checks when copying the files and it makes sense
to add the file checksum from active storage to the response so we:

- Calculate the checksum only once on PMPY
- Have different source for the checksum


lagoan:
This is a very small change that I don't think warrants a new CHANGELOG entry.

PR #1512, opened 2020-03-02
Rename methods to be more informative, fix bug with double rendering …

murny:
…page image url

Noticed this bug when working on webpacker. It works fine on an item/community, but broken on the default image url. Here is what this page_image helper out puts for default image url in production currently:

image

Notice the root url and image url being appended together which is not a valid url. 

Part of the problem i think is the confusion between `url` and `path`. So cleaned up those model methods to be more informative.

### DIFF
--- a/app/models/concerns/draft_properties.rb
+++ b/app/models/concerns/draft_properties.rb
@@ -46,7 +46,7 @@ module DraftProperties
     end
 
     # compatibility with the thumbnail API used in Items/Theses and Communities
-    def thumbnail_url(args = { resize: '100x100', auto_orient: true })
+    def thumbnail_path(args = { resize: '100x100', auto_orient: true })
       return nil unless thumbnail.present? && thumbnail.blob.present?
 
       Rails.application.routes.url_helpers.rails_representation_path(thumbnail.variant(args).processed)
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -26,7 +26,7 @@
     <meta property="og:locale"       content="en_US">
     <meta property="og:title"        content="<%= page_title %>">
     <meta property="og:description"  content="<%= page_description %>">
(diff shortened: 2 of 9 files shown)


app/models/concerns/draft_properties.rb
```diff
@@ -46,7 +46,7 @@ def thumbnail
     end
 
     # compatibility with the thumbnail API used in Items/Theses and Communities
-    def thumbnail_url(args = { resize: '100x100', auto_orient: true })
+    def thumbnail_path(args = { resize: '100x100', auto_orient: true })
       return nil unless thumbnail.present? && thumbnail.blob.present?
 
       Rails.application.routes.url_helpers.rails_representation_path(thumbnail.variant(args).processed)
```
murny: As you notice here we are returning the `rails_representation_path` instead of the `rails_representation_url` in this `thumbnail_url` helpers. Lets rename these to `thumbnail_path` as this is what its actually returning

app/views/layouts/application.html.erb
```diff
@@ -26,15 +26,15 @@
     <meta property="og:locale"       content="en_US">
     <meta property="og:title"        content="<%= page_title %>">
     <meta property="og:description"  content="<%= page_description %>">
-    <meta property="og:image"        content="<%= request.base_url + page_image %>">
+    <meta property="og:image"        content="<%= page_image_url %>">
```
murny: Move this `request.base_url` concatenation into the helper itself. This was causing the bug on the default images. Update the helper name to be more informative (`page_image` vs `page_image_url`)

test/helpers/page_layout_helper_test.rb
```diff
@@ -2,6 +2,16 @@
 
 class PageLayoutHelperTest < ActionView::TestCase
 
+  attr_reader :request
```
murny: We don't have access to `request` inside helpers. Some setup to fake this

PR #1518, opened 2020-03-03
Thesis description optional prior 2009

ConnorSheremeta:
If graduation date is prior to 2009 the description is now optional. Added system and model tests.

### DIFF
--- a/app/models/draft_thesis.rb
+++ b/app/models/draft_thesis.rb
@@ -34,9 +34,7 @@ class DraftThesis < ApplicationRecord
 
   validates :description, presence: true,
                           if: [:validate_describe_thesis?,
-                               lambda do |draft_thesis|
-                                 draft_thesis.graduation_year.nil? || (draft_thesis.graduation_year >= 2009)
-                               end]
+                               :description_required?]
 
   validate :communities_and_collections_presence,
            :communities_and_collections_existence,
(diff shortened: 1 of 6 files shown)


app/views/admin/theses/draft/describe_thesis.html.erb
```diff
@@ -54,7 +54,7 @@
 
     <%= f.input :description,
                 as: :text,
-                required: true,
+                required: @draft.graduation_year.nil? || (@draft.graduation_year >= 2009),
```
mbarnett: We should pull this logic into the model, I think? `@draft.description_required?` or something

app/models/draft_thesis.rb
```diff
@@ -29,10 +29,15 @@ class DraftThesis < ApplicationRecord
...
   belongs_to :institution, optional: true
 
-  validates :title, :description, :creator,
-            :member_of_paths, :graduation_year,
+  validates :title, :creator, :member_of_paths, :graduation_year,
             presence: true, if: :validate_describe_thesis?
 
+  validates :description, presence: true,
+                          if: [:validate_describe_thesis?,
+                               lambda do |draft_thesis|
```
mbarnett: We could pull this into a method that can be shared with the view, below.

PR #1519, opened 2020-03-04
Change docker-compose defaults, update Docker compose, Dockerfile, tr…

murny:
…avis, and README accordingly

Big changes:
- Remove all fedora containers from our docker setups
- Switched `docker-compose.lightweight.yml` and `docker-compose.yml` around.   Now `docker-compose.yml` and `docker-compose.development.yml` respectfully 
- Removed hacks to run test suite in `docker-compose.development.yml` (previously `docker-compose.yml`)
- Updated our Dockerfile to run again. Now runs Ruby 2.6, Bundler 2.1.4. Includes other minor fixes and updates
- `docker-compose.yml` (previously  `docker-compose.lightweight.yml` )  now runs all data stores by default (redis/postgres/solr).
- Postgres containers now running v11 instead of v9.6
- Renamed `docker-compose.deployment.yml` to `docker-compose.production.yml`
- TravisCI now only runs solr docker container, instead of full  `docker-compose.lightweight.yml` setup

How to run new simplified `docker-compose.yml`:
- Make sure you clear out old and existing volumes. `docker-compose down -v`
- If you have postgres/redis running locally, turn them off (e.g: On Linux: `sudo service postgresql stop`/`sudo service redis stop`  On Mac: `brew services stop postgresql`/`brew services stop redis`). As we will now use postgres/redis via docker.
- Remove `export JUPITER_DATABASE_URL=postgresql://jupiter:mysecretpassword@127.0.0.1?pool=5` as no longer needed
- Now just run as normal... `docker-compose up -d`, prepare db, go to localhost:3000. All should be good to go! 

Tested both `docker-compose.yml` and `docker-compose.development.yml` all works great locally.


.travis.yml
```diff
@@ -17,19 +17,17 @@ cache: bundler
...
-  - bundle exec danger
-  - docker version
-  - docker-compose version
-
-  - docker-compose -f docker-compose.lightweight.yml up -d
-
-  - docker ps
-
+  - >
+    docker run -d -p 8983:8983
```
murny: We only care about solr docker for travis, so just run this separately (outside of docker-compose)

README.md
```diff
@@ -99,31 +96,11 @@ Now everything is ready, you can go and view Jupiter! Just open your favorite br
 
 (Note: ip address may be different if you are using `docker-machine`)
 
-## Want to run the test suite in docker?
```
murny: I really don't see much value here with having someone being able to run the test suite in the development docker setup.  The vision I have for the `development` docker compose is just a quick way for someone to set up Jupiter and be able to run demo it which it does a perfect job at doing this. 

To get the test suite running in a full docker setup is quite tricky and requires a lot of workarounds (complicated hacks around test code, hacks around selenium and VCR) for very little gain.  I doubt anyone actually uses this full docker setup for testing? (and if you do I'd push you to use the normal one, as its very easy to install rails/ruby on your local). 

So as a result, removed all hacks around this and removed the documentation around this as well. Which greatly simplifies our docker setup and test suite. But I am open to opinions.

config/database.yml
```diff
@@ -5,15 +5,19 @@
 # `sudo -u postgres alter user ${USER} superuser
 default: &default
   adapter: postgresql
-  encoding: utf8
-  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  encoding: unicode
```
murny: Not sure if i can change this? Unicode is the default for rails. But for some reason we have it set as `utf8` 🤔

config/database.yml
```diff
@@ -22,13 +26,6 @@ test:
   <<: *default
   database: jupiter_test
 
-  # TODO: Review this later
```
murny: No longer needed 🎉

Dockerfile
```diff
@@ -1,28 +1,34 @@
...
+# Install standard gems
+COPY Gemfile* /app/
+RUN bundle config --global frozen 1 && \
+    bundle install -j4 --retry 3
 
 # *NOW* we copy the codebase in
 COPY . $APP_ROOT
 
+# Add user
+RUN addgroup -g 1000 -S app \
```
murny: This should fix all files created in full docker-compose being owned by `root` and all the issues around that (pids being left around, temp files not being able to get cleared, etc etc)

.travis.yml
lagoan: I have very basic knowledge on docker so I would like to take this as a learning opportunity. What do you mean by only caring about solr docker for travis ? I see that we are still referencing the image in docker-compose.

docker-compose.yml
```diff
@@ -1,82 +1,30 @@
...
-      - solr
-      - selenium
-
-  # Rails server
-  web:
-    <<: *rails
-    command: bin/docker-start
     ports:
-      - "3000:3000"
+      - '6379:6379'
```
lagoan: Why did you change the port?

.travis.yml
murny: Previous we were running `docker-compose` using the `docker-compose.lightweight.yml` file (which by default only ran fedora/solr).

Here we don't need to run the `docker-compose` (as we have also changed it, so it would try to setup redis/postgres which would conflict with travis`) we only need to run solr docker container. So I am JUST using docker here. No docker-compose at all. I am pulling down this image https://github.com/docker-solr/docker-solr and configuring it myself outside of docker-compose.

docker-compose.yml
murny: Confusing diff. If you look its referencing the redis image. So nothing actually changed (other then I am opening the port on redis)

README.md
mbarnett: Bit of a niggle, but if developers are mostly expected to use `docker-compose.yml` and `docker-compose.yml` and `docker-compose.development.yml` is more for running a full demo environment that isn't usefull for development because you can't easily run tests, should we rename the file to something like `docker-compose.demo_env.yml` to avoid confusion when onboarding?

murny: Yeah we could do that, call it `docker-compose.demo.yml` or something? I'm terrible with naming.

murny: Updated the name of this and added more info for a new developer to set up their machine to the README (needs more love though 😬)

config/database.yml
pgwillia: Unicode is just an alias for UTF8 in [Postgresql](https://www.postgresql.org/docs/9.1/multibyte.html) so I think this change would be okay.  I ran into an issue where the character set/encoding was different in the rails config than the database in Mysql and Discovery and it wouldn't run migrations correctly.  If you look at the history this project was initialized with Mysql and at the time the default was utf8 (now utf8mb4).

test/application_system_test_case.rb
```diff
@@ -3,14 +3,7 @@
...
 
-  if ENV['CAPYBARA_NO_HEADLESS']
-    # Set options if you have a special selenium url (like if your running selenium in a docker container)
-    # Otherwise just use the defaults by providing empty hash
-    options = ENV['SELENIUM_URL'].present? ? { url: ENV['SELENIUM_URL'] } : {}
-    driven_by :selenium, using: :chrome, screen_size: [1400, 1400], options: options
-  else
-    driven_by :selenium_chrome_headless
-  end
+  driven_by :selenium, using: :headless_chrome, screen_size: [1400, 1400]
```
pgwillia: Aw man!  I use CAPYBARA_NO_HEADLESS for debugging tests.

README.md
```diff
@@ -77,72 +79,87 @@ cd jupiter
...
-(See other sections of this README for more information about developing in a rails project environment)
+## Step 3: Open and view Jupiter!
+Now everything is ready, you can go and view Jupiter! Just open your favorite browser and go to the following url:
+
+[localhost:3000](http://localhost:3000)
+
+(Note: ip address may be different if you are using `docker-machine`)
 
-## For deployment (on UAT environment)
+## For production deployment (on UAT environment)
```
pgwillia: Did you deploy to UAT?

config/database.yml
murny: Yeah I confirmed with matt before merging this. Matt found the same thing that its just an alias 👍 . He also mentioned the same thing about Mysql which it was quite special with locales 😆. Just glad we can make the switch without any issues 😓

test/application_system_test_case.rb
murny: I have no problems bringing this back! Should I put up a PR for this?

README.md
murny: Nope! Never did (don't think I can ssh from home either). I tried not to change much for the deployment part. Nothing changed from the docker-compose yml. Only the underlying Dockerfile was updated. So hopefully no breaking changes here

PR #1521, opened 2020-03-04
Refactor javascript code for fallback thumbnail, use something more simplier

murny:
So this was an issue I was fighting with for awhile on the webpacker PR. 

The problem is we want this `default_thumbnail()` method setup as a global method and exposed to the entire app (e.g you can call it from browser console, or any html). This goes against Webpackers defaults and is a major pain to get something like this to work in webpacker. I tried to replace this with jquery and watching for `.error/.load` events on the image but it works half the time (works on turbolinks page loads, but on a hard refresh it fails as image fails to load before javascript loads and fires). Also looked into using a library like https://github.com/desandro/imagesloaded but didn't get much success either. Other option is opening a `<script></script` tag and throwing this function inside to by pass webpacker...but thats messy as well.

Instead of spending a ton of time on this, I just took a simple approach. Just fallback to the ERA logo without any text on it. We do this for community thumbnails already, so figured this be okay here as well. 

Tested this and it all works quite nicely. If we cannot load an image we just get the ERA logo without text as such:

image

### DIFF
--- a/app/assets/javascripts/jupiter/thumbnail.js
+++ b/app/assets/javascripts/jupiter/thumbnail.js
@@ -1,8 +1,0 @@
-// Replace thumbnail with default thumbnail on error
-function default_thumbnail($element) {
-  if ($element.parentElement.getElementsByClassName("img-thumbnail").length > 1) return;
-  $element.error = null;
-  $element.style.display = 'none';
-  $thumbnail_html = "<div class='text-muted text-center img-thumbnail p-3'><i class='fa fa-file-o fa-5x'></i></div>";
-  $element.insertAdjacentHTML('beforebegin', $thumbnail_html);
-}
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/jupiter/thumbnail.js
```diff
@@ -1,8 +0,0 @@
-// Replace thumbnail with default thumbnail on error
-function default_thumbnail($element) {
```
murny: Some future feedback on this javascript. 

It's a best practice to create variables with `$` in them, when they are jquery selectors.

So for example,  `const $imageSelector = $('img');` is a good practice. Then by quickly looking at the variable you know if you can use Jquery methods on them or not!

But these `$element` and `#thumbnail_html` variables are not a jquery selector, so we shouldn't be naming them as such

Issue #1480, opened 2020-02-14
Add authentication for Jupiter's APIs


lagoan:
**Is your feature request related to a problem? Please describe.**
Jupiter needs to provide a way to access its APIs from different clients. The simplest way to accomplish this could be with a system user account and a token. 

**Describe the solution you'd like**
Basic http authentication and sessions would be the easiest approach.

**Describe alternatives you've considered**
We could also go with JWTs or PASETOs but sometimes their implementations have security problems and have added complexity

**Additional context**
From comment https://github.com/ualbertalib/jupiter/pull/1370#issuecomment-561799351

mbarnett:
I think we can close this thanks to #1522? (feel free to reopen if I'm wrong)

PR #1522, opened 2020-03-04
Add system authentication for use with API clients

lagoan:
## Context
Jupiter will provide a way to access its APIs to clients using accounts with special privileges. For now these accounts will only use local system authentication to simplify things. This PR addresses #1480
## Changes
+ Add password capability to Identity model
+ System authentication following omniauth code structure
+ Rake task to create local accounts
## Ongoing work
- [x] Change method of adding system user from task to migration
- [x] Move password information to User model
- [x] Change mentions of ``api`` to ``system``
- [x] Test only system accounts can access API content


db/migrate/20200227182711_add_password_digest_to_identity.rb
```diff
@@ -0,0 +1,5 @@
+class AddPasswordDigestToIdentity < ActiveRecord::Migration[6.0]
+  def change
+    add_column :identities, :password_digest, :string, null: false, default: ''
```
murny: Curious if we can simplify this a bit. Identities are for OAuth providers(omniauth). It would be nice if we don't mix password/api_key accounts with Omniauth/OAuth identities. 

What if we just made one column on the user model and call it `api_key_digest`. This replaces the `password_digest` column. Then we can also replace the `api` boolean column by simply checking if that column is present or null to see if a user has an API key?

lagoan: I started working on this by using the omniauth-identity gem which adds password_digest to the Identity model. 

Using the omniauth-identity gem proved more trouble than it was worth since I had to document code changes that had to do with the gem itself assuming it was the only authentication method, also I was removing more things than I was using. 

I continued in this direction though so that our authentication mechanisms did not split into different ways, I believe this split can cause problems down the line. This way, we can still keep in line with omniauth and change things down the road if we ever need to.

murny: Omniauth is a very special authentication mechanism (its authentication via another provider/identity whether thats facebook/google/github/linkedin/saml/etc. Those buttons on websites that say "Login with your google account!"). Authentication via an API token (which is what you are doing) or email/password authentication is not Omniauth/OAuth.  Hence why I am pushing for you to keep this logic out of omniauth (identities table/model) as it has nothing to do with Omniauth. 

This is a very popular pattern in Rails. If you ever used Devise, it is best practice to keep these two authentication patterns separate. Such as keeping the password stuff on the user model and all the omniauth stuff in a seperate table (e.g identities table).

The big reason behind this is being future proof. If tomorrow we decide we want to go to email/password authentication in ERA and remove our omniauth SAML provider....we can't just remove this code (and I've worked on a lot of projects in the past that rolled out omniauth only to remove it down the road). We will be in an awkward phase of trying to revert your API token authentication stuff back onto the user model, before we can proceed with ripping out omniauth. If you just kept these two types of authentication seperate from the beginning, then its super simple to rip out omniauth (remove identities table/model, remove omniauth controller/views/routes, done!).

app/policies/application_policy.rb
```diff
@@ -40,6 +40,10 @@ def admin?
     user.try(:admin?)
   end
 
+  def api?
```
mbarnett: I suggest we use the terminology `system` rather than `api` anywhere it's used, because I can see some other use-cases for the system-user that aren't necessarily going to involve API calls but might still need to interact with policies, or the DB column, etc

db/migrate/20200227182711_add_password_digest_to_identity.rb
mbarnett: I think Shane makes a compelling argument here; both in the sense that the last time I left the university, the predominant CCID authentication pattern was binding to LDAP, it's SAML right now, and it could be something else entirely in a year or two if IST moves into the cloud, and in the sense that Rails auth frameworks turn over pretty frequently – we're likely to need to replace Omniauth long before we replace Jupiter entirely.

I'm also not sure that it makes sense that we'd want a system user to have multiple points of log-in access via identities; it seems like it would be easy to accidentally open it up to a human log-in that an API key directly on the user table wouldn't enable.

It also just generally gets us away from confusing 'passwords' with 'API keys', when it's probably better to treat them as distinct things with distinct usecases.

lib/tasks/jupiter.rake
```diff
@@ -144,6 +144,71 @@ namespace :jupiter do
     puts 'done!'
   end
 
+  desc 'creates local api system account which will not be able to log into the system by other means'
+  task create_local_system_account: :environment do
```
mbarnett: I don't like this approach.

1) A task means that we can create multiple system accounts, when the assumption right now is that we just want one. We're opening ourselves up for human error where someone accidentally runs this when one already exists.

2) We're introducing a manual step into the DB setup process, which complicates container setup and teardown, general dev environment setup, etc, as code gets written that depends on the system user existing.

3) We have a specific email address already in-mind for the system user, so asking the sysadmins to manually input it just opens up room for error.

4) Human API token generation opens up room for an overly weak token, and again, human error if someone typos it here.

My suggestion would be to replace this rake task with a migration that does a find_or_create for a system user, with a hard-coded email address and pulls the token out of secrets.yml. We can then generate a UUID or something sufficiently opaque & random and have the sysadmins set it up in an env variable.

lagoan: I see your point on simplifying deployment. I considered using a task in case we ever needed more users, and it could potentially be used at some other time if we need to. For now it does make sense to use a migration and not introduce more complexity.

app/policies/application_policy.rb
lagoan: Makes sense. Just so we are on the same page, what other use-cases for the system-user do you foresee? I ask this to make sure we want to change the naming from ``password`` to ``api-key`` as @murny recommends.

murny: I think API KEY would be a bit more special then what you are currently doing. You would generate a random 32 hash and give that to a user to use as an API KEY which they can pass into their requests (`era.com/users.json?api_key=ewoK3yJpNF&2tSnHwt$%sKT%7@Sw6Vt`). This is quite different from what you are doing. As you are basically implementing password authentication (having user remember a plaintext password which we take and hash it and check if its the same one we store in our DB). 

So I'm totally cool with transition your stuff to "System password" instead of "API Key".  And as Matt suggested the authorization around a system user is much more special then a normal API user. So makes far more sense this way 👍

db/migrate/20200227182711_add_password_digest_to_identity.rb
lagoan: @murny I get your point on separating the authentication mechanisms and it would make it easier to pull out omniauth if we ever need to as how @mbarnett suggest might happen.

@murny Just to be clear, at this moment I am literally using email/password authentication on the implementation side, I think it is just semantics. We could change this to public key authentication or something else if there is ever a need for it but right now that is out of our scope.

test/controllers/sessions_controller_test.rb
```diff
@@ -147,4 +147,37 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_nil session[:admin_id]
   end
 
+  test 'should log in as local system account' do
+    user = users(:system_user)
+    uid = 'ditech@ualberta.ca'
+    password = 'correct horse battery staple'
```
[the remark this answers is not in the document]
lagoan: Good catch! :smiley:

db/migrate/20200305210657_add_system_account_if_needed.rb
```diff
@@ -0,0 +1,17 @@
...
+        email: 'ditech@ualberta.ca',
+        admin: false,
+        password_digest: BCrypt::Password.create(
+          Rails.application.secrets.system_user_password
+        )
+      })
+    end
+  end
+
+  def down; end
```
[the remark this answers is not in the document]
murny: We should remove the user if we want to rollback to the previous state of the app? Migrations are meant to be rollback (meaning the `down` method always reverses the work done in the `up` method). 

If we push this code out into production and everything fails, we should be able to restore the state of the application back to the last working state. With not implementing a `down`, when we rollback the application, we would be leaving changes in the DB. 

It's not a big deal in this case as its just a user being left in our user model. And totally up to you to implement this. But hopefully you can understand the importance of implementing the `down`part of a migration.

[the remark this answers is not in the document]
lagoan: On the matter of implementing ``down`` method and the ``up`` method using ``find_or_create_by(email: 'ditech@ualberta.ca')``

If we assume that the user can exist before the migration, when we rollback the migration and delete the entry, then the state would not be the same.

This is not a big deal as you mentioned. I think the best way to go about this is to follow the ``find_or_create``method and set a password and not removing the user on ``down``.

I see what you mean about the problems seeding the database with migrations.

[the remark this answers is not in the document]
murny: Yep 👍 good call on not implementing down if user does exist in up (as that be problematic 😅 ) 

Think if you just made that change on the up to handle if they exist then that's pretty good from my point of view

.rubocop.yml
```diff
@@ -121,6 +121,15 @@ Style/FormatStringToken:
 Style/FrozenStringLiteralComment:
   Enabled: false
 
+Style/HashEachMethods:
```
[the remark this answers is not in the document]
lagoan: Still need to figure out how to address this issue.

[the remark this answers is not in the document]
pgwillia: Sometimes when things get wonky I start a new branch same as the old branch and cherry-pick.
```
git checkout integration_postmigration
git branch -D feature/basic-authentication
git checkout -b feature/basic-authentication
git cherry-pick 89fdbb8bbb322f6a3476024c90018ee3e11b7570
git cherry-pick ...
```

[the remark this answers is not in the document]
lagoan: Thank you @pgwillia , will take a look!

lagoan:
Rewrote history to remove rebase/merge problems.

Gemfile
```diff
@@ -29,6 +29,7 @@ gem 'redis', '~> 4.1'
 gem 'rsolr'
 
 # Authentication
+gem 'bcrypt', '~> 3.1.13'
```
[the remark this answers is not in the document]
lagoan: You bring up a good point. What do you think about keeping it pessimistic with '~> 3.1' ?

app/policies/application_policy.rb
```diff
@@ -40,6 +40,10 @@ def admin?
     user.try(:admin?)
   end
 
+  def system?
+    user.try(:password_digest?)
+  end
```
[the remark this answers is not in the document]
lagoan: I see your point on the clarity of keeping the ``system`` boolean

app/controllers/sessions_controller.rb
```diff
@@ -67,4 +67,16 @@ def logout_as_user
     redirect_to admin_user_path(original_user), notice: t('.flash', original_user: original_user.name)
   end
 
+  def system_login
+    email = params[:email]
+    password = params[:password]
```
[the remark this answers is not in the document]
lagoan: I think we should keep the password name. From an implementation point of view this is a password not an api key https://github.com/ualbertalib/jupiter/pull/1522#discussion_r388442234

Also, as you suggest, there may be other use-cases that do not involve API calls https://github.com/ualbertalib/jupiter/pull/1522#discussion_r388013725
 . Keeping the name password would give us clear flexibility in the future on how we use it instead of forcing the use based on the name; I hope I am being clear on this.

That previous conversation also includes @murny talking about how api keys and passwords are different https://github.com/ualbertalib/jupiter/pull/1522#discussion_r388440280 . It makes sense for the code to be named for what it is. 

I see what you mean on making it easily readable for future developers, this can be addressed with a comment on the model describing how the system password works at this time.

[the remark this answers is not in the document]
mbarnett: I'm sorry Omar, but I don't find the argument in https://github.com/ualbertalib/jupiter/pull/1522#discussion_r388442234 convincing.

1. That you're using password infrastructure to build it does not make it a password, and if we feel that it does, we should back out this work and rebuild it as an API key from the ground up, because what I as the person whose job is to be responsible for the larger architecture of all of our applications am asking you to build here is an API key mechanism for PMPY that I can potentially leverage in other services down the road, not an interactive login functionality.

2. A comment "in the module" doesn't really help in the long run – first of all, which module? This is referred to by the name "password" in SessionsController, the User model, the ApplicationPolicy, two different migrations, and an integration test. If we need to put a comment in 6 different places to clarify to future readers what a variable name actually means, then we need to face the fact that the variable is badly named and we should name it more clearly.

3. Comments don't reach into the database, where the column is named `password_digest` instead of `api_key_digest`. That means both that in an emergency a sysadmin might do the wrong thing at 3 am while working directly with the database* and that given that databases tend to outlive the applications that originally supported them, when someone goes to work with this data years from now, we've done a poor job of self-labeling our data for them.

* it's happened in the past – imagine what would happen if an overtired Admin is asked to turn off or reset logins due to a security issue saw the password column and decided that issuing a `UPDATE users SET password_digests = "";` – and its not unreasonable for them to expect that to work given the naming. We always want to be building our systems in a way that lends itself to being understood by other people, not misunderstood unless they have the context

Gemfile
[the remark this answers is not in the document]
mbarnett: Experience has taught us that the best thing to do is let the versions float as much as possible; we should trust our tests to catch API breakages, not rely on pessimism in the Gemfile. 

'~> 3.1' means that Dependabot won't let us know when 4.0 comes out since it doesn't meet our specification, but I'd prefer to have Dependabot open a PR for 4.0, so that we can get a sense of what tests it breaks and therefore a rough scope of work that will need to be addressed to keep us up to date with the ecosystem. If we put arbitrary caps on the versions we accept, we never know how far behind we are unless someone happens to manually check.

app/controllers/sessions_controller.rb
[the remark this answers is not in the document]
lagoan: Responses to your comments follow:

1. I think I misunderstood your comment on how to give API access to PMPY by adding a system user https://github.com/ualbertalib/jupiter/pull/1370#issuecomment-561799351 . Should we switch gears to an API key mechanism or renaming the password field on the user table be enough? 

2. Sorry if I was not clear, I was talking about the User model, as the logic is here I figured this is the first place a developer would look.

3. I do see your point on how the comments don't reach into the database.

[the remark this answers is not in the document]
mbarnett: Thanks Omar. I’m happy with your implementation as is if we can change the naming to indicate the intent that it’s serving as an API key. 

We can always refactor the implementation down the road if we decide it doesn’t quite match the intent we communicate with the naming.

app/models/user.rb
```diff
@@ -17,6 +18,24 @@ class User < ApplicationRecord
...
+  #             |        |    X    | False
+
+  # Check if the api key is there only if the user is a system account
+  validates :api_key_digest,
+            presence: { message: 'must be present if System value is true' },
+            if: -> { system && api_key_digest.blank? }
+
+  # Check that the api key is not present if it is not a system account
+  validates :api_key_digest,
+            absence: { message: 'must be blank if System value is false' },
```
murny: Looks pretty good!

Minor, but Perhaps move these strings to I18n (as they could be potentially front facing to users)?

test/models/user_test.rb
```diff
@@ -90,4 +90,45 @@ class UserTest < ActiveSupport::TestCase
...
+  end
+
+  test 'should not validate if it has an api key and it is not a system account' do
+    user = User.new(
+      name: 'User name',
+      email: 'valid@example.com',
+      api_key: '70d800e9-5fe8-49e4-86ed-eefc11ebfa52'
+    )
+
+    assert_not user.valid?
```
murny: Maybe just enhance these tests to double check the error message coming back and its correct?

e.g:
`assert_equal user.errors[:api_key_digest].first, "can't be blank"`

lagoan: In your example you included the string in English. I see the benefits to this. However, would it be problematic when testing with different locales ?

murny: Yeah was just an example. Since these are custom error messages and from my other comment, it be nice if they are defined in I18n, you can reuse that for your test. (So that if we ever change it, its changed in both places)

Example from a real test in jupiter:

```
 assert_includes collection.errors[:community_id],
                    I18n.t('activerecord.errors.models.collection.attributes.community_id.community_not_found',
                           id: community_id)
```

mbarnett: Testing with locales is a bit ... tricky. But you can assume the tests run in the English locale.

The benefit of testing for the message is that you avoid a false positive where the user isn't valid, but it's not because of the test you expect, but because of some unrelated validation that  just happens to have failed.

As for hardcoding the message in the test versus an i19n lookup – the school of thought I subscribe to is that the tests should look for the concrete expectation, not duplicate the logic of the code you're testing. So if you're testing a method that does some math, the test should compare the result to an expected hardcoded number, not a math expression. Similar idea with things like this, we compare to a hardcoded string instead of comparing to lookup logic.

mbarnett: @murny would you do that? to my mind you're just introducing a potential copy paste error on the i18n key. Doesn't that test just check that rails is sane?

murny: I can see it both ways. I'm okay with either way personally (In my own experience I typically hardcode the model errors but reuse I18n for system tests/controller tests (e.g: click on this Link (name comes from I18n instead of hardcoding it). No particular reason why I do this. I think It's probably view code changes far far far more then model code and model code you really really want to have everything locked down). So doesn't matter to me, I see pros and cons of both.

You are correct though if we hardcode the expected message in tests this makes like you said, guaranteed and prevents false positives. Like if we accidentally edit i18n, then you would never catch that. But if you explicitly said the error message, then you would have a failing test to catch that. Plus its more self documenting... Reading my example `I18n.t('activerecord.errors.models.collection.attributes.community_id.community_not_found',  id: community_id)` It's not super clear what this error message is in this context of the test. You would have to go to the `en.yml` and reading it to find out what it actually is. If you explicitly hard code it then its right there.

My counter argument for if it just checks if rails is sane...I think its a bit different in this case, as we still checking that hey...this string from i18n is being displayed when this validation fires. So we still checking something more then just if ruby/rails did what its suppose to. (testing if `a = b` then `assert a == b` is what I consider just checking rails (i guess ruby in this case) is sane). 

Whatever the team wants to do I'm okay. Think the biggest thing is just remaining consistent with the approach we choose throughout the test suite. (Sounds like we want to hardcode this and that's okay!)

mbarnett: Do you know if Rails will raise an exception if a translation is missing in test mode? I'm wondering if tests along the lines of `assert_equal user.errors[:api_key_digest].first, I18n.t('path.to.locale.message')` could false positive with "translation missing" == "translation missing"

murny: Yep, good catch. We have enabled this in `environments/test.rb`
https://github.com/ualbertalib/jupiter/blob/integration_postmigration/config/environments/test.rb#L48-L49

By default its turned off

mbarnett: Eh, I'm probably fine with continuing this pattern, then. Like you said, we're not quite duplicating the same logic, and it is a bit less fragile than hardcoding when random wording changes come in.

app/controllers/sessions_controller.rb
```diff
@@ -29,6 +29,9 @@ def create
       user.identities.create(provider: auth_hash.provider, uid: auth_hash.uid)
     end
 
+    # System user accounts should not be able to access application through
+    # omniauth authentication
```
mbarnett: Nice 👍

app/models/user.rb
```diff
@@ -17,6 +18,25 @@ class User < ApplicationRecord
...
   validates :name, presence: true
 
+  # User validation for sysatem and api key values
+  #             | System | Api key | Valid |
+  #             ----------------------------
+  # Presence(X) |   X    |    X    | True  |
+  #             |        |         | True  |
+  #             |   X    |         | False |
+  #             |        |    X    | False |
+  #             ----------------------------
```
mbarnett: Nice, I really like the ASCII truth table for making this understandable.

lagoan:
Excellent

PR #1527, opened 2020-03-05
Fix problem with postgress requiring password

lagoan:
New postgres image requires a super user to be set up. This reverts to
old behavior.


murny:
Thanks for the quick fix Omar! But lets just set a password instead!

https://github.com/ualbertalib/jupiter/pull/1528

Issue #1445, opened 2020-01-30
Post Fedora Automated Test Cleanup


murny:
From #1430, I'm not able to parallelize our test suite until we refactor our current way of doing tests.

Biggest refactoring changes I see are the following:
- [x] Remove `before_all`/`after_all` setup blocks (We should not need this in post fedora world and I'd argue we can remove the `minitest-hooks` gem which gives us this ability)
- [x] Use fixtures for all data needs and turn on transactional tests.  Includes cleaning up things like this:
```
  # Transactional tests were creating a problem where a collection defined as a
  # fixture would only be found sometimes (a race condition?)
  self.use_transactional_tests = false
```

More info and context from Matt's Fedora Removal PR (#1273) :

> System Test Flapping
> This PR also makes the flapping problem in the system tests much worse. After digging into this for a couple of days, I've realized that the system tests are actually very badly written as they currently exist. Rails expects the system tests to operate off of fixtures that are loaded once, before all of the tests run, and expects that any extra data created by the system tests will be cleaned up after the test (or at least, that subsequent tests won't care).
> 
> Existing system tests create data constantly, don't clean up after themselves, and do break if data from another test is still around. Example: This test file is a big offender. It creates 11 Communities, and then tests that those are the only 11 being given in search results -- BUT other tests also create communities, so depending on what random order tests run in there may be as many as 15 search results, which causes some assertions to fail. Conversely, if you decide to try to fix this by wiping the database between tests, These tests will fail, because fixtures only load once at the start of all system tests, and will not reload again afterwards (I spent a good deal of time trying to hack around this, to no avail). With the database wiped, there is no user record for these tests to run against.
> 
> This all (accidentally) wasn't so bad when most of the tests were based on ActiveFedora, and got wiped between tests, but with the data in ActiveRecord it becomes much more frequent that a random ordering of tests will leave stale data in a way that some other test breaks. Wiping postgresql between tests doesn't work, as it breaks the one-shot loading of fixtures that Rails does automatically, leading to different tests breaking.
> 
> The only fix here is to re-write these tests properly. As this is a pre-existing problem with the way tests are written that is just exacerbated by the different rules of ActiveRecord, that's out of scope for this PR, so I've tried to fix little bits where I could, marked a couple of the worst offenders as skip, and left the rest flapping. Fixing this with proper rewrites of these tests would be a good idea at some point.

Also my approval comment:
[...]

PR #1534, opened 2020-03-10
#1445: Refactor admin controller tests. Remove all before_all blocks,…

murny:
… and other refactors

#1445 

First of many PRs. This one just tackles our admin/controller tests.

A quick example of how minor this change is amd helps illustrate how quick setup and fixtures are:

Using `before_all` to create objects before all tests which are then used for each test.

```
➜  jupiter (integration_postmigration) ✗ rails test test/controllers/admin/theses/draft_controller_test.rb
Run options: --seed 8096

# Running:

..............

Finished in 3.027713s, 4.6240 runs/s, 15.8536 assertions/s.
14 runs, 48 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /Code/jupiter/coverage. 1404 / 2670 LOC (52.58%) covered.
```

Using `setup` and fixtures, to tear down the DB and reprovision the DB with all the fixture data before each and every test:
```
➜  jupiter (integration_postmigration) ✗ rails test test/controllers/admin/theses/draft_controller_test.rb
Run options: --seed 53600
# Running:

..............

Finished in 3.037624s, 4.6089 runs/s, 15.8018 assertions/s.
14 runs, 48 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /Code/jupiter/coverage. 1398 / 2670 LOC (52.36%) covered.
```

Difference is hardly even noticeable

NOTE: Travis is hating on this PR. Looking into this. Seems to have been fixed...maybe caching issue? will have to keep an eye out for this.

### DIFF
--- a/test/controllers/admin/communities_controller_test.rb
+++ b/test/controllers/admin/communities_controller_test.rb
@@ -2,15 +2,9 @@ require 'test_helper'
 
 class Admin::CommunitiesControllerTest < ActionDispatch::IntegrationTest
 
-  def before_all
-    super
-    @admin = users(:admin)
-    @community = Community.new(title: 'Nice community',
-                               owner_id: @admin.id)
-    @community.save!
-  end
-
   def setup
+    @admin = users(:admin)
+    @community = communities(:books)
     sign_in_as users(:admin)
   end
 
@@ -26,6 +20,12 @@ class Admin::CommunitiesControllerTest < ActionDispatch::IntegrationTest
 
   test 'should show community' do
     get admin_community_url(@community)
+    assert_response :success
+  end
+
+  test 'should show community for js response' do
+    get admin_community_url(@community), xhr: true
+
     assert_response :success
   end
 
--- a/test/controllers/admin/items_controller_test.rb
+++ b/test/controllers/admin/items_controller_test.rb
@@ -21,4 +21,5 @@ class Admin::ItemsControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to root_path
     assert_equal I18n.t('admin.items.destroy.deleted'), flash[:notice]
   end
+
 end
--- a/test/controllers/admin/items_controller_test.rb
+++ b/test/controllers/admin/items_controller_test.rb
@@ -6,6 +6,11 @@ class Admin::ItemsControllerTest < ActionDispatch::IntegrationTest
     @item = items(:fancy)
     @admin = users(:admin)
     sign_in_as @admin
+
+    File.open(file_fixture('text-sample.txt'), 'r') do |file|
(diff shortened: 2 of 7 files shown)


test/controllers/admin/communities_controller_test.rb
```diff
@@ -29,6 +23,12 @@ def setup
     assert_response :success
   end
 
+  test 'should show community for js response' do
```
murny: Saw we were missing a test for this. Added one.

test/controllers/admin/theses/files_controller_test.rb
```diff
@@ -30,10 +24,9 @@ def before_all
   end
 
   test 'should be able to attach files to a draft thesis' do
-    # TODO: Why this failing?
```
murny: Seems to be working now? Maybe now that we are off monkey patched ActiveStorage or something?

test/fixtures/collections.yml
```diff
@@ -2,12 +2,18 @@ fantasy_books:
   title: 'Fantasy Books'
   owner: admin
   community: books
+  record_created_at: <%= Time.zone.now - 1.hour %>
```
murny: These are required fields for Community/Collections. Added them to fixtures

test/controllers/admin/items_controller_test.rb
```diff
@@ -46,8 +14,6 @@ def setup
   end
 
   test 'should destroy item and its derivatives' do
-    # TODO: 'ActiveStorage::Attachment.count' || 'ActiveStorage::Blob.count'
-    # wish I could test the thumbnail deletion but flaky on travis-ci
```
pgwillia: I think we should leave this in because the TODO hasn't been addressed yet.

murny: Yeah I could go either way. I choose the simple solution. We don't need attachments for these test as we are basically just doing simple smoke test to see if, we can load the index and that we can delete an item successfully. 

If deleting an item cleanups its attachments is important (which i say it is) than that should probably be well tested in the model. 

But its not much more effort to double check that here either. So I could add that back. I noticed I never updated the test name `should destroy item and its derivatives` so regardless of what I choose, I should do that 😆

murny: This is fixed now. Works locally, will keep an eye on it for TravisCI in case it starts flaking

murny:
Too many flakey tests, not able to get this passing. #1558 fixes these flakey tests completely. So closing this for https://github.com/ualbertalib/jupiter/pull/1558 instead. Moving additional fixes I made over to that PR

PR #1539, opened 2020-03-13
Can now go through wizard with an old license

ConnorSheremeta:


app/models/draft_item.rb
```diff
@@ -191,13 +201,18 @@ def license_as_uri
     return nil if license == 'license_text'
 
     code = LICENSE_TO_URI_CODE.fetch(license.to_sym)
-    CONTROLLED_VOCABULARIES[:license].send(code)
+
+    begin
+      CONTROLLED_VOCABULARIES[:license].send(code)
+    rescue RuntimeError
```
murny: Is there a way to make this more specific? What error does `CONTROLLED_VOCABULARIES[:license].send(code)` actually throw when its a bad code?

Got similar thinkings as rubocop here around this: https://docs.rubocop.org/en/stable/cops_style/#styleimplicitruntimeerror

If we accidentally had a spelling mistake or something, this error would be silenced by this `rescue RuntimeError` and no one would be any wiser about it

test/system/edit_item_old_license_test.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    visit item_path item
+    click_on I18n.t('edit')
+
+    click_on I18n.t('items.draft.save_and_continue')
+    click_on I18n.t('items.draft.save_and_continue')
+    click_on I18n.t('items.draft.save_and_continue')
+    click_on I18n.t('items.draft.save_and_deposit_edits')
+    assert_text I18n.t('items.draft.successful_deposit')
+
```
murny: Nice job adding a test 👍 . 

Curious if we can verify that the edited item on the show page actually has a an old license attached to it?

My worry is this test just checked that we can walk through the wizard. What if the wizard drops the license or fallbacks to a default and just allowed you to save? But if we can assert the final item after its been edited that it still has the correct license (an old license in this case), then we can be sure it all worked successfully 👍

app/models/draft_item.rb
ConnorSheremeta: The error thrown is specified now.

ConnorSheremeta:
Not sure how I requested review from you three right now, its not required as Shane as already approved

PR #1551, opened 2020-03-24
Fix/1408 - Cleanup null values for DraftItem and DraftThesis is_published_in_era attribute

lagoan:
Change Draft entities' boolean attribute is_published_in_era to not null.
    
Both entities DraftItem and DraftThesis had boolean columns is_published_in_era that where nullable. This created a three-valued logic problem when null values started to creep in.
    
Columns is_published_in_era are no longer nullable and null entries are set to their default false value.
    
The draft scopes that checked for nil values have been simplified by only checking for false.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -2,7 +2,7 @@ class Collection < JupiterCore::Depositable
 
   acts_as_rdfable
 
-  scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
+  scope :drafts, -> { where(is_published_in_era: false) }
 
   has_solr_exporter Exporters::Solr::CollectionExporter
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -2,8 +2,6 @@ class Collection < JupiterCore::Depositable
 
   acts_as_rdfable
 
-  scope :drafts, -> { where(is_published_in_era: false) }
-
   has_solr_exporter Exporters::Solr::CollectionExporter
 
   belongs_to :owner, class_name: 'User'
--- a/db/migrate/20200320164053_change_is_published_in_era_column_to_not_null.rb
+++ b/db/migrate/20200320164053_change_is_published_in_era_column_to_not_null.rb
@@ -1,0 +1,6 @@
+class ChangeIsPublishedInEraColumnToNotNull < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :draft_items, :is_published_in_era, false, false
+    change_column_null :draft_theses, :is_published_in_era, false, false
+  end
+end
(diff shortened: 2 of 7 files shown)


app/models/collection.rb
```diff
@@ -2,7 +2,7 @@ class Collection < JupiterCore::Depositable
 
   acts_as_rdfable
 
-  scope :drafts, -> { where(is_published_in_era: false).or(where(is_published_in_era: nil)) }
+  scope :drafts, -> { where(is_published_in_era: false) }
```
mbarnett: I think you can actually delete this scope entirely here and on Community. The corresponding column doesn't exist – I think this was some terrible copy-and-paste error on my part.

db/migrate/20200320164053_change_is_published_in_era_column_to_not_null.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeIsPublishedInEraColumnToNotNull < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :draft_items, :is_published_in_era, false, false
```
murny: As mentioned in slack, wondering what people thought of renaming this to just `published`? Then AR gives you a `published?` method?

`is_published_in_era` is quite verbose and not great from rails naming standards 🤷‍♂

mbarnett: I think the concern, originally, was that there's a separate attribute on Items, `publication_status` that indicates whether or not certain kinds of items have been published in the sense of like, it's an Article and it was published in a Journal somewhere instead of being a preprint or whatever.

So if we have one attribute called `publication_status` and one called `published`, it's not super clear that one of these is indicating a status that has a big impact on how the thing behaves in ERA and one of these is just an enum that doesn't impact anything other than how the item is described.

I'm open to changing the name, but I think the names need to be distinct enough that we don't need to consult a doc somewhere to not accidentally make items public or something.

mbarnett: Or maybe that's less of a concern after Drafts ended up being a separate table from Items proper. Open to arguments.

murny: Yeah I guess it gets confusing (or at least I am confused). Maybe we need to take a step back and think about this a little more. 

Currently we have a status enum on draft items/thesis models which have 3 values: inactive, active, archived...then we also have a scope which we considered unpublished items are `active items without a uuid`  (or you can potentially assume the opposite is published?)

`scope :unpublished, -> { where(status: :active).where('uuid IS NULL') }`

Now we created another boolean column for `published`

Then we also have a Items table which you could also considered to be published items 🤔 

Like I said maybe I am confused 😆

mbarnett: Oh wow, this is even more confusing than I thought.

Here's my initial thought: we go forward with the column name at is, and I create a ticket to figure out and document this mess so we can untangle this a bit, because it's probably a bad sign that we have competing ideas of un/published right now.

lagoan: How about a small change to ``published_in_era``? Keeps it closer to rails naming standards and is still easy to understand what it means.

@murny The issue you ring up where Jupiter is suppose to be “era” still remains though https://ualbertalib-di.slack.com/archives/CSEHP0L0K/p1585002487381500

PR #1558, opened 2020-03-24
Post Fedora Automated Test Cleanup #1445

murny:
- Removed before_all blocks (delete `gem minitest-hooks`)
- Use `setup do` over `def setup`
- Truncate Solr index after every test (dirty hack, should probably track down tests that pollute solr and clean it up there)

Lots of test cleanup probably still needs to happen, especially in "search" tests where we seeding 40-50 objects. Lots more opportunities to use fixtures over seeding data via `.create` but I've already spend far to long on this. This is already a huge improvement.

Will create another PR for attempting parallel testing which should be the last item for  #1445 issue


test/controllers/search_controller_test.rb
```diff
@@ -2,100 +2,64 @@
...
 
   test 'should get results in relevance order when a query is present' do
-    get search_url, as: :json, params: { search: 'Moose' }
+    get search_url, as: :json, params: { search: 'Item' }
     assert_response :success
-    data = JSON.parse(response.body)
-    assert_equal([@item2.id, @item5.id, @item4.id], data.map { |result| result['id'] })
+    results = JSON.parse(response.body).map { |result| result['id'] }
+
+    # TODO: is there a way to set relevance score?
```
pgwillia: I don't think you can set relevance score without messing with the Solr config.  You could game it, though.  Repeating the search term should boost it's relevancy. (Hmm, the inverse is true, but you get the point)

For example notice the difference in scores:
```
"response":{"numFound":3,"start":0,"maxScore":0.41904876,"docs":[
      {
        "id":"1",
        "title_tesim":["Item"],
        "_version_":1662173232014295040,
        "timestamp":"2020-03-25T21:32:00.790Z",
        "score":0.41904876},
      {
        "id":"2",
        "title_tesim":["Item Item"],
        "_version_":1662173472089964544,
        "timestamp":"2020-03-25T21:35:49.742Z",
        "score":0.37039027},
      {
        "id":"3",
        "title_tesim":["Item Item Item"],
        "_version_":1662173485349208064,
        "timestamp":"2020-03-25T21:36:02.387Z",
        "score":0.36290687}]
  }
```

test/integration/collection_show_test.rb
```diff
@@ -3,27 +3,14 @@
 # TODO: This test would be better as an System Acceptance test instead?
 class CollectionShowTest < ActionDispatch::IntegrationTest
 
-  # TODO: add tests involving non-public items
```
pgwillia: Did the non-public items get tested?

test/integration/collection_show_test.rb
```diff
@@ -71,7 +58,7 @@ def before_all
   end
 
   test 'visiting the show page for a collection as a regular user' do
-    user = users(:regular)
+    user = users(:regular_two)
```
pgwillia: Why did the user change?

test/integration/community_edit_test.rb
```diff
@@ -2,17 +2,16 @@
...
-                            filename: 'image-sample.jpeg', content_type: 'image/jpeg'
-
-    # A community with no collections, no logo
-    @community2 = Community
-                  .create!(title: 'Empty community', owner_id: users(:admin).id)
+  setup do
+     # A community with a logo
+     @community1 = Community
+     .create!(title: 'Logo Community', owner_id: users(:admin).id)
+@community1.logo.attach io: File.open(file_fixture('image-sample.jpeg')),
```
pgwillia: Something up with the spacing.

test/integration/sitemap_test.rb
```diff
@@ -88,7 +52,7 @@ def file.original_filename
     assert_select 'loc', item_url(@item)
     assert_select 'lastmod', @item.updated_at.iso8601
     # not show private items
-    assert_select 'url', count: 2
+    assert_select 'url', count: 3
```
pgwillia: Hmm.  There should be 2 items because the private item shouldn't be visible in the sitemap.

test/models/thesis_test.rb
```diff
@@ -26,17 +21,16 @@ def before_all
       unlocked_thesis.save!
     end
     assert thesis.valid?
-    assert_not_equal 0, Thesis.public_items.count
```
pgwillia: Just curious why you dropped the Thesis.public_items tests?

murny: Yeah great question let me describe my thinking. 

First this fails as we now have fixtures being loaded (previous we were destroying all thesis objects before running the test). So I could update this just to say... `assert_not_equal 2, Thesis.public_items.count` (as this is how many public thesis fixtures we have and we want to make sure its not still 2, aka a save was successful). 

The line you highlighted here is actually valid though, and I think we should be asserting something along the lines of what this assertion was previously doing but not using a "magic number" (0 or 3). It be better if we wrap the creation in a block like so:
```
    assert_difference('Thesis.count') do
       # thesis new stuff
    end
```

So thanks for making a comment here. I can update that.

The other deletions are kinda bad tests. We testing the validations on a non persisted object, then asserting it doesn't magically save? 

```
    thesis = Thesis.new
    assert_nil thesis.visibility
    assert_equal 0, Thesis.public_items.count
```

Here we basically testing if rails internals works....making sure `.new` is NOT creating an object in the DB. 

I'd say we don't have to test this and we can be confident that Rails/ActiveRecord is working properly.

test/integration/sitemap_test.rb
murny: I added a "authenticated_item" to the items fixtures which get listed in the sitemap. So reason for the increment on the number here

test/integration/community_edit_test.rb
murny: Yuck! will fix!

test/integration/collection_show_test.rb
murny: Good catch. The reason is the `fancy_item = items(:fancy)` fixture is actually created/owned by `users(:regular)`, and we have an assertion in this test that we  should not see a "Edit" button on this page. But since the user we are logging in as, owns the item, they do see it! 

Simple solution is just login with a different user

test/integration/collection_show_test.rb
murny: Can re-add the TODO 👍  (will look into adding these tests if their simple as well)

test/controllers/redirect_controller_test.rb
```diff
@@ -2,16 +2,17 @@
 
 class RedirectControllerTest < ActionDispatch::IntegrationTest
 
-  def before_all
-    super
-
+  setup do
```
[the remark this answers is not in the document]
murny: Yeah absolutely, I need to refactor these to use fixtures If I can 👍 . I just wanted to rip out `before_all` before doing too much damage (heavily refactoring setup code)

test/fixtures/items.yml
```diff
@@ -23,22 +24,11 @@ admin:
   item_type: <%= CONTROLLED_VOCABULARIES[:item_type].book %>
   subject: ['Ownership']
   member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
-edit_item_history:
```
murny, line 26: Biggest thing here is try to reuse fixtures where possible. If you can reuse a fixture then do so. This one is pretty much exactly the same as the `admin` fixture...so lets just use the admin one

test/fixtures/thesis.yml
```diff
@@ -5,29 +5,15 @@ nice:
...
   record_created_at: <%= 5.day.ago.to_s(:db) %>
-  member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:thesis, :uuid)}/#{ActiveRecord::FixtureSet.identify(:thesis, :uuid)}"] %>
-legacy:
-  visibility: <%= JupiterCore::VISIBILITY_PUBLIC %>
-  owner: regular
-  title: 'Nice Item'
-  dissertant: 'Joe Blow'
-  graduation_date: '2000'
-  record_created_at: <%= 5.day.ago.to_s(:db) %>
+  abstract: 'This is my abstract'
```
murny, line 18: Fixtures must be valid in the eyes of our DB (easy as fixture will fail when it attempts to create) and Models (harder...as it will only fail when you attempt to "resave" the object...which is hard to debug)validations. All these Thesis' require an "Abstract" unless its "legacy" (grad date is older then 2009)

test/system/thesis_description_optional_test.rb
```diff
@@ -2,30 +2,33 @@
 
 class ThesisDescriptionOptionalTest < ApplicationSystemTestCase
 
-  test 'cannot go through wizard step without description on thesis after 1990' do
+  test 'cannot go through wizard step without description on thesis after 2009' do
     admin = users(:admin)
     thesis = thesis(:nice)
-    thesis.save
```
murny, line 8: This save is useless, Fixtures have already been saved

test/system/thesis_description_optional_test.rb
```diff
@@ -2,30 +2,33 @@
...
     login_user admin
 
     visit item_path thesis
     click_on I18n.t('edit')
 
+    fill_in I18n.t('admin.theses.draft.describe_thesis.description'), with: ''
```
murny, line 13: This has been added, as previously the fixture is not valid...we can't have this. Fixtures must be valid in terms of DB/Model validations. We can make this invalid via the form here instead.

test/system/thesis_description_optional_test.rb
```diff
@@ -2,30 +2,33 @@
...
     click_on I18n.t('admin.theses.draft.save_and_continue')
     assert_text I18n.t('admin.theses.draft.please_fix_errors')
 
     logout_user
   end
 
-  test 'can go through wizard step without description on thesis before 1990' do
+  test 'can go through wizard step without description on thesis before 2009' do
     admin = users(:admin)
-    thesis = thesis(:legacy)
```
murny, line 22: Example of how we can reuse a fixture. 

Instead of creating a brand new fixture just for a single test. We can reuse an existing one. Here instead of having the fixture already be in the state we want, we can just update the form before we save the item. This should still give us the validation failure we are looking for.

test/controllers/admin/items_controller_test.rb
```diff
@@ -46,16 +18,12 @@ def setup
   end
 
   test 'should destroy item and its derivatives' do
-    # TODO: 'ActiveStorage::Attachment.count' || 'ActiveStorage::Blob.count'
-    # wish I could test the thumbnail deletion but flaky on travis-ci
-    assert_difference(['Item.count'], -1) do
+    assert_difference(['Item.count', 'ActiveStorage::Attachment.count'], -1) do
```
pgwillia, line 21: 👏

test/system/search_test.rb
```diff
@@ -109,7 +109,7 @@ def setup
     end
   end
 
-  def teardown
+  teardown do
```
pgwillia, line 112: Did you figure out why this is still necessary?  Should be getting this from the test_helper, ya?

murny: Ah yeah, we can remove this in favor of the global one from test_helper. Good catch 👍

murny: Removed this and any other ones I found 👍

pgwillia: 👏

PR #1562, opened 2020-03-25
Set up admin view of papertrail edit history

ConnorSheremeta:
- Created partial and helper
- Added differ gem
- Added to fixtures
- Added tests


app/decorators/humanized_change_set.rb
```diff
@@ -0,0 +1,241 @@
+class HumanizedChangeSet
+
+  TIMEZONE = 'Mountain Time (US & Canada)'.freeze
```
murny: Future: We should use a gem like `local_time`(https://github.com/basecamp/local_time) where we just output this in UTC and then on client side we convert this into their browser/systems timezone. 

But for now this is good (considering this will only affect admin users who are also were travelling/outside of MST and viewing ERA which would be very rare)

app/javascript/src/application/index.js
```diff
@@ -1,6 +1,5 @@
 import './communities_collections_searchbox';
 import './item_draft';
-import './item_show';
```
murny: Might have to be careful here. We still need some of the things in `item_show.js` like the multi download functionality and hidding/showing more info? This is probably all still valid and still needs to be here.

app/javascript/src/application/item_show.js
```diff
@@ -1,15 +0,0 @@
-document.addEventListener('turbolinks:load', () => {
```
murny: Yeah dont think this should be removed?

app/javascript/styles/jupiter.scss
```diff
@@ -53,3 +55,16 @@ body {
   background-position: center 10%;
   background-size: cover;
 }
+
```
murny: This file needs some cleanup (should probably be renamed `layout.scss`, etc). But for now lets not continue making this a dumping ground for random CSS. 

Can we move this into a `changeset.scss` or whatever you want to call it and include it in application.scss?

ConnorSheremeta:
There are still a few issues I'm still working through. One being that the admin js (/app/javascript/src/admin) not loading for admins, I think it might be related to now being on webpacker?

murny:
@ConnorSheremeta 
Look into adding your JS file to this file:
https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/javascript/src/admin/index.js

ConnorSheremeta:
Hmm didn't quite do it, will continue to investigate

app/views/items/_edit_history.html.erb
```diff
@@ -0,0 +1,40 @@
+<% if current_user.present? && current_user.admin? %>
+  <% content_for(:head) do %>
+    <%= javascript_include_tag 'admin' %>
```
murny: Tried to slack you. Here's the javascript problem.

This needs to be webpacker now:
```
<% content_for(:head) do %>
  <%= javascript_pack_tag 'admin' %>
<% end %>
```

And then you also need to import your items_show.js in the admin.js src file, as this is what this is going to bring in here.

ConnorSheremeta: Thank you! Still having problems, it wouldn't catch the event if the listener was in the admin namespace. I moved it out and it works, its not ideal as only admins need this js but it works now. Not sure if I should spend more time trying to get it perfect or just leave it as is

Issue #1435, opened 2020-01-24
[ERA] Thumbnail choice resets between saves 


ConnorSheremeta:
**Describe the bug**
When an item is saved with a thumbnail other than the first attachment, another edit and save will set the thumbnail to be the first attachment again

**To Reproduce**
1. Have an item/thesis with more than 1 attachment
2. Click on edit
3. Set the thumbnail to be the second attachment
4. Save
5. Go back to edit
6. Save
7. Notice that the thumbnail is now different than what was set previously

**Expected behavior**
Thumbnail choice persists between saves

PR #1565, opened 2020-03-26
Thumbnail choice no longer resets between saves

ConnorSheremeta:
#1435 We were using attachment ids on drafts when we should have been using the blob ids, they persist between saves! We weren't actually pulling anything from the item on draft creation regarding the thumbnail id which has been fixed.


app/models/draft_item.rb
```diff
@@ -123,7 +124,8 @@ def update_from_fedora_item(item, for_user)
       citations: item.is_version_of,
       source: item.source,
       related_item: item.related_link,
-      is_published_in_era: false
+      is_published_in_era: false,
+      thumbnail_id: thumbnail_file.nil? ? nil : thumbnail_file.blob_id
```
lagoan: Would 
```
thumbnail_id: thumbnail_file&.blob_id
```
achieve the same?

app/models/draft_thesis.rb
```diff
@@ -88,7 +89,8 @@ def update_from_fedora_thesis(thesis, for_user)
       specialization: thesis.specialization,
       departments: thesis.departments,
       supervisors: thesis.supervisors,
-      committee_members: thesis.committee_members
+      committee_members: thesis.committee_members,
+      thumbnail_id: thumbnail_file.nil? ? nil : thumbnail_file.blob_id
```
lagoan: Same as above

murny:
Can we get a PR description detailing the bug? Looks like we were using attachment ID when we should have been using the blob ID? is that correct?

ConnorSheremeta:
Sorry, GitHub keeps automatically adding review requests. Don't feel the need to review this.

app/models/draft_item.rb
ConnorSheremeta: It does. Good cal!

ConnorSheremeta:
> Can we get a PR description detailing the bug? Looks like we were using attachment ID when we should have been using the blob ID? is that correct?

Added a description and linked to the issue. That is correct.

test/fixtures/items.yml
```diff
@@ -49,6 +49,18 @@ fancy_private:
   publication_status: <%= [CONTROLLED_VOCABULARIES[:publication_status].published] %>
   subject: ['Fancy things']
   member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
+thumbnail_resetting:
```
murny: Could we not reuse an existing fixture? This looks exactly the same as the "admin" item fixture, just with a slight couple of changes? For the test case I feel like we could just reuse that one instead of creating a new one?

ConnorSheremeta: Yes we can. I haven't been leveraging existing fixtures, I should be!

Issue #1566, opened 2020-03-26
[Continuous Integration] Migrate to Github Actions from TravisCI


murny:
TravisCI has started to become problematic for our team. Some developers have lost access to the organization and no longer able to restart builds/make changes/etc. Also TravisCI has been extremely flakey for our test suite.

Maybe its time to move to Github Action: https://github.com/features/actions

We can probably run them in parallel for a bit, then once Github Actions is up and running and we have confidence in it, we can then rip out TravisCI.

PR #1567, opened 2020-03-26
1566 enable GitHub actions

murny:
#1566 

All seems to work. Github Actions is now building and running our unit tests, system tests, danger and eslint/rubocop.

Had to reuse our Docker compose solution as I wasn't able to use Github Actions Services as Solr is a bit painful (chicken and egg problem....we need our code pull down as we use some of it to configure solr....but Services are created and setup before we checkout our code..)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,6 +9,7 @@ jobs:
       postgres:
         image: postgres:11-alpine
         ports: ["5432:5432"]
+        options: -e="POSTGRES_PASSWORD=mysecretpassword"
 
       redis:
         image: redis:alpine
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -5,25 +5,11 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
-    services:
-      postgres:
-        image: postgres:11-alpine
-        ports: ["5432:5432"]
-        options: -e="POSTGRES_PASSWORD=mysecretpassword"
-
-      redis:
-        image: redis:alpine
-        ports: ["6379:6379"]
-
-      solr:
-        image: solr:6.6
-        ports: ["8983:8983"]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 # Jupiter
 
 [![Build Status](https://travis-ci.org/ualbertalib/jupiter.svg?branch=master)](https://travis-ci.org/ualbertalib/jupiter)
```
mbarnett: Should we take travis off the page? Or do you want to pull out everything travis related in a separate PR once this is working?

murny: Yeah I wanted to run them parallel for a bit. Just to make sure Github Actions is working really well. Then we can remove Travis in another PR

mbarnett: Makes sense to me!

.github/workflows/push.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build the docker-compose stack
+        run: docker-compose up -d
+
+      - name: Set up Ruby 2.6
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
```
pgwillia: Is it possible to do this with a matrix of Ruby versions, à la travis?  It'd be nice to have it check against current (or whatever is deployed in production) and the next version of Ruby at the same time.

murny: Yep I'd assume it would just be a completely separate Job (currently everything is under a "test" job, we could just split this out for different version or just run certain jobs). But maybe for this we should just set this for the Ruby version we want to make sure it works against so maybe we should downgrade this to 2.5.x instead 🤔(Keep forgetting we on 2.5 😆)

pgwillia: It's nice to know we **could** be on 2.6.

Issue #1581, opened 2020-04-06
Enable New Rubocops


pgwillia:
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file:
 - Lint/RaiseException (0.81)
 - Lint/StructNewOverride (0.81)
For more information: https://docs.rubocop.org/en/latest/versioning/

pgwillia:
The StructNewOverride cop warns that for https://github.com/ualbertalib/jupiter/blob/03691d9211db9010901af6a846a6a73edb1b2dfd/app/models/jupiter_core/solr_services/facet_result.rb#L3
Lint/StructNewOverride: :count member overrides Struct#count and it may be unexpected.

murny:
This is now merged. Closing this issue

PR #1583, opened 2020-04-07
Simply running bundle update and yarn upgrade to fix some security concerns

murny:
Hopefully fixes acorn and minimist versions to make github security alerts go away

#trivial


.rubocop.yml
```diff
@@ -36,6 +36,12 @@ Layout/LineLength:
 Layout/RescueEnsureAlignment:
   Enabled: false
 
+Lint/RaiseException:
```
murny, line 39: New cops. 

https://rubocop.readthedocs.io/en/latest/cops_lint/#lintraiseexception

and

https://rubocop.readthedocs.io/en/latest/cops_lint/#lintstructnewoverride

app/models/jupiter_core/solr_services/facet_result.rb
```diff
@@ -1,6 +1,8 @@
 class JupiterCore::SolrServices::FacetResult
 
+  # rubocop:disable Lint/StructNewOverride
```
murny, line 3: We overriding `count` here. This is probably fine. Just whitelisting this

pgwillia:
Resolves #1581

app/models/jupiter_core/solr_services/facet_result.rb
pgwillia: I wasn't sure.  Trust your judgement on this.

murny: Yeah i'll let @mbarnett  chime in here. 

We can either call this `count` something else so we not overriding it...`results_count` or what not. Then we don't need to add this exception. Or just leave it as is. 

Seems like a silly cop? I think the default `count` of a Struct.New is just the number of Struct Members? (or is that `.length`?)

lagoan: From https://rubocop.readthedocs.io/en/latest/cops_lint/#lintstructnewoverride

count in Struct new overrides [Enumerable#count](https://ruby-doc.org/core-2.7.1/Enumerable.html#method-i-count)

I would say it is best call it something else as you suggest @murny

mbarnett: Yeah, just to clarify: we're intentionally overriding Enumerable#count here. This lets the FacetResult play nicely with things that build on the Enumerable machinery (pagination, etc) without having to expose the underlying collection directly.

As for why Rubocop objects to basic encapsulation, ¯\_(ツ)_/¯

PR #1595, opened 2020-04-16
Remove todo for parallelization. Currently not worth the effort

murny:
## Context

I attempted to add parallelization testing to Jupiter. But it seems way more work than its worth which I hope to explain here.

The original idea is to have each process/thread name space each index (e.g: "ArItem_0", "ArItem_1"). This whole idea is a similar to how some ElasticSearch gems have achieved this (e.g: https://github.com/ankane/searchkick#parallel-tests). This would let each process have its own "instance" of Solr where it should be able to run tests without clobbing other indexes in different processes. But this also means we have to change how we Search/Read Solr and truncate/destroy Solr indexes to handle this name spacing by process. In order to achieve this, we end up with a ton of hacks within Solr code (and its still not perfect...lots of flaking tests where I am probably still clobbing Solr indexes somewhere 😢). 

Now the big thing is, after getting this to work, there seemed to be no improvement in test speed times. So it appears we have some kind of bottleneck in our codebase? My guess is its Solr? But honestly have no idea 🤔. As an example our system tests still take around 2 minutes to run even when running 4 processes: 

```
➜  jupiter (enable-test-parallelization) ✗ rails test:system
Run options: --seed 19435
# Running:
Capybara starting Puma...
* Version 4.3.3 , codename: Mysterious Traveller
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:53258
Capybara starting Puma...
* Version 4.3.3 , codename: Mysterious Traveller
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:53289
Capybara starting Puma...
* Version 4.3.3 , codename: Mysterious Traveller
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:53309
Capybara starting Puma...
* Version 4.3.3 , codename: Mysterious Traveller
* Min threads: 0, max threads: 4
* Listening on tcp://127.0.0.1:53324
......................................S....
Finished in 152.109324s, 0.2827 runs/s, 3.0833 assertions/s.
43 runs, 469 assertions, 0 failures, 0 errors, 1 skips
You have skipped tests. Run with --verbose for details.
Coverage report generated for Minitest to /Users/murny/Code/jupiter/coverage. 1163 / 2835 LOC (41.02%) covered.
```

And similar results exist with our unit tests which show no speed improvements. So with the amount of work to get this to work, which just adds a ton of extra complexity to our codebase and increases our flakiness in our testsuite by a huge margin. All this for basically the exact same amount of test speed this makes no sense to keep pursuing. 

## What's New

Instead of continuing to pursue parallelization of our test suite, lets just remove this TODO and put this on hold for time being.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2443,7 +2443,7 @@ debug@2.6.9, debug@^2.2.0, debug@^2.3.3, debug@^2.6.9:
   dependencies:
     ms "2.0.0"
 
-debug@^3.0.0, debug@^3.1.1, debug@^3.2.5, debug@^3.2.6:
+debug@^3.0.0, debug@^3.1.1, debug@^3.2.5:
   version "3.2.6"
   resolved "https://registry.yarnpkg.com/debug/-/debug-3.2.6.tgz#e83d17de16d8a7efb7717edbe5fb10135eee629b"
   integrity sha512-mel+jf7nrtEl5Pn1Qx46zARXKDpBbvzezse7p7LqINmdoIk8PYP5SySaxEmYv6TZ0JyEKA1hsCId6DIhgITtWQ==
@@ -2478,11 +2478,6 @@ deep-equal@^1.0.1:
     object-is "^1.0.1"
     object-keys "^1.1.1"
     regexp.prototype.flags "^1.2.0"
-
-deep-extend@^0.6.0:
-  version "0.6.0"
-  resolved "https://registry.yarnpkg.com/deep-extend/-/deep-extend-0.6.0.tgz#c4fa7c95404a17a9c3e8ca7e1537312b736330ac"
-  integrity sha512-LOHxIOaPYdHlJRtCQfDIVZtfw/ufM8+rVj649RIHzcm/vGwQRXFt6OPqIFWsm2XEMrNIEtWR64sY1LEKD2vAOA==
 
 deep-is@~0.1.3:
   version "0.1.3"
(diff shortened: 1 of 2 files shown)


lagoan:
Would it be a good idea to do some profiling to find if Solr is indeed our bottleneck ? Future work of course.

pgwillia:
Thanks for your work on this. Good call.

One question. Does simplecov not work with parallelized tests? 
> 1163 / 2835 LOC (41.02%) covered

yarn.lock
```diff
@@ -5180,35 +5108,14 @@ normalize-url@^3.0.0:
...
-  integrity sha512-gqkfgGePhTpAEgUsGEgcq1rqPXA+tv/aVBlgEzfXwA1yiUJF7xtEt3CtVwOjNYQOVknDk0F20w58Fnm3EtG0fA==
-  dependencies:
-    npm-normalize-package-bin "^1.0.1"
-
-npm-normalize-package-bin@^1.0.1:
-  version "1.0.1"
-  resolved "https://registry.yarnpkg.com/npm-normalize-package-bin/-/npm-normalize-package-bin-1.0.1.tgz#6e79a41f23fd235c0623218228da7d9c23b8f6e2"
-  integrity sha512-EPfafl6JL5/rU+ot6P3gRSCpPDW5VmIzX959Ob1+ySFUuuYHWHekXpwdUZcKP5C+DS4GEtdJluwBjnsNDl+fSA==
-
-npm-packlist@^1.1.6:
```
pgwillia, line 5195: I'm assuming the yarn.lock changes are on purpose but minor.

murny:
Yeah thats a good catch. Simplecov doesn't work out of the box with parallelized tests. Theres some workarounds like merging the results of each process into a final output (more info here: https://github.com/colszowka/simplecov#merging-results) but mileage may vary/

Yarn lock changes was just me running/updating via  `yarn install`

Issue #1493, opened 2020-02-24
Remove Matomo tracking on Jupiter


weiweishi:
As an early experiment to see if we can replace GA because of privacy concerns, we implemented Matomo. But without a systematic review on what's been collected, and a lack of framework on our requirements for web stats, we would like to decommission Matomo at this stage, and will reevaluate our needs and requirements in the future, and assess if there will be a need for an in-house web analytics tool. 

Please remove the Matomo tracker we have implemented on Jupiter.

pgwillia:
Was introduced by this PR https://github.com/ualbertalib/jupiter/pull/602/files

murny:
Believe this is now done and merged in. Closing issue.

PR #1601, opened 2020-04-17
Removed Matomo analytic tracking

ConnorSheremeta:
#1493

Context
We no longer need Matomo tracking

Question: Does anything need to be mentioned to the sys admin regarding these changes on deployment?


ConnorSheremeta:
@murny Do you know if anything needs to be mentioned to the sys admin regarding these changes on deployment?

murny:
@ConnorSheremeta they will probably just have to clean up their scripts which override our secrets.yml to just remove Matomo vars on their end (but even if this isn't done, its not a big deal as setting these vars won't do anything now in the code base)

Issue #1585, opened 2020-04-08
[ERA] Fix rake tasks


ConnorSheremeta:
Post fedora two rake tasks are not working, fix them (reindex & preserve_all_items_and_theses). Remove 'turn existing filesets into attachments' rake task as it is no longer needed. Remove recover.rake file as it is no longer needed. Use find_each to retrieve AR objects in batches

PR #1605, opened 2020-04-21
Fix rake tasks

ConnorSheremeta:
#1585

Post fedora two rake tasks are not working, fix them (reindex & preserve_all_items_and_theses). Remove 'turn existing filesets into attachments' rake task as it is no longer needed. Remove recover.rake file as it is no longer needed. Use find_each to retrieve AR objects in batches


lib/tasks/recover.rake
```diff
@@ -1,65 +0,0 @@
-namespace :jupiter do
-  desc 'recover complete Solr index from data stored in Fedora, in the event of catastrophic loss of Solr data'
```
murny, line 2: Curious if we need something like this still? Recover/Repopulate Solr from AR objects?

ConnorSheremeta:
> This should all be fine, but I wonder if we shouldn't add some way to configure the batch size for all these `find_each` calls, so that we can tune this thing up and down on the fly to avoid overwhelming anything (the default batch size is going to bring back 1000 objects, which may or may not be too extreme?)

I have a hunch that 1000 objects shouldn't be too extreme but I'll make it configurable just in case. Better safe than sorry!

PR #1620, opened 2020-04-27
Remove TravisCI in favor of Github Actions

murny:
Now that we been using Github Actions for close to a month now and seems to be working great, lets pull the plug for TravisCI


lagoan:
This is a really good move. Will be looking into making the same changes on PMPY!

Quick question, will we be using danger ? Right now we get ``Not a GitHubActions Pull Request - skipping `danger` run.``

murny:
Yeah seems theres two events on Github Actions one for `pushes` and one for `pull requests`. So enabled the `pull requests` one as well which should trigger Danger properly. Also updated ruby/checkout while i was doing this

test/system/search_test.rb
```diff
@@ -236,7 +236,7 @@ class SearchTest < ApplicationSystemTestCase
   end
 
   test 'anybody should only see some facet results by default, with a "show more" button' do
-    skip 'This test continues to flap on CI that should be investigated ASAP' if ENV['TRAVIS']
```
murny, line 239: This would have mean these were passing originally on Github Actions (no flakey tests). But lets disable this for now. Update to use Github Actions ENV VAR instead.

Issue #1557, opened 2020-03-24
Validate the AIP with the metadata team


lagoan:
**Is your feature request related to a problem? Please describe.**
The AIP API returns information based on prefixes assuming that the required data will be there. As it stands there is missing information.

**Describe the solution you'd like**
The [documentation](https://docs.google.com/spreadsheets/d/1tavdlL_z9i8Rib7qUoSHmPJY7HRbMQs2x7x7nSnoI3Y/edit#gid=1733779686) provided by the metadata team outlines the required data. We need to double check the information available on Jupiter and use it where we can.

The work required for this is 
- [ ] Create sample AIPs to check current content  
- [ ] Compare predicates on Jupiter with the required data and check if it can be used to complete the missing data
- [ ] Consult with the metadata team on next steps to take

PR #1623, opened 2020-04-28
[WIP] Finalize AIP data #1557

lagoan:
## Context

We need to make sure the information provided for AIP creation is complete. This is a work in progress with the metadata team to make sure the data from the AIP API is complete and correct.

Related to #1557

## What's New

- Manually add predicate values that were not part of the initial RDF annotations
Some values are simply not on the RDF table annotations and they don't need to be there so they are now added on the controller
- Change n3 render method to show namespaces
The previous way the graph was rendered outputted the full predicates, we now have prettier graphs using the namespaces
- Address problems with the fixity report
The output had incorrect predicates, they are now correct
- Add Collections controller for AIP API
Entity graphs will show to which collection they belong to so now we have an entry point to show the collection information. This implementation has the most basic information there and will need to be developed at a later time
- Add radioactive Item to seeds    
In order to test the output from the AIP API we need to add entities that have values for all available predicates and make sure the resulting graph has these values


lagoan:
Closing this PR in favor of one coming from a branch.

PR #1624, opened 2020-04-28
Feature/finalize Item entity AIP data #1557

lagoan:
## Context

We need to make sure the information provided for AIP creation is complete. This is a work in progress with the metadata team to make sure the data from the AIP API is complete and correct.

Related to #1557

## What's New

- Manually add predicate values that were not part of the initial RDF annotations
    Some values are simply not on the RDF table annotations and they don't need to be there so they are now added on the controller
- Change n3 render method to show namespaces
    The previous way the graph was rendered outputted the full predicates, we now have prettier graphs using the namespaces
- Address problems with the fixity report
    The output had incorrect predicates, they are now correct
- Add Collections controller for AIP API
    Entity graphs will show to which collection they belong to so now we have an entry point to show the collection information. This implementation has the most basic information there and will need to be developed at a later time
- Add radioactive Item to seeds
    In order to test the output from the AIP API we need to add entities that have values for all available predicates and make sure the resulting graph has these values


app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,55 +210,21 @@ def file_paths
...
-    prefixes.each do |prefix|
-      result = result.or rdfable_entity.rdf_annotations
-                                       .where('predicate like :prefix',
-                                              prefix: "#{prefix}%")
+  def entity_institutional_repository_name
+    case @entity.class.to_s
+    when 'Item'
+      return 'IRItem'
+    when 'Thesis'
+      return 'IRThesis'
```
lagoan, line 218: If the IR names are used elsewhere these values should be moved to their respective model.

app/controllers/aip/v1/collections_controller.rb
```diff
@@ -0,0 +1,34 @@
+class Aip::V1::CollectionsController < ApplicationController
```
lagoan, line 1: This controller will grow when we get more data from the metadata team.

It could have been added to the aip_controller(now entity_controller) but using it there required special cases like checking if the entity had files which will cause complexity issues down the road.

For this reason this PR has entity and collection controllers.

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -1,4 +1,6 @@
-class Aip::V1::AipController < ApplicationController
+class Aip::V1::EntitiesController < ApplicationController
```
lagoan, line 1: As mentioned above, the aip controller could deal with collections and communities, but it could grow in complexity, separating this makes it simpler.

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -49,8 +53,32 @@ def show_entity
...
+    graph << rdf_type_statement(RDF::Vocab::PCDM.Object)
+
+    # To set the value for the predicate http://pcdm.org/models#memberOf we use
+    # the data from column member_of_paths, strip the collection id, and
+    # concatenate the url for the collection. This should be done cleaner
+
+    @entity.member_of_paths.each do |community_collection|
+      collection_id = community_collection.split('/')[1]
+      aip_base_url = request.url.split('/')[0..-3].join('/')
+      collection_url = "#{aip_base_url}/collections/#{collection_id}"
```
lagoan, line 65: As we will need the value from community and collection, these values should be retrieved from the model instead of splitting them here.

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,80 @@
...
+      subject: subject,
+      predicate: rdf_predicate,
+      object: stringed_value
+    )
+  end
+
+  def owner_email_statement
+    prepare_statement(
+      subject: self_subject,
+      predicate: RDF::Vocab::BIBO.owner,
```
lagoan, line 62: Currently the triplestore only has the RDF::Vocab::BIBO.owner information as id.

config/routes.rb
```diff
@@ -123,27 +123,35 @@
   # AIP API v1
 
   namespace :aip,
-            constraints: EntityConstraint.new,
```
lagoan, line 126: Instead of defining two times the same namespace one for entities with the EntityConstraint and another one for collections without the constraint I added the constraint to each call definition explicitly. Will be adding calls for communities at a later time.

db/seeds.rb
```diff
@@ -391,6 +391,104 @@ def file2.original_filename
...
+    #
+    # In order to set embargo values the visibility value needs to be set to 
+    # Item::VISIBILITY_EMBARGO
+    # embargo_history: 'An expired embargo was deactivated on 2000-01-01T00:00:00.000Z.  Its release date was 2000-01-01T00:00:00.000Z.  Visibility during embargo was restricted and intended visibility after embargo was open',
+    # embargo_end_date: '2000-01-01T00:00:00.000Z',
+    # visibility_after_embargo: JupiterCore::VISIBILITY_PUBLIC,
+  }
+
+  Item.new(    
+    base_radioactive_values.merge(id: 'e2ec88e3-3266-4e95-8575-8b04fac2a679')
```
lagoan, line 439: Adding separate entries for radioactive items with a defined id for ease of testing.

app/controllers/aip/v1/entities_controller.rb
mbarnett: What's the context here? There's some history involved that might help figure out if this is needed:

With the original version of Jupiter based on Fedora, the model names we went with for the "primary" models in the codebase, after discussion with metadata and the repository managers, were "Item" and "Thesis" – BUT we needed to auto-generate ActiveFedora versions of these models and we did it by prefixing "IR-" to the primary model names for the auto-generated models (we couldn't maintain the same names because models exist in a global namespace, and we didn't want to use a jargon-y name for the primary models, so that the repository manager and developers were using the same name to mean the same thing, for clear communication).

Internally, ActiveFedora had a special triple it used to record the model-name a record was associated with so that it new what model to instantiate when it read a record back, and would record "IRItem" or "IRThesis" because from ActiveFedora's view of the world, those were the models it was saving and instantiating. We tried to communicate to the metadata team at that time that that triple and its values were purely an artefact of ActiveFedora's internal implementation and not something that could or should be relied upon.

With the removal of Fedora (and therefore ActiveFedora) we've removed any autogenerated models with names with IR- prefixes. I don't know that there's a good reason to keep faking these names in the AIPs? They no longer have any meaning or relation to anything, and I'm not sure that it makes sense to preserve false information just based on what it once might have misleadingly seemed to be?

lagoan: @mbarnett I see what you mean, I thought the IR prefixes were a domain specific definition. 

@sfarnel what sort of names would be useful for describing the models?

mbarnett: Ah, gotcha. Yeah, we can keep them or use the actual model names or do something else entirely based on whatever makes sense for Metadata folks; those specific names were just kind of an accident of history.

sfarnel: Thanks both. Good material for discussions around terminology

app/models/collection.rb
```diff
@@ -16,16 +16,6 @@ class Collection < JupiterCore::Depositable
...
 
-  acts_as_rdfable do |config|
-    config.title has_predicate: ::RDF::Vocab::DC.title
-    config.fedora3_uuid has_predicate: ::TERMS[:ual].fedora3_uuid
-    config.depositor has_predicate: ::TERMS[:ual].depositor
-    config.community_id has_predicate: ::TERMS[:ual].path
-    config.description has_predicate: ::RDF::Vocab::DC.description
-    config.restricted has_predicate: ::TERMS[:ual].restricted_collection
-    config.creators has_predicate: ::RDF::Vocab::DC.creator
-  end
```
lagoan, line 27: Old code that had not been removed previously.

app/models/community.rb
```diff
@@ -22,14 +22,6 @@ class Community < JupiterCore::Depositable
     self.visibility = JupiterCore::VISIBILITY_PUBLIC
   end
 
-  acts_as_rdfable do |config|
-    config.title has_predicate: ::RDF::Vocab::DC.title
-    config.fedora3_uuid has_predicate: ::TERMS[:ual].fedora3_uuid
-    config.depositor has_predicate: ::TERMS[:ual].depositor
-    config.description has_predicate: ::RDF::Vocab::DC.description
-    config.creators has_predicate: ::RDF::Vocab::DC.creator
-  end
```
lagoan, line 31: Sames as in collection, this is just old code that needed to be removed.

db/seeds.rb
```diff
@@ -391,6 +402,123 @@ def file2.original_filename
     )
   end
 
+  # Radioactive entities
+  # TODO: Add our own radioactive community and collection
+
+  community_with_collection = Community.joins(:collections).first
+
+  base_radioactive_values = {
```
lagoan, line 410: This is the base values for the radioactive item provided for the metadata team. 

There are 5 items that pass the different validations. The base is what I am using as default and the other 4 are variations on it.

test/controllers/aip/v1/aip_collections_controller_test.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+require Rails.root.join('test/support/aip_helper')
+
+class Aip::V1::CollectionsControllerTest < ActionDispatch::IntegrationTest
```
lagoan, line 4: These are just stub tests to be filled out when we get radioactive examples from the metadata team.

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -59,17 +60,88 @@ class Aip::V1::ItemsControllerTest < ActionDispatch::IntegrationTest
...
     sign_in_as_system_user
     get aip_v1_entity_url(
       entity: @entity,
-      id: @public_item
+      id: items(:radioactive_item_base)
     )
     assert_response :success
 
     graph = generate_graph_from_n3(response.body)
-    assert_equal true, graph.graph?
```
lagoan, line 72: The original test that just checked if the putput was a valid graph has now been split into 5 to check the different validations from each radioactive item specialization/

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -29,28 +38,46 @@ def show_entity
...
+    )
 
-    owners_email = owner_email_statement
-    graph << owners_email unless owners_email.nil?
+    # Handle special case where predicate http://projecthydra.org/ns/auth/acl#embargoHistory
+    # needs to maintain the order in which it was entered. The original nodes
+    # the predicate are removed and a new rdf list is inserted instead with the
+    # predicate
+
+    if graph.has_predicate?(::TERMS[:acl].embargo_history)
```
lagoan, line 69: The value for embargo_history needs to maintain the order. Because order is not guaranteed, for this reason we remove all embargo_history statements and add a single one with its values in an rdf list that does maintain order.

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,108 @@
+module GraphCreation
```
lagoan, line 1: This behavior was defined before in the aip entity controller. I moved it to this concern to be used on communities, collections, etc.

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -1,4 +1,12 @@
-class Aip::V1::AipController < ApplicationController
+class Aip::V1::EntitiesController < ApplicationController
+
+  include GraphCreation
+
+  ITEM_CLASS_NAME = 'Item'.freeze
+  THESIS_CLASS_NAME = 'Theses'.freeze
+  ITEM_INSTITUTIONAL_REPOSITORY_NAME = 'IRItem'.freeze
+  THESIS_INSTITUTIONAL_REPOSITORY_NAME = 'IRThesis'.freeze
+  ENTITY_INSTITUTIONAL_REPOSITORY_NAME = 'IREntity'.freeze
```
lagoan, line 9: For now we are keeping the IR names for all entities.

db/migrate/20200529212705_annotate_item_with_creator_predicate.rb
```diff
@@ -0,0 +1,9 @@
+class AnnotateItemWithCreatorPredicate < ActiveRecord::Migration[6.0]
+  def change
+    add_rdf_table_annotations for_table: :items do |t|
+      # This change removes the predicate RDF::Vocab::BIBO.authorList defined in
+      # 20190926213733_annotate_tables_with_rdf
+      t.creators has_predicate: ::RDF::Vocab::DC.creator
```
lagoan, line 6: Removing the authorList predicate that used to save the creators in a stringified JSON array. Now we keep the creator statements and add a separate authorList predicate to hold the values in an rdf list.

test/fixtures/items.yml
```diff
@@ -67,3 +67,166 @@ authenticated_item:
   publication_status: <%= [CONTROLLED_VOCABULARIES[:publication_status].published] %>
   subject: ['Secret things']
   member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
+
+radioactive_item_base:
```
lagoan, line 71: These values are copied from the seed entries.

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,100 @@
...
+  )
+
+    {
+      statements_to_insert: derivate_list_values(
+        rdfable_entity,
+        subject,
+        rdf_original_predicate
+      )
+    }
+  end
```
mbarnett, line 99: This is where the split vertical param declarations, along with the vertical hash declaration, become incredibly confusing visually to me.

```ruby
def delete_insert_list_values_statements(rdfable_entity, subject, rdf_original_predicate)
  {
    statements_to_insert: derivate_list_values(rdfable_entity, subject, rdf_original_predicate)
  }
end
```

to me, it's much easier to tell at a glance what the scope is, what the params are, and that the hash only has a single member, not 4.

db/seeds.rb
```diff
@@ -38,6 +38,17 @@
...
+  # Seed system user for API requests
+  User.create(
+    email: 'ditech@ualberta.ca',
+    name: 'System user',
+    admin: false,
+    api_key_digest: BCrypt::Password.create(
+      Rails.application.secrets.system_user_api_key
+    ),
+    system: true
+  )
```
mbarnett, line 50: This is kind of another style deviation I'm noticing a lot in this PR (I'm not going to go through and comment on each one, but there's a bunch).

When method calls are too long for 120 character lines, we almost universally (to the point that I'm surprised we weren't enforcing this in Rubocop somehow considering how much else it likes to yell at us about) have the first param on the same line as the opening parentheses, and the closing parentheses after the last param. To give just one example:

https://github.com/ualbertalib/jupiter/blob/master/test/system/admin_users_show_test.rb#L163-L171

Occasionally with really long method names even the first param has to drop down to the second line, but I'd prefer to avoid that where possible, and having the closing parentheses stand on its own is a bit messy visually.

The upshot is that this again makes it a lot easier to tell literal hashes from param lists at a glance

```ruby
User.create(email: 'ditech@ualberta.ca',
            name: 'System user',
            admin: false,
            api_key_digest: BCrypt::Password.create(Rails.application.secrets.system_user_api_key),
            system: true)
```

vs

```ruby
{
            email: 'ditech@ualberta.ca',
            name: 'System user',
            admin: false,
            api_key_digest: BCrypt::Password.create(Rails.application.secrets.system_user_api_key),
            system: true
}
```

I'd really like to stick with the already dominant style, and we should probably see if this is something Rubocop could be configured to help with.

mbarnett: Noticing on further inspection that seeds.rb was already kind of inconsistent with the rest of the codebase. I'm going to blame Chris because he kept trying to stick to 80 character lines and was very much an ex-C guy 😉. Now's probably as good a time as any to clean this up though.

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -1,5 +1,6 @@
 require 'test_helper'
 require 'json-schema'
+require 'rdf/isomorphic'
```
murny, line 3: Not needed. If you want to leave this, then add `require: false` on the gemfile (e.g: `gem 'rdf-isomorphic', '~> 3.1.0', require: false`)

app/controllers/aip/v1/collections_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      object: 'IRCollection'
+    )
+
+    render plain: graph.to_n3, status: :ok
+  end
+
+  protected
+
+  def load_and_authorize_collection
+    authorize @collection = Collection.find(params[:id])
```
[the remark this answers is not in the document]
murny: Yeah I'd echo this. Should be two lines:

```
@collection = Collection.find(params[:id])
authorize @collection
```

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,100 +209,123 @@ def file_paths
...
-  def create_graph(rdfable_entity, prefixes, subject = nil)
-    subject = self_subject if subject.nil?
-    graph = RDF::Graph.new
-    annotations = get_prefixed_predicates(rdfable_entity, prefixes)
-
-    annotations.each do |rdf_annotation|
-      column = rdf_annotation.column
-      value = rdfable_entity.send(column)
+  def entity_institutional_repository_name
+    case @entity.class.to_s
```
murny, line 213: Tiny style suggestion instead of using "returns" could just use a proper case statement?
```ruby
def entity_institutional_repository_name
  case @entity.class.to_s
  when ITEM_CLASS_NAME
     ITEM_INSTITUTIONAL_REPOSITORY_NAME
  when THESIS_CLASS_NAME
     THESIS_INSTITUTIONAL_REPOSITORY_NAME
  else
     ENTITY_INSTITUTIONAL_REPOSITORY_NAME
  end
end
```

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -1,4 +1,12 @@
-class Aip::V1::AipController < ApplicationController
+class Aip::V1::EntitiesController < ApplicationController
+
+  include GraphCreation
+
+  ITEM_CLASS_NAME = 'Item'.freeze
+  THESIS_CLASS_NAME = 'Theses'.freeze
```
murny, line 6: How come thesis is pluralized here and item is not? (`Item` vs `Theses`)?

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,100 +209,123 @@ def file_paths
...
 
-  def get_prefixed_predicates(rdfable_entity, prefixes)
-    result = RdfAnnotation.none
-    ActsAsRdfable.add_annotation_bindings!(rdfable_entity)
-    prefixes.each do |prefix|
-      result = result.or rdfable_entity.rdf_annotations
-                                       .where('predicate like :prefix',
-                                              prefix: "#{prefix}%")
+  def load_and_authorize_entity
+    case params[:entity]
```
murny, line 231: Minor nit, if theres just two "cases' then typically a if, elsif is better? 
```ruby
@entity = If params[:entity] == Item.name.underscore.pluralize
  Item.find(params[:id])
elsif  params[:entity] == Thesis.name.underscore.pluralize
  Thesis.find(params[:id])
end
```


(Unless you expect more "cases" in the future?)

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,100 +209,123 @@ def file_paths
...
+    # There is a routing constraint specifying which models are available
+    # through the url. We need to update the routing constraint whenever a new
+    # entity is made available
+    when Item.name.underscore.pluralize
```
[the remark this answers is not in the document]
murny: Also got questions about this 🤔

Isn't `Item.name` or `Thesis.name` just "Item" or "Thesis"?

Therefore, the `underscore` is doing nothing for us (it actually causing "Item"/"Thesis" to be lowercased as a side effect).

`Item.name.underscore` => "item"
`Item.name.underscore.pluralize` => "items"

If "items" is what we actually want, then I'd how about we use `.downcase` instead of `.underscore`?

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,100 +209,123 @@ def file_paths
...
   end
 
-  def file_statements
+  def entity_file_statements
     file_statements = []
-    @entity.files.each do |file|
-      fileset_url = "#{request.original_url}/filesets/#{file.fileset_uuid}"
-      statement = prepare_statement(
+
+    @entity.try(:files).try(:each) do |file|
```
murny, line 257: I question the potentially overuse of ".try" here? Wouldn't all ".files" method on an entity return an array? (doesn't respond to .each method?). Shouldn't all entities have a ".files" method? If the answer to these questions is yes, then we won't need these ".try" methods 🤔

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -172,100 +209,123 @@ def file_paths
...
+      object: entity_institutional_repository_name
+    )
+  end
+
+  def entity_type_statements
+    statements = [
+      rdf_type_statement(RDF::Vocab::PCDM.Object)
+    ]
+
+    statements << rdf_type_statement(RDF::Vocab::BIBO.Thesis) if @entity.class.to_s == THESIS_CLASS_NAME
```
murny, line 326: If you look at my comment regarding `THESIS_CLASS_NAME = "Theses".frozen` I question if this is working correctly. 

Maybe need some test coverage here for when @entity is thesis and making sure this is working?

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,100 @@
...
+      values.each do |value|
+        next if value.blank?
+
+        statement = RDF::Statement(
+          subject: subject,
+          predicate: RDF::Vocabulary::Term.new(rdf_annotation.predicate),
+          object: value
+        )
+
+        graph << statement unless statement.nil?
```
murny, line 29: can statement be nil here 🤔 ? That seems really odd if RDF::Statement returns a nil when you giving it a valid subject, object (as we check if value is blank) and a predicate?

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    graph
+  end
+
+  def get_prefixed_predicates(rdfable_entity, prefixes)
+    result = RdfAnnotation.none
+    ActsAsRdfable.add_annotation_bindings!(rdfable_entity)
+    prefixes.each do |prefix|
+      result = result.or rdfable_entity.rdf_annotations
```
murny, line 40: Use more brackets? Seems weird to be jumping between no brackets and brackets within the same assignment

e.g use this
```ruby
result.or(something.where(somethingelse))
```
instead of:
```ruby
result.or something.where(somethingelse)
```

murny, line 40: Also I assume prefixes can be multiple? Since we `.each` and looping through them. So we loop through them and we only really care about the last iteration of what "result" will be. We then return this "result". I question why we loop through them all if we only care about the last prefix's result?

app/policies/collection_policy.rb
```diff
@@ -4,4 +4,11 @@ def index?
     true
   end
 
+  # This policy is used for the AIP V1 API. Pundit does not allow use of
```
murny, line 7: What does "Pundit does not allow use of namespaces" mean? Just curious. Maybe I can help or offer suggestions of what you trying to do.

app/policies/collection_policy.rb
```diff
@@ -4,4 +4,11 @@ def index?
...
+  # namespaces
+
+  def show_collection?
```
murny, line 10: This is name is a bit misleading? This is the collection policy.

So we have a `show?` method already (inherited from application policy) which is set to true (as everyone can see the show page of a collection).

Now we have a "show_collection?" method which is saying... show collection only if your system or admin user. 

Hopefully you can see the confusion. Can we come up with a better name here? 🤔

app/policies/collection_policy.rb
```diff
@@ -4,4 +4,11 @@ def index?
     true
   end
 
+  # This policy is used for the AIP V1 API. Pundit does not allow use of
+  # namespaces
```
murny, line 8: I suppose your comment gives more info with helping with the naming problem (and maybe I see what you mean by pundit does not allow namespaces, as you want it to be "AIP")

But I think this may be getting outside of the area of Pundit and Authorization. Seems like what you want is Authentication?

You want to lock down all these AIP Routes to just System and Admin users correct? Then this should be done via authentication and not authorization (look at how we did Admin area for ideas?).

To put it another way, Pundit does everything based of Models (not controllers, or routes). Should everyone be able to see the details of a collection? Yes. Here we saying "yes". But  we then say "no" if that request is being served under a certain controller (AIP API). Which is where this falls out of Pundits duties.

This goes for the same thing regarding the polices found in item/thesis for these AIP APIs

db/seeds.rb
```diff
@@ -391,6 +402,123 @@ def file2.original_filename
...
   end
 
+  # Radioactive entities
+  # TODO: Add our own radioactive community and collection
+
+  community_with_collection = Community.joins(:collections).first
+
+  base_radioactive_values = {
+    # Set id on each new Item so we can find it easily when testing
+    # id: 'e2ec88e3-3266-4e95-8575-8b04fac2a679',
```
murny, line 412: Maybe instead of commenting out data all over this default hash. Just set up your default one. Then override it when you want to. 

Can override the id, and any other values you want to change?

db/seeds.rb
```diff
@@ -391,6 +402,123 @@ def file2.original_filename
...
+      visibility: Item::VISIBILITY_EMBARGO,
+      embargo_end_date: '2000-01-01T00:00:00.000Z',
+      embargo_history: [
+        'acl:embargoHistory1$ An expired embargo was deactivated on 2000-01-01T00:00:00.000Z.  Its release date was 2000-01-01T00:00:00.000Z.  Visibility during embargo was restricted and intended visibility after embargo was open', 
+        'acl:embargoHistory2$ An expired embargo was deactivated on 2000-01-01T00:00:00.000Z.  Its release date was 2000-01-01T00:00:00.000Z.  Visibility during embargo was restricted and intended visibility after embargo was open'
+      ],
+      visibility_after_embargo: CONTROLLED_VOCABULARIES[:visibility].public
+    )
+  ).tap do |uo|
+    uo.add_to_path(community_with_collection.id, community_with_collection.collections[0].id)
```
murny, line 505: Curious if we should file an issue in the backlog to clean up all this `add_to_path` abuse? It would remove all the "tap"ing we do in seeds/tests/etc. This was fine in Fedora days...but now its just painful to see.

We even still using the "uo" iterative name 😕 

E.g these are the same thing in ActiveRecord:
```ruby
# current way
Item.new(
  something...
  somethingelse...
).tap do |uo|
  uo.add_to_path(community_id, collection_id)
  uo.save!
end

# Better way
Item.create(
  something...
  somethingelse...
  member_of_paths: ["#{community_id}/#{collection_id}"]
)

File: test/fixtures/items.yml (hunk 17 above)
Comment (murny) [inline test/fixtures/items.yml:71]: I want to put a lot of caution here. Do we need these things as a fixtures? We use all of these new fixtures, once. The one time we are using these are for just retrieving its "Id" value, which is hardcoded (we may be using them indirectly via controller code to pull them from the DB but this isn't the point).

Having this as a fixture impacts EVERY single test in Jupiter. These 5 new items will be created before every single test we have! So I can't stress this enough, we all have to put some thought in when we are adding new fixtures to the codebase. We just doubled our item fixtures right here 😓.  We also have to guarantee these new items will be valid and meet all validation we have in our model. Considering they are "radioactive" and are not using our Controlled Vocab we have, this makes me very nervous. 

So I highly question the purpose of all these fixtures. Some questions we should be asking before adding new fixtures. What are we actually trying to test? Can I reuse a fixture? If not, maybe I can reuse a fixture, but tweak it a bit for my purpose? If not, then would the rest of Jupiter benefit from having this as a fixture (aka would it be reusable and not just for my test case?)?  If not, then it might server to just add the data to my test and not use a fixture.

File: test/fixtures/rdf_annotations.yml [hunk 35]
```diff
@@ -67,213 +67,213 @@ predicate_16:
   :column: subject
```
murny, line 67: Same comment goes for these rdf_annotations fixtures? Do we need 100 rdf_annotations fixtures ran before every single test in Jupiter? What are we actually trying to test here? Why we seeding all this data? Is every single AIP test need all 100 rdf annontations loaded into the DB before it can run it's assertions? Or does a single test just need a couple of these? or some none at all?

config/routes.rb
```diff
@@ -123,27 +123,35 @@
...
     namespace :v1 do
+      get '/collections/:id',
+          to: 'collections#show_collection',
+          as: 'collection'
       get '/:entity/:id',
-          to: 'aip#show_entity',
-          as: 'entity'
+          to: 'entities#show_entity',
+          as: 'entity',
+          constraints: EntityConstraint.new
```
murny, line 134: I kinda question the use of a constraint here. 

This logic should be handled on the controller layer, not the router layer. Why? If someone passes in a bad "entity" param, our constraint will handle this and return false, which makes the error code a 404 (route doesn't exist).

What should happen instead is you get a proper error message back with a proper error code. This is an API we our building after all, so we should be nicer to the consumers of our API if they accidentally pass in an invalid "entity"

config/routes.rb
```diff
@@ -123,27 +123,35 @@
   # AIP API v1
 
   namespace :aip,
-            constraints: EntityConstraint.new,
             defaults: { format: :n3 } do
     namespace :v1 do
+      get '/collections/:id',
+          to: 'collections#show_collection',
```
murny, line 129: We should be RESTful here and using rails standards? These `show_X` actions should just be `show` and follow normal rails standards. `/collections/:id` is literally the show action. Why try to create our own standard of naming it show_collection instead?

We could just use resource route helper here instead and we don't need to name these routes (`as: entity`)

app/controllers/concerns/graph_creation.rb
lagoan: I get how concerns can get problematic if we are not careful and services will push us towards separating the logic. Will look into switching it to a service.

app/controllers/aip/v1/entities_controller.rb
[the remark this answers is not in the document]
lagoan: @mbarnett you are right, table_name is better. As for the value of the param, I believe urls are case sensitive in unix, we should be consistent with the rest of the paths.

app/controllers/concerns/graph_creation.rb
```diff
@@ -0,0 +1,100 @@
...
+  def self_subject
+    RDF::URI(request.url)
+  end
+
+  def derivate_list_values(
+    rdfable_entity,
+    subject,
+    rdf_original_predicate,
+    rdf_list_predicate = nil
+  )
```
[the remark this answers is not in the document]
lagoan: There is a lot to be said about the benefits and disadvantages of 80, 100, and 120 character limits when programming, the internet is full of posts about this subject. I find it hard to read code with 120 character limit; Your example is an extreme case but for me, new line definitions are clearer.

Having said this, I won't argue for changing the full codebase for this and will make the changes to match the rest of the code.

db/seeds.rb
lagoan: I prefer ``new_line``. Maybe the rest of the team can provide input on preferences.

app/controllers/aip/v1/entities_controller.rb
lagoan: Good catch!

app/controllers/aip/v1/entities_controller.rb
lagoan: Yes, we are expecting more cases in the future.

app/controllers/aip/v1/entities_controller.rb
lagoan: Definite overuse of ```.try```. This is leftover code from a special case that is no longer a problem. Good catch!

app/controllers/aip/v1/entities_controller.rb
lagoan: Have not added the radioactive thesis examples so that is why this was not causing a problem. This PR is big enough as it is and will add the rest of the Thesis code in another one. Will make the fixes you suggest here though.

app/controllers/concerns/graph_creation.rb
lagoan: This must be leftover code as well, need to be more careful with these errors in the future.

app/controllers/concerns/graph_creation.rb
lagoan: I should add documentation here as this .or is kind of wierd, your comment made me doubt myself. This ```.or``` is compounding the results not just saving the last element.

app/policies/collection_policy.rb
lagoan: Will take a look at the Admin area and see if it is something worth changing.

db/seeds.rb
lagoan: I see what you mean, I can move these comments to each instantiation where I am adding the changes to values for non default items

db/seeds.rb
pgwillia: I prefer `new_line` mostly because the opening and closing braces are lined up.

# bad
[
  :a,
  :b ]

# bad
[ :a,
  :b ]

# good
[ :a,
  :b
]

# good
[
  :a,
  :b
]

db/seeds.rb
lagoan: If there is no other bennefit of using the add_to_path method i can make the changes you suggest. The ```tap``` would still be needed for ingesting files, no?
```
.tap do |uo|
    # Attach a file
    File.open(Rails.root + 'test/fixtures/files/image-sample.jpeg', 'r') do |file1|
      uo.add_and_ingest_files([file1])
    end
    uo.set_thumbnail(uo.files.first) if uo.files.first.present?

    uo.save!
  end
```

test/fixtures/items.yml
lagoan: We should be using the radioactive examples rather than the minimal examples we have right now as they include all of the possible data.

I see the benefit of having the data as provided by the metadata team as it is what they expect to be there, I can change the seeds to use the controlled vocabulary.

db/seeds.rb
mbarnett: Did a quick survey of some major ruby codebases in the spirit of "lacking some other good argument otherwise, we should default to what the larger community does":

1) Rails – doesn't seem to have this issue (no setting for rubocop, I couldn't find any examples of a split-line call). They don't cap line length and seem to prefer to keep method calls on a single line of whatever length is necessary.

2) Discourse and Shopify both use new_line.

app/controllers/concerns/graph_creation.rb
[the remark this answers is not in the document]
mbarnett: For what it's worth, we had a long conversation about this when we started the project. Data in favor of 80 column lines seems entirely based on historical punch card limitations with some weak support from readability studies on prose books where the text starts in the first column without any leading indent and little intra-line whitespace.

120 was chosen based on wanting to keep consistent with the industry, drawing on previous experience in other Rails shops and Shopify using 120 in their standards (and Rails taking the position that lines should be as long as they need to be).

test/fixtures/rdf_annotations.yml
lagoan: We need them for all AIP tests. We fetch the predicates and their values to form the resulting graphs. The information that comes as output in is just as important as the information that does not.

config/routes.rb
lagoan: I understand what you mean by the ```show_collection``` name. It was to do with the comment you made (before)[https://github.com/ualbertalib/jupiter/pull/1624#discussion_r441287315] and how pundit handles things.

Will take a look at using resource route helper.

config/routes.rb
lagoan: The idea of using a constraint rather than having the logic in the model came from a discussion with @mbarnett where we talked how "a constraint will bounce a request for an unknown model faster than catching it on the AIP API controller."

For now I suggest we keep the current implementation and discuss the benefits of a friendly api vs an efficient api on another occasion.

test/fixtures/items.yml
murny: If you want to beef up he current item fixtures we have, so they include all of the possible data then be my guess, I'd be totally for this (assuming the are using our controlled vocab and are valid items). 

But we shouldn't be adding 5 brand new fixtures for a couple tests with no real justification for them that I can see? So my previous comment is still valid and my concerns are still there.

mbarnett: I can see some value in beefing up existing fixtures with some of the data from _a_ radioactive item, to make the fixtures a bit more realistic, but I think Shane's right: there needs to be a balance here, and it doesn't make sense to add these 5 different kinds of radioactive items as fixtures when 99% of the tests in the codebase never use these fixtures but will have to pay the price for loading them into the DB.

Fixtures are a shared concern, like gem dependencies, where we always want to be really thoughtful and make sure the entire team is onboard with and benefiting from any changes we're making to the files. They also play a big role in test suite performance.

Changes I'd like to see here:

1) Use some of this data to beef up existing fixtures, but using controlled vocab etc to keep us in sync with the models. Maybe we pick one of the radioactive fixtures and use it to flesh out any values missing from the other fixtures.

2) Move the rest of this data into only the tests that are using them. This eliminates both the unneeded loading for all of the tests that don't need them, and ensures that any "radioactive" data that doesn't match the model validations doesn't accidentally break tests elsewhere by being malformed vs what other code paths are expecting.

config/routes.rb
murny: I could be wrong, but I think the performance/efficiency of having the check in controller or route is almost negligible. And on top of that, we already doing the check in the controller anyways (so you could probably argue, it  doesn't matter on performance since we doing the check in the controller anyways and by removing the route constraint will probably make it faster since its less code 😄)

But the difference now, is either you get a 404 error if you use the constraint, or a proper error message that tells the user what actually happened (if we make sure we handling it properly) for almost zero effort. Plus less code and simpler routes. Seems like a win, win.

mbarnett: Remember that this is an internal API, so there aren't really any end users to see the error message anyways. 

The routing constraint was my idea. Previously iirc you could stick any "entity" name into the API call and the request would make it into the controller and into the initial logic of trying to map the entity name to a model class, and perform the load if the name was "handled" in the cases.

That feels...fishy? To me? I'd probably have baked the model names into the routes themselves to ensure that unhandled model "routes" simply didn't exist, but the constraint approach is basically the same in the end. 

Otherwise these are kind of "god" routes that handle an infinite number of requests, ie) without the constraint, /aip/v1/foobar/12345 is technically a valid route that gets rejected at the controller level, whereas with the constraint approach the only valid routes are /aip/v1/items/ and /aip/v1/theses, just as if they were hardcoded.

murny: That makes sense. 

I guess I recommend we not have these "god" routes at all that attempt to handle all "entities" (currently 2, could be a few more)? I rather go more RESTful especially since we broke away from this current model with "collections" already within this PR (we added a separate route for it). Which probably means we will do the same for "communities" too. "Entity" is not a common terminology in Jupiter, and this can cause confusion as well. We also never did this approach on the non API routes (Theses are considered Items and are handled under Items routes and if we did split them out, or add newspapers, they would get their own route namespace).

So feel like we should stay with Rails standards (which would fix my "show_collection" comment for the "show" page) and It's far easier to just do this (even if its a bit more repeated code):

```ruby
resources :collections, only: [:show]
resources :theses, only: [:show] do
   resources :filesets do
     member do
       get :fixity
       get :original_path
   end
   resources :file_paths
end
resources :items, only: [:show] do
   resources :filesets do
     member do
       get :fixity
       get :original_path
   end
   resources :file_paths
end
```

vs our current code:

```ruby
 get '/collections/:id',
          to: 'collections#show_collection',
          as: 'collection'
      get '/:entity/:id',
          to: 'entities#show_entity',
          as: 'entity',
          constraints: EntityConstraint.new
      get '/:entity/:id/filesets',
          to: 'entities#file_sets',
          as: 'entity_filesets',
          constraints: EntityConstraint.new
      get '/:entity/:id/file_paths',
          to: 'entities#file_paths',
          as: 'entity_file_paths',
          constraints: EntityConstraint.new
      get '/:entity/:id/filesets/:file_set_id',
          to: 'entities#file_set',
          as: 'entity_file_set',
          constraints: EntityConstraint.new
      get '/:entity/:id/filesets/:file_set_id/fixity',
          to: 'entities#fixity_file',
          as: 'entity_fileset_fixity',
          constraints: EntityConstraint.new
      get '/:entity/:id/filesets/:file_set_id/original_file',
          to: 'entities#original_file',
          as: 'entity_fileset_original_file',
          constraints: EntityConstraint.new
```

So its actually less code as of right now but very easy to read and understand. Keeps us being Restful where possible. If we add more "entities" then it might grow bigger which isn't a huge problem. If it gets really large, then there's solutions we can tackle to clean this up then. For example we can create a method which gives the inside part and reuse that to help reduce the amount of code (or move it into its own route file which gets loaded here (see gitlabhq for a major app that split their routes into files as an example), etc):

```ruby
def entity_filesets_routes
resources :filesets do
     member do
       get :fixity
       get :original_path
   end
   resources :file_paths
end

resource :collections
resource :communities
resource :items do 
  entity_filesets_routes
end
resource :items do 
  entity_filesets_routes
end
resource :newpapers do 
  entity_filesets_routes
end
```

Just a suggestion though. From looking at this code I see we are trying to pre optimize and building for ERA in 30 years out (we could have 10 entities! we need to build this general solution that will fit everything!). But probably realistically, we aint going to need it (YAGNI) and just making our code far more complicated in the short term. But I am missing this entire context so I could be out to lunch 😅

lagoan: Those routes do look prettier! I tried them but call to filesets is no longer directed to the items controller as it is now, but it expects a Fileset controller.

On the name entity: entities is a domain term passed by the metadata team that will be used more in developing jupiter, it describes everything that can be ingested and saved into a collection. Item, Thesis, and everything to come will be called entities. I don't thing we need to use quotations when referring to them.

When I started working on this Item and Thesis controller they had the same behaviors and were described the same so that is why I implemented them as an entity controller. This decision was not to done for optimization, but rather to avoid code duplication as this would be prone to errors.

As we continue describing entities differences in Item and Thesis are starting to come up, so now it makes sense to have them separate as more differences will likely come up now, as well as when we add more entities. At some point you mentioned that having dumb controllers was OK, you were in fact correct in this case.

As it stands right now, fixing this is looking more like refactoring, and as we should be getting closer to release I propose we set up a ticket to come back to this issue and try to release on the current proposed timeline. What do you think about this @mbarnett ?

mbarnett: Yeah, I think maybe setting up a larger refactor of our approach here makes sense while going with this in the short term. 

In the last couple of weeks I've gotten a better understanding of how people see the newer data models in terms of how related they are to the existing one, and they're sounding less and less like something we'd be able to share much logic with in terms of AIPs anyways, so I think we're just learning that just like there's differences between Item & Thesis, there's going to be a lot more differences with the digitization content, so for the longer-term approach it might make sense to go back to the drawing board and maybe taking an approach somewhat like the Solr approach might be worth considering (`exporters/aip` or something). It's something we can capture and discuss on the new ticket.

app/policies/aip_policy.rb
```diff
@@ -0,0 +1,7 @@
+class AipPolicy < ApplicationPolicy
+
+  def access?
```
murny, line 3: Nice cleanup here with the policy stuff 👍

test/support/aip_helper.rb
```diff
@@ -94,4 +93,214 @@ def create_entity(
...
+                                    predicate: TERMS[:ual].thesis_level)
+    RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'title', predicate: RDF::Vocab::DC.title)
+    RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'unicorn', predicate: TERMS[:ual].unicorn)
+    RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'visibility_after_embargo',
+                                    predicate: TERMS[:acl].visibility_after_embargo)
+    RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'visibility',
+                                    predicate: RDF::Vocab::DC.accessRights)
+  end
+
+  def seed_all_rdf_annotations
```
murny, line 299: Nice solution to get around using fixtures 🎉

app/policies/aip_policy.rb
lagoan: It is definitely cleaner. Thanks for the suggestion!

test/support/aip_helper.rb
lagoan: Thanks! I was worried that instantiating all those fixture for each AIP API test would be slower than adding them on every test through fixtures since those are just entered in the database. However, I did a few manual checks and did not see a difference with either approach. This can of course change as we add more test that do not need the annotations though so we can load them only when needed.

Issue #1449, opened 2020-01-31
Subject strings concatenated


sfarnel:
**Describe the bug**
Some items in ERA have all of the subject terms concatenated into a single string.

**To Reproduce**
For example, https://era.library.ualberta.ca/items/0d9000f4-3c7e-4221-9f26-f5faf0a1d1bb

**Expected behavior**
Each subject word or phrase should be a single item in the list

**Additional context**
A list of impacted items (based on triplestore query) is attached. @weiweishi  This had happened for another collection previously and was cleaned up via script so hoping the same can be done again.
[ERA_subject_issue.txt](https://github.com/ualbertalib/jupiter/files/4141023/ERA_subject_issue.txt)
Thanks!

pgwillia:
Something like 
```
namespace :jupiter do
  desc 'Find Subjects that have been concatenated together and split on comma seperator'
  task unconcatenate_subjects: :environment do
    puts 'Unconcatenating subjects...'
    (Item.all + Thesis.all).each do |item|
        print '.' 
        next unless item.subject.any? {|subject| subject.include? ','}
        
        print '-'
        item.subject = item.subject.first.split(%r{,\s*}) 
        item.save!
    end
    puts 'Subjects unconcatenated!'
  end
end
```

will look at all the Items and Theses and un-concatenate Subjects.  This assumes that there aren't any `,` purposefully embedded in subjects.  @sfarnel is there a way to know if this assumption is correct.  Alternately I could just use the provided ERA_subject_issue.txt to get the uuid's that you've mentioned.

I could prepare this as a one of rake task (in master or integration_postmigration) or run it in the rails console.  Not certain what we've done in the past, or want to do in the future.  Would appreciate some guidance or history from @mbarnett @weiweishi

sfarnel:
Thanks @pgwillia!
Many items and theses in ERA have Library of Congress Subject Headings which do make legitimate use of commas in many instances, so I think safer to use the list provided to do a targeted fix. We can then run another query and find any additional that may have sneaked in.

pgwillia:
@weiweishi said
> Sorry I didn't have any formalized script I could share. It has always been a direct Rails Console fix in the past
> Most of the batches we need to work on are smaller than 1,000 items and I've never run into any issues.
> I have always output the IDs, title, and a link and some basic information as a base report for metadata team to review.

I gave @henryzhang87 that I was working on this I'll continue with this in mind.

My plan is to practice, practice and then execute.

[...]

PR #1627, opened 2020-05-01
add migration to fix concatenated subjects (part 1)

pgwillia:
This will take a subject like ["Cat, Dog, Rabbit"] and make it ["Cat", "Dog", "Rabbit"] in a new deconcatenated_subject column. It will also create a log to track the identified ids that looks like
```
I, [2020-04-30T18:02:17.788696 #5414]  INFO -- : 7ee24835-4820-434d-911b-0a293ab456ba, Former embargo of Chad: the day the librarians were free, ["Librarian, Freedom"], ["Librarian", "Freedom"]
I, [2020-04-30T18:02:18.133287 #5414]  INFO -- : 8bc8353c-aace-4b27-90c4-1b763b85650f, Impact of non-admin users on librarians, ["Librarian, Equality"], ["Librarian", "Equality"]
I, [2020-04-30T18:02:18.318830 #5414]  INFO -- : 9593c260-033d-46d2-bae9-5b39069d6f75, The complete effects of Double Bastard Ale on librarians, ["Librarian, Mondo"], ["Librarian", "Mondo"]
```
@sfarnel does this suit your need for review?

We're doing this in two parts to make data cleanup a little more sane than edits in production.

This assumes that at some point in the future this new column will be reviewed and replace the Item's subject.

#1449 

## What's New

A migration to create a deconcatenated_subject column and copy data over for review.


db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -0,0 +1,21 @@
+class DeconcatenateSubjectInItemAndThesis < ActiveRecord::Migration[6.0]
+
+  def change
+    add_column :items, :deconcatenated_subject, :json
+    add_column :theses, :deconcatenated_subject, :json
+
+    concatenated_subject_ids = File.readlines('db/migrate/ERA_subject_issue.txt').collect(&:strip)
+    report = Logger.new('log/concatenatedsubjects.log')
+    (Item.all + Thesis.all).each do |item|
```
mbarnett, line 9: (Item.all + Thesis.all) would end up instantiating about 50,000 objects up front and holding them all in memory until the end of the migration – net result is that this would probably be pretty hard on the production servers. Probably better to use something like: https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -0,0 +1,21 @@
+class DeconcatenateSubjectInItemAndThesis < ActiveRecord::Migration[6.0]
+
+  def change
+    add_column :items, :deconcatenated_subject, :json
+    add_column :theses, :deconcatenated_subject, :json
+
+    concatenated_subject_ids = File.readlines('db/migrate/ERA_subject_issue.txt').collect(&:strip)
```
mbarnett, line 7: I'd probably store this file somewhere else – in config/ or temp/ or just at the Rails root, rather than mix it in with the migrations

db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -0,0 +1,21 @@
...
+    report = Logger.new('log/concatenatedsubjects.log')
+    (Item.all + Thesis.all).each do |item|
+      item.deconcatenated_subject = if concatenated_subject_ids.include? item.id
+                                      deconcatenated_subject = item.subject.map { |subject| subject.split(%r{,\s*}) }.flatten
+                                      report.info("#{item.id}, \"#{item.title}\", #{item.subject}, #{deconcatenated_subject}")
```
mbarnett, line 12: Really just a small nit, but it might be worth labelling what each field is for ease of review eg:

```id: 12345, title: "This is my title", original subject: ["Cat, dog, Mouse"], deconcatenated subjects: ["Cat", "dog", "mouse"]```

File: db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb (hunk 2 above)
Comment (murny) [inline db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb:7]: This is minor but worth mentioning. We can merge as is, but thought I point out the following:

The schema change (`add_column`) above is fine here, but I'm really not a huge fan of bringing in non schema changes (seeding/editing data, data cleanup, etc) into Rails Migrations. So wondering if this data editing stuff should be in a one time rake task instead? 

Is this a big deal? Not at all. But from past experience, I recall we did something similar to this in the past (`MigrateAttachmentGids` i think?) which ended up in a big production mess and downtime. The migration we added in our codebase loaded a ton of DraftItem/Community/Item/Thesis models and ended up taking henry hours to run (which ended up timing out on his VPN as he did it from his home network) and left production in a mess. The big problem however was it made Unix team lose confidence in DB Migrations. Which from that point forward they  they now had a fear of running them and were hesitant having them being ran automatically in Ansible. Could this happen again? Perhaps (but should be much quicker if we not on fedora though)?  As matt said, we could be loading 50k objects which could result in some similar surprises again if we never notified Henry beforehand when running `rails db:migrate`. I think having it as a Rake task might give Unix team more control over this process and communicate that this is not a normal migration. Mainly, hopefully it reduces the amount of surprises or lost of confidence in normal rail production operations. 

Also with current approach, if everything looks good in Part 1, and we decide to remove `subject` column and rename `deconcatenated_subject` to `subject` in part 2. Then this migration will now fail for any developer running migrations from a fresh DB (as this migration is referencing `item.deconcatenated_subject` which doesn't exist anymore). So as part of your cleanup, you will have to edit this migration now and remove all this code anyways. [Editing migrations that have been committed and ran in production typically are not a good idea](https://edgeguides.rubyonrails.org/active_record_migrations.html#changing-existing-migrations). If its a rake task, then when we finish part 2, we can just delete it and the `ERA_subject_issue.txt`.

Comment (mbarnett) [reply db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb:7]: In general, I agree, but I don't think the rake task idea works here:

The plan, essentially, is:

1. Create the new column
2. Copy old column to new column, modifying some specific items, otherwise directly copy the value
3. (in part 2, after manual verification of modified values), drop old column and rename old column to new.

If we do the copy in a rake task, then we need every environment and developer to do step 1 and run the rake task manually to prep them all for step 2, whereas if it happens in the migration it's automatic.

> we decide to remove subject column and rename deconcatenated_subject to subject in part 2. Then this migration will now fail for any developer running migrations from a fresh DB (as this migration is referencing item.deconcatenated_subject which doesn't exist anymore).

This shouldn't be an issue? Migration 0 (some time in the past) creates old column. Migration 1 (this one) creates new column and references old column. Migration 2 (the step 2 one down the road) drops old column and renames new->old.

This should always run correctly from scratch because 1 always runs after 0 and before 2, so old column always exists. It just needs to correctly handle the missing file if we take it out later.

Comment (murny) [reply db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb:7]: Sounds like we put some thought into this, so this works for me 👍. 

Not seeing the every environment and developer thing being that big of a concern though? UAT is disposable and same with developers machines (Both are getting data from Seeds. So we probably won't have these string concation problems and if they do...then we need to fix our Seed data). As for the environments we care about, whatever we run in Prod should run in Preprod. So if we give Henry a rake task to run, he should run it in both environments  (and if that's not happening then thats a big whole other issue). Just having it manual just gives him the idea that this is special and not a normal Rails Migration (if it ends up taking hours and hours and he runs it from his home VPN again, he can't blame rails migrations)

Yeah your correct on that second part. This won't fail in future as you described (thinking it was a model method for some reason) 👍!

But all sounds good 💯

File: db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb [hunk 4]
```diff
@@ -0,0 +1,21 @@
+class DeconcatenateSubjectInItemAndThesis < ActiveRecord::Migration[6.0]
+
+  def change
+    add_column :items, :deconcatenated_subject, :json
+    add_column :theses, :deconcatenated_subject, :json
+
+    concatenated_subject_ids = File.readlines('db/migrate/ERA_subject_issue.txt').collect(&:strip)
+    report = Logger.new('log/concatenatedsubjects.log')
```
murny, line 8: Should we make this absolute? `Rails.root.join("log/concatenatedsubjects.log")`

If you run this DB Migration from another folder that's not root will it just create a log folder in that folder?

pgwillia:
I think I addressed the feedback.  I really appreciate the discussion about whether this is the best approach.  I'm looking forward to reducing the terror about live data cleaning.  Migrations and Rake tasks are both better than that.  

I like that this approach is still behind the scenes -- it won't affect any data visible to the user but allow us to look at it before making it live.

The scope of what is being touched is widened by this approach -- every item is touched vs just the 626 that were identified.  It sounds like this size of a batch hasn't been a problem even with Fedora.  But I think we've mitigated that with Matt's suggestion about using the find_each batches of 1000 (default).

@mbarnett you mentioned removing the ERA_subjects_issue.txt at some point.  I thought that having it in the migrations directory provided a hint that it was linked to a migration.  I've moved it because I don't have a strong opinion about that.  I am curious how you'd go about removing it if changing existing migrations are frowned upon?

mbarnett:
@pgwillia my main thought is that rather than have this file of IDs hang around forever, we can modify this migration to check if it's present, or wrap the file open in some error handling and use an empty array if the file open fails, so that in a later commit we can just delete ERA_subjects_issue.txt and this migration becomes essentially a no-op that just copies unmodified data in all cases to the new column

db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -0,0 +1,33 @@
...
+  def change
+    add_column :items, :deconcatenated_subject, :json
+    add_column :theses, :deconcatenated_subject, :json
+
+    @concatenated_subject_ids = File.readlines(Rails.root.join('ERA_subject_issue.txt')).collect(&:strip)
+    @report = Logger.new(Rails.root.join('log/concatenated_subjects.log'))
+
+    Item.find_each do |item|
+      item.deconcatenated_subject = deconcatenated_subject(item)
+      item.save!
```
mbarnett, line 12: Just occured to me that here, and on the Thesis save, it's probably worth using https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/update_attribute or https://apidock.com/rails/v6.0.0/ActiveRecord/Persistence/update_columns because we need this write to succeed in all cases, even ones where the item or thesis would otherwise fail validations

db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -0,0 +1,26 @@
+class DeconcatenateSubjectInItemAndThesis < ActiveRecord::Migration[6.0]
+
+  def change
+    add_column :items, :deconcatenated_subject, :json, array: true
+    add_column :theses, :deconcatenated_subject, :json, array: true
+
+    @concatenated_subject_ids = File.readlines(Rails.root.join('ERA_subject_issue.txt')).collect(&:strip)
```
mbarnett, line 7: I think we want to wrap this in the ability to catch an exception if the file doesn't exist, and then not do the rest of the migration and carry on. That way we can delete the file `ERA_subject_issue.txt` once this is all fixed instead of having to keep it around forever to avoid breaking migrations.

pgwillia: Good thinking.  I've added the error handling and fixed up another typo.

db/migrate/20200430205827_deconcatenate_subject_in_item_and_thesis.rb
```diff
@@ -4,7 +4,11 @@ def change
     add_column :items, :deconcatenated_subject, :json, array: true
     add_column :theses, :deconcatenated_subject, :json, array: true
 
-    @concatenated_subject_ids = File.readlines(Rails.root.join('ERA_subject_issue.txt')).collect(&:strip)
+    @concatenated_subject_ids = if File.exist?(Rails.root.join('ERA_subject_issue.txt'))
+      File.readlines(Rails.root.join('ERA_subject_issue.txt')).collect(&:strip)
+    else
+      []
+    end
```
pgwillia, line 11: I was considering using a guard clause (see suggestion) but I think that in order for this migration to make sense the `deconcatenated_subject` column needs to be filled.
```suggestion
    return unless File.exist?(Rails.root.join('ERA_subject_issue.txt'))
    @concatenated_subject_ids = File.readlines(Rails.root.join('ERA_subject_issue.txt')).collect(&:strip)
```

Issue #1625, opened 2020-04-28
Refactor tests around checking differences of `ActiveStorage.Attachment.count`


murny:
Our code for checking file attachments in our test suite:

`assert_difference('ActiveStorage::Attachment.count', 1) do`

Seems to be resulting in very flakey behaviour for a few team members.

Let's try to improve this by checking the associations count instead, e.g: `draft_items.files.count` instead of `ActiveStorage::Attachment.count`

Some of the problem test files:

- Items::FilesControllerTest#test_should_be_able_to_attach_files_to_a_draft_item

- Admin::Theses::FilesControllerTest#test_should_be_able_to_attach_files_to_a_draft_thesis

murny:
Closing out issue as this has now been merged into master

PR #1629, opened 2020-05-01
#1625: Check attachment association count instead of ActiveStorage model count

murny:
#1625

Hopefully this help reduce flakey test behaviour for team members

### DIFF
--- a/test/controllers/admin/theses/files_controller_test.rb
+++ b/test/controllers/admin/theses/files_controller_test.rb
@@ -24,9 +24,9 @@ class Admin::Theses::FilesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should be able to attach files to a draft thesis' do
-    assert_difference('ActiveStorage::Attachment.count', 1) do
-      post admin_thesis_files_url(@draft_thesis), params: { file: @file_attachment }, xhr: true
-    end
+    assert_equal 1, @draft_thesis.files.count
+
+    post admin_thesis_files_url(@draft_thesis), params: { file: @file_attachment }, xhr: true
 
     assert_response :success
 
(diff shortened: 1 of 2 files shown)


test/controllers/admin/theses/files_controller_test.rb
```diff
@@ -24,9 +24,9 @@ class Admin::Theses::FilesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should be able to attach files to a draft thesis' do
-    assert_difference('ActiveStorage::Attachment.count', 1) do
-      post admin_thesis_files_url(@draft_thesis), params: { file: @file_attachment }, xhr: true
-    end
+    assert_equal 1, @draft_thesis.files.count
```
murny, line 27: `assert_difference` is actually smarter than I thought.

We could just write `assert_difference('@draft_thesis.files.count', 1) do` and this does what you expect it does.

However since we already checking the final count below:

`assert_equal 2, @draft_thesis.files.count` on Line 36

Lets just check the before count here. This accomplishes the same thing 👍

mbarnett: It might be (slightly?) better to assert a difference rather than equality?

I'm just thinking in terms of if, in the future, another test got added that attached (or removed) another file to the same draft thesis (since it's set up as available to every test in this file), then depending on the order the tests run in this could need to be looking for a 1 or a 3 instead? 

I know assuming an exact count has been a source of flakey tests a few times now.

murny: Yeah I'm okay with either or. I can refactor and remove both `assert_equal X, @draft_thesis.files.count` checks in favor of `assert_difference`

But your right, if someone came in and added/removed attachment in the setup block, then we will have to update all these hardcoded numbers.

So I will go update these 👍

PR #1630, opened 2020-05-01
add migration to fix concatenated subjects (part 2) DONT MERGE YET 

pgwillia:
## Context
In the first part of this data clean up we created a new column `deconcatenated_subject` and copied all subjects here, cleaning up the ones we've identified.

Now we want to replace the subjects column with these values.

#1449 #1627 

## What's New

We remove the subject column and rename `deconcatenated_subject` to `subject`.  Also clean up the ERA_subjects_issue.txt file and reference in the previous migration.


db/schema.rb
```diff
@@ -231,6 +230,7 @@
     t.string "aasm_state"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.json "subject"
```
murny, line 233: Noticed the other json columns and old one are `array: true`. Is this needed? (I assume this has something to do with AR returning this field as an Array instead of as an JSON blob 🤔 )

pgwillia: I noticed this too.  I'm fixing now!  Thanks for the second set of eyes on this.

Issue #989, opened 2019-01-17
Bring in ERBLint (rubocop for ERB files)


murny:
Would be nice to bring in another linter (yeah I know...I love linters...) for our ERB files: https://github.com/Shopify/erb-lint

Catches and autofixes a bunch of stuff for ERB files using your rubocop rules (line lengths, spacing, alignment, all that good stuff)! With some other nice things like checking for hardcoded strings vs using I18n, etc.

murny:
This has now been merged. Closing issue

PR #1646, opened 2020-05-12
Add erblint

murny:
#989 Brings in ERBLint for Jupiter

Majority of the fixes are white space diffs (will document when that is not the case)

I also added this to CI, but seems to be breaking (looks like its linting our Gems which we should be ignoring 🤔 )


.erb-lint.yml
```diff
@@ -0,0 +1,63 @@
...
+    # Project custom linters
+    SelfClosingTag:
+      enforced_style: 'never' # We want HTML5 style
+
+    HardCodedString:
+      enabled: true
+      exclude:
+        - '**/*.en.html.erb'
+
+    # ErbSafety:
```
murny, line 40: This basically runs https://github.com/Shopify/better-html. So it be nice to have this. Currently with default config its complaining about our usages of `raw` or `.html_safe` in a few places (like kaminari pagination views). So might provide our own config to disable this offenses so we continue to use.

.github/workflows/push.yml
```diff
@@ -60,6 +60,9 @@ jobs:
           DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
         run: bundle exec danger
 
+      - name: Run ERBLint
```
murny, line 63: Need to look into this. Currently failing with lint failures that I don't see on local. My guess is its not picking up our `.erb-lint.yml` or something

app/views/application/_analytics.html.erb
```diff
@@ -2,7 +2,9 @@
   <!-- Google Analytics -->
   <script async src="https://www.googletagmanager.com/gtag/js?id=<%= Rails.application.secrets.google_analytics_token %>"></script>
   <%# TODO: javascript_tag, nonce: true do %>
-  <%= javascript_tag do %>
```
murny, line 5: They really don't like `javascript_tag`. I really couldn't find out why either? From looking into their source It seems like they are not able to valid ERB inside them so they prefer you don't have them in your project. They autocorrect it to just replace it with what it would of rendered anyways (a <script> tag)

.erb-lint.yml
```diff
@@ -0,0 +1,63 @@
...
+    #   enabled: true
+
+    Rubocop:
+      enabled: true
+      rubocop_config:
+        inherit_from:
+          - .rubocop.yml
+        Layout/InitialIndentation:
```
murny, line 48: Using their recommendation of turning off these cops. We could maybe look into this more

.erb-lint.yml
murny: Added a code comment here and left this turned off. We will leave this as disabled for time being. I tried to add an `.better-html.yml` config, but doesn't seem to be as fleshed out as erblint or rubocop configs. Doesn't seem like you can exclude cops, only change their values. Very little documentation about this as well. Better-html brings in only ~5 cops and we will probably turn off 2 of them, so not seeing if this is worth the hassle. Will leave it off. Maybe in future better documentation will be there (this is probably the reason why this is disabled as a default as well).

Issue #932, opened 2018-11-27
Move Logic from SearchController into ItemSearch Concern 


murny:
Issue discovered and explained in https://github.com/ualbertalib/jupiter/pull/931#issue-234028680

SearchController logic should be moved into ItemSearch Concern so all item search pages (items, collections, and users) can receive and share the same logic everywhere.

Thus we could handle range limit better, by keeping them on the originating page, instead of redirecting them off to the /search page instead.

murny:
This has been merged. Closing issue

PR #1656, opened 2020-05-20
Move Logic from SearchController into ItemSearch Concern

murny:
#932 

Main problem was Search Controller got far more attention and development work which never made it back into ItemSearch concern. This ItemSearch concern is being used on all item search screens (user profile, user show, collections show, admin collections show, etc) except for search controller.

This PR does a few things:
- Move fixes and improvements from search controller to item search concern
- Makes search controller use item search concern (now all item search screens are using it)
- Refactor ItemSearch concern into a SearchQueryIndexService class instead
- Remove hardcoded url for Range Facets so it doesn't redirect to /search page

Profile page and collection page showing Range Facet now working:
Screen Shot 2020-05-21 at 11 43 09 PM

Screen Shot 2020-05-21 at 11 31 09 PM

Tested various searches and facets on the different search pages and they all seem to be working.


app/controllers/concerns/item_search.rb
```diff
@@ -3,28 +3,68 @@ module ItemSearch
```
murny, line 3: I'm kinda leaning towards ditching the concern approach and making this a service class (PORO- Plain old ruby object)

This works fine, but I find its a bit too much magic with leaking instance variables (@results/@search_models) and helper methods (results/search_models) into controller/views where a PORO would be far more explicit (you would call @results = Class.new().get_search_query_results() and @search_models = Class.new().get_search_models, etc within the controller action). 

But I'll leave this as is, unless someone else has strong feelings about it.

We basically took all the logic within the Search Controller and moved it down into this concern so all search pages get it for free.

mbarnett: I actually like this idea more than the concerns, because you're pointing out the same issue that we really saw a lot in Sufia and Blacklight – lots of magical leaking of state from the controller into the concern, and even worse, when a controller had multiple concerns there was a lot of magical leakage of states _between_ concerns that led to them needing to them needing to be included in a specific non-obvious order just to avoid mysterious errors.

The only thing I'd suggest doing would be to make the `search_query_results` etc methods class methods, so that you can just do `SearchQueryService. search_query_results` (or don't repeat the service name in the method name and just call it `results`, idk) without having to call `new` if there's no instance state that needs to be created/maintained.

murny: Yeah exactly, same thoughts on issues we saw in Sufia/Blacklight 👍. Well said. Probably easier to test too if its a Service class then Concern 🤔. If you like me to refactor this, I can do this!

I was leaning towards instantiation of a class instead of class level methods as we need to pass in the arguments into it, and then `search_models` would be simply a getter method on the class to get the argument value back out. But definitely a few ways to skin this cat. So open to suggestions

```
search_query_service =  SearchQueryService.new(
  base_restriction_key: Item.solr_exporter_class.solr_name_for(:owner, role: :exact_match),
  value: @user.id,
  search_models: [Item, Thesis]
)

@results = search_query_service.results
@search_models = search_query_service.search_models
```

mbarnett: Ah yeah that makes a lot of sense, I'd say we might as well go ahead with that refactor.

app/views/application/_range_facet_result.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="card mt-3">
   <div class="card-header"><%= range_facet_result.category_name %></div>
-  <%= form_tag search_path, method: :get, class: 'form-group' do %>
+  <%= form_tag search_params_hash, method: :get, class: 'form-group' do %>
```
pgwillia, line 3: Was this a bug?  I'm not sure why this change.

murny: We hardcoded the `search_path` (/search) on this partial (which is used on any of the item search screens). So if you were looking at items on a collection show page (or users/profile) and you do a range facet it will redirects you to `/search` which is a bit confusing. This fixes this. It will keep you on the original page and then do a range facet of the items.

app/services/search_query_index_service.rb
```diff
@@ -0,0 +1,71 @@
+class SearchQueryIndexService
```
mbarnett, line 1: Niggle: Maybe we could change the name slightly? `SearchQueryIndexService` sounds like something we could use throughout the app to interact with Solr/searching, but this is really a collaborator object for, specifically, controllers that relies on the params being passed coming from the front end and makes sorting and pagination assumptions (which totally makes sense for controllers, but wouldn't if you tried to use this in a model or worker or something).

Maybe a name like `UserSearchService` or similar would make it clear that it's user/view/controller-oriented in nature?

app/controllers/profile_controller.rb
```diff
@@ -1,13 +1,19 @@
...
     @user = current_user
     @draft_items = @user.draft_items.unpublished
     @draft_theses = @user.draft_theses.unpublished
-    restrict_items_to(Item.solr_exporter_class.solr_name_for(:owner, role: :exact_match), @user.id)
+
+    search_query_index = SearchQueryIndexService.new(
+      base_restriction_key: Item.solr_exporter_class.solr_name_for(:owner, role: :exact_match),
+      value: @user.id,
+      params: params,
+      current_user: current_user
```
mbarnett, line 13: This was a problem in the original code, too, but we're switching back and forth between using `@user` (2 lines up) and `current_user` (here), and they're the same. We should probably just stick to one or the other?

app/services/search_query_index_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  QUERY_MAX = 500
+  # How many facets are shown before it says 'Show more ...'
+  MAX_FACETS = 6
+
+  attr_reader :search_models
+
+  def initialize(base_restriction_key: nil, value: nil,
+                 search_models: [Item, Thesis], params: nil, current_user: nil)
```
mbarnett, line 10: Should we leave off a default `nil` for `current_user` and force the dev using this to pass it? I'm thinking someone could try to build a new search page, forget to pass a function, everything looks like it's working because nil/"guest" users see public stuff, but there's a bug in that logged in users are treated like guests because we forgot to pass the user into this.

app/controllers/profile_controller.rb
murny: Ohh good catch. Can change this to `@user` to remain consistent

app/services/search_query_index_service.rb
murny: Yeah your right, looks like this is mandatory from our use cases. Will make it as such 👍

app/services/search_query_index_service.rb
murny: Good call, can update to `UserSearchService` 👍 (Naming is hard 😆 )

app/services/user_search_service.rb
```diff
@@ -0,0 +1,71 @@
+class UserSearchService
```
pgwillia, line 1: 👍

app/controllers/profile_controller.rb
```diff
@@ -1,13 +1,19 @@
...
     @user = current_user
     @draft_items = @user.draft_items.unpublished
     @draft_theses = @user.draft_theses.unpublished
-    restrict_items_to(Item.solr_exporter_class.solr_name_for(:owner, role: :exact_match), @user.id)
+
+    search_query_index = UserSearchService.new(
+      base_restriction_key: Item.solr_exporter_class.solr_name_for(:owner, role: :exact_match),
+      value: @user.id,
+      params: params,
+      current_user: @user
```
pgwillia, line 13: 👍

app/services/user_search_service.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  QUERY_MAX = 500
+  # How many facets are shown before it says 'Show more ...'
+  MAX_FACETS = 6
+
+  attr_reader :search_models
+
+  def initialize(base_restriction_key: nil, value: nil,
+                 search_models: [Item, Thesis], params: nil, current_user:)
```
pgwillia, line 10: 👍

Issue #1599, opened 2020-04-17
Merge PR #1590 once JQuery v3.5.1 is released


lagoan:
From @murny:

Looks like unintended breaking change in JQuery v3.5.0 which is breaking bootstrap and few other libraries we have. Breaking change is documented here: https://github.com/jquery/jquery/issues/4665

Sounds like this will be fixed in v3.5.1 (once this PR gets merged: https://github.com/jquery/jquery/pull/4666). So once v3.5.1 gets released, we can upgrade.

_Originally posted by @murny in https://github.com/ualbertalib/jupiter/pull/1590#issuecomment-615022773_

murny:
This has now been upgraded. Closing issue

PR #1662, opened 2020-05-25
Upgrade jquery to 3.5.1

murny:
#1599 (fixes security concern)

Forced a upgrade of our yarn lockfile which fixed a few weird things with easy autocomplete library. All tests are passing now

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9,28 +9,28 @@
   dependencies:
     "@babel/highlight" "^7.8.3"
 
-"@babel/compat-data@^7.8.6", "@babel/compat-data@^7.9.0":
-  version "7.9.0"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.9.0.tgz#04815556fc90b0c174abd2c0c1bb966faa036a6c"
-  integrity sha512-zeFQrr+284Ekvd9e7KAX954LkapWiOmQtsfHirhxqfdlX6MEC32iRE+pqUGlYIBchdevaCwvzxWGSy/YBNI85g==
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -4278,11 +4278,16 @@ jquery-multidownload@^4.0.1:
...
+jquery@*:
   version "3.4.1"
   resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.4.1.tgz#714f1f8d9dde4bdfa55764ba37ef214630d80ef2"
   integrity sha512-36+AdBzCL+y6qjw5Tx7HgzeGCzC81MDDgaUP8ld2zhx58HdqXGoBd+tHdrBMiyjGQs0Hxs/MLZTu/eHNJJuWPw==
 
+jquery@^3.5.1:
+  version "3.5.1"
+  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.5.1.tgz#d7b4d08e1bfdb86ad2f1a3d039ea17304717abb5"
+  integrity sha512-XwIBPqcMn57FxfT+Go5pzySnm4KWkT1Tv7gjrpT1srtf8Weynl6R273VJ5GjkRb51IzMp5nbaPjJXMWeju2MKg==
+
```
lagoan: Not familiar with Yarn but, do both versions need to be added?

lagoan: Never mind, saw your fix in next commit.

murny: Yeah was doing something weird here. So I hard upgraded yarn via `$ yarn upgrade` (trying to upgrade just jquery wasn't doing anything)

Seemed to fix everything now 👍

Issue #1405, opened 2020-01-07
Feature Image on Item show page need to be centered align within column


murny:
Noticed this today:
<img src="https://user-images.githubusercontent.com/1930474/71913103-f1520980-3133-11ea-8117-c16c52a7f914.png" width="400px">
<img src="https://user-images.githubusercontent.com/1930474/71913119-f911ae00-3133-11ea-8baf-da984024347a.png" width="400px">

Images on the left column are left aligned, but the rest of the content is centered align. We should probably keep consistent and have the images centered align.

Can simply wrap the image in a div and put some styling on it like so:
<img src="https://user-images.githubusercontent.com/1930474/71913230-34ac7800-3134-11ea-9a94-c401675b49c4.png" width="400px">

murny:
This has been merged. Closing issue

PR #1666, opened 2020-05-28
#1405: Center align feature images within sidebar

murny:
#1405 

Screen Shot 2020-05-27 at 10 54 17 PM

Couple of questions:

- We displaying this feature image as a "thumbnail" which has the following css:
  ```css
  .j-thumbnail {
      max-width: 135px;
      max-height: 175px;
  }
  ```
  But we have more real-estate to work with here? Should we increase the size of these feature 
  images on the show page, so users can view the image better? 

  This actually appears to be a regression. We were orginally styling them as feature images 
  before: https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/javascript/styles/image.scss#L6-L9. But when we added the fallback image, we changed everything to `j-thumbnail`

- For objects with multiple files, the files section looks quite a bit different then everything else in this sidebar. It be nice to keep this consistent with community/collections and usage cards and put the files section into a card component (card header is files and body contains all file objects with a download all button). Think this would greatly improve the look of the sidebar and keep it consistent with the facets sidebar on search index.

Both of the above questions have been broken out as issues now ( #1675  & #1676 )

### DIFF
--- a/app/views/items/_files_section_sidebar.html.erb
+++ b/app/views/items/_files_section_sidebar.html.erb
@@ -1,88 +1,92 @@
-<%= render partial: 'feature_image', locals: { object: @item } %>
+<div class="mb-3">
+  <div class="d-flex justify-content-center">
+    <%= render partial: 'feature_image', locals: { object: @item } %>
+  </div>
 
-<% if policy(@item).download? %>
-  <% if @item.files.count == 1 %>
-    <% file = @item.files.first %>
-    <% if file.present? && file.fileset_uuid.present? %>
-      <div class="d-flex justify-content-center mt-3">
-        <div class="p-1">
-          <%= link_to(t('.view'),
-                      file_view_item_url(id: file.record.id,
-                                         file_set_id: file.fileset_uuid,
-                                         file_name: file.filename.to_s),
-                      class: 'btn btn-outline-primary',
-                      rel: 'noopener noreferrer',
-                      target: :_blank) %>
+  <% if policy(@item).download? %>
+    <% if @item.files.count == 1 %>
+      <% file = @item.files.first %>
+      <% if file.present? && file.fileset_uuid.present? %>
+        <div class="d-flex justify-content-center mt-3">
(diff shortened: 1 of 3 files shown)


app/views/items/_files_section_sidebar.html.erb
```diff
@@ -1,88 +1,92 @@
-<%= render partial: 'feature_image', locals: { object: @item } %>
+<div class="mb-3">
+  <div class="d-flex justify-content-center">
+    <%= render partial: 'feature_image', locals: { object: @item } %>
+  </div>
 
-<% if policy(@item).download? %>
```
murny, line 3: Rest of the diff is just whitespace. I moved the wrapping `<div class="mb-3">` into this partial to keep the show view a bit more cleaner as a result everything in this partial has been tabbed over

pgwillia:
> Couple of questions:
> 
>     * We displaying this feature image as a "thumbnail" which has the following css:
>       ```css
>       .j-thumbnail {
>           max-width: 135px;
>           max-height: 175px;
>       }
>       ```
>       
>       
>       But we have more real-estate to work with here? Should we increase the size of these feature
>       images on the show page, so users can view the image better?
>       This actually appears to be a regression. We were orginally styling them as feature images
>       before: https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/javascript/styles/image.scss#L6-L9. But when we added the fallback image, we changed everything to `j-thumbnail`

Nice catch!  I think it would look better to have these j-feature-image sized.  I think the thumbnail is a major cue to whether or not this is the thing I was looking for and I can't think of a good argument for them being smaller.

> 
>     * For objects with multiple files, the files section looks quite a bit different then everything else in this sidebar. It be nice to keep this consistent with community/collections and usage cards and put the files section into a card component (card header is files and body contains all file objects with a download all button). Think this would greatly improve the look of the sidebar and keep it consistent with the facets sidebar on search index.

I think this would be a nice enhancement, both for the consistency and organization of the page.

PR #1667, opened 2020-05-28
Revert "Bump vcr from 5.0.0 to 6.0.0"

pgwillia:
Reverts ualbertalib/jupiter#1665

@murny a while back we chatted about if we are able to use VCR beyond 5.0 because it changed the the MIT-HIP licence. #1469

I should have told dependabot to ignore major versions too until we got legal feedback about this, but I only specified minor versions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,7 +110,7 @@ group :test do
 
   gem 'json-schema', '~> 2.8.1'
   gem 'launchy'
-  gem 'vcr', '6.0.0', require: false
+  gem 'vcr', '5.0', require: false
   gem 'webdrivers', '~> 4.3'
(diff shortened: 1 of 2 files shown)


mbarnett:
Sorry, I missed the initial dependabot PR.

Yeah, the consensus at the time was that the new license probably required ERA to add "ERA may not be used for X purposes" language to the site, which would necessitate a larger legal discussion about whether or not the Libraries could impose those kinds of usage restrictions on the rest of the University, and questions about whether or not some research could be construed as violating that clause.

The bandwidth anyone has for this legal discussion just so that we can upgrade our test framework is, probably unsurprisingly, essentially nil. We probably need to find another gem.

Issue #1586, opened 2020-04-08
[ERA] Refactor thesis ingest


ConnorSheremeta:
Thesis ingest is currently a in a [PR](https://github.com/ualbertalib/jupiter/pull/925). This will need to be merged once it is all refactored for post fedora. Will need an example thesis ingest csv for testing.

PR #1670, opened 2020-05-29
Theses ingest rewrite

ConnorSheremeta:
#925
#1586


lib/tasks/batch_ingest.rake
```diff
@@ -199,74 +176,89 @@ rescue StandardError => e
...
+      unlocked_obj.embargo_end_date = Date.strptime(thesis_data[:date_of_embargo], '%m/%d/%Y')
+    end
 
     # Handle rights
     unlocked_obj.rights = thesis_data[:license]
 
     # Additional fields
-    unlocked_obj.date_accepted = thesis_data[:approved_date].to_date if thesis_date[:approved_date].present?
-    unlocked_obj.date_submitted = thesis_data[:submitted_date].to_date if thesis_date[:submitted_date].present?
+    # Assumes the data received for approved_date and submitted_date follow the pattern of "D/M/Y".
```
ConnorSheremeta, line 224: This is the assumption I made

ConnorSheremeta: No I do not. The format is currently not captured in the csv but I think it would be a good idea to have it captured somewhere. I'm not completely sure what the process is but I believe we're given the csv from another department so as long as we communicate this to them. @mbarnett do you know who to communicate this date format to in order to ensure it is always followed?

Issue #1343, opened 2019-10-29
Centralize Abstraction for Thumbnail Generation


mbarnett:
Right now we call ActiveStorage's `variant`  and `preview` methods in a variety of places , with inconsistent error handling. This is leading to issues where an invalid or corrupted JPEG or other unknown format issue can lead to a page or search 500ing when it attempts to display a thumbnail for the file.

Arguably, calling into ActiveStorage for variants through the items and theses and communitiers and drafts that own these files is a law of demeter violation, and it's that anti-pattern that's at the root of our inability to correctly handle these errors. We should set up a single "clearing house" method for thumbnails across models that can abstract away the variant generation and handle errors consistently and easily.

If you're going to work on this please grab @mbarnett so that we can whiteboard out approaches and come up with a good, scalable solution that will serve us well down the road.

pgwillia:
Some related tickets https://github.com/ualbertalib/jupiter/issues/1129 and https://github.com/ualbertalib/jupiter/issues/898

pgwillia:
https://github.com/ualbertalib/jupiter/blob/6c44b3208ae36ba2d4bed819e01ab6d9256c5609/app/models/community.rb#L51-L56

https://github.com/ualbertalib/jupiter/blob/2877fcb802d291918421d44e7903832b0d6dec10/app/models/jupiter_core/depositable.rb#L99-L118

https://github.com/ualbertalib/jupiter/blob/be07bdabf2cc2e801d9c1ac3e5af75ebdc4ad099/app/models/concerns/draft_properties.rb#L53-L64

https://github.com/ualbertalib/jupiter/blob/f9c878aaf9acd10d00e00137bf087405117bb43a/app/views/items/draft/_thumbnail.html.erb#L2-L17
https://github.com/ualbertalib/jupiter/blob/f9c878aaf9acd10d00e00137bf087405117bb43a/app/views/admin/theses/draft/_thumbnail.html.erb#L2-L17

pgwillia:
Here's the rough approach:
- Community, Depositable and DraftProperties all get an instance method #logo 
- ImagesHelper gets a new method #logo_path which does the error_handling and rollbar reporting
- views call `logo_path(item.logo)`

pgwillia:
Merged into postmigration branch.

PR #1681, opened 2020-06-02
centralize abstraction for thumbnail generation

pgwillia:
## Context
Similar logic appeared in `Depositable`, `DraftProperties`, `Community` and the thumbnail/feature_image partials. By refactoring the `thumbnail_path` into a helper and being able to call `thumbnail_file` on each instance we can centralize the logic and have each class only deal with the things that concern them.

Related to #1343 

## What's New
Refactor so no new functionality.  Added #thumbnail_path to page_layout_helper and updated anything that called the old way.  Also added tests for the logic in the helper.

In investigating this I learned a few new things.  ImageMagick is required for variant on images, but a pdf utility like Poppler is required to preview PDFs and ffmpeg is required to preview video. All three are requirements for our thumbnails to work correctly.

I tried a few different things in order to test the helper.  I tried creating a dummy `@logo` which implemented only the surface area that the `thumbnail_path` touched.  With this approach I got stuck on the `rails_representation_path` with the pdf and images case.  Then I tried creating fixtures (turns out that active_storage fixtures are supported out of the box) for these objects.  I ran into a similar road block because the fixtures wouldn't respond to `variant` and `preview`. Finally I tried a hybrid approach using real objects for the pdf and image and fixtures for the error cases. Success!

I would have liked to also test that a useful warning was logged in the error cases but couldn't find a approach that worked for me short of using the minitest-logger gem.  Maybe that is the best way forward.

### DIFF
--- a/test/helpers/page_layout_helper_test.rb
+++ b/test/helpers/page_layout_helper_test.rb
@@ -97,32 +97,23 @@ class PageLayoutHelperTest < ActionView::TestCase
 
   # thumbnail_path
 
+  def mupdf_exists?
+    ActiveStorage::Previewer::MuPDFPreviewer.mutool_exists?
+  end
+
+  def poppler_exists?
+    ActiveStorage::Previewer::PopplerPDFPreviewer.pdftoppm_exists?
+  end
+
   test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
-    community = communities(:books)
-    collection = collections(:fantasy_books)
-    @item = Item.new.tap do |uo|
-      uo.title = 'Fantastic item'
-      uo.owner_id = users(:admin).id
-      uo.creators = ['Joe Blow', 'Smokey Chantilly-Tiffany', 'Céline Marie Claudette Dion']
-      uo.visibility = JupiterCore::VISIBILITY_PUBLIC
-      uo.created = '1999-09-09'
-      uo.languages = [CONTROLLED_VOCABULARIES[:language].english]
-      uo.license = CONTROLLED_VOCABULARIES[:license].attribution_4_0_international
-      uo.item_type = CONTROLLED_VOCABULARIES[:item_type].article
-      uo.publication_status = [CONTROLLED_VOCABULARIES[:publication_status].draft,
-                               CONTROLLED_VOCABULARIES[:publication_status].submitted]
-      uo.subject = ['Items']
-      uo.add_to_path(community.id, collection.id)
-      uo.save!
-    end
-    Sidekiq::Testing.inline! do
-      # Attach a file to the item so that it provide preview
-      File.open(file_fixture('pdf-sample.pdf'), 'r') do |file|
-        @item.add_and_ingest_files([file])
-      end
+    skip 'this test requires the dependency MuPDF or Poppler to be installed' unless mupdf_exists? || poppler_exists?
+
+    item = items(:fancy)
+    File.open(file_fixture('pdf-sample.pdf'), 'r') do |file|
+      item.add_and_ingest_files([file])
     end
 
-    logo = @item.files.first
+    logo = item.reload.files.first
     expected = Rails.application.routes.url_helpers.rails_representation_path(
       logo.preview(resize: '100x100', auto_orient: true).processed
     )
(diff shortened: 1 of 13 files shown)


pgwillia:
The failing test is because Poppler is not installed in the CI test environment

test/helpers/page_layout_helper_test.rb
```diff
@@ -93,9 +92,94 @@ def base_url
...
-    assert_equal page_image_url, request.base_url + @community.thumbnail_path
+    assert_equal page_image_url, request.base_url + thumbnail_path(@community.thumbnail_file)
+  end
+
+  # thumbnail_path
+
+  test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
+    community = communities(:books)
+    collection = collections(:fantasy_books)
+    @item = Item.new.tap do |uo|
```
murny, line 103: Could probably clean up some of this by using an Item fixture. E.g:

```ruby
    @item = items(:fancy)
    File.open(file_fixture('pdf-sample.pdf'), 'r') do |file|
      @item.add_and_ingest_files([file])
    end
```

test/helpers/page_layout_helper_test.rb
```diff
@@ -93,9 +92,94 @@ def base_url
...
+    expected = Rails.application.routes.url_helpers.rails_representation_path(
+      logo.preview(resize: '100x100', auto_orient: true).processed
+    )
+    assert_equal expected, thumbnail_path(logo)
+  end
+
+  test 'thumbnail_path should return preview for image (Variable)' do
+    community = communities(:books)
+    collection = collections(:fantasy_books)
+    @item = Item.new.tap do |uo|
```
murny, line 135: Same here, just use an item fixture and add an attachment to it

(dont even think you need the `Sidekiq::Testing.inline!` anymore. Can look at admin items controller and else where for examples)

```ruby
    @item = items(:fancy)
    File.open(file_fixture('text-sample.txt'), 'r') do |file|
      @item.add_and_ingest_files([file])
    end
```

test/helpers/page_layout_helper_test.rb
```diff
@@ -93,9 +92,94 @@ def base_url
...
+      uo.item_type = CONTROLLED_VOCABULARIES[:item_type].article
+      uo.publication_status = [CONTROLLED_VOCABULARIES[:publication_status].draft,
+                               CONTROLLED_VOCABULARIES[:publication_status].submitted]
+      uo.subject = ['Items']
+      uo.add_to_path(community.id, collection.id)
+      uo.save!
+    end
+    Sidekiq::Testing.inline! do
+      # Attach a file to the item so that it provide preview
+      File.open(file_fixture('pdf-sample.pdf'), 'r') do |file|
```
murny, line 120: Is installing poppler easy on CI/Dev machines? If not then maybe we should keep it as an "optional" dependency? Maybe this test only runs if Poppler dependency is defined 🤔 ?

test/helpers/page_layout_helper_test.rb
```diff
@@ -93,9 +92,94 @@ def base_url
     @community.logo.attach io: File.open(file_fixture('image-sample.jpeg')),
                            filename: 'image-sample.jpeg', content_type: 'image/jpeg'
 
-    assert_equal page_image_url, request.base_url + @community.thumbnail_path
+    assert_equal page_image_url, request.base_url + thumbnail_path(@community.thumbnail_file)
+  end
+
+  # thumbnail_path
+
+  test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
```
murny, line 100: Is installing poppler easy on CI/Dev machines? If not then maybe we should keep it as an "optional" dependency? Maybe this test only runs if Poppler dependency is defined 🤔 ?

test/helpers/page_layout_helper_test.rb
pgwillia: Yeah, I was hoping that would work too.  Maybe I'm using it wrong
```
test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
    @item = items(:fancy)
    File.open(file_fixture('pdf-sample.pdf'), 'r') do |file|
      @item.add_and_ingest_files([file])
    end

    logo = @item.files.first
    expected = Rails.application.routes.url_helpers.rails_representation_path(
      logo.preview(resize: '100x100', auto_orient: true).processed
    )
    assert_equal expected, thumbnail_path(logo)
  end
```
gives me
```
Error:
PageLayoutHelperTest#test_thumbnail_path_should_return_preview_for_pdf_(Invariable_but_Previewable):
NoMethodError: undefined method `preview' for nil:NilClass
    test/helpers/page_layout_helper_test.rb:108:in `block in <class:PageLayoutHelperTest>'
```

test/helpers/page_layout_helper_test.rb
pgwillia: It's pretty easy in Ubuntu `sudo apt-get install poppler-utils`  I'll see if I can get this to work like Clamby. https://github.com/ualbertalib/jupiter/blob/be07bdabf2cc2e801d9c1ac3e5af75ebdc4ad099/app/models/concerns/draft_properties.rb#L153

test/helpers/page_layout_helper_test.rb
murny: Hmm, my guess is you might have to `.reload` the item?

Take a peak at downloads controller test. Heres the relevant snippet:

```ruby
   item = items(:fancy)

    File.open(file_fixture('text-sample.txt'), 'r') do |file|
      item.add_and_ingest_files([file])
    end

    @file = item.reload.files.first
```

test/helpers/page_layout_helper_test.rb
```diff
@@ -93,9 +92,66 @@ def base_url
...
+  def mupdf_exists?
+    ActiveStorage::Previewer::MuPDFPreviewer.mutool_exists?
+  end
+
+  def poppler_exists?
+    ActiveStorage::Previewer::PopplerPDFPreviewer.pdftoppm_exists?
+  end
+  
+  test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
+    skip 'this test requires the dependency MuPDF or Poppler to be installed' unless mupdf_exists? || poppler_exists?
```
pgwillia, line 109: Some information about the dependencies here https://edgeguides.rubyonrails.org/active_storage_overview.html#previewing-files

test/helpers/page_layout_helper_test.rb
pgwillia: That was exactly what I was missing.  Thanks!

test/helpers/page_layout_helper_test.rb
pgwillia: https://github.com/ualbertalib/jupiter/blob/b667a368a2eac539161f23bceca04e828e39e427/.github/workflows/push.yml#L6

https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu2004-README.md

Has ImageMagick installed but not the PDF dependencies.

Issue #447, opened 2018-02-08
Improvements for Files display on Item detail view page


murny:
Files should show their name, type and size on the item view screen?
Currently for a single file it looks like:
![image](https://user-images.githubusercontent.com/1930474/35989226-d95f3ca4-0cbd-11e8-8779-5578bec7cfed.png)

and multiple file looks like:
![image](https://user-images.githubusercontent.com/1930474/35989115-981c89d6-0cbd-11e8-96bc-256e74d741ba.png)

Should be kinda like Deposit Review screen where at a glance you know exactly what the file looks like without having to download it:
![image](https://user-images.githubusercontent.com/1930474/35989079-7d602fda-0cbd-11e8-9adf-ac71e300ed55.png)

Would hate for someone to accidentally download a 1gb file on their iphone over their data network because you have no idea how big a file is

weiweishi:
@sfbetz and @leahvanderjagt Please review this ticket, and comment what are your thoughts on this. Thank you.

sfbetz:
Makes sense to me. Let's include the info as displayed in the "review and update item" screen just above the view download buttons. eg filename.pdf 613KB

pbinkley:
Note that currently very long filenames overrun the left column and push into the main column, e.g. https://era-beta.library.ualberta.ca/items/6ede136f-683e-4e68-a403-c55d39fb0412

![image](https://user-images.githubusercontent.com/272816/37543931-db09a7ce-2928-11e8-808d-a37a1a99de6e.png)

cwant:
Interesting, I don't get that behaviour for that item (in fact, I had added some CSS to prevent that from happening while addressing #401). Maybe a browser specific thing? Here is what I get:

![files-long-name](https://user-images.githubusercontent.com/672104/37544168-b3daa33c-2929-11e8-964f-38fd329613bb.png)

pbinkley:
I get it on Firefox, but I just checked Chrome and I don't get it there (both on Ubuntu). Sorry for not crosschecking before I posted. Hopefully it's a small css tweak to bring Firefox into line.

cwant:
Changing this from `break-word` to `break-all` fixes it on firefox for me:

https://github.com/ualbertalib/jupiter/blob/master/app/assets/stylesheets/items.scss#L11

I'll leave it to whomever takes this ticket to follow up.

weiweishi:
Break this into two issues:
[...]

PR #1686, opened 2020-06-04
#1676: Style improvements for sidebar section such as showing files within card component

### LINKED ISSUE #1676: Style "Files" section as a card to keep consistent with rest of sidebar on item show page
From #1666 

For objects with multiple files, the files section looks quite a bit different then everything else in this sidebar. It be nice to keep this consistent with community/collections and usage cards and put the files section into a card component (https://getbootstrap.com/docs/4.5/components/card/). Card header would be "Files" and body contains all file objects with a download all button. Think this would greatly improve the look of the sidebar and keep it consistent with the facets sidebar on search index.

Example here:
image

Notice how Files section looks quite different and out of place to Community/Collection and Usage cards. We should keep everything consistent where possible.

A quick mockup might looked like the following:
image

murny:
Fixes #1676 

What this looks like:
Screen Shot 2020-06-04 at 12 32 53 AM

Fixed a few minor spacing issues as well to keep everything consistent

### DIFF
--- a/app/views/items/show.html.erb
+++ b/app/views/items/show.html.erb
@@ -58,7 +58,7 @@
         </div>
       </div>
 
-      <div class="card mt-2">
+      <div class="card mt-3">
         <div class="card-header">
           <%= t('.usage') %>
         </div>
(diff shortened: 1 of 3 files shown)


app/views/items/show.html.erb
```diff
@@ -58,7 +58,7 @@
         </div>
       </div>
 
-      <div class="card mt-2">
+      <div class="card mt-3">
```
murny, line 61: Space between files and community sections is 1 rem, but here between community and usage sections it was 0.75 rem. So this looked a bit funny especially now when files section is now also a card component like community/usage. Lets make this 1 rem as well (aka use mt-3 instead of mt-2)

lagoan:
This is more of a design issue but could we show thumbnails for the files instead of the file names  in the sidebar?

murny:
Yeah thats a great suggestion and probably very doable now that we on ActiveStorage.

Another suggestion we've had in the past was to show "file sizes" on this screen (as currently there is no way to know if that file your downloading is 2 GB or 200 KB without first downloading it). Issue is here https://github.com/ualbertalib/jupiter/issues/447)

We can probably add a note for adding thumbnails to the above issue

Issue #661, opened 2018-04-10
[Jupiter] Bad proportions for thumbnail css in search results


pbinkley:
Landscape-mode thumbnail are displayed with their true proportions, but portrait-mode ones are squashed to a square. Presumably we want them to appear in their true proportions.

Thumbnail:
![p1020101-20--20copy jpg](https://user-images.githubusercontent.com/272816/38587828-3ece8390-3ce0-11e8-8198-f1a7a1448b49.jpg)

Screenshot of search results:
![thumbnailsinsearchresults](https://user-images.githubusercontent.com/272816/38587933-8a343744-3ce0-11e8-8505-b579bb9c225f.png)

pbinkley:
This problem exists in the main item page as well as in search results.

weiweishi:
Confirmed with @leahvanderjagt and @abombak that the views (both search and item view) should:
- recognize the orientation of the thumbnail image and present accordingly in landscape/portrait mode
- critical to have thumbnails for reports/documents are presented properly with their letter-size proportion (350x270 or 4x3)
![image](https://user-images.githubusercontent.com/2047614/47049802-9bbec500-d15b-11e8-8866-548c8912037b.png)
- landscape thumbnails should be presented as is. 
- for cases that the images are out of proportion, can crop/trim the image to fit into the maximum length/width allowed.

pbinkley:
So "out of proportion" is defined as having the short dimension < 270 when the long dimension is converted to 350? And cropping should be centered?

weiweishi:
it would be more like
- long dimension would be a maximum of 350 (or on the view pages, a fragment of 350, such as 175px)
- if the short dimension exceeds 270, (or relative to the long dimension, such as 135 px), trim/crop the short dimension to fit 135 px.

weiweishi:
Update this file: https://github.com/ualbertalib/jupiter/blob/master/app/assets/stylesheets/image.scss#L2-L3
with 175x135 or 4x3 dimensions.

pgwillia:
I found an alternate solution.  https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
The results look like this:
![screenshot from 2018-11-28 15-09-32](https://user-images.githubusercontent.com/1220762/49185639-a1ddbf00-f31f-11e8-8699-0e7850683046.png)
[...]

PR #1690, opened 2020-06-05
#1675: Fix feature images being styled as thumbnails

### LINKED ISSUE #1675: Feature Images on Item show page are being styled as Thumbnails
From #1666 

We displaying this feature image as a "thumbnail" on Item show page which has the following css:

```css
.j-thumbnail {
    max-width: 135px;
    max-height: 175px;
}
```

But we have more real-estate to work with here? Should we increase the size of these feature
images on the show page, so users can view the image better?

This actually appears to be a regression. We were originally styling them as feature images
before using the following css: https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/javascript/styles/image.scss#L6-L9. But when we added the fallback image about 9 months ago, we changed everything (including the feature images) to be styled as thumbnails (`.j-thumbnail`)

murny:
Fixes #1675

Feature Image now takes up far more real estate and easier to see. What this looks like:

image

Question: We resize our images (via activestorage/imagemagick) and set html width/height to be 350x350 for feature images and 100x100 for thumbnails. However for CSS we set a max-width/max-height for these differently as 270x350 for feature images and 135x175 for thumbnails. So not sure why these are different? We set these max widths as a solution for this issue : https://github.com/ualbertalib/jupiter/issues/661. But curious since we are forcing a resize via imagemagick if this is doing what we think it is (if the images are always going to be 100x100 or 350x350 since we resize them, then our max-width/max-height CSS will force them into different dimensions which could hurt the image quality)? 

Also in Rails 6 there appears to be better/improved resize methods we should maybe look into (#resize_to_fit, #resize_to_fill)

### DIFF
--- a/app/helpers/images_helper.rb
+++ b/app/helpers/images_helper.rb
@@ -26,7 +26,7 @@ module ImagesHelper
   end
 
   def safe_thumbnail_tag(thumbnail, image_tag_options)
-    image_tag_options[:class] = 'j-thumbnail img-thumbnail'
+    image_tag_options[:class] ||= 'j-thumbnail img-thumbnail'
     image_tag_options[:onerror] = "this.onerror=null;this.src='#{
       asset_pack_url('media/images/era-logo-without-text.png')
     }';"
--- a/app/views/application/_feature_image.html.erb
+++ b/app/views/application/_feature_image.html.erb
@@ -1,6 +1,9 @@
 <% if policy(object).thumbnail? %>
   <% if thumbnail_path(object&.thumbnail_file).present? %>
-    <%= safe_thumbnail_tag(thumbnail_path(object&.thumbnail_file, resize: '350x350', auto_orient: true), alt: '', size: '350x350') %>
+    <%= safe_thumbnail_tag(thumbnail_path(object&.thumbnail_file, resize: '350x350', auto_orient: true),
+                           alt: '',
+                           size: '350x350',
+                           class: 'j-feature-image img-thumbnail') %>
   <% else %>
     <div class="d-flex justify-content-center align-items-center img-thumbnail p-5">
       <%= icon('far', file_icon(object.thumbnail_file&.content_type), class: 'fa-5x text-muted') %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 - Can now go through wizard with an old license [#1539](https://github.com/ualbertalib/jupiter/pull/1539)
 - Fixed rake tasks [#1585](https://github.com/ualbertalib/jupiter/issues/1585)
 - Style "Files" section as a card to keep consistent with rest of sidebar on item show page [#1676](https://github.com/ualbertalib/jupiter/issues/1676)
+- Feature Images on Item show page are being styled as Thumbnails [#1675](https://github.com/ualbertalib/jupiter/issues/1675)
 
 ### Security
 - add `noopener noreferrer` when opening a link in a new tab [PR#1344](https://github.com/ualbertalib/jupiter/pull/1344)


app/helpers/images_helper.rb
```diff
@@ -26,7 +26,7 @@ def file_icon(content_type)
   end
 
   def safe_thumbnail_tag(thumbnail, image_tag_options)
-    image_tag_options[:class] = 'j-thumbnail img-thumbnail'
+    image_tag_options[:class] ||= 'j-thumbnail img-thumbnail'
     image_tag_options[:onerror] = "this.onerror=null;this.src='#{
       asset_pack_url('media/images/era-logo-without-text.png')
     }';"
```
pgwillia, line 32: Nice!  Would there be any advantage to memoizing the `:onerror ` option as well?

app/views/application/_feature_image.html.erb
```diff
@@ -1,6 +1,9 @@
 <% if policy(object).thumbnail? %>
   <% if thumbnail_path(object&.thumbnail_file).present? %>
-    <%= safe_thumbnail_tag(thumbnail_path(object&.thumbnail_file, resize: '350x350', auto_orient: true), alt: '', size: '350x350') %>
+    <%= safe_thumbnail_tag(thumbnail_path(object&.thumbnail_file, resize: '350x350', auto_orient: true),
+                           alt: '',
+                           size: '350x350',
+                           class: 'j-feature-image img-thumbnail') %>
```
pgwillia, line 6: 👍

app/helpers/images_helper.rb
murny: I can't really image a usecase where we would want to override this? I guess if we do have a usecase, we can do it then is my thinking 🤔

pgwillia:
I think you're right about that what we're doing could hurt the image quality.  [Reading the docs](https://github.com/janko/image_processing/blob/master/doc/minimagick.md#methods) it looks like what we really want is #resize_to_fit or #resize_to_limit rather than just resize.

murny:
Makes me wonder if `.resize` is doing the correct thing? Seems like its a valid method if your using 
`minimagick` gem but not for `image_processing` (which jupiter is currently using)?

So definitely think we need to switch to one of the methods from `image_processing`. Just the question is which one? 

They all seem pretty similar with minor differences:

- `#resize_to_limit` resizes only when its larger
- `#resize_to_fit` will downsize if image is larger and upsize if smaller
- `#resize_to_fill` will crop the image in the larger dimension (default is crop center)

So probably we want `#resize_to_limit` 🤔

app/helpers/images_helper.rb
pgwillia: I just realized that I didn't understand the significance of this before.  It enables the other change you made so that sometimes it can be `j-feature-image`.  

Then I was thinking that maybe it would be useful for addressing the different cases where we do this https://github.com/ualbertalib/jupiter/blob/d804d8c7c2856feffa84c5758c87421325708acd/app/views/application/_thumbnail.html.erb#L7-L12 but then I realized that this isn't the same thing.  We want the javascript there to clean up the image 500 errors.

pgwillia: In conclusion, there is no advantage to memoizing the `:onerror` option.

mbarnett:
I would imagine that what we would want is `resize_to_limit` (or `resize_to_fit` if we happen to have a bunch of files smaller than 270x350? Also agree that how we gen the thumbnails should probably match the CSS, or we're likely running into artifacts?

murny:
Makes sense 👍. 
I'll submit a different PR for converting our `resize` to `resize_to_limit`. 

I will also change our css a tad. My thinking is thumbnails should match html/resize, so lets make max-width/max-height the same, so 100px for each?

For Feature images we have to be a bit more careful due to responsiveness nature of the sidebar (the width will dynamically change on screen size). So max-width im curious what the best solution. I'm leaning towards using `max-width: 100%` then setting a `max-height: 350px`.  The other option is defining breakpoints (on Large screens its max-width: 350px but on any screen smaller its 100%?). We will have to play around with this though 🤔 

Will merge this PR as is though. Thanks for thoughful feedback 👍

PR #1699, opened 2020-06-09
Recent RDF vocab version renamed filename to resourceFilename

murny:
This was breaking migration after we updated rdf-vocab gem.

Change is made in this giant commit: https://github.com/ruby-rdf/rdf-vocab/commit/d07779dbcc4cca5c34a96d6118c4aa8ee23673a9

For whatever reason they changed the name from  `EBUCore::filename` to `EBUCore::resourceFilename`


mbarnett:
I guess one take-away is that we'll need to be careful merging rdf-vocab pull requests from dependabot, since they violated Semantic Versioning by doing this in a bump from 3.1.4 to 3.1.5.

By the Semantic Versioning assumptions that dependabot and rubygems make, anything that broke a method name (even something generated from an upstream source) should have bumped the gem up to 4.0.0 ¯\_(ツ)_/¯

Issue #1685, opened 2020-06-03
Data Migration Testing & Fixup


mbarnett:
### Tasks

- [x] Give Henry heads up that we're going to be testing this
- [x] Connor sends me his public key so I can set up the tunnel
- [x] Fix known date issue on stage1 branch (see below). 
- [x] Fix record_created_at issue on stage1 branch (see below). 
- [x] Address https://github.com/ualbertalib/jupiter/pull/1272#discussion_r364950912 (this will also need to be fixed on postmigration)
- [x] Create tunnel for Connor
    - [x] SSH keys (gen new on my machine, Connor hand off to Henry)
- [x] Tag a new release of stage1 with the fix and get Henry to deploy it to the new staging server
- [x] Get into the new staging server
- [x] Ensure stage1 is deployed on server
- [x] Run migration rake task
- [x] If it errors out, fix any new errors
- [x] Get Henry to restore Solr to its pre-migration state
- [x] Clean out Postgresql
- [x] Re-Run migration
- [x] Rinse & Repeat
- [x] Once we have this working, tag a new release of integration_postmigration and get him to deploy it to the new staging app server. Migrations will need to be run
- [x] All data should now be migrated, thoroughly test that everything is working in the new environment

### Known Date Issue

Legacy theses have a four-digit `date_accepted` field instead of a full date, which blows up when we try to migrate those theses into the postgres column, which is type date. The decision from metadata was that this is an optional field, so for this legacy data we can simply blank this attribute out during the migration, if we detect that the thesis has a four character `date_accepted` (or more broadly, any invalid `date_accepted`). Probably the best way to fix this would be in `from_thesis`, here: https://github.com/ualbertalib/jupiter/blob/integration_migration_phase1/lib/tasks/jupiter.rake#L111.

### Known record_created_at issue

We track when a record was first created in Jupiter on a recordCreatedInJupiter predicate that corresponds to the attribute `record_created_at`.  We set that here: https://github.com/ualbertalib/jupiter/blob/4a220575d1dbb45af4e3f3a2b03dde5fa9ea6f6c/app/models/jupiter_core/depositable.rb#L140. Normally this is fine, but when we're migrating this data we don't want to override the data coming out of ActiveFedora with a new "recordCreateInJupiter" time for their ActiveRecord version; we need to preserve the original date.

**Note that any fixes to the above will need to merge into the stage 1 branch and then separately into postmigration if we want to keep them around post-deployment.**

[...]

PR #1700, opened 2020-06-09
Now ignoring strings of length 4 for dates in coerce_value

ConnorSheremeta:
#1685
Decided to just simply ignore strings of length 4 when coercing dates. Should resolve our invalid date_accepted entries. Initially I had a plpgsql query that checked if the date was invalid but it was decided that that is overkill as this should be enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


murny:
Did you want to merge this into `integration_migration_phase1` instead of `integration_postmigration`? 

Reason why TravisCI is showing up and is failing

ConnorSheremeta:
> Did you want to merge this into `integration_migration_phase1` instead of `integration_postmigration`?
> 
> Reason why TravisCI is showing up and is failing

Yes, it needs to be on stage 1 for when the rake task is run.

Issue #1669, opened 2020-05-28
Investigate "This file is processing and will be available shortly "


pgwillia:
**Describe the bug**
As an admin or a non-admin user, a successful deposit results in "This file is processing and will be available shortly" 
![image](https://user-images.githubusercontent.com/1220762/83202219-5e2a6800-a104-11ea-848d-6eb6d5a4721a.png)

It appears that this doesn't resolve after the sidekiq queue has been cleared or any amount of waiting.

**To Reproduce**
1. Go to 'Deposit'
2. Fill out the wizard
3. Add some files
4. Click on "Save"
5. See "This file is processing and will be available shortly"  for each of the files you've added

**Expected behavior**
The view and download links should be available when the files are completed processing.

**Additional context**
If `fileset_uuid` is not present then the processing message will be shown https://github.com/ualbertalib/jupiter/blob/d8d0f1fbf1196a15bc5fa61cb141ecc0de375b4a/app/views/items/_files_section_sidebar.html.erb#L6
The `fileset_uuid` should be set here so I suspect that we're not reaching line 134 https://github.com/ualbertalib/jupiter/blob/d8d0f1fbf1196a15bc5fa61cb141ecc0de375b4a/app/models/jupiter_core/depositable.rb#L124-L137

Rails 6 update made a few changes to ActiveStorage which may or may not be related https://edgeguides.rubyonrails.org/6_0_release_notes.html#active-storage

murny:
Should be fixed with above PR being merged. Closing issue.

PR #1701, opened 2020-06-10
#1669: Make sure we are setting fileset_uuid whenever we add files for items/theses

murny:
#1669

This should fix this issue for "files are processing" which was a result of `fileset_uuid` not being set on items/theses. 

Seems like we only set `fileset_uuid` via `add_and_ingest_files` method but this is only being called when we seed items/theses or batch ingest items/theses.

So when we actually create or edit items in Jupiter, these files are not being given a `fileset_uuid`. My changes fix this so now that when we add files from create/edit items, `fileset_uuid` is getting set correctly.

What is `fileset_uuid` being used for? Its essentially just legacy and is for decorating the URL we use when serving these files. When you view or download a file in ERA you get a URL like the following:

/items/{{id}//download/{{fileset_uuid}}

Example: https://era.library.ualberta.ca/items/d8d783c1-b0bb-42cf-8266-497dfee4e1f1/download/dd3414be-7cdb-417e-84a5-a2d245cad896

### DIFF
--- a/app/views/items/_files_section_sidebar.html.erb
+++ b/app/views/items/_files_section_sidebar.html.erb
@@ -26,18 +26,6 @@
                         download: file.filename) %>
           </div>
         </div>
-      <% else %>
-        <h4 class="mt-3"><%= t('.header') %></h4>
-        <div class="list-group item-files">
-          <div class="list-group-item list-group-item-action d-flex flex-column">
-            <div class="item-filename text-center pb-3"><%= file.filename %></div>
-            <div class="d-flex justify-content-center">
-              <div class="text-center pb-3">
-                <%= t('.processing') %>
-              </div>
-            </div>
-          </div>
-        </div>
       <% end %>
     <% else %>
       <div class="card mt-3">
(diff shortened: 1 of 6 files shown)


app/views/items/_files_section_sidebar.html.erb
```diff
@@ -26,18 +26,6 @@
...
           </div>
         </div>
-      <% else %>
-        <h4 class="mt-3"><%= t('.header') %></h4>
-        <div class="list-group item-files">
-          <div class="list-group-item list-group-item-action d-flex flex-column">
-            <div class="item-filename text-center pb-3"><%= file.filename %></div>
-            <div class="d-flex justify-content-center">
-              <div class="text-center pb-3">
-                <%= t('.processing') %>
```
murny, line 36: If the image has no fileset_uuid then we can't view/download them (as we can't build the urls we use) so this is quite problematic.

However the current messaging isn't correct anymore as we aren't processing these anymore via background jobs. Not sure if these fallback error messages are really needed anymore or adds any value to end users?  So I'm just removing them.

PR #1704, opened 2020-06-10
Beef up AR migrations with null:false

ConnorSheremeta:
#1685 from https://github.com/ualbertalib/jupiter/pull/1272#discussion_r364950912 Beefed up AR migrations by stating that certain attributes cannot be null

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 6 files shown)


pgwillia:
Same comment as Shane left on the other one
> So we might have to be careful here editing existing migrations like this (i assume we are probably okay here, as these haven't been ran in production yet). So if we all good, then this looks good 👍 (otherwise we will have to add seperate migrations for adding these as null: false)

It looks like the tests aren't passing in the branch, but I think that's expected.

So I'm okay with this being merged 👍

PR #1705, opened 2020-06-10
Beef up AR migrations with null:false

ConnorSheremeta:
#1685 from https://github.com/ualbertalib/jupiter/pull/1272#discussion_r364950912 Beefed up AR migrations by stating that certain attributes cannot be null

Also in PR https://github.com/ualbertalib/jupiter/pull/1704


ConnorSheremeta:
I've added a constraint on the model, the only one that was missing was the collection title as far as I can tell. I have the schema changes in https://github.com/ualbertalib/jupiter/pull/1704 but I didn't add them here, I thought it would cause problems... I can definitely run through the migration process and commit the new schema db file here if there aren't any issues in doing so. My knowledge of migrations and impact on the schema file doesn't run too deep. I believe there are other changes to the schema that occur during the migrations and rake task that aren't in integration_postmigration that I could add as well, this might be better off in another PR.

ConnorSheremeta:
It has been decided to not commit the changes to schema.db here

Issue #1706, opened 2020-06-10
`content_tag` is deprecated, we should work to replace this with `tag.<name>` but the autocorrect was not helpful


pgwillia:
Rails/ContentTag: Use tag instead of content_tag. (rails/rails#25195, https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag)

`content_tag` is deprecated, we should work to replace this with `tag.<name>` but the autocorrect was not helpful

_Originally posted by @pgwillia in https://github.com/ualbertalib/jupiter/pull/1697#issuecomment-642278419_

lagoan:
Fix merged into integration_postmigration

lagoan:
Fix merged into integration_postmigration

PR #1715, opened 2020-06-16
Replace deprecated content_tag method

lagoan:
## Context

content_tag is deprecated. We are now using tag.<name>. The cop checking for content_tag use has been activated.

Related to #1706 

## What's New

Replaced content_tag with tag.

### DIFF
--- a/app/views/admin/theses/draft/_progress_bar.html.erb
+++ b/app/views/admin/theses/draft/_progress_bar.html.erb
@@ -17,12 +17,13 @@
     </ul>
   </nav>
   <div class="progress">
-    <%= content_tag :div,
-                    progress_bar_text,
-                    class: "progress-bar w-#{progress_bar_percentage}",
-                    role: 'progressbar',
-                    'aria-valuenow' => progress_bar_percentage,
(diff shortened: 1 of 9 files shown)


app/views/admin/theses/draft/_progress_bar.html.erb
```diff
@@ -17,12 +17,13 @@
...
-                    progress_bar_text,
-                    class: "progress-bar w-#{progress_bar_percentage}",
-                    role: 'progressbar',
-                    'aria-valuenow' => progress_bar_percentage,
-                    'aria-valuemin' => '0',
-                    'aria-valuemax' => '100' %>
+    <%= tag.div progress_bar_text,
+                class: "progress-bar w-#{progress_bar_percentage}",
+                role: 'progressbar',
+                aria: {
```
murny, line 23: Oh neat that you can pass these aria values as a hash 👍

lagoan: tag actually breaks if you don't do it this way (as opposed to content_tag)

PR #1726, opened 2020-06-19
Cut 1.2.18 release

ConnorSheremeta:
### Changed
- Beefed up AR migrations by stating that certain attributes cannot be null [PR#1704](https://github.com/ualbertalib/jupiter/pull/1704)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+## [1.2.18] - 2020-06-19
+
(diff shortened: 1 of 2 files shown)


pgwillia:
I'm not sure that this is the correct tag.  Previously Weiwei had tagged https://github.com/ualbertalib/jupiter/tree/2.0.0.pre2 so I think we should riff on that.

Also we should get the CHANGELOG up to date so that it represents the current history.

I'd like @mbarnett to help plan this.  He should be back next week.

mbarnett:
Yeah, Tricia is correct, we should use 2.0.0.pre.whatever for the tags off of the integration branches

PR #1738, opened 2020-06-29
Cds/bump sidekiq unique jobs

ConnorSheremeta:
Our old version was yanked!


ConnorSheremeta:
Tests are supposed to fail as this is on integration_migration_phase1

PR #1742, opened 2020-07-02
Now catching and ignoring ArgumentError in coerce_value

ConnorSheremeta:
We had many invalid dates in our production data ('1994', '[2000]', 'c1984', and even '[1977?]'). This should prevent the rake task from choking on those by catching the exception.

### DIFF
--- a/app/models/jupiter_core/locked_ldp_object.rb
+++ b/app/models/jupiter_core/locked_ldp_object.rb
@@ -504,8 +504,12 @@ class JupiterCore::LockedLdpObject
       when :json_array
         JSON.parse(value)
       when :date
-        if value.is_a?(String) && value.length != 4
-          Time.zone.parse(value)
+        if value.is_a?(String)
+          begin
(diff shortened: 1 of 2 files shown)


ConnorSheremeta:
Tests are supposed to fail as this is on integration_migration_phase1

app/models/jupiter_core/locked_ldp_object.rb
```diff
@@ -504,8 +504,12 @@ def coerce_value(value, to:)
       when :json_array
         JSON.parse(value)
       when :date
-        if value.is_a?(String) && value.length != 4
-          Time.zone.parse(value)
+        if value.is_a?(String)
+          begin
+            Time.zone.parse(value)
+          rescue ArgumentError
+            # Invalid date value, ignore.
```
mbarnett, line 511: let's put an explicit `nil` here. It does that implicitly anyways, I just think it's a bit clearer when it's explicit.

Issue #1120, opened 2019-05-03
Bring in StyleLint NPM package


pgwillia:
https://github.com/ualbertalib/discovery/blob/1b1904c69d07ef1528f48b3862fece839e2274d9/Gemfile#L50

>The Sass core team is now building Sass in Dart instead of Ruby, and will no longer be maintaining the Ruby implementation unless a maintainer steps up to help. Since the SCSS-Lint project relies on the Ruby Sass implementation, this means it will eventually not support the latest Sass features and bug fixes.

>Two alternatives worthy of consideration are stylelint and sass-lint, with sass-lint having closer feature parity to SCSS-lint.

murny:
We are now on Webpacker, I have removed SCSS LInt gem. But we should bring in something like Stylelint from NPM.

More info: https://stylelint.io/

murny:
Jupiter is now using stylelint.io so we can now close this.

PR #1753, opened 2020-07-10
#1120: Bring in styelint.io, cleanup css, and add to CI

murny:
Resolve this tech debt issue: #1120 

https://stylelint.io/


murny:
Yeah no consequences, they both do the same thing in Yarn:
https://yarnpkg.com/cli/run

`yarn run command` is just more verbose (also coming from NPM, only `npm run command` works i believe, so would help people familiar with that tool)

Issue #1680, opened 2020-06-02
Jupiter UAT – What's Broken, What Needs Fixing


mbarnett:
Specific questions I've thought of:

- Is it working at all?
- Are queues pumping?
- Are file uploads working?
- What data is in UAT (synthetic stuff from the seeds, or an old dump?)
- Is it still the Fedora version
- Should we blow the data away and start over with post-Fedora, or try upgrading it?

pgwillia:
> 
> 
> Specific questions I've thought of:
> 
>     * Is it working at all?
Yes, most of it is working.
- [x] search
- [x] download
- [x] view
- [x] deposit - sort of
- [x] admin

![image](https://user-images.githubusercontent.com/1220762/84316923-34703880-ab29-11ea-8602-73a5f0cef58a.png)

> 
>     * Are queues pumping?
No.  This deposit from three months ago is still enqued http://uat.library.ualberta.ca/items/b78fbf20-9519-4b31-a455-df7a6475d542
![image](https://user-images.githubusercontent.com/1220762/84317175-9761cf80-ab29-11ea-99d7-403013673427.png)

> 
>     * Are file uploads working?
Not exactly.  The wizard show's the upload, 
![image](https://user-images.githubusercontent.com/1220762/85075788-3024ca80-b17c-11ea-833c-d55d2c1eef00.png)
sidekiq shows the job in the queue
![image](https://user-images.githubusercontent.com/1220762/85075884-682c0d80-b17c-11ea-85cd-cc65a5f41ecb.png)
But they don't finish processing
![image](https://user-images.githubusercontent.com/1220762/85075929-8265eb80-b17c-11ea-947b-dfbdc2b6301c.png)
> 
>     * What data is in UAT (synthetic stuff from the seeds, or an old dump?)

The data is mostly seeds but there are three 'user' deposited items including the one I created
![image](https://user-images.githubusercontent.com/1220762/85076083-bfca7900-b17c-11ea-9979-d711dcdd013f.png)

> 
>     * Is it still the Fedora version
This is the Fedora version `<meta name="generator" content="Jupiter 1.2.17 - https://github.com/ualbertalib/jupiter">`
![image](https://user-images.githubusercontent.com/1220762/84316979-4c47bc80-ab29-11ea-898c-c11c4f9dcbdb.png)

> 
>     * Should we blow the data away and start over with post-Fedora, or try upgrading it?
[...]

PR #1761, opened 2020-07-16
docker image can be built and deployed

pgwillia:
## Context

This is to deploy the integration_postmigration branch as a docker image on uat.library.ualberta.ca

Related to #1680

## What's New

A bump to the "built with bundler" version in Gemfile.lock required a newer version of bundler. Fixed this in Dockerfile.deployment

Changed the predeploy script to look to this branch for getting the docker-compose with the correct docker image.

When `assets:precompile` happens Redis may not be running so ignore the Oaisys initializer that tries to create a new redis connection.

Added webpacker (to replace sprockets) so needed some tweaks to work with the new /public/packs setup.

Bumped the version to 2.0.0.pre so that when/if this is deployed we'll know which version is running.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,7 @@ gem 'rollbar'
 
 # OAI-PMH
 gem 'builder_deferred_tagging', github: 'ualbertalib/builder_deferred_tagging', tag: 'v0.01'
-gem 'oaisys', github: 'ualbertalib/oaisys', ref: 'd0e3f515472b304c6b728c01ddbda99a50af91a8'
+gem 'oaisys', github: 'ualbertalib/oaisys', ref: '68affbf4ca98bd81928b4650a85d0b5230f3d47e'
 
 # Seeds
 group :development, :test, :uat do
--- a/bin/deploy
+++ b/bin/deploy
@@ -4,7 +4,7 @@ docker rmi $(docker images -f "dangling=true" -q)
 
 docker-compose pull web
 
-docker-compose up -d
+docker-compose up -d --remove-orphans
 
 docker-compose run --rm web rake db:migrate
 docker-compose run --rm web rake assets:precompile
--- a/bin/predeploy
+++ b/bin/predeploy
@@ -10,12 +10,12 @@ set -o nounset
 [[ -n "${TRACE:-}" ]] && set -o xtrace
 
 
-readonly branch_name="master"
+readonly branch_name="integration_postmigration"
 readonly install_directory="/home/deploy/jupiter"
 
 mkdir -p "${install_directory}"
 curl -o "${install_directory}/deploy" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/deploy"
-curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.deployment.yml"
+curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.production.yml"
 
 mkdir -p "${install_directory}/config"
 curl -o "${install_directory}/config/nginx.conf" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/config/nginx.conf"
--- a/config/initializers/oaisys.rb
+++ b/config/initializers/oaisys.rb
@@ -11,5 +11,5 @@ Rails.application.config.after_initialize do
   Oaisys::Engine.config.top_level_sets_model = Community
   Oaisys::Engine.config.set_model = Collection
   Oaisys::Engine.config.redis_url = Rails.application.secrets.redis_url
-  Oaisys::Engine.config.redis = Oaisys::RedisConnection.new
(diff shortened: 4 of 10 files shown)


config/initializers/oaisys.rb
```diff
@@ -11,5 +11,5 @@
   Oaisys::Engine.config.top_level_sets_model = Community
   Oaisys::Engine.config.set_model = Collection
   Oaisys::Engine.config.redis_url = Rails.application.secrets.redis_url
-  Oaisys::Engine.config.redis = Oaisys::RedisConnection.new
+  Oaisys::Engine.config.redis = Oaisys::RedisConnection.new unless ENV['RAILS_ENV'] == 'uat'
```
pgwillia, line 14: Need a good way to ignore this line for `rake asset:precompile` in the Dockerfile but then do it for real later.

murny: Whats the real reason? Is it raising a ConnectionError here from this line? https://github.com/ualbertalib/oaisys/blob/b8dbfc1ce6f0568ecff58425aee46991d6e782e8/lib/oaisys/redis_connection.rb#L12

Probably could raise a bug in oaisys repo? We probably shouldn't be throwing an error if we cant connect to redis when initializing the class? We check for connection in every other instance method and raise a ConnectionError there so we should be good (seems like redis gem does something similar). Maybe Matt/Connor can chime in.

pgwillia: That's exactly what's happening.  Good thinking about fixing the source of the problem rather than working around it.

bin/deploy
```diff
@@ -4,7 +4,7 @@ docker rmi $(docker images -f "dangling=true" -q)
 
 docker-compose pull web
 
-docker-compose up -d
+docker-compose up -d --remove-orphans
```
pgwillia, line 7: Maybe it's better to leave this as a one of command that we do as needed.

pgwillia:
favicons_postmigration
Still need to figure out why the public assets/favicons aren't being served.

docker-compose.production.yml
```diff
@@ -55,7 +55,7 @@ services:
   web:
     <<: *rails
     volumes:
-      - assets:/app/public/assets/
+      - assets:/app/public/packs/
```
murny, line 58: Random guess, but do we need both assets and packs? Might be why favicons/logos/etc aren't serving 🤔

pgwillia: Close.  We need to share the whole `/app/public` directory so that the favicons/etc are available to nginx.

Gemfile
```diff
@@ -65,7 +65,7 @@ gem 'rollbar'
 
 # OAI-PMH
 gem 'builder_deferred_tagging', github: 'ualbertalib/builder_deferred_tagging', tag: 'v0.01'
-gem 'oaisys', github: 'ualbertalib/oaisys', ref: 'd0e3f515472b304c6b728c01ddbda99a50af91a8'
+gem 'oaisys', github: 'ualbertalib/oaisys', ref: '68affbf4ca98bd81928b4650a85d0b5230f3d47e'
```
pgwillia, line 68: This includes the fix to oaisys that doesn't throw an error on intialization.

bin/predeploy
```diff
@@ -10,12 +10,12 @@ set -o nounset
...
 
 
-readonly branch_name="master"
+readonly branch_name="integration_postmigration"
 readonly install_directory="/home/deploy/jupiter"
 
 mkdir -p "${install_directory}"
 curl -o "${install_directory}/deploy" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/deploy"
-curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.deployment.yml"
+curl -o "${install_directory}/docker-compose.yml" "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/docker-compose.production.yml"
```
pgwillia, line 18: The name of the docker compose file for deployment had been changed.

config/nginx.conf
```diff
@@ -27,7 +27,7 @@ server {
     error_log /var/log/nginx.error.log;
   }
 
-  location ^~ /assets/ {
+  location ^~ /packs/ {
```
pgwillia, line 30: This is for the change to webpacker.

config/webpacker.yml
```diff
@@ -83,6 +83,30 @@ test:
...
+  # Production depends on precompilation of packs prior to booting for performance.
+  compile: false
+
+  # Extract and emit a css file
+  extract_css: true
+
+  # Cache manifest.json for performance
+  cache_manifest: true
+
+staging:
```
pgwillia, line 98: Added staging as well.  This was a hurdle for avalon and library-cms getting to our staging environment.

docker-compose.production.yml
```diff
@@ -55,7 +55,7 @@ services:
   web:
     <<: *rails
     volumes:
-      - assets:/app/public/assets/
+      - assets:/app/public/
```
pgwillia, line 58: To include the static assets in the public directory (favicons/etc) but outside of the packs directory.

PR #1762, opened 2020-07-17
Feature/finalize thesis original fileset aip

lagoan:
## Context

Ongoing work to make sure the data sent from the AIP API is complete. This PR adds the data from the radioactive Thesis and it is now complete.

There are also changes to the output for the ```file_set``` and ```original_file``` methods but this is work in progress, we are waiting for radioactive samples but the current version should be close to the final result.

Related to #1557 

## What's New

+ Add radioactive thesis example to seeds and fixtures
The seeds for the Thesis model have been expanded to include valid examples of radioactive instances. A Thesis fixture has been expanded to use the base radioactive example.
+ Change entity output to complete the information for Thesis model
The output for the entity view has been checked against valid radioactive examples.
+ Add tests to confirm validity of Thesis AIP API output
Tests have been manually verified so that the output conforms with the definition from the metadata team.
+ Change output for the ```file_set``` view
This is work in progress and are waiting for radioactive example to do a final check.
+ Change output for the ```original_file``` view
This is work in progress and are waiting for radioactive example to do a final check.
+ Add tests for the ```file_set``` view output
+ Add tests for the ```original_file``` view output


app/controllers/aip/v1/entities_controller.rb
```diff
@@ -64,10 +65,21 @@ def show
...
+    rdf_graph_creator.copy_predicate_to_sorted_rdf_list(
+      self_subject,
+      TERMS[:ual].department,
+      TERMS[:ual].department_list
+    )
+    rdf_graph_creator.copy_predicate_to_sorted_rdf_list(
+      self_subject,
+      TERMS[:ual].supervisor,
+      TERMS[:ual].supervisor_list
+    )
```
lagoan, line 78: Added a special case for Theses, even more of a reason to separate the controllers.

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -162,7 +211,31 @@ def original_file
...
+        object: self_subject.parent / 'fixity'
+      ),
+      # We need to change the value for the predicate RDF::Vocab::PREMIS.hasMessageDigest so that it includes the
+      # uniform resource name with the algorightm used to create the checksum. In this case, we know that Active storage
+      # uses MD5
+      RDF::Statement(
+        subject: self_subject,
+        predicate: RDF::Vocab::PREMIS.hasMessageDigest,
+        object: RDF::URI.new('urn:md5') / @file.blob.checksum
+      ),
```
lagoan, line 228: As mentioned in the comment, here we do not have another stetement specifying the checksum algorithm, instead we specify it on the urn.

db/seeds.rb
```diff
@@ -553,6 +553,107 @@ def file2.original_filename
 
 end
 
+base_radioactive_thesis_values = {
```
lagoan, line 556: Repeating same method for radioactive theses as we did for radioactive items.

test/controllers/aip/v1/aip_theses_controller_test.rb
```diff
@@ -24,6 +24,10 @@ class Aip::V1::ThesesControllerTest < ActionDispatch::IntegrationTest
         file_fixture('text-sample.txt')
       ]
     )
+
+    # Don't forget to load your rdf annotations!
+    seed_active_storage_blobs_rdf_annotations
+    seed_theses_rdf_annotations
```
lagoan, line 30: Right now we are just adding the statements we need in the tests. We should be testing the correct output from multiple sources on the acts_as_rdfable gem.

test/support/aip_helper.rb
```diff
@@ -249,7 +250,7 @@ def seed_theses_rdf_annotations
                                     predicate: RDF::Vocab::DC.dateSubmitted)
     RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'degree', predicate: RDF::Vocab::BIBO.degree)
     RdfAnnotation.create_or_find_by(table: thesis_table_name, column: 'departments',
-                                    predicate: TERMS[:ual].department_list)
+                                    predicate: TERMS[:ual].department)
```
lagoan, line 253: Changes required due to the changes in rdf annotations on the migration on [db/migrate/20200709231443_change_rdf_thesis_annotations_to_be_listified.rb](https://github.com/ualbertalib/jupiter/pull/1762/files#diff-90d238d88729eda3bde0a8994593b438)

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -112,10 +124,49 @@ def file_set
...
+      RDF::Statement(
+        subject: self_subject,
+        predicate: ::TERMS[:fedora].has_model,
+        object: FILESET_INSTITUTIONAL_REPOSITORY_NAME
+      ),
+      RDF::Statement(
+        subject: self_subject,
+        predicate: ::TERMS[:ual].sitemap_link,
+        object: RDF::URI('rs:ln') + " href=\"#{file_view_uri.path}\" rel=\"content\" hash=\"md5:#{@file.checksum}\" " \
```
mbarnett, line 163: Rather than using `""` for a string and then having to escape all of the internal "s, it's generally better (more readable) to use the `%Q()` form:

`%Q( href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}" )`

app/services/rdf_graph_creation_service.rb
```diff
@@ -11,6 +11,18 @@ def insert(*nodes)
...
+  end
+
+  def delete_insert(deletes, inserts)
+    @graph.delete_insert(deletes, inserts)
+  end
+
+  def query(pattern, **options)
+    @graph.query(pattern, **options)
+  end
+
```
mbarnett, line 25: I get the _idea_ here from a Law of Demeter standpoint, but since we're just re-exporting `@graph`'s API verbatim, I think this is kind of a false improvement: we haven't actually isolated downstream clients from the internals, and anytime the details of `@graph`'s API changes, this API would need to change in lockstep with it and force the changes downstream anyways. There's no actual encapsulation here.

If we don't have a better API to wrap around RDF Graph's, we might as well just expose @graph as an `attr_accessor` and let the client code touch it directly, which will at least eliminate a layer of indirection when people are looking for the _real_ implementation of these methods.

app/controllers/aip/v1/entities_controller.rb
lagoan: That is indeed more readable! Thanks.

db/seeds.rb
```diff
@@ -553,6 +553,107 @@ def file2.original_filename
...
+  doi: 'doi:10.7939/R3V980074',
+  fedora3_handle: 'http://hdl.handle.net/10402/era.40349',
+  ingest_batch: '6395w734s',
+  rights: 'dc:rights1$ Some license terms',
+  sort_year: '2015',
+  is_version_of: ['dcterms:isVersionOf1$ Lartey, S., Cummings, G. G., & Profetto-McGrath, J. (2013). Interventions that promote retention of experienced registered nurses in health care settings: A systematic review. Journal of Nursing Management. doi: 10.1111/jonm.12105'],
+  member_of_paths: ["#{community_with_collection.id}/#{community_with_collection.collections[0].id}"],
+  subject: ['dc:subject1$ Some subject heading', 'dc:subject2$ Some subject heading', 'dc:subject3$ Some subject heading'],
+  abstract: 'dcterms:abstract1$ Arabic ناتيومرلبسفأعدقحكهجشطصزخضغذثئةظؤىءآإ Greek αβγδεζηθικλμνξοπρςστυφχψω ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩ Cyrillic абвгдеёжзийклмнопрстуфхцчшщъыьэюя АБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ Lao ກ-ໝ Thai ก-๎ Burmese က-ၙ Khmer ក-៹ Korean 가-힣 Bengali অ-ৱ // Spanish áéíóúüñ French àâçèéêëîïôùûü Portuguese àáâãçéêíóôõú Hindi ऄ-ॿ Pujabi ਅ-ੴ Mandarin 海萵苣白菜冬瓜韭菜竹筍生菜大頭菜豆薯銀甜菜莧菜豌豆蒲公英蔥豌豆苗亞羅婆羅門參西葫蘆。小豆辣根土豆 Japanese アオサメロンキャベツニラ竹シュートレタスルタバガのクズイモ銀ビートアマランスエンドウタンポポねぎ',
```
murny, line 574: Can we make sure these lines conform to 120 length?

Can split these long strings with a `\` char. Ruby compiler treats it as the same string (not concatenation) 

E.g: 
abstract: 'reallly loooong text'\
  'continue of long text'\
  'more continue'

test/controllers/aip/v1/aip_items_contoller_test.rb
```diff
@@ -221,16 +221,34 @@ class Aip::V1::ItemsControllerTest < ActionDispatch::IntegrationTest
 
   test 'should get item file set metadata graph with n3 serialization' do
     sign_in_as_system_user
+
+    radioactive_item = items(:admin)
+    radioactive_item.id = 'e2ec88e3-3266-4e95-8575-8b04fac2a679'.freeze
```
murny, line 226: Why we freezing this? Any particular reason?

lagoan: My thinking is that the id should not be modified. It is a small test but still.

Why do you think it should not be frozen?

murny: Because we doing testing here. If you want to confirm your code is doing the right thing, then you should be asserting that and your tests should pass or fail accordingly. Currently, we will raise an exception which will crash the entire test suite which is not a great way to achieve testing.

So if you are concern that your code should not be modifying the item's ID then I'd assert that as a test case. E.g:

```
assert radioactive_item.id, 'e2ec88e3-3266-4e95-8575-8b04fac2a679'
```

But is there any value in this? Why would your code be changing the ID or why is there a concern that this could happen? 

Also we are inconsistent in your tests, sometimes you freeze the ID and other times you don't

lagoan: You bring up good points. Will change the id assignments accordingly.

app/controllers/aip/v1/entities_controller.rb
murny: This is marked as resolved but code hasn't changed?

lagoan: I think your comment might be from an older version. The changes are here https://github.com/ualbertalib/jupiter/blob/a35b8238ec0664b0780630707833aa18ecc8888e/app/controllers/aip/v1/entities_controller.rb#L163-L165

murny: Ah! My bad, I see them now 👍 . Thanks!

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -112,11 +124,50 @@ def file_set
...
+      ),
+      RDF::Statement(
+        subject: self_subject,
+        predicate: ::TERMS[:fedora].has_model,
+        object: FILESET_INSTITUTIONAL_REPOSITORY_NAME
+      ),
+      RDF::Statement(
+        subject: self_subject,
+        predicate: ::TERMS[:ual].sitemap_link,
+        object: RDF::URI('rs:ln') + %Q( href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}" ) +
```
murny, line 163: Minor and we can merge without this. But it be nice to just use string interpolation for this instead of using both string concatenation and string interpolation

For example just one `%Q(...)` with everything inside it, instead of 3 of them concatenated together 
```
        object: %Q(#{RDF::URI('rs:ln')} href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}" length="#{@file.byte_size}" type="#{@file.content_type}")               
```

lagoan: I see what you mean here in how it would be nice to just use one interpolation but the results are different.

The first concatenation is the more important, we are concatenating to an URI (making the end result an URI) which when rendered as such in n3 it is  ``<rs:ln href="/items...`` if we instead use interpolation it is treated as a literal value which is rendered as ``"rs:ln href=\"/items`` where the quotes are escaped among others and it does not start with the ``<`` character; we need to handle it as an URI.

I use the second concatenation is just to stay within the 120 character limit.

PR #1766, opened 2020-07-20
Skip broken item causing fedora migration rake task to fail

ConnorSheremeta:
Skip broken item (4b3e0c51-33f7-4de9-97ad-d8bd298a8e92) which was causing the jupiter:get_me_off_of_fedora rake task to fail.


ConnorSheremeta:
Tests are going to fail as this is on integration_migration_phase1

Issue #1288, opened 2019-09-09
Consolidate Validations into Validation Plugins


mbarnett:
We were kind of abusing concerns to share logic between models that defined ad-hoc validations. With the removal of LockedLDPObject, those validations are largely duplicated between model instances (on the theory that concerns weren't really the "right" place for them, and that the concerns were starting to turn into messy ad-hoc inheritance grab-bags).

We should clean this up and establish a better pattern going forward by turning these into proper, reusable validations a la https://github.com/ualbertalib/jupiter/blob/master/app/validators/uri_validator.rb that eliminate the need to define validations like [these](https://github.com/ualbertalib/jupiter/blob/master/app/models/concerns/item_properties.rb#L110-L127) in models

See: https://guides.rubyonrails.org/active_record_validations.html#custom-validators for how to do this

mbarnett:
Specific validations I think are worth targetting:

- visibility_must_be_known (in Depositable.rb) could be a visibility_validator
- publication_status_compound_uri in Item (could be something like a compound_uri validator ala uri_validator)
- publication_status_presence & publication_status_absence in Item (could these just use the normal presence and absence validator? not sure why these are custom at first glance)
- license_xor_rights_must_be_present in Item (could be a custom license_or_rights_validator)
- communities_and_collections_must_exist in Item & Thesis (could be a community_collection_validator)
- visibility_after_embargo_must_be_valid in Item & Thesis (could be a after_embargo_visibility_validator)

lagoan:
Completed validators:
- [ ] visibility_must_be_known (in Depositable.rb) could be a visibility_validator
- [ ] publication_status_compound_uri in Item (could be something like a compound_uri validator ala uri_validator)
- [ ] publication_status_presence & publication_status_absence in Item (could these just use the normal presence and absence validator? not sure why these are custom at first glance)
- [ ] license_xor_rights_must_be_present in Item (could be a custom license_or_rights_validator)
- [ ] communities_and_collections_must_exist in Item & Thesis (could be a community_collection_validator)
- [ ] visibility_after_embargo_must_be_valid in Item & Thesis (could be a after_embargo_visibility_validator)
- [ ] community_validations in collection.

lagoan:
Addressed this issue on #1768 . Closing.

PR #1768, opened 2020-07-21
Consolidate Validations into Validation Plugins

lagoan:
## Context

Clean up validators defined in models an use reusable validators instead.

Related to #1288 

## What's New

Changed the following validators

+ visibility_must_be_known (in Depositable.rb) could be a visibility_validator
+ publication_status_compound_uri in Item (could be something like a compound_uri validator ala uri_validator)
+ publication_status_presence & publication_status_absence in Item (could these just use the normal presence and absence validator? not sure why these are custom at first glance)
+ license_xor_rights_must_be_present in Item (could be a custom license_or_rights_validator)
+ communities_and_collections_must_exist in Item & Thesis (could be a community_collection_validator)
+ visibility_after_embargo_must_be_valid in Item & Thesis (could be a after_embargo_visibility_validator)
+ community_validations in collection.


app/models/collection.rb
```diff
@@ -9,7 +9,7 @@ class Collection < JupiterCore::Depositable
 
   validates :title, presence: true
   validates :community_id, presence: true
-  validate :community_validations
+  validates :community_id, community: true
```
murny, line 12: Curious if this is even needed now that we on ActiveRecord? 

```
validates :community_id, presence: true
belongs_to :community
```

Both add validations for community presence, and the community id must exist in the community table due to reference integrity of the DB (so no need to look up the community and check if it exists)

But out of scope and needs verification/testing to ensure this 🤔

lagoan:
@mbarnett This PR touches ```app/models/jupiter_core/``` so you need to give a final OK.

app/models/item.rb
```diff
@@ -31,21 +31,24 @@ class Item < JupiterCore::Doiable
...
+  validates_with LicenseXorRightsValidator
 
   validates :embargo_end_date, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
   validates :embargo_end_date, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
   validates :visibility_after_embargo, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
   validates :visibility_after_embargo, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
   validates :member_of_paths, presence: true
-  validate :communities_and_collections_must_exist
-  validate :visibility_after_embargo_must_be_valid
+  validates :member_of_paths, community_collection: true
```
mbarnett, line 49: maybe `validates :member_of_paths, community_and_collection_existence: true`?

app/models/item.rb
```diff
@@ -31,21 +31,24 @@ class Item < JupiterCore::Doiable
...
+  validates :publication_status, presence: { message: :required_for_article },
+                                 if: ->(item) { item.item_type == CONTROLLED_VOCABULARIES[:item_type].article }
+  validates :publication_status, absence: { message: :must_be_absent_for_non_articles },
+                                 if: ->(item) { item.item_type != CONTROLLED_VOCABULARIES[:item_type].article }
+  validates :publication_status, compound_uri: true, if: lambda { |item|
     item.item_type == CONTROLLED_VOCABULARIES[:item_type].article && item.publication_status.present?
   }
-  validate :license_xor_rights_must_be_present
+
+  validates_with LicenseXorRightsValidator
```
mbarnett, line 42: maybe `LicenseXorRightsPresenceValidator`?

app/models/jupiter_core/depositable.rb
```diff
@@ -8,7 +8,8 @@ class JupiterCore::Depositable < ApplicationRecord
                                 CONTROLLED_VOCABULARIES[:visibility].draft,
                                 CONTROLLED_VOCABULARIES[:visibility].public].freeze
 
-  validate :visibility_must_be_known
+  validates :visibility, visibility: true
```
mbarnett, line 11: This is a tricky one. Maybe `validates :visibility, known_visibility: true` would read less confusingly?

app/models/collection.rb
mbarnett, line 12: Maybe `validates :community_id, community_existence: true`?

(Shane raises a good question but I agree it's probably out of scope)

app/validators/visibility_validator.rb
```diff
@@ -0,0 +1,15 @@
+class VisibilityValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attr, value)
+    return if value.blank?
+
+    value = [value] unless value.is_a?(Array)
```
mbarnett, line 6: I don't think we ever expect visibility to be an array, do we? This is always a singular value in the DB and nothing can handle it being multivalued (it would break so many assumptions that I think we should actively try to make sure everything possible doesn't work with it rather than accidentally accommodate it somehow).

app/models/item.rb
```diff
@@ -31,21 +31,24 @@ class Item < JupiterCore::Doiable
...
   validates :publication_status, uri: { in_vocabulary: :publication_status }
-  validate :publication_status_presence,
-           if: ->(item) { item.item_type == CONTROLLED_VOCABULARIES[:item_type].article }
-  validate :publication_status_absence, if: ->(item) { item.item_type != CONTROLLED_VOCABULARIES[:item_type].article }
-  validate :publication_status_compound_uri, if: lambda { |item|
+  validates :publication_status, presence: { message: :required_for_article },
+                                 if: ->(item) { item.item_type == CONTROLLED_VOCABULARIES[:item_type].article }
+  validates :publication_status, absence: { message: :must_be_absent_for_non_articles },
+                                 if: ->(item) { item.item_type != CONTROLLED_VOCABULARIES[:item_type].article }
+  validates :publication_status, compound_uri: true, if: lambda { |item|
```
mbarnett, line 38: Can we make `compound_uri` more generic? ie) could we pass in the compound URI we want to check for as an option, rather than hard-coding it in the validator (which limits its usefulness to pretty much this one case).

app/models/item.rb
```diff
@@ -31,21 +31,24 @@ class Item < JupiterCore::Doiable
...
   validates :embargo_end_date, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
   validates :embargo_end_date, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
   validates :visibility_after_embargo, presence: true, if: ->(item) { item.visibility == VISIBILITY_EMBARGO }
   validates :visibility_after_embargo, absence: true, if: ->(item) { item.visibility != VISIBILITY_EMBARGO }
   validates :member_of_paths, presence: true
-  validate :communities_and_collections_must_exist
-  validate :visibility_after_embargo_must_be_valid
+  validates :member_of_paths, community_collection: true
+  validates :visibility_after_embargo,
+            after_embargo_visibility: { visibilities_after_embargo: VISIBILITIES_AFTER_EMBARGO }
```
mbarnett, line 51: Maybe we could combine this with the general visibility validator below, to cut down on some redundancy in naming/logic?

Something like `validates :visibility_after_embargo, known_visibility: { only: VISIBILITIES_AFTER_EMBARGO }`

Then in the KnownVisibilityValidator you're just looking to see if options[:only] is set and checking to see if `options[:only].include?(value)` otherwise doing the record lookup it currently does?

app/validators/after_embargo_visibility_validator.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def validate_each(record, attr, value)
+    return if value.blank?
+
+    visibilities_after_embargo = options[:visibilities_after_embargo]
+    if visibilities_after_embargo.empty?
+      raise ArgumentError, "#{attr} must specify the list of visibilities after embargo to check against!"
+    end
+
+    value = [value] unless value.is_a?(Array)
```
mbarnett, line 11: `visibility_after_embargo` should always be a single value (same as regular visibility below) so rather than coercering it into an Array you can just assume it will always be singular.

app/validators/compound_uri_validator.rb
```diff
@@ -0,0 +1,18 @@
+class CompoundURIValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attr, value)
+    return if value.blank?
+
+    compounds = options[:compounds]
```
lagoan, line 6: Naming things is hard (```compounds```). happy to change the name to something else if it makes more sense.

test/models/item_test.rb
```diff
@@ -126,7 +126,7 @@ class ItemTest < ActiveSupport::TestCase
 
     assert_not item.valid?
     assert item.errors[:visibility_after_embargo].present?
-    assert_includes item.errors[:visibility_after_embargo], 'is not recognized'
+    assert_includes item.errors[:visibility_after_embargo], 'whatever is not a known visibility'
```
lagoan, line 129: Changed the error to the one outputted by the previous know visibility validator, which adds the visibility that was not part of the expected set. The other validator was outputting the more generic ```:is_not_recognized``` error.

app/validators/compound_uri_validator.rb
mbarnett: I think `compounds` works! Hard to think anything much better.

PR #1770, opened 2020-07-21
Cute 2.0.1pre1 release

ConnorSheremeta:
Called the release 1.3.0 when it should've been called 2.0.1.pre1.


ConnorSheremeta:
Alright, will do! I've also updated the PR to make it consistent.

Issue #1707, opened 2020-06-10
[Spike] Refactor Jupiter Routes/Controllers/Layout to Be Able To Serve Multiple Sites


mbarnett:
# Overview

Spike out some ideas for implementing #1684, to see how this approach might work and figure out any pros & cons.

I'm leaning towards serving both sites out of the same underlying webprocess here; rather than trying to make the controllers super-generic and parameterize the models they deal with via config, just having different controllers handle each site and moving shared logic into services. This seems _easier_ overall to test than the alternative – you can test controllers directly without worrying about how the subdomain interacts with routing, etc. With the config route we'd really need to run tests twice, once with each config set.

# Basic tasks as I see them (but add or change as needed)

- Wrap parts of routes in an ERA subdomain constraint:
    - items, sitemap, robots, static, redirects
    - can leave admin, oai, aip, and auth stuff exposed to all subdomains?
    - search & profile look like we can leave it out of the subdomain restriction, see below

    - bunch of ways we can access this in dev, but the dirt simplest way is just to add '127.0.0.1 era.library.dev' to your hosts file? (may need to see also https://gist.github.com/indiesquidge/b836647f851179589765)

- replace `root to:` with an era_root declaration instead. swap out existing calls to root_url or root_path for era_root_url & era_root_path
- create an intermediary controller class (EraController or something?) between ApplicationController and the other controllers related to ERA materials:
collections_controller, communities_controller, items_controller, search_controller, robots_controller, sitemap_controller, static_pages_controller, welcome_controller
    - I don't _think_ we need to namespace them, because the new models with have different names? Open to doing it, but we'd need to move views around, etc
        - could consider just adding an era_ prefix to some of the more generic controller names, like Search and Welcome etc.
    - parametrize parts of application layout? eg) rename navbar to era_navbar and add the ability for EraController to pick that navbar, so 
    that a PeelController could set different navbar contents? Same deal with the footer.
    - need to be able to switch between "themes" via EraController and a PeelController too. Basically meaning swap between
    css files that maybe set a different color for navbar and buttons, maybe? basically keep the same overall visual feel but give
    them a slightly different identity via changing the dominant color
    - search can have its `@search_models` determined by the parent controller; the way we've implemented it, the actual results rendering is determined by what models it brings back, so that should all just work™ if we tell it to search something else. 
[...]

PR #1786, opened 2020-07-29
Subdomains

murny:
This PR wraps all current routes under an "ERA" subdomain. Which sets up the foundation for allowing multiple front doors going forward.


test/controllers/aip/v1/aip_collections_controller_test.rb
```diff
@@ -1,11 +1,11 @@
 require 'test_helper'
-require Rails.root.join('test/support/aip_helper')
+require 'support/aip_helper'
 
 class Aip::V1::CollectionsControllerTest < ActionDispatch::IntegrationTest
 
   include AipHelper
 
-  def setup
+  setup do
```
murny, line 8: Slight difference here. `setup do` hooks into all setup blocks in the inheritance. This was flakey if this test ran first as our host configuration never ran yet.

test/controllers/aip/v1/aip_collections_controller_test.rb
```diff
@@ -1,11 +1,11 @@
 require 'test_helper'
-require Rails.root.join('test/support/aip_helper')
+require 'support/aip_helper'
```
murny, line 2: use a relative require here instead (no different then `require 'test_helper'`)

test/fixtures/files/n3/2107bfb6-2670-4ffc-94a1-aeb4f8c1fd81-prev-embargoed.n3
```diff
@@ -6,11 +6,11 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
-<http://localhost/aip/v1/items/2107bfb6-2670-4ffc-94a1-aeb4f8c1fd81> a pcdm:Object;
+<<%= url %>/aip/v1/items/2107bfb6-2670-4ffc-94a1-aeb4f8c1fd81> a pcdm:Object;
```
murny, line 9: url is no longer `http://localhost`, so need to override this url in for all these files

test/controllers/aip/v1/aip_collections_controller_test.rb
lagoan: I tried that so many times with no luck ¯\_(ツ)_/¯

murny: Haha yeah not sure. Seems to work on my local and CI likes it 🤷

README.md
```diff
@@ -159,7 +170,7 @@ Now everything should be up and running!
 ## Step 3: Open and view Jupiter!
 Now everything is ready, you can go and view Jupiter! Just open your favorite browser and go to the following url:
 
-[localhost:3000](http://localhost:3000)
+[era.ualberta.localhost:3000](http://era.ualberta.localhost:3000)
 
 (Note: ip address may be different if you are using `docker-machine`)
```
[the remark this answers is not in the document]
murny: That's a really good point. If someone's using docker-machine they would probably have to add their edit their own `/etc/hosts` to map to their ip address that docker-machine is providing.

But this is probably old instructions and I doubt anyone is using docker-machine? Since docker has come along way since v1.12 and  Docker Desktop is the best way to run docker on Mac/Windows now. 

We probably should just delete this note 🤔

lib/jupiter.rb
```diff
@@ -1,3 +1,4 @@
 module Jupiter
   PRODUCTION_URL = 'https://era.library.ualberta.ca'.freeze
+  TEST_URL = 'http://era.ualberta.localhost'.freeze
```
pgwillia, line 3: Is this module just for tests? Is it related to `Rails.application.routes.default_url_options`?

README.md
```diff
@@ -211,7 +222,7 @@ modifying the config files, set the environment variable `RUN_FITS_CHARACTERIZAT
 
 * Update `secrets.yml` (and maybe `omniauth.rb`) for the SAML implementation (you may need to generate a certificate/key for certain environments)
 * Give IST's Identity Provider (uat-login or login) the metadata for our service provider
-  * Quick way to view this metadata is to the start the Rails server and navigate to `http://localhost:3000/auth/saml/metadata` (feel free to edit this metadata accordingly for example adding Organization and ContactPerson metadata)
+  * Quick way to view this metadata is to the start the Rails server and navigate to `http://era.ualberta.localhost:3000/auth/saml/metadata` (feel free to edit this metadata accordingly for example adding Organization and ContactPerson metadata)
```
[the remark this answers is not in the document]
murny: It shouldn't. URL for production/staging is unchanged, so should be no impact. 

This is just instructions on retrieving the metadata on your local development environment which we need to update as everything in jupiter now is nested under "era" subdomain

config/environments/development.rb
```diff
@@ -67,4 +67,6 @@
 
   # Action on unpermitted parameters
   config.action_controller.action_on_unpermitted_parameters = :raise
+
+  config.hosts << 'era.lvh.me'
```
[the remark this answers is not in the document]
murny: This line is for whitelisting this host. As documented in the spike: 

https://github.com/ualbertalib/jupiter/issues/1707#issuecomment-653675718 under the "Whitelisting hosts"

Without this, `era.lvh.me:3000` will end up with a 500 error except.

But you raise a very interesting question about our `default_url_options` in environments and might answer some of my troubles I was having in our test suite 🤔. Will look into maybe a better way of doing things now that I've seen this, so thank you for raising that point 👍

config/routes.rb
```diff
@@ -5,153 +5,155 @@
...
-  post '/auth/system', to: 'sessions#system_login'
-
-  # Sidekiq panel
-  if Rails.env.development?
-    mount Sidekiq::Web => '/sidekiq'
-  else
-    mount Sidekiq::Web => '/sidekiq', constraints: AdminConstraint.new
-  end
+    post '/logout_as_user', to: 'sessions#logout_as_user'
+    match '/auth/:provider/callback', to: 'sessions#create', via: [:get, :post], as: :login
```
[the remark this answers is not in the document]
murny: Yeah absolutely. Not sure why we never named it before 😆. But should allow us to reuse this elsewhere instead of hardcoding the path

config/secrets.yml
```diff
@@ -32,6 +32,7 @@ shared:
   fits_path: <%= ENV['FITS_PATH'] || 'fits.sh' %>
   rack_attack_safelisted_ips: <%= ENV['RACK_ATTACK_SAFELISTED_IPS'] || '""' %>
   system_user_api_key: <%= ENV['SYSTEM_USER_API_KEY'] %>
+  tld_length: <%= ENV['TLD_LENGTH'] || 1 %>
```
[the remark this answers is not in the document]
murny: This will only need to be overridden when we have host/domains that have multiple subdomains.

So UAT may need it? I assume UAT might be `era.uat.library.ualberta.ca` in which case this value needs to be set for `3`  as we have 3 subdomains.

We will need this for production especially since our url is `era.library.ualberta.ca` and we will need to set this value to be `2`

lib/jupiter.rb
murny: This module is a generic place to store global configuration. 

Examples of other apps in the wild with similar modules:
https://github.com/gitlabhq/gitlabhq/blob/master/lib/gitlab.rb
https://github.com/discourse/discourse/blob/master/lib/discourse.rb

I just need a place to stick this TEST_URL constant without having to copy and paste it all around the codebase. This TEST_URL is used in various places in our test suite to configure the tests to run on this url (so code works as code is now under an "era" subdomain)

Is there a better place for this? Maybe 🤔

test/test_helper.rb
```diff
@@ -79,3 +79,13 @@ def logged_in?
   OmniAuth.config.test_mode = true
 
 end
+
+class ActionDispatch::IntegrationTest
+
+  setup do
+    host = URI(Jupiter::TEST_URL).host
+    host! host
```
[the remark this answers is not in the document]
murny: This sets the host that our tests run on to `era.ualberta.localhost`. So when we `get root_url` the `@request.host` is being set as `era.ualberta.localhost`. Which allows our code to work as the `constraints(subdomain: 'era') do` returns true, allowing traffic to our routes.

The next line, configures our url_helpers to set our host properly. But as you highlighted in your review, we are setting this in `environment/test.rb`, so may not need this anymore if I just set it properly in `test.rb` 🎉 . Will play around with this

pgwillia (dismissed):

mbarnett (dismissed):

PR #1788, opened 2020-07-29
Fix issue with reporting preservation errors for expected behaviour

mbarnett:
## Context

I was watching several hundred the "Could not preserve" warnings roll in on Rollbar earlier when I happened to notice Tricia comment:

> Changed this to warning because of the noise it generates without necessarily being an error. See the PMPY logs for the error source

> This error occurs when the item is already in the preservation queue. This could be a succession of multiple saves within a short period of time, ...

PMPY was designed to use a sorted set for the queue, with the idea that when you added something that was already in the queue, the higher timestamp would "push it to the back" of the queue instead of adding it a second time, ensuring that a succession of quick multiple saves only resulted in one preservation event. So why are we throwing an error in Rollbar for something that we designed the system to handle?

## What's New

The semantics of the Redis `ZADD` command are that it returns the number of items it added to the queue. In particular, if an ID is already in the queue, ZADD updates the ID's score with the new timestamp we send (meaning it moves to the back of the queue, because time is ever-increasing), and returns 0, because it didn't add anything new to the queue. We were then taking that return value and comparing it to false, and since 0 is falsey, that test was true, so we sent an error to Rollbar to complain that we had successfully done what we set out to do in the first place.

Jim looks at the camera

This fixes that.

### DIFF
--- a/app/models/jupiter_core/depositable.rb
+++ b/app/models/jupiter_core/depositable.rb
@@ -178,21 +178,8 @@ class JupiterCore::Depositable < ApplicationRecord
     self.embargo_history << embargo_history_item
   end
 
-  def push_item_id_for_preservation
-    if preserve == false
-      Rails.logger.warn("Could not preserve #{id}")
-      Rollbar.error("Could not preserve #{id}")
-    end
-
-    true
-  rescue StandardError => e
-    # we trap errors in writing to the Redis queue in order to avoid crashing the save process for the user.
-    Rollbar.error("Error occured in push_item_id_for_preservation, Could not preserve #{id}", e)
-    true
-  end
-
   # rubocop:disable Style/GlobalVars
-  def preserve
+  def push_item_id_for_preservation
     queue_name = Rails.application.secrets.preservation_queue_name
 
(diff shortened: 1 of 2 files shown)


app/models/jupiter_core/depositable.rb
```diff
@@ -178,21 +178,8 @@ def add_to_embargo_history
...
-
-    true
-  rescue StandardError => e
-    # we trap errors in writing to the Redis queue in order to avoid crashing the save process for the user.
-    Rollbar.error("Error occured in push_item_id_for_preservation, Could not preserve #{id}", e)
-    true
-  end
-
   # rubocop:disable Style/GlobalVars
-  def preserve
```
mbarnett, line 195: I also took out this second level of wrapper method?

Logic previously looked like:

```ruby
def push_item_id_for_preservation
   preserve
rescue StandardError => e
....
end

def preserve
.....
rescue StandardError => e
...
end 
```

There was nothing for `push_item_id_for_preservation` to really worry about rescuing, even though it had better & more specific error messages, because everything was already being rescued in `preserve`. So I just flattened the two methods into one.

PR #1790, opened 2020-07-30
add tmp/cache to docker ignore

pgwillia:
## Context

to avoid `Errno::EACCES: Permission denied @ rb_sysopen - /app/tmp/cache/bootsnap-load-path-cache.1.18857.tmp` error don't copy in that cache!

Related to #1680

## What's New

Ignore tmp/cache directory when building docker images.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 *.md
 
 tmp
+tmp/cache
 !tmp/pids/.keep
 log
 coverage/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+### Added
+- tmp/cache to docker ignore [#1680](https://github.com/ualbertalib/jupiter/issues/1680)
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -3,6 +3,7 @@
 *.md
 
 tmp
```
murny, line 5: Shouldn't `tmp` here cover everything in `tmp` directory? including cache? 🤔

pgwillia:
@murny we were talking about adding a user/group but it turns out that the problem was in the .dockerignore after all.  The documentation suggests that the `tmp` should be recursive but it doesn't seem to be in practice.

```
$ docker -v
Docker version 19.03.12, build 48a66213fe
$ docker-compose -v
docker-compose version 1.26.2, build eefe0d31
```

.dockerignore
pgwillia: I'll try building this branch from docker hub to see if it has the same results I'm seeing locally.  But I didn't have to remove volumes and everything else as before after using the image I built with just this change.

pgwillia: Confirmed the docker hub image has the same result.

PR #1799, opened 2020-08-05
Feature/merge file set original file

lagoan:
## Context

The information for original_file and file_set data for AIP had a lot of overlap. After discussing this with the metadata team it was decided that they should be merged in a single entry. The term file_set is used throughout the code base, so even if it not the same concept, we will be keeping the name now for legacy reasons. 

Related to  #1557

## What's New

Tests and routes for original_file have been removed and the information provided by them are now part of file_set entry point.

I modified the test fixtures for file_set so that we only pass the information for fileset id and collection id, removing all the file information passed to the test since we know exactly what will be the output for the file.


app/models/jupiter_core/depositable.rb
```diff
@@ -125,6 +125,8 @@ def add_and_ingest_files(file_handles = [])
   end
 
   def set_record_created_at
+    return if record_created_at.present?
+
     self.record_created_at = Time.current.utc.iso8601(3)
   end
```
lagoan, line 131: I changed this method so that it adds the record_created_at date only when it is not provided. This has to do with the predicate ``::TERMS[:ual].record_created_in_jupiter`` linked to the value ``record_created_at``. I added the predicate on another PR based on the name of the variable and it may not be the case that they are related.

https://github.com/ualbertalib/jupiter/blob/44df0635fe3002ddc1279519f1e5f25d6129c8d4/db/migrate/20200529230303_annotate_tables_with_date_ingested_and_created_predicates.rb#L5

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 - Beefed up AR migrations by stating that certain attributes cannot be null [PR#1704](https://github.com/ualbertalib/jupiter/pull/1704)
 - Finalize Item AIP data [#1557](https://github.com/ualbertalib/jupiter/issues/1557)
 - Finalize Thesis AIP data [#1557](https://github.com/ualbertalib/jupiter/issues/1557)
+- Merge file_set and original_file AIP API entry points [#1557](https://github.com/ualbertalib/jupiter/issues/1557)
```
murny, line 47: Minor, this should go up in the `[Unreleased]` section. `[2.0.1.pre1]` version has already been cut

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -142,10 +134,13 @@ def file_set
...
       RDF::Statement(subject: self_subject, predicate: RDF::Vocab::Fcrepo4.hasParent, object: self_subject.parent),
+      # We need to change the value for the predicate RDF::Vocab::PREMIS.hasMessageDigest so that it includes the
+      # uniform resource name with the algorightm used to create the checksum. In this case, we know that Active storage
+      # uses MD5
       RDF::Statement(
         subject: self_subject,
-        predicate: RDF::Vocab::PCDM.hasFile,
-        object: self_subject / 'original_file'
+        predicate: RDF::Vocab::PREMIS.hasMessageDigest,
+        object: RDF::URI.new('urn:md5') / @file.blob.checksum
```
murny, line 143: Probably stupid question but are we doing division here? Or this is a string or something?

`object: something / something`

lagoan: Not a stupid question at all. The ```/``` operator joins parts of the URI taking into account their context. 

The first case it is an URL so it joins them with ```/```
The second case it is an URN so it joins them with ```:```

Documentation is here https://www.rubydoc.info/github/ruby-rdf/rdf/RDF/URI#%2F-instance_method

murny: Ah so as long as the first part is using `RDF::URI.new` then it will use this "smart separator". 

Kinda weird they overrode a ruby std lib method (`#/` division) for this. Seems like they have a few alternative methods like `#+` or `.join()` which makes more sense but seems like they are not quite the same (slight differences)

Thanks for the info 👍

lagoan:
@mbarnett When you get the chance please take a look at this PR. It makes a small change to jupiter_core that should be double checked to make sure the intention of use for the ```record_created_at``` value is correct.

app/controllers/aip/v1/entities_controller.rb
mbarnett: I'm going to suggest we link to the documentation in the couple of spots we use it, just because it's a bit of a surprising behaviour to people who aren't familiar with the URI library (which is probably going to be nearly everyone, and I am absolutely going to forget this in a month 😉 )

app/controllers/aip/v1/entities_controller.rb
```diff
@@ -308,17 +267,18 @@ def entity_file_statements
       file_statements << RDF::Statement(
         subject: self_subject,
         predicate: RDF::Vocab::PCDM.hasMember,
-        object: RDF::URI.new("#{request.original_url}/filesets/#{file.fileset_uuid}")
+        # RDF::URI Overrides the #/ operator to be used as a smart separator when building URIs.
+        # More information here: https://www.rubydoc.info/github/ruby-rdf/rdf/RDF/URI#%2F-instance_method
+        object: RDF::URI.new(request.original_url) / 'filesets' / file.fileset_uuid
```
lagoan, line 272: Changed for consistency when using the ```#/``` operator

Issue #1722, opened 2020-06-18
Automate deployment to UAT upon push to master


pgwillia:
We want to be able to merge a PR, move the issue to 'On UAT' and email the ticket owner to seek their feedback.  To optimize the feedback cycle we would want UAT to be within a couple minutes of the latest changes.

I looked at [watchtower](https://containrrr.dev/watchtower/) for keeping our deployment up to date
> With watchtower you can update the running version of your containerized app simply by pushing a new image to the Docker Hub or your own image registry. Watchtower will pull down your new image, gracefully shut down your existing container and restart it with the same options that were used when it was deployed initially.

It can even be used in a docker-compose.yml
```
version: "3"
services:
  cavo:
    image: index.docker.io/<org>/<image>:<tag>
    ports:
      - "443:3443"
      - "80:3080"
  watchtower:
    image: containrrr/watchtower
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /root/.docker/config.json:/config.json
    command: --interval 30
```
One problem with this approach might be 
>By default, watchtower will watch all containers. However, sometimes only some containers should be updated.
but there is a workaround to [include or exclude using a label](https://containrrr.dev/watchtower/arguments/#filter_by_enable_label)

An alternative to this which I used on the discovery deployment is having a cron job update
```
[deploy@uatsrv01 ~]$ crontab -l
@daily cd /home/deploy/discovery && docker-compose pull && docker-compose up -d
```
This has been relatively successful `Discovery 3.2.2` is the current version and the current version deployed. However this is done daily rather than within minutes of changes.

_Originally posted by @pgwillia in https://github.com/ualbertalib/jupiter/issues/1680#issuecomment-646336872_

pgwillia:
My investigation into whether having watchtower in one docker-compose would impact all docker containers (even those not defined in the same docker-compose).  The answer is that it will try to update all running containers
```
watchtower_1  | time="2020-08-06T00:44:12Z" level=info msg="Starting Watchtower and scheduling first run: 2020-08-06 00:44:42 +0000 UTC m=+31.410428201"
watchtower_1  | time="2020-08-06T02:39:39Z" level=info msg="Unable to update container /jupiter_nginx_1. Proceeding to next."
watchtower_1  | time="2020-08-06T02:39:39Z" level=info msg="Unable to update container /jupiter_worker_1. Proceeding to next."
[...]

PR #1802, opened 2020-08-06
script to launch watchtower

pgwillia:
## Context

We want to start practicing Continuous Deployment.  Watchtower is a docker container that polls docker hub for updated images for the containers that are running locally.  This script will be responsible for running the watchtower container.

Related to #1722

## What's New
Added watchtower script and using predeploy script to retrieve it.

### DIFF
--- a/bin/watchtower
+++ b/bin/watchtower
@@ -1,0 +1,12 @@
+## this file comes from the Jupiter project  "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/watchtower"
+## it will start the watchtower container which polls docker hub for the latest image of the running docker containers
+## to review watchtower activity use `docker logs watchtower`
+
+docker stop watchtower && docker rm watchtower
+
+docker run -d \
+    --name watchtower \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    containrrr/watchtower \
+    --interval 30 \
+    --cleanup
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ### Added
 - tmp/cache to docker ignore [#1680](https://github.com/ualbertalib/jupiter/issues/1680)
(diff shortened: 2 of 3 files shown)


bin/watchtower
```diff
@@ -0,0 +1,12 @@
+## this file comes from the Jupiter project  "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/watchtower"
+## it will start the watchtower container which polls docker hub for the latest image of the running docker containers
+## to review watchtower activity use `docker logs watchtower`
+
+docker stop watchtower && docker rm watchtower
+
+docker run -d \
+    --name watchtower \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    containrrr/watchtower \
```
murny, line 10: is that a spelling mistake? `containrrr`? Or that's on purpose?

pgwillia: That's on purpose https://hub.docker.com/r/containrrr/watchtower

bin/watchtower
```diff
@@ -0,0 +1,12 @@
+## this file comes from the Jupiter project  "https://raw.githubusercontent.com/ualbertalib/jupiter/${branch_name}/bin/watchtower"
```
murny, line 1: We should maybe ensure we have a shebang comment at the top of this file as I assume this is a bash executable file?

`#!/usr/bin/env bash`

bin/watchtower
```diff
@@ -0,0 +1,12 @@
...
+## to review watchtower activity use `docker logs watchtower`
+
+docker stop watchtower && docker rm watchtower
+
+docker run -d \
+    --name watchtower \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    containrrr/watchtower \
+    --interval 30 \
+    --cleanup
```
murny, line 12: Minor, typical best practice for all bash files should have a newline at the end of the file. As there's some weird things that can happen if you do not have this with how files get parse/etc (e.g if you use cat or diff on this file)

bin/watchtower
murny: Haha sounds good 👍

pgwillia:
Discovered a better approach.  Closing and pursuing that instead.

PR #1803, opened 2020-08-06
Upgrade Rubocop/Erblint and fix cop violations

murny:
Replaces #1749 

Upgrade rubocop and erblint gems to latest versions

Fix code to make new cops happy, mainly `Style/ArrayCoercion` and `Style/StringConcatenation`


.rubocop.yml
```diff
@@ -93,6 +93,12 @@ Rails:
 Rails/FindEach:
   Enabled: false
 
+Rails/SkipsModelValidations:
+  Exclude:
+    - app/controllers/aip/v1/collections_controller.rb
```
murny, line 98: This cop wants us to avoid using `.insert` because it skips validations. But it's a false positive as we using `.insert` from our rdf graph object.

```
app/controllers/aip/v1/collections_controller.rb:24:31: C: Rails/SkipsModelValidations: 
Avoid using insert because it skips validations. 
(https://guides.rubyonrails.org/active_record_validations.html#skipping-validations)

      rdf_graph_creator.graph.insert(RDF::Statement(statement_definition))
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

app/decorators/facets/member_of_paths.rb
```diff
@@ -7,7 +7,7 @@ def display
     community_id, collection_id = @value.split('/')
     community_title = Community.find(community_id).title
     collection_title = if collection_id
-                         '/' + Collection.find(collection_id).title
+                         "/#{Collection.find(collection_id).title}"
```
murny, line 10: Rubocop prefers string interpolation over string concatenation. Updates all our violations of this.

```
app/decorators/facets/member_of_paths.rb:10:26: C: 
Style/StringConcatenation: Prefer string interpolation to string concatenation. 
(https://rubystyle.guide#string-interpolation)
                         ‘/’ + Collection.find(collection_id).title
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

app/models/exporters/solr/base_exporter.rb
```diff
@@ -180,7 +180,7 @@ def indexed_model_name(name)
     # the basic DSL for declaring Solr indexes who will take their contents from attributes
     # declared on the objects we will export
     def index(attr, role:, type: :string)
-      role = [role] unless role.is_a? Array
+      role = Array(role)
```
murny, line 183:
```
app/models/exporters/solr/base_exporter.rb:183:7: C: 
Style/ArrayCoercion: Use Array(role) instead of explicit Array check. 
(https://rubystyle.guide#array-coercion)
      role = [role] unless role.is_a? Array
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Interesting new cop. Essentially it argues you should Use Array() instead of explicit Array check or [*var], when dealing with a variable you want to treat as an Array, but you’re not certain it’s an array.

So switched any violations over to new syntax

mbarnett: Going to suggest that we disable this cop/revert these changes (I vaguely think there might have been an older cop we disabled for the same reason?). `Array()` is kind of a subtle foot-gun, and I don't think Batsov understands that...

The difference is this:

```ruby
irb(main):001:0> a = {:a => 1, :b => 2}
=> {:a=>1, :b=>2}
irb(main):002:0> array = [a] unless a.is_a? Array
=> [{:a=>1, :b=>2}]
irb(main):003:0> different_array = Array(a)
=> [[:a, 1], [:b, 2]]
```

Instead of putting a hash in an array, it blows it up into keywise paired subarrays, which generally isn't what people expect, making it not semantically equivalent to the `[x] unless x.is_a? Array` pattern.

murny: Makes sense. We can play it safe here and revert these 👍

Issue #1782, opened 2020-07-25
Collections and communities not showing up after fedora migration


ConnorSheremeta:
After following the fedora migration steps the collections and communities do not show up:

![download (1)](https://user-images.githubusercontent.com/14242562/88445277-cd5bca00-cdde-11ea-8749-8462d54819cd.jpeg)

A reindex seems to resolve the issue:
`Community.find_each(&:save!) && Collection.find_each(&:save!)`

This needs to be fixed with changes needed to be made on the phase1 branch. There needs to be arcollection and arcommunity exporters in [app/models/exporters/solr](https://github.com/ualbertalib/jupiter/tree/integration_postmigration/app/models/exporters/solr)

The ar models also need to reference that exporter like in [app/models/item](https://github.com/ualbertalib/jupiter/blob/integration_migration_phase1/app/models/item.rb) and [app/models/thesis](https://github.com/ualbertalib/jupiter/blob/integration_migration_phase1/app/models/thesis.rb):

`has_solr_exporter Exporters::Solr::xxxExporter`

PR #1809, opened 2020-08-10
added two solr exporters for ar community and ar collection

ConnorSheremeta:
#1782

Added missing solr exporters so that after migration the collections and communities show up.
After:
Screenshot from 2020-08-10 14-44-24
Screenshot from 2020-08-10 14-44-16


ConnorSheremeta:
Tests will fail as this is the phase 1 branch

Issue #1781, opened 2020-07-25
Rails migrations failing on production data


ConnorSheremeta:
Our migrations are failing on the current integration_postmigration branch. This is due to attachments not having corresponsing blobs. I have the code I used to resolve this below:

```
ActiveStorage::Attachment.all.each do |asa|
  asb = ActiveStorage::Blob.find_by(id: asa.blob_id)
  asa.delete unless asb.present?
end
```
This should be put in the get_me_off_of_fedora rake task in the phase 1 branch. 

I'm curious if the logo_id is cleared out of the item/thesis if one of these attachments missing a blob is a thumbnail. There is also an activestorageattachment.purge I believe (might be worth looking into the differences).

relates to #1685

PR #1811, opened 2020-08-11
Fixed failing migrations due to missing blobs

ConnorSheremeta:
Fixed failing migrations due to missing blobs [#1781](https://github.com/ualbertalib/jupiter/issues/1781)


ConnorSheremeta:
Tests will fail as this is on phase 1

mbarnett:
Those do two different things: The rake task deletes Blobs that are no longer pointed to by any Attachment records – Connor's adjustment cleans up stale attachment records that point at non-existent Blobs.

pgwillia:
> 
> 
> Those do two different things: The rake task deletes Blobs that are no longer pointed to by any Attachment records – Connor's adjustment cleans up stale attachment records that point at non-existent Blobs.

Oh.  I see now.

PR #1826, opened 2020-08-20
Skip failing oaisys tests

ConnorSheremeta:
#1871


test/integration/oaisys_list_identifiers_test.rb
```diff
@@ -16,6 +16,7 @@ class OaisysListIdentifiersTest < ActionDispatch::IntegrationTest
   end
 
   test 'list identifers items xml' do
+    skip('skipping until bug regarding path helper is fixed')
```
lagoan, line 19: Maybe we could add a link to the issue for the bug to track its status

Issue #1689, opened 2020-06-04
Test failing intermittently


pgwillia:
https://github.com/ualbertalib/jupiter/blob/f6fceeaff72249202d749b21103b066ff7a52643/test/system/search_test.rb#L359

Run options: --seed 64065

Failure:
SearchTest#test_admin_should_be_able_to_filter_the_public_and_private_items [/home/runner/work/jupiter/jupiter/test/system/search_test.rb:359]:
expected to find css "a" but there were no matches

https://github.com/ualbertalib/jupiter/pull/1687/checks?check_run_id=738021996

PR #1827, opened 2020-08-21
Add tie breaker for solr queries by title

lagoan:
## Context

Multiple tests were flapping when results from solr queries had the same score, since they results were not always order the same.

Related to #1689 

## What's New

Added a simple method to add a tie breaker on ``JupiterCore::SolrServices::DeferredFacetedSolrQuery`` for solr query results. This tie breaker assumes results from solr will always have a ``title`` value that can be used for sorting.

### DIFF
--- a/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
@@ -46,22 +46,25 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
       next if attr.blank?
 
       attr = attr.to_sym
-      criteria[:sort] << begin
+      sort_attribute = begin
                         if attr == :relevance
                           # We change the sort attribute name because Solr uses score when sorting by relevance.
                           :score
                         else
                           solr_exporter.solr_name_for(attr, role: :sort)
                         end
-                      # Ignore current sort attribute if we could not find its solr name. It is safe to ignore the
-                      # pairing sort order
                       rescue ArgumentError, NameManglingError
-                        next
+                        nil
                       end
+
+      # Ignore current sort attribute if we could not find its solr name. It is safe to ignore the pairing sort order
+      next if sort_attribute.nil?
+
+      criteria[:sort] << sort_attribute
 
       order = orders[idx]&.to_sym
 
-      criteria[:sort_order] << if criteria[:sort].last == :score
+      criteria[:sort_order] << if sort_attribute == :score
                                  # When sorting by score it only makes sense to use :desc order from the user
                                  # perspective so we ignore the order if one is given.
                                  :desc
@@ -69,7 +72,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
                                  order
                                else
                                  # We could not find the order so we switch to default sort direction
-                                 sort_direction_index = solr_exporter.default_sort_indexes.index(criteria[:sort].last)
+                                 sort_direction_index = solr_exporter.default_sort_indexes.index(sort_attribute)
                                  solr_exporter.default_sort_direction[sort_direction_index]
                                end
     end


lagoan:
You are right in that in that we wont be sorting by relevance there. Sorry for not being clear on what I meant; I was talking about something simpler than that, where minor changes still needed to be made to ``DeferredFacetedSolrQuery`` since the sort method was being used in other areas. 

In the current implementation the sort method makes sure we are dealing with arrays to avoid special cases.

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -32,33 +32,42 @@ def offset(num)
...
+    criteria[:sort_order] = if (order - [:asc, :desc]).empty?
+                              # When we order by score we need to sort in descending order to show the highest scores
+                              # first
+                              order[criteria[:sort].index(:score)] = :desc if criteria[:sort].include?(:score)
+                              order
                             else
-                              solr_exporter.default_sort_direction
+                              Array.new(criteria[:sort].length, solr_exporter.default_sort_direction)
                             end
+
```
mbarnett, line 70: I found this approach to be really confusing, and had to read it a few times to really understand what was going on here, and it feels like there are a few corner cases that aren't being handled (for instance, what happens if `order` consists of only `:asc` and `:desc` but is _shorter_ than `attributes`? At least some of those attributes should have a `default_sort_direction` order, but I don't think they will here?)

How about a much more straight-forward map approach instead:

```ruby
criteria[:sort_order] = criteria[:sort].map.with_index do |sort_attr, idx|
  order = orders[idx]
  if order.present? && [:asc, :desc].include?(order.to_sym)
    order
  elsif sort_attr == :score
    :desc
  else
    solr_exporter.default_sort_direction
  end
end

File: app/services/user_search_service.rb [hunk 2]
```diff
@@ -28,12 +28,22 @@ def results
...
 
-    # sort by relelvance if a search term is present and no explicit sort field has been chosen
-    sort_field = @search_params[:sort]
-    sort_field ||= :relevance if query.present?
+    # Sort by relevance if a search term is present and no explicit sort field has been chosen
+    sort_fields = [@search_params[:sort]]
+    sort_fields ||= [:relevance] if query.present?
+
+    # Handle default sort order as ascending
+    sort_orders = @search_params[:direction] ? [@search_params[:direction]] : [:asc]
```
mbarnett, line 36: Where is the idea that default sort order should be ascending coming from?

Default sort direction is currently set per-model, which is by design. Overriding that here breaks the semantics of the default sort DSL.

app/services/user_search_service.rb
```diff
@@ -28,12 +28,22 @@ def results
...
+
+    # Add sort by title for breaking ties in results
+    unless sort_fields.include?('title')
+      sort_fields << :title
+      # We want to perform the tie breaker in the same sort order as the original query
+      sort_orders << sort_orders.first
+    end
```
mbarnett, line 43: I don't think we want to tie-break by title _any_ time title isn't part of a search. I think the goal here was _specifically_ to tie-break by title when relevance happens to be insufficient.

In other cases we may wish to tie-break by something else, eg relevance if a date sort is insufficiently granular, or something. In yet other cases we may want to surface the lack of granularity so that we can improve the data quality, rather than hiding it behind the unrequested title-sort. Better to restrict this to the relevancy case and if it's a consistent issue with other sort attributes, based on consulting with the service manager and metadata we can look at extending the attribute DSL to allow specifying tie-breakers per-attribute.

app/services/user_search_service.rb
```diff
@@ -28,12 +28,22 @@ def results
     search_options = { q: query, models: search_models, as: @current_user,
                        facets: facets, ranges: @search_params[:ranges] }
 
-    # sort by relelvance if a search term is present and no explicit sort field has been chosen
-    sort_field = @search_params[:sort]
-    sort_field ||= :relevance if query.present?
+    # Sort by relevance if a search term is present and no explicit sort field has been chosen
+    sort_fields = [@search_params[:sort]]
+    sort_fields ||= [:relevance] if query.present?
```
mbarnett, line 33: Should be `[:relevance, :title]`.

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -32,33 +32,42 @@ def offset(num)
...
     criteria[:sort] = solr_exporter.default_sort_indexes if criteria[:sort].blank?
 
-    # Note the elsif: if no explicit order was passed from the user, and we're ordering by score, we default
-    # to sorting scores descending rather than ascending, as is otherwise used when eg) title is the default sort field
-    criteria[:sort_order] = if order.present? && [:asc, :desc].include?(order.to_sym)
-                              [order]
-                            elsif criteria[:sort] == [:score]
-                              [:desc]
+    order = [order] unless order.is_a?(Array)
+    order.map! { |x| x ? x.to_sym : nil }
```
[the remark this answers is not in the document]
mbarnett: Forgot to mention this could also be simplified to just `order&.to_sym` inside the block

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -32,33 +32,42 @@ def offset(num)
...
+        solr_name = if attr == :relevance
                       :score
                     else
                       solr_exporter.solr_name_for(attr, role: :sort)
                     end
-                  rescue ArgumentError
-                    nil
-                  end
-      criteria[:sort] = [solr_name] if solr_name.present?
+
```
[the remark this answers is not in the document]
lagoan: Right! Thanks for catching this

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
lagoan: That solution is really clean! However, if I understand correctly there will be an issue with the ``:score`` attribute, as it will occur that ``:score`` has a matching order of ``:asc``. If this is the case then it will not be set to ``:desc`` as it should, this is why I did not handled it on an ``elsif`` but a normal case where it should just be changed to ``:desc``. This is a simple change that can be done easily.

Please let me know if I missed something.

app/services/user_search_service.rb
lagoan: I think I may have misunderstood this comment:
https://github.com/ualbertalib/jupiter/blob/dcab7a53c1d0307a84a3a7eccd56695bb07bdbbc/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb#L53-L54

I understood it as, when no explicit order is passed, then use ascending, as is otherwise used.

The otherwise used made me think this was the default for everything else. If it is not I can change it to the one set per-model.

app/services/user_search_service.rb
lagoan: I find it may be really useful to add different tie breakers.

The idea of using title as always a tie breaker (or default at least after changing it to be definable) so we have consistent results.

My reasoning is this: If we sort by date, and we have items with the same date, we would want the user to be able to easily read the results that should come up in a deterministic manner.

So if I have items A and B, that have the same date (01/01/2000), it makes sense to me that they should always appear 

- A (01/01/2000)
- B (01/01/2000)

Which may not be the case right now as we've seen when sorting by ``:score``

This is why I am ordering it in the same direction as the first parameter, for when we switch the date order, we get them sorted in reverse order by title as I still easily see them reversed when I flip the direction:

- B (01/01/2000)
- A (01/01/2000)

I picked title because that is what I am looking at in the list.

app/services/user_search_service.rb
lagoan: I addressed them separately for the reason mentioned here https://github.com/ualbertalib/jupiter/pull/1827#discussion_r477404243

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
lagoan: Just to add to this, I am not saying we will be sorting by ``:asc`` order for ``:score`` but solr needs to have paired ``sort_attr`` and ``order``, so ``:score`` will need a defined ``order``

mbarnett: Right, good catch! The if and elsif should be reversed so that `:sort, :desc` takes precedence

app/services/user_search_service.rb
mbarnett: Sorry, that's my bad for leaving the stale comment in there. I think at one point it was true that :score got descending and everything else got ascending, but it ended up becoming "score gets descending, everything else gets whatever the default on the model is (which is what `solr_exporter.default_sort_direction` gives you, as the `solr_exporter varies by the model being searched). Hence any :sort argument lacking an order gets that as the default :order

app/services/user_search_service.rb
mbarnett: I agree that it might be useful in some sense, but the problem is that it becomes a major user-facing change that we haven't discussed or tested with anyone.

I think we want to keep this PR, as much as possible, narrowly focused on extending the ability of the search subsystem to accomodate tie-breakers and put an end to test-flapping, *without* unduly changing the semantics of existing usages of search in ways that are user-visible.

As weird as the current Date sorting results are, they behave consistently. We have a plan outlined in that other ticket to make them behave in a way that stakeholders agree will best match their expectations for date sorting. Changing the results it in the meantime is just going to cause surprise and consternation among people who have built workflows around the existing ordering, without actually being something that defensible as being "better" because it doesn't properly solve the problem, it just gives a different improper ordering.

Lets stick to the minimally-invasive approach.

mbarnett (approved):
well done

PR #1829, opened 2020-08-24
Disabling Collection checking of Communities

mbarnett:
Disabling due to issue with many communities in Prod having become corrupted in Fedora, blocking Collection creation. Unnecessary once Fedora is migrated away from. 

Also disables updating and editing of Communities since that's currently broken for any of the impacted Communities, and deleting of Communities since that seems related to the issue in the first place (although the bug seems unlikely to have come through the interactive layers).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to Jupiter project will be documented in this file. Jupiter 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and releases in Jupiter project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.2.22] – 2020-08-24
+- disable Collection usage of Communities for a PCDM triple in order to work around data corruption in Prod
 
 ## [1.2.21] – 2020–05-28
 - bump kaminari from 1.1.1 to 1.2.1 [CVE-2020-11082](https://github.com/advisories/GHSA-r5jw-62xg-j433)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to Jupiter project will be documented in this file. Jupiter 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and releases in Jupiter project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ## [1.2.22] – 2020-08-24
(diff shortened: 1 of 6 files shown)


ConnorSheremeta (dismissed):
LGTM

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to Jupiter project will be documented in this file. Jupiter
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and releases in Jupiter project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
```
pgwillia, line 7: Minor-est of nits but we could keep the Unreleased section around. Most of the recent updates have just been quick releases anyways so there hasn't been a build up of unreleased content.  But even as a place marker maybe it's useful.

mbarnett: Whoops!

test/integration/community_show_test.rb
```diff
@@ -40,15 +40,6 @@ def before_all
...
-    # Community delete, edit and create new collection buttons should be shown
-    assert_select 'a[href=?]',
-                  admin_community_path(@community1),
-                  text: I18n.t('delete')
-
-    assert_select 'a[href=?]',
-                  edit_admin_community_path(@community1),
-                  text: I18n.t('edit')
-
     assert_select 'a[href=?]',
```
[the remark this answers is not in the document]
mbarnett: No, the `before_save~ is actually what's keeping them from being created on Production right now.

What's happening is that a bunch of Communities are deleted in Fedora but still exist in Solr.

This actually _mostly_ doesn't break anything, because most of the code that deals with Communities, like 

https://github.com/ualbertalib/jupiter/blob/22b8ecb952768aa2fbf7373b28173a7505ed94c8/app/models/collection.rb#L25-L30

just deals with Solr. BUT, that `before_save` in Collection tries to unlock the Community (ie, read it from Fedora) in order to update the PCDM triple:

https://github.com/ualbertalib/jupiter/blob/22b8ecb952768aa2fbf7373b28173a7505ed94c8/app/models/collection.rb#L67

and it's that one action that's blowing things up anytime someone tries to create a Collection right now.

CHANGELOG.md
mbarnett: Put that back, now.

test/integration/community_show_test.rb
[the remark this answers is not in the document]
mbarnett: The consequence is that we're going to briefly stop tracking that triple, but it doesn't really matter because we track it in a different way anyways, and it's going away with the removal of Fedora anyways.

PR #1830, opened 2020-08-25
Destroy attachments now and call save without validations in rake task

ConnorSheremeta:
#1685


mbarnett:
Was thinking about it and the attachments on drafts should probably bypass validations too.

Issue #1723, opened 2020-06-18
Can we get https enabled for UAT?


pgwillia:
I’ve sometimes run across instances where a “thing” works in http but fails under https

_Originally posted by @pgwillia in https://github.com/ualbertalib/jupiter/issues/1680#issuecomment-646341599_

pgwillia:
```
# config/nginx.conf
server {

  listen 80 ssl;

  client_max_body_size 4G;
  keepalive_timeout 10;

  ssl_certificate     library.ualberta.ca.bundle.crt;
  ssl_certificate_key library.ualberta.ca.key;
  ssl_protocols       TLSv1 TLSv1.1 TLSv1.2;
  ssl_ciphers         HIGH:!aNULL:!MD5;

  error_page 500 502 504 /500.html;
  error_page 497  301 =307 https://$host:$server_port$request_uri;
...
}
```

```
# docker-compose.yml
nginx:
    restart: always
    image: nginx
    depends_on:
      - web
    env_file: .env_deployment
    volumes:
      - ./config/nginx.conf:/etc/nginx/conf.d/default.conf
      - ../UALcert/library.ualberta.ca.key:/etc/nginx/library.ualberta.ca.key
      - ../UALcert/library.ualberta.ca.crt:/etc/nginx/library.ualberta.ca.crt
      - ../UALcert/library.ualberta.ca.crt:/etc/nginx/library.ualberta.ca.bundle.crt
      - assets:/app/public/
    ports:
      - "80:80"
```
```

PR #1831, opened 2020-08-25
Enable SSL for UAT

pgwillia:
## Context
I’ve sometimes run across instances where a “thing” works in http but fails under https. We want to be able to see this working on UAT.

Related to #1723 

## What's New

References to certs. Added [X-Forwarded-For headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) to nginx configuration.  Port forwarding for nginx 443 in the docker-compose. Updated the default_url options. Force ssl.

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -2,6 +2,7 @@ upstream jupiter {
   server web:3000;
 }
 
+  return 301 https://$server_name$request_uri;
 
 server {
 
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -5,7 +5,7 @@ upstream jupiter {
 
 server {
 
-  listen 80 ssl;
+  listen 443 ssl;
 
   client_max_body_size 4G;
   keepalive_timeout 10;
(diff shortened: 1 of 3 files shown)


config/nginx.conf
```diff
@@ -5,23 +5,36 @@ upstream jupiter {
 
 server {
 
-  listen 80;
+  listen 443 ssl;
```
lagoan, line 8: We should add to the configuration so that nginx redirects requests made to port 80 to port 443.

config/nginx.conf
```diff
@@ -5,23 +5,36 @@ upstream jupiter {
...
 
   client_max_body_size 4G;
   keepalive_timeout 10;
 
+  ssl_certificate     /etc/nginx/certs/library.ualberta.ca.crt;
```
lagoan, line 13: Does it make a difference if we use library.ualberta.ca.bundle.crt vs library.ualberta.ca.crt ? I think the bundle includes information from the intermediate certificate authority.

mbarnett: It depends on the exact details of the cert, but generally you almost always want to include the intermediate certs

config/nginx.conf
```diff
@@ -2,26 +2,44 @@ upstream jupiter {
   server web:3000;
 }
 
+server {
+  listen 80;
+  server_name $HOSTNAME;
+  return 301 https://$server_name$request_uri;
+}
```
pgwillia, line 9: Added redirect for port 80 to https

config/nginx.conf
```diff
@@ -2,26 +2,44 @@ upstream jupiter {
...
 
 server {
 
-  listen 80;
+  listen 443 ssl;
 
   client_max_body_size 4G;
   keepalive_timeout 10;
 
+  ssl_certificate /etc/nginx/certs/library.ualberta.ca.bundle.crt;
```
pgwillia, line 18: Use the one with the information about the intermediate certificate authority.

Issue #1816, opened 2020-08-13
[Oaisys] Resolve testing issues


ConnorSheremeta:
Full details on issues here: https://github.com/ualbertalib/jupiter/issues/1712

## GetRecord
Get record does not work fully right now.
Issue: Cant find a matching record with that identifier due to the prefixed oai:era.library.ualberta.ca: on the identifiers. It works when the identifier is just the uuid.

## ListRecords
List records works for the metadata prefix oai_dc but not oai_etdms.
Issues with oai_etdms:

- 500 from object.departments.first from app/decorators/metadata/oai_etdms/thesis_decorator.rb:47:in discipline'`.
- Publisher field is never displayed.
- Degree name/grantor is sometimes not displayed.

The first issue in ListRecords can be resolved by using the safe navigator operator.

Once we've worked out all of the low-hanging bugs, reach out to Metadata and ask for their help verifying the output.

ConnorSheremeta:
@sfarnel @danydvd 
Regarding OAI:

We've noticed that some theses do not contain a degree name or grantor, also all theses do not contain a publisher. LAC has expressed that these fields are mandatory.

What should be a placeholder for all publisher fields? Also, what should the placeholders be if the degree name or degree grantor is missing?

PR #1834, opened 2020-08-27
Opened up oai routes, modified and added decorators, updated gemfile …

ConnorSheremeta:
…tags

Will still need to bundle this once the new versions are tagged in oaisys and acts_as_rdfable
https://github.com/ualbertalib/jupiter/issues/1816


ConnorSheremeta:
Tests will pass once merged https://github.com/ualbertalib/jupiter/commit/9fac3679283208fe53fc958ede55a584a8826513

app/decorators/metadata/oai_etdms/thesis_decorator.rb
```diff
@@ -44,11 +44,22 @@ def degree_level
...
 
   def discipline
-    object.departments.first
+    object.departments&.first
   end
 
   def institution
-    h.humanize_uri(:institution, object.institution)
+    # TODO: Replace Unknown with appropriate tag from metadata team
+    object.institution.present? ? h.humanize_uri(:institution, object.institution) : 'Unknown'
```
mbarnett, line 52: Does `humanize_uri` just return nil if the URI is unknown? If so you could just do:

```ruby
h.humanize_uri(:institution, object.institution) || 'Unknown'
```

ConnorSheremeta: Yes it does, this is much simpler than using a ternary operator!

PR #1861, opened 2020-09-08
uat configuration to accept proxy from upstream nginx-proxy

### LINKED ISSUE #1724: Distinct server names or routes to applications on UAT
Distinct server names or routes to applications instead of port ‘x’ is Jupiter, port ‘y’ is avalon, etc. Personally, I find names easier to remember than numbers. If the applications are containerised, a reverse proxy could handle the routing. [Envoy](https://www.envoyproxy.io/) or [Traefik](https://containo.us/traefik/)?

Originally posted by @pgwillia in #1680 ([comment](https://github.com/ualbertalib/jupiter/issues/1680#issuecomment-646341599))

pgwillia:
## Context

Names are so much easier to remember that port numbers.  Let's use subdomains of *.uat.library.ualberta.ca and [nginx-proxy](https://github.com/nginx-proxy/nginx-proxy)/docker-gen to do this proxying to different docker containers in our different projects.
https://github.com/ualbertalib/di_internal/pull/96

Closes #1724

## What's New
Don't do the port mapping anymore and add the environment variables for nginx-proxy

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -21,8 +21,10 @@ services:
   solr:
     restart: always
     image: solr:6.6
+    environment:
+      - VIRTUAL_HOST=solr.era.uat.library.ualberta.ca
     ports:
-      - "8983:8983"
+      - "8983"
     volumes:
       - solr:/opt/solr/server/solr/mycores
       - ./solr/config:/config
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -70,5 +70,5 @@ services:
       - ../UALcert:/etc/nginx/certs
       - assets:/app/public/
     ports:
-      - "80:80"
-      - "443:443"
+      - "80"
+      - "443"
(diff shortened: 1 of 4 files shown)


docker-compose.production.yml
```diff
@@ -70,5 +70,5 @@ services:
       - ../UALcert:/etc/nginx/certs
       - assets:/app/public/
     ports:
-      - "80:80"
-      - "443:443"
```
murny, line 74: Just curious why we don't to specify the HOST port here?

I assume we can't because if we running multiple docker composes with the same config (Avalon/etc) we will be fighting over the same host ports as each other. 

Does WatchTower do something fancy here when docker chooses a. random ephemeral host port?

murny (dismissed):
LGTM. Just tiny question to help me understand the port stuff

docker-compose.production.yml
murny: Ah helps me if I read your DI Internal PR 😆 . Nice 👍

murny (approved):

murny (approved):

Issue #1838, opened 2020-08-28
Read-Only Mode


mbarnett:
We want a simple way to enable and disable a read-only mode for ERA, in response to events like emergencies or upgrades that require us to freeze the data in a certain state, but don't require us to take the site down altogether.

I'm thinking we:

- Create a very small one-value table and model to indicate whether or not ERA is in read-only mode. Model just needs a couple of class methods (`ReadonlyMode.enabled?` and `disabled?`. Can probably make creating instances private and bury setting and clearing the mode behind class methods as well, for simplicity)
- Create two rake tasks to set the value to true or false
- Shortcut out of https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/controllers/application_controller.rb#L33 by returning nil immediately, and https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/controllers/application_controller.rb#L53 by doing nothing, if the read-only mode is true.
- Cache the check for, say, 1 minute so-as not to introduce more DB traffic on every single request. 
- When the read-only mode is enabled, remove the log in button from the site 
- When the read-only mode is enabled, display a banner on the site indicating that logins are temporarily disabled. This should work _in addition to_ any notifications an admin has chosen to post (so both should be able to exist one atop the other simultaneously, or either by themselves)

lagoan:
What if instead of having a single one value table we have a site wide configuration table (JupiterConfiguration or whatever) Where readOnlyMode is a value in that table.

This would allow to add more configuration values down the line and there is trivial extra work done to achieve this.

We could have a #readOnlyEnabled? method that can set the default value if it does not exist on the database.

mbarnett:
In the spirit of YAGNI I really think that would be getting ahead of ourselves? 

Right now we only have one use-case for this, and my philosophy is that you always want to have a minimum of two competing use-cases for something before you try to design a generic API around it. 

I'd also be concerned about the lack of clarity between a "generic configuration" table and our general philosophy of config being something that is contributed by the environment (the 12 factor app approach, more or less). The need to make a read-only mode "dynamic configuration" is as much as anything a work-around for the extremely long deployment times imposed upon us any time we want to make a simple change to configuration; I don't think we should let this corner case inform our more general approach to configuration by building out a mechanism that in some sense encourages doing the wrong thing?

ConnorSheremeta:
What would be the best message for the banner on the top of the page for this? I have:

'Logins are currently disabled due to system maintenance, please try again later.'

[...]

PR #1862, opened 2020-09-09
add read only mode feature

ConnorSheremeta:
#1838

relates to #1875


db/migrate/20200831181118_create_read_only_modes.rb
```diff
@@ -0,0 +1,7 @@
+class CreateReadOnlyModes < ActiveRecord::Migration[6.0]
+  def change
+    create_table :read_only_modes do |t|
+      t.boolean :enabled
```
murny, line 4: Should force this as `null: false` and `default: false`?

app/views/application/_navbar.html.erb
```diff
@@ -94,7 +94,7 @@
           </li>
         <% else %>
           <li role="presentation" class="nav-item">
-            <%= link_to t('.links.login'), '/auth/saml', class: 'nav-item nav-link text-nowrap', method: :post %>
+            <%= link_to t('.links.login'), '/auth/saml', class: 'nav-item nav-link text-nowrap', method: :post if ReadOnlyMode.disabled? %>
```
murny, line 97: Are you still able to login with dev mode (below) if ReadOnlyMode is disabled? If not, then we should wrap both of these login buttons in this if condition

test/controllers/sessions_controller_test.rb
```diff
@@ -173,16 +173,6 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_not logged_in?
   end
 
-  test 'should not log in from a saml account that was changed to system' do
```
murny, line 176: Is this no longer valid? You can login from saml with system user now?

app/models/user.rb
```diff
@@ -37,6 +39,11 @@ class User < ApplicationRecord
             absence: { message: :blank_if_system_false },
             if: -> { !system && api_key_digest.present? }
 
+  # Ensure only one system user exists
+  # rubocop:disable Rails/Validation
+  validates_uniqueness_of :system, if: :system
```
murny, line 44: I would say we use what rubocop wants? Instead of disabling it

```
# bad
validates_uniqueness_of :foo

# good 
validates :foo, uniqueness: true
```

app/models/read_only_mode.rb
```diff
@@ -0,0 +1,36 @@
+class ReadOnlyMode < ApplicationRecord
+
+  validate :only_one_record_exists, on: :create
```
murny, line 3: Should throw a quick validation on the enabled boolean field. e.g:

```

validates : enabled, inclusion: { in: [true, false] }

```

config/locales/en.yml
```diff
@@ -166,6 +166,11 @@ en:
               present_if_system_true: 'must be present if System value is true'
               blank_if_system_false: 'must be blank if System value is false'
 
+        read_only_mode:
+          attributes:
+            enabled:
+              only_one_record_exists: 'Only one read_only_mode record may exist'
```
murny, line 172: Minor but since we talking about the model here, should maybe be `ReadOnlyMode` instead of `read_only_mode`

app/views/application/_navbar.html.erb
```diff
@@ -94,12 +94,12 @@
...
           </li>
           <% if Rails.env.development? || Rails.env.uat? %>
             <%# TODO: Temporary login form for local development or UAT %>
             <li role="presentation" class="nav-item">
-              <%= link_to t('.links.login_as_dev'), '/auth/developer', class: 'nav-item nav-link text-nowrap', method: :post %>
+              <%= link_to t('.links.login_as_dev'), '/auth/developer', class: 'nav-item nav-link text-nowrap', method: :post if ReadOnlyMode.disabled? %>
```
murny, line 102: Minor but I'd throw the if logic around the entire html. Why? Cause this creates extra markup when false.

E.g when ReadOnlyMode is enabled, then this markup still gets rendered in the HTML:
```
<li role="presentation" class="nav-item">
</li>
<li role="presentation" class="nav-item">
</li>
```

If you surrounded this markup with an if block instead:

```
<% if ReadOnlyMode.disabled? %>
  ...contents
<% end %>
```
Then no markup gets rendered when ReadOnlyMode is enabled.

app/models/read_only_mode.rb
```diff
@@ -0,0 +1,37 @@
+class ReadOnlyMode < ApplicationRecord
```
murny, line 1: Since we adding a new model can we add a new model test for this class?

And just check some simple things like...can we create more then one record? can we create a record without setting enabled? Does our class level methods work properly (enable/disable/etc) etc

murny (changes requested):
Looks really good. Just would like to see a model test for the new ReadOnlyMode model 👍. Then I'll approve 🥇

app/controllers/application_controller.rb
```diff
@@ -51,6 +53,8 @@ def current_user
 
   # Signs in the given user.
   def sign_in(user)
+    return if ReadOnlyMode.enabled?
```
mbarnett, line 56: I'm going to suggest we create a method in ApplicationController to wrap `ReadOnlyMode.enabled?` and hoist the `Rails.cache.fetch('read_only_mode.first.enabled', expires_in: 1.minute) do` caching policy up into the method in ApplicationController, rather than keeping it in the model.

There's a couple of reasons for this: one I'll get to below, but the other one is just that caching feels like application policy and what you want for the web application isn't necessarily what you want for other use-cases, like rake tasks (like, if we wanted a task to check the status, we'd want it to report the real-time state, not the cached value. that's harder if the model itself is hiding the caching)

app/views/application/_navbar.html.erb
```diff
@@ -94,12 +94,12 @@
           </li>
         <% else %>
           <li role="presentation" class="nav-item">
-            <%= link_to t('.links.login'), '/auth/saml', class: 'nav-item nav-link text-nowrap', method: :post %>
+            <%= link_to t('.links.login'), '/auth/saml', class: 'nav-item nav-link text-nowrap', method: :post if ReadOnlyMode.disabled? %>
```
mbarnett, line 97: This is the other reason to wrap this call in a method in ApplicationController: Views shouldn't talk directly to models, we want a method in the controller layer in between to act as a "fire-break" to keep them from getting to familiar with one another. It helps keep things more cleanly separated, which makes any refactoring down the road a lot easier.

app/models/read_only_mode.rb
```diff
@@ -0,0 +1,37 @@
...
+    read_only_mode.save!
+
+    Announcement.new(message: I18n.t('announcement_templates.read_only_mode'), user: User.system_user).save!
+  end
+
+  def self.disable
+    read_only_mode = ReadOnlyMode.first
+    read_only_mode.enabled = false
+    read_only_mode.save!
+    Announcement.where(user: User.system_user.id).destroy_all
```
mbarnett, line 28: Because we have all of the scopes declared properly, you can just use the ActiveRecord join helpers! `User.system_user.announcements`

app/models/read_only_mode.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
```
mbarnett, line 29: @murny want to get your thoughts on this, too: 

We have some fairly tight coordination between 2 (3, if you count the magic system user) different models here, and depending on who you ask, this tight inter-model coupling of state is a bad idea. Options would be to just let the models touch each other like this, move to something like the old Observer system, or just pull `enable` and `disable` up into a Service and let it handle the coordination? So you could check values on the model, but actually changing it requires using the service?

mbarnett (changes requested):

app/models/read_only_mode.rb
murny: Current code isn't bad by any means (but if we add more things to this Model in the future, then it could become troubling). But I like where your going with this. I'd probably lean towards a service object as well instead of these Class level methods that are touching multiple models. This makes our ReadOnlyMode model super simple and straightforward.

 This logic is perfect candidate for a Service object (Ruby module which encapsulates the logic for completing an action). A simple ReadOnlyService that completes the action of turning on or off Read Only Mode for Jupiter 👍

mbarnett:
Forgot to mention: remember this needs to be opened against master, too, because we need it in prod for the deployment.

test/services/read_only_service_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class ReadOnlyServiceTest < ActiveSupport::TestCase
+
+  include ActiveJob::TestHelper
```
murny, line 5: Probably not needed? Doesn't look like your checking for jobs being queued/etc in the tests

murny (dismissed):
Very nice work on service and adding tests 👏.

Just a small extra include I saw in your service tests. Otherwise looks good 👍

murny (approved):

mbarnett (approved):

PR #1863, opened 2020-09-09
Fix rubocop performance cop issue #trivial

lagoan:
This small change fixed the rubocop problem introduced when merging PR #1827 

## Context

PR #1827 introduced a problem with cop [Performance/CollectionLiteralInLoop](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancecollectionliteralinloop)

## What's New

Small change replacing an array with literals with a local variable.

### DIFF
--- a/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
@@ -41,6 +41,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
 
     criteria[:sort] = []
     criteria[:sort_order] = []
+    possible_sort_orders = %i[asc desc]
 
     attributes.each_with_index do |attr, idx|
       next if attr.blank?
@@ -68,7 +69,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
                                  # When sorting by score it only makes sense to use :desc order from the user
                                  # perspective so we ignore the order if one is given.
                                  :desc
-                               elsif order.present? && [:asc, :desc].include?(order)
+                               elsif order.present? && possible_sort_orders.include?(order)
                                  order
                                else
                                  # We could not find the order so we switch to default sort direction
--- a/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
@@ -4,6 +4,10 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
   include Kaminari::PageScopeMethods
   include Kaminari::ConfigurationMethods::ClassMethods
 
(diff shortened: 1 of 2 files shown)


app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -41,6 +41,7 @@ def sort(attributes, orders = nil)
 
     criteria[:sort] = []
     criteria[:sort_order] = []
+    possible_sort_orders = %i[asc desc]
```
murny, line 44: Minor but think we prefer `[:asc, :desc]` over the percent versions `%i[asc desc]`, but Matt can confirm that.

I suppose maybe instead of turning these off, we should explicit say we want the other version (might be a reason why we did this though):

https://github.com/ualbertalib/jupiter/blob/integration_postmigration/.rubocop.yml#L154-L162

lagoan: You are right in that it is more consistent with the rest of the code base.

mbarnett: Yeah, I like `[:asc, :desc]` better myself, and I agree with Shane that we might need to tell Rubocop we explicitly want that approach vs `%i`. One of my many complaints about rubocop is that it tends to want you to use obscure corner-case syntax like `%i` that just tends to confuse people because of how rarely it comes up...

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -41,6 +41,7 @@ def sort(attributes, orders = nil)
 
     criteria[:sort] = []
     criteria[:sort_order] = []
+    possible_sort_orders = [:asc, :desc]
```
murny, line 44: Another nitpick, is we are defining this way up here (outside of the for loop). So we defining and allocating this way up here but a good chance we may never use it. Since where its being actually used is nested in a ton of conditional logic.

Might be better to move this where we are using it?

murny: I guess you could argue then that we are creating this variable on each iteration. So nevermind, let's leave it as is 👍

murny (approved):
LGTM 👍

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
mbarnett: While we're micro-optimizing – I think rather than move it closer like Shane's suggesting I'd do the opposite – since there's only ever going to be two sort orders there's no reason for this to be mutable at all, so you could just declare it as a constant at the start of the class, and only ever do the allocation once globally.

mbarnett (approved):

murny (approved):

PR #1865, opened 2020-09-09
Enable Skylight on Staging

mbarnett:
## Context

We definitely have some asset-related performance issues on Staging, and need to do some profiling to see what's up.

## What's New

Enable Skylight in the staging Rails environment. Also removes "UAT" as an enabled environment, since we were never actually giving it the skylight tokens on UAT, and the performance measures are less likely to be representative.


murny (approved):
:shipit:

mbarnett:
Shouting Rubocop! at the sky and shaking my fist

murny (approved):
gif

PR #1875, opened 2020-09-15
add read only mode

ConnorSheremeta:
#1838

almost the same as #1862 but in master instead of integration_postmigration. No system user in this branch so I assumed it by email (the user with that email becomes a system user in a later migration.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,7 +7,7 @@ class ApplicationController < ActionController::Base
 
   protect_from_forgery with: :exception
 
-  helper_method :current_announcements, :current_user
+  helper_method :current_announcements, :current_user, :read_only_mode_enabled?, :read_only_mode_disabled?
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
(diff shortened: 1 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -107,4 +111,14 @@ def current_announcements
     Announcement.current
   end
 
+  def read_only_mode_enabled?
+    Rails.cache.fetch('read_only_mode.first.enabled', expires_in: 1.minute) do
+      ReadOnlyMode.first.enabled
+    end
+  end
+
+  def read_only_mode_disabled?
```
mbarnett, line 120: Really tiny niggle, but I found the view below with the conditional login link a bit confusing at first, I think because this method name forces the check to be phrased in terms of a negative, eg) `elsif read_only_mode_disabled? `.

Maybe the methods should be both phrased in the positive, eg) `read_only_mode_enabled?` and `logins_enabled?` as its opposite, so that the views can read more like `elsif logins_enabled? <render login link>`

mbarnett (commented):
Looks really good other than I would maybe tweak the one method name for readability.

mbarnett (approved):

PR #1881, opened 2020-09-18
Resolve N+1 on Attachments in Search Results

mbarnett:
## Context

With the switch from Fedora to PostgreSQL, we're increasing pressure on the DB server, and consequently turning up some old inefficiencies that can now be addressed.

Previously, search results involved making:
1) one Solr query containing all of the data needed to render the results minus information about the attachments
2) 10 queries to Solr to fetch the attachments for the 10 results on the page.

Now, we are making:

1) one Solr query, giving us the IDs of the resulting model instances
2) 10 queries for the model instances
3) 10*N more queries for the model instances' attachments, where N is the number of attachments each instance has

This is excessive. 

This PR aims to address point 3 (point 2 may be addressed separately at some point as per https://github.com/ualbertalib/jupiter/blob/291ebcc6bdb17049523941df25a44c31831b6baa/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb#L119 but note the significant challenge here is that not all Solr results will necessarily come from the same DB table, so batching it up into a single query is likely impossible unless we unify `Item` and `Thesis`)

A complicating factor here is that the ActiveStorage API names the eager-load scope according to the attached model so, for instance, `Item`(and `Thesis`), which `has_many_attached :files`, has an eager-loading scope of `with_attached_files`, whereas `Community`, which `has_one_attached :logo`, has an eager-loading scope of `with_attached_logo`. And Collection has no eager-loading scope at all, because it lacks any attachments.

`DeferredFacetedSolrQuery` potentially deals with results containing all of the above models at runtime, with no knowledge of which scope to call when making a query for the record in the result:

https://github.com/ualbertalib/jupiter/blob/291ebcc6bdb17049523941df25a44c31831b6baa/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb#L120-L121

## What's New

We introduce two new Class methods in the `Depositable` superclass (which all Solr-searchable models must inherit from).

1) A protected method, `self.eager_attachment_scope`, which should return a model-specific call to the correct scope for the kind of attachment specific to the model (so eg. `Item` returns `self. with_attached_files`). By default this method blows up noisily when called, to force us to remember to override it correctly in all subclasses. Models like `Collection` simply make this method a no-op by choosing to return `self`.

2) A public method, `self.with_eagerly_loaded_attachments`, which simply gives the models' client code a nicer DSLy method for accessing the scope.

`DeferredFacetedSolrQuery` is then altered to use the scope on all classes during reification of search results.

(unrelatedly, I also shifted the methods around in `Item` and `Thesis` to group all instance methods and class methods together, because they were kind of a random mess)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 ### Fixed
 - bump rubocop and fix cop violations [PR#1845](https://github.com/ualbertalib/jupiter/pull/1845)
 - bump rubocop-performance and fix cop violations [PR#1850](https://github.com/ualbertalib/jupiter/pull/1850)
+- N+1 query issue with attachments to models in search results [PR#1881](https://github.com/ualbertalib/jupiter/pull/1881)
 
 ## [2.0.1.pre2] - 2020-09-01
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -17,6 +17,9 @@ class Collection < JupiterCore::Depositable
     self.visibility = JupiterCore::VISIBILITY_PUBLIC
   end
 
+  # We have no attachments, so the scope is just the class itself.
+  def self.eager_attachment_scope; self; end
+
   def path
     "#{community_id}/#{id}"
   end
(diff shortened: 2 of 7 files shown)


murny (approved):
Looks good to me 👍. Nice work.

Might be nice to use `bullet` gem (https://github.com/flyerhzm/bullet) to help us discover these in the future since we now on ActiveRecord 🤔

mbarnett:
Good idea! I'll take that for a spin locally and try it out.

PR #1886, opened 2020-09-23
Disable PCDM member of collections manipulation on Items/Theses

mbarnett:
## Context

This is continuing to cause problems in Prod w/ 404s being thrown on ingest during deposit. I think the root cause here may be a bug in ActiveFedora which sometimes writes (or returns?) triples related to the actual item in question instead of PCDM proxy objects when calls to `member_of_collections` are made. When these items are subsequently deleted, the PCDM manipulation code then calls delete not upon the proxies, but on the containing collection itself (which bypasses all validation safeguards because that's how ActiveFedora works for `delete` instead of `destroy`)

## Changes

Rip out everything related to this code. All of it is going away shortly during the migration anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+## [1.2.23] – 2020-09-23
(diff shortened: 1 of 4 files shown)


murny (approved):
gif

Issue #1739, opened 2020-06-30
Give admins the ability to reset/request DOIs for Items/Theses where they are missing or broken


mbarnett:
## User Story

 As a Developer, I would like to get out of the business of manually fixing issues with DOIs.

## Context

Occasionally (phase of the moon? random server reboots? timeouts on datacite's end? idk. I suspect we might be hitting some kind of rate limit on Datacite's end, as it seems to happen to things that are ingested in a large batch...) items are getting stuck with their DOI state in `:unavailable` or `:unminted`. Dropping the state and manually touching off a call to `handle_doi_states` has resolved all of these so far, but I've seen 4 or 5 such issues come to me since May, and having me do this manually is an expensive way to solve this problem.

## Pitch

Let's just create a button in the interface, only available to admins, on Items and Theses, that resets the state machine and re-requests the DOI, so that helpdesk and metadata stakeholders can fix this themselves rather than need our assistance.

ConnorSheremeta:
Clicking on the button/link should call [remove_doi](https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/models/jupiter_core/doiable.rb#L62) and the subsequently call [handle_doi_states](https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/models/jupiter_core/doiable.rb#L41)

These can be called from within a method in a doi_controller in the admin directrory.

The link/button can be made a partial and can be rendered in [_item.html.erb](https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/views/items/_item.html.erb) along with [_thesis.html.erb ](https://github.com/ualbertalib/jupiter/blob/integration_postmigration/app/views/items/_thesis.html.erb).


Screenshots of mock-ups:

### 1. Under an admin panel
If we think three will be more things in here in the future this might be the way to go? I think it would be nice for all admin abilities to be in one spot.
![Screenshot from 2020-07-15 14-16-39](https://user-images.githubusercontent.com/14242562/87592916-e0b4ba00-c6a7-11ea-8b95-d17cee989ab7.png)
___
### 2. As a button below the DOI
This might make the most sense as it is right below the 'broken' DOI
![Screenshot from 2020-07-15 14-13-03](https://user-images.githubusercontent.com/14242562/87592917-e0b4ba00-c6a7-11ea-9775-d1fc114cfd0e.png)
___
### 3. As a link beside the DOI
This is not a nice looking as a button in my opinion but is still a valid approach
[...]

PR #1891, opened 2020-09-25
Implement doi resetting for admins

ConnorSheremeta:
#1739

87592916-e0b4ba00-c6a7-11ea-8b95-d17cee989ab7


config/routes.rb
```diff
@@ -78,6 +78,7 @@
   match '/auth/failure', to: 'sessions#failure', via: [:get, :post]
   match '/logout', to: 'sessions#destroy', via: [:get, :post]
   post '/auth/system', to: 'sessions#system_login'
+  get '/admin/reset_doi/:id', to: 'admin/doiable#reset_doi', as: '/admin/reset_doi'
```
murny, line 81: Couple things here. First a `get` request is suppose to be `idempotent` (same operation applied multiple times yields the same result). This is not idempotent as you are changing data. Also you have to remember how a browser operates. Why is chrome so fast? It thinks ahead. It will click on links before you even get to them. Having this link on the page that is a get request, guess what? Chrome will click on it for you even if you have no intention of issuing a request for resetting the DOI. This is a problem!

Second, we should preferably get nested under admin namespace or you should add the `constraints: AdminConstraint.new` to this. So this only gets exposed and gives responses to admin users.

Easy way to get it nested under admin namespace is to be more RESTful.

```
  resources :doi, only: [] do
    match do
      post :reset
    end
  end
```

Or is reset much different then a create? If it is, we can go full REST just do this:
`resources :doi, only: [:create]`

app/views/items/_admin_sidebar.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="card mt-3 admin-sidebar">
+  <div class="card-header">
+    <%= t('.admin') %>
+  </div>
+  <div class="card-body p-2">
+    <ul class="list-group">
+      <div class="d-flex justify-content-center">
+        <div class="p-1">
+          <%= link_to(t('.reset_doi'), admin_reset_doi_path(@item),
```
murny, line 9: Should we throw a confirmation dialog on this too? Just to add some caution to it?

Like "Do you want to proceed to reset the DOI? Yes / No"

```ruby
<%= link_to(t('.reset_doi'), admin_reset_doi_path(@item),
class: 'btn btn-outline-primary',
data: { confirm: t('.are_you_sure_you_want_to_reset') }) %>
```

app/helpers/items_helper.rb
```diff
@@ -54,4 +54,12 @@ def humanize_date(dateish)
     6 => I18n.t('items.thesis.graduation_terms.spring'),
     11 => I18n.t('items.thesis.graduation_terms.fall')
   }.freeze
+
+  def display_admin_sidebar(item)
+    aasm_not_available = item.aasm_state == 'not_available'
```
mbarnett, line 59: AASM gives you a bunch of helper methods for the declared states, eg) `item.not_available?`

config/routes.rb
mbarnett: On top of Shane's good point, I'm thinking this route and the corresponding method would probably make more sense under `Admin::ItemsController`? We're kind of creating a single-purpose controller to handle a change to one small part of a model that already has its own controller – it would make more sense to me to fold it into the main admin controller for these models.

mbarnett (changes requested):

test/integration/doi_reset_test.rb
```diff
@@ -0,0 +1,53 @@
+require 'test_helper'
+
+class DoiResetTest < ActionDispatch::IntegrationTest
+
+  setup do
+    @admin = users(:admin)
+    @user = users(:regular)
+    # A community with two collections
```
murny, line 8: We don't need all this setup here? Should be able replace all this setup from Line 8 to Line 29 with just 

`@item = items(:fancy)` 

Should give you the exact same results (especially since we never access @community/@collection1/@collection2 in any of the test code)

config/routes.rb
```diff
@@ -71,6 +71,8 @@
     end
 
     resources :announcements, only: [:index, :destroy, :create]
+
+    patch '/reset_doi/:id', to: 'items#reset_doi', as: '/reset_doi'
```
murny, line 75: Minor but its probably easier to maintain if we keep this route with the rest of the admin `items` routes on Line 68:

`resources :items, only: [:index, :destroy]`

Can just become:

```
    resources :items, only: [:index, :destroy] do
       member do
         patch :reset_doi
       end
    end

```

which gives you the exact same kind of functionality as your code here. There will be slight differences but it's more RESTful.

For example the route name will now be `reset_doi_admin_items` instead of `admin_reset_doi` and the url will be `admin/items/:id/reset_doi` instead of `admin/reset_doi/:id`

test/integration/doi_reset_test.rb
```diff
@@ -0,0 +1,53 @@
+require 'test_helper'
+
+class DoiResetTest < ActionDispatch::IntegrationTest
```
murny, line 3: Nice work on this test, looks good 👍. 

Curious if we can get some more test coverage for our new `reset_doi` action at the controller level? Essentially add a test in Admin::ItemsControllerTest that probably just sends a patch to `reset_doi` acton, check if a DOICreateJob gets enqueued in our job queue, and finally checks if we are successfully redirected back

Other than these 3 minor things, PR is looking really good 🎉

test/integration/doi_reset_test.rb
ConnorSheremeta: I created the item in the setup block because it lets the test check that the 'Reset DOI' button is not present within the first minute of the item's creation. I just realised that I can still use fixtures and test that, just need to do some time travelling.

test/integration/doi_reset_test.rb
ConnorSheremeta: We have[ DOI minting disabled in test](https://github.com/ualbertalib/jupiter/blob/d7d5fe5ea66a28e628e9731d71cf9788c2867afc/config/secrets.yml#L24).

Do either of you know if enabling this (just for the test env) will have any negative repercussions? Should I try to enable it just for this particular test? @mbarnett @murny

murny: Hmm, it shouldn't impact things as we don't run a real ActiveJob queue in testing (its just a mocked array). And I assume all of our DOI minting stuff runs as background jobs. So we should be safe turning it on in testing?

Worse case we can just leave it as is, and just leave out the `check if DOICreateJob gets enqueued in our job queue` part of the test (really just checking if the controller action returns a successful response code (redirect)).

I'll let Matt decide 😆

PR #1892, opened 2020-09-25
post-update run some rake update commands

pgwillia:
## Context

It is possible to execute pre/post-check and pre/post-update commands inside every container [updated by watchtower](https://containrrr.dev/watchtower/lifecycle-hooks/).

We used to do this by running the deploy script manually for each update.  Now that we're using watchtower the update
happens automagically but we still need these follow on tasks completed.

Related to https://github.com/ualbertalib/jupiter/issues/1722

## What's New

A script which runs `rake db:migrate` and `rake assets:precompile` and the label which tells watchtower's hook what to run.


murny (approved):
Nice work! 👍

Issue #1800, opened 2020-08-05
Investigate how to highlight terms in search results descriptions


murny:
From EPIC: https://github.com/ualbertalib/jupiter/issues/1610

When my search results are returned, I expect to be able to see where the keyword shows up and highlighted within the results.

Example: 
![image](https://user-images.githubusercontent.com/1930474/89468431-ac776b00-d734-11ea-80cb-e78a581ffb7b.png)

![image](https://user-images.githubusercontent.com/1930474/89468452-b4cfa600-d734-11ea-800c-67de7d0ffaed.png)

I know elasticsearch has a "highlighting" feature which gives us exactly this behaviour (more info: https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html). Solr must have something similar we can look into.

pgwillia:
Check out https://lucene.apache.org/solr/guide/6_6/highlighting.html

PR #1895, opened 2020-09-27
#1800: Add solr highlight for description/abstract fields on search r…

murny:
#1800 

Adds solr highlighting to item description/thesis abstract fields on search results (for search, profile and collection search pages)

Highlighting in action:

Search page:
image
image

Profile page:
image

Collection show page:
image

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -8,7 +8,11 @@ class CollectionsController < ApplicationController
       base_restriction_key: Item.solr_exporter_class.solr_name_for(:member_of_paths, role: :pathing),
       value: @collection.path,
       params: params,
-      current_user: current_user
+      current_user: current_user,
+      highlight_fields: [
+        Item.solr_exporter_class.solr_name_for(:description, role: :search),
+        Thesis.solr_exporter_class.solr_name_for(:abstract, role: :search)
+      ]
     )
     @results = search_query_index.results
     @search_models = search_query_index.search_models
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -9,10 +9,7 @@ class CollectionsController < ApplicationController
       value: @collection.path,
       params: params,
       current_user: current_user,
-      highlight_fields: [
-        Item.solr_exporter_class.solr_name_for(:description, role: :search),
-        Thesis.solr_exporter_class.solr_name_for(:abstract, role: :search)
-      ]
+      fulltext: true
     )
     @results = search_query_index.results
     @search_models = search_query_index.search_models
--- a/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
@@ -8,7 +8,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
 
   private_constant :POSSIBLE_SORT_ORDERS
 
-  def initialize(q:, qf:, fq:, facet_map:, facet_fields:, ranges:, restrict_to_model:, highlight_fields:)
+  def initialize(q:, qf:, fq:, facet_map:, facet_fields:, ranges:, restrict_to_model:, fulltext_fields:)
     criteria[:q] = q
     criteria[:qf] = qf # Query Fields
     criteria[:fq] = fq # Facet Query
@@ -18,7 +18,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
     criteria[:restrict_to_model] = restrict_to_model
     criteria[:sort] = []
     criteria[:sort_order] = []
-    criteria[:highlight_fields] = highlight_fields
+    criteria[:fulltext_fields] = fulltext_fields
   end
 
   def criteria
@@ -243,7 +243,7 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
       rows: limit,
       start: criteria[:offset],
       sort: sort_clause,
-      highlight_fields: criteria[:highlight_fields] }
+      fulltext_fields: criteria[:fulltext_fields] }
   end
 
   def raw_model_to_model(raw_model)
--- a/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
+++ b/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
@@ -114,12 +114,12 @@ class JupiterCore::SolrServices::DeferredFacetedSolrQuery
     each do |object|
       fulltext_hits = []
 
-      if @highlights.any? && object.solr_exporter_class.fulltext_searchable_field.present?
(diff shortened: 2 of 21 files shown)


app/services/user_search_service.rb
```diff
@@ -6,15 +6,18 @@ class UserSearchService
 
   attr_reader :search_models
 
-  def initialize(current_user:, base_restriction_key: nil, value: nil,
-                 search_models: [Item, Thesis], params: nil)
-    raise ArgumentError, 'Must supply both a key and value' if @base_restriction_key.present? && @value.blank?
```
murny, line 11: Oops this is a bug. `@base_restriction_key` and `@value` will always be blank here as we don't define these until later. We want the arguments that have been passed in instead (e.g `base_restriction_key` and `value`)

app/services/user_search_service.rb
```diff
@@ -6,15 +6,18 @@ class UserSearchService
 
   attr_reader :search_models
 
-  def initialize(current_user:, base_restriction_key: nil, value: nil,
-                 search_models: [Item, Thesis], params: nil)
```
murny, line 10: Params must be present here, we shouldn't default to nil. We take these params and pass it into our `search_params` function below:

`@search_params = search_params(params)`

`search_params` will crash if `nil` is given for the params value.

So let's make params be a required argument (like current_user)

mbarnett (changes requested):
(I haven't looked at this in depth yet, I'm just blocking the merge so that we don't accidentally import this into the about-to-become-master branch before we tag our final version of new prod. Will review in a bit)

app/controllers/collections_controller.rb
```diff
@@ -8,7 +8,11 @@ def show
       base_restriction_key: Item.solr_exporter_class.solr_name_for(:member_of_paths, role: :pathing),
       value: @collection.path,
       params: params,
-      current_user: current_user
+      current_user: current_user,
+      highlight_fields: [
+        Item.solr_exporter_class.solr_name_for(:description, role: :search),
+        Thesis.solr_exporter_class.solr_name_for(:abstract, role: :search)
```
mbarnett, line 14: Should we expand the DSL around Solr to let this be declared in the exporter rather than in the controller? Might centralize this knowledge in a SPOT rather than risk bugs where we add it in one context but not another?

I'm thinking something like https://github.com/ualbertalib/jupiter/blob/bcc71dadee525771cd36754f3b4d42f1765e882a/app/models/exporters/solr/item_exporter.rb#L65

maybe 

```ruby
full_text_searchable :description
```

?

Then we could change the API for SearchService a bit to just take a `fulltext: true` argument and default it to `false`, and then not use it in the current controllers like here, but opt into it in the digitization controller?

mbarnett (commented):
Overall I really like this!

Added some thoughts on how we could tweak the API to be a bit DRYer, and make it opt-in because current ERA controllers won't want to leverage this immediately.

Will give a bit more thought to the results rendering API tomorrow but this seems good to start with.

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -109,6 +110,11 @@ def each_facet_with_results(first_facet_categories = [])
     end
   end
 
+  def highlights
+    reify_result_set
+    @highlights || {}
```
mbarnett, line 115: I'm thinking we should pre-bake the results here rather than returning the raw structure we get from RSolr and processing it in the helper.

What if we added something like an `each_with_fulltext_results` method?

Then this: https://github.com/ualbertalib/jupiter/blob/bcc71dadee525771cd36754f3b4d42f1765e882a/app/views/application/_results.html.erb#L81-L85

becomes something like
```ruby
<% results. each_with_fulltext_results do |result, fulltext_hits| %>
    <%= render result, highlights: fulltext_hits %>
  </li>
<% end %>
```

and the implementation looks something like:

```ruby
def each_with_fulltext_results
  each do |obj|
    fulltext_solr_field = obj.solr_exporter_class.fulltext_searchable_solr_field
    fulltext_hits = @highlights[obj.id][fulltext_solr_field].map { |hit|  new FulltextResult(highlight_text: hit)}
    yield obj, fulltext_hits
  end
end
```

Baking it like this eliminates the need for the view to hardcode a fieldname that matches the search configuration elsewhere, and wrapping the results in a little utility class will let us simplify the views a lot (see comment there)

app/views/application/_item.html.erb
```diff
@@ -45,10 +45,18 @@
...
+      <% highlight_matches = highlight_matches(object: item, field: :description, highlights: local_assigns[:highlights]) %>
+      <% if highlight_matches.empty? %>
+        <p>
+          <%= jupiter_truncate item.description %>
+        </p>
+      <% else %>
+        <% highlight_matches.each do |match| %>
+          <p><%= match.html_safe %></p>
+        <% end %>
+      <% end %>
```
mbarnett, line 59: if we assume that we changed the API and wrapped the matches in a little utility class, `highlights` is now an array of instances of the utility class, and you can replace these 9 lines with

```erb
<p>
<%= render highlights || jupiter_truncate item.description %>
</p>
```

and add a little `_fulltext_result.html.erb` partial to render its contents. Rails' collection partial rendering returns nil if the collection is empty, making the fallback easy.

tl;dr basically prebaking the results and wrapping them in a custom class drops a lot of logic out of the views.

mbarnett (commented):
Added some thoughts on how we could make some changes to the results API to DRY up the views having to know the field that was fulltext indexed/searched, and how wrapping the raw RSolr result would drop a bunch of logic out of the views

app/controllers/collections_controller.rb
murny: Yeah I would love to do this 🎉 . 

Sunspot leaves it like I did at the search level. 

But popular elasticsearch gems (like SearchKick) does exactly what your proposing. See https://github.com/ankane/searchkick#highlight for an example.  On the model level they denote what fields are highlight-able. Then when they search they pass in a `highlight: true` flag and can get the results with `results.with_highlights.each` (which is basically what your proposing here and your view cleanup proposal below)

I just might need to bug you on best way to do this? Like we set `full_text_searchable :description` and that becomes exposed in  `app/models/jupiter_core/search.rb` 🤔 . I think I'm getting it, just need to spend some time fleshing this out

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
murny: Yeah I like where you going with this 💯 . Much cleaner (and much similar to Searchkick as noted above). Will see what I can do 👍

app/controllers/collections_controller.rb
mbarnett: I've got some time today, I can add this and PR it to your branch

murny: Sure, that be great! Just enough to get me started 👍

PR #1910, opened 2020-10-02
Change ActiveStorage::Blobs to use UUID ids

mbarnett:
Switch the post-migration branch to use UUID Ids for `ActiveStorage::Blob`s. Resolves an issue where PDF or video blobs would fail to correctly track their attached `preview_image`, because `ActiveStorage::Attachment`'s `record_id` expects a UUID. 

This was causing an enormous amount of regeneration of thumbnails (4x per thumbnail) on the new staging environment (a typical search result page was taking ~8 seconds to render).

You will need to recreate and remigrate your DB after this is merged.

### DIFF
--- a/app/helpers/page_layout_helper.rb
+++ b/app/helpers/page_layout_helper.rb
@@ -46,6 +46,7 @@ module PageLayoutHelper
 
   def thumbnail_path(logo, args = { resize_to_limit: [100, 100], auto_orient: true })
     return nil if logo.blank?
+
     Rails.application.routes.url_helpers.rails_representation_path(logo.representation(args).processed)
   rescue StandardError => e
     logger.warn("#{logo.record_type} with id: #{logo.record_id} and thumbnail #{logo.name} threw an error.")
--- a/app/helpers/page_layout_helper.rb
+++ b/app/helpers/page_layout_helper.rb
@@ -46,24 +46,10 @@ module PageLayoutHelper
 
   def thumbnail_path(logo, args = { resize_to_limit: [100, 100], auto_orient: true })
     return nil if logo.blank?
-
-    # images have variants
-    Rails.application.routes.url_helpers.rails_representation_path(logo.variant(args).processed)
-  rescue ActiveStorage::InvariableError
-    begin
-      # pdfs and video have previews
-      Rails.application.routes.url_helpers.rails_representation_path(logo.preview(args).processed)
-    # ActiveStorage::UnpreviewableError and sometimes MiniMagick::Error gets thrown here
-    rescue StandardError => e
-      logger.warn("#{logo.record_type} with id: #{logo.record_id} and thumnail #{logo.name} \
-      threw an error after ActiveStorage::InvariableError.")
-      Rollbar.warn("#{logo.record_type} with id: #{logo.record_id} and thumnail #{logo.name} \
-      threw an error after ActiveStorage::InvariableError.", e)
-      nil
-    end
+    Rails.application.routes.url_helpers.rails_representation_path(logo.representation(args).processed)
   rescue StandardError => e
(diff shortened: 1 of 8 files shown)


test/fixtures/active_storage/attachments.yml
```diff
@@ -2,4 +2,4 @@ logo:
   name: 'files'
   record_type: 'Item'
   record_id: <%= ActiveRecord::FixtureSet.identify(:fancy) %>
-  blob_id: <%= ActiveRecord::FixtureSet.identify(:jpeg) %>
\ No newline at end of file
+  blob_id: 'e34bbd2e-c819-491b-9f60-e0b1c9b7b2ee'
```
mbarnett, line 5: For no discernible reason, ` <%= ActiveRecord::FixtureSet.identify(:jpeg) %>` won't work when Blob has a UUID ID column, whether or not I hardcode it in `blobs.yml` or let ActiveRecord generate it.

db/migrate/20190826221814_change_activestorage_records_to_uuid_part1.rb
```diff
@@ -7,5 +7,10 @@ def change
     add_index :draft_theses, :upcoming_id, unique: true
     add_column :draft_items_languages, :upcoming_draft_item_id, :uuid
     add_index :draft_items_languages, :upcoming_draft_item_id
+
+    add_column :active_storage_blobs, :upcoming_id, :uuid, null: false, default: 'uuid_generate_v4()'
+    add_column :active_storage_attachments, :upcoming_blob_id, :uuid
+    add_column :draft_items, :upcoming_thumbnail_id, :uuid
+    add_column :draft_theses, :upcoming_thumbnail_id, :uuid
   end
 end
```
mbarnett, line 16: Remember that when you're not remigrating a DB from scratch, the `jupiter:get_me_off_of_fedora` rake task runs right here after this migration to copy the new ids over to the tables that reference the old one, hence the two step "create a new column with new ids" and "drop the old id column and rename the one from the previous migration" process.

app/helpers/page_layout_helper.rb
```diff
@@ -47,23 +47,10 @@ def page_description(description = nil)
...
-      Rails.application.routes.url_helpers.rails_representation_path(logo.preview(args).processed)
-    # ActiveStorage::UnpreviewableError and sometimes MiniMagick::Error gets thrown here
-    rescue StandardError => e
-      logger.warn("#{logo.record_type} with id: #{logo.record_id} and thumnail #{logo.name} \
-      threw an error after ActiveStorage::InvariableError.")
-      Rollbar.warn("#{logo.record_type} with id: #{logo.record_id} and thumnail #{logo.name} \
-      threw an error after ActiveStorage::InvariableError.", e)
-      nil
-    end
+    Rails.application.routes.url_helpers.rails_representation_path(logo.representation(args).processed)
```
mbarnett, line 50: The ActiveStorage API is a lot nicer here than it used to be.

db/schema.rb
```diff
@@ -19,15 +19,14 @@
 
   create_table "active_storage_attachments", force: :cascade do |t|
     t.string "name"
-    t.bigint "blob_id", null: false
     t.datetime "created_at"
     t.string "record_type"
     t.uuid "fileset_uuid"
     t.uuid "record_id"
-    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
```
murny, line 27: Should we add this index back on the new blob_id column?

mbarnett: Yes, good catch!

mbarnett: Done.

murny (approved):
LGTM 👍

PR #1916, opened 2020-10-06
bump selfsigned

pgwillia:
## Context

Addresses CVE-2020-7720 which impacts node-forge dependency.

Related to https://github.com/ualbertalib/jupiter/network/alert/yarn.lock/node-forge/open

## What's New

bumped the version in yarn.lock


murny (approved):
LGTM 👍

pgwillia:
@murny it looks like this came when we moved over to webpack but I don't see it used anywhere.  Do you know why we have it?

murny:
@pgwillia  Looks like `selfsigned` is a dependency of `webpack-dev-server` which is a devDependency of Jupiter. So this would never impact production. But worth bumping regardless

PR #1917, opened 2020-10-06
Add spring gem back to jupiter

murny:
Noticed when we were debugging Matt's Attachment issue last week we no longer have Spring in our gemfile. Spring is a default rails gem. 

This is a PR to add this back or at least start a discussion. If we decide we don't want it, let's completely remove spring from Jupiter (we still have it's binstub in jupiter, etc).

What is spring? 

> Spring is a Rails application preloader. It speeds up development by keeping your application running in the background so you don't need to boot it every time you run a test, rake task or migration.

More info: https://github.com/rails/spring

On jupiter I see around 5 seconds in savings when running rake/test/rails commands.

As an example I ran a few `time rails test` with and without spring. It seems on average with spring (when spring is warm) its around 67 seconds and without spring around 73 seconds to run the test suite.

Note:
One caveat with adding spring back is spring and simplecov do not play nicely together at all! I tried the tips simplecov suggests, but ultimately couldn't get it to work. As a result simplecov will incorrectly report very low test coverage when spring is enabled as shown:

`Coverage report generated for Minitest to /Users/e88838/Code/jupiter/coverage. 46 / 4667 LOC (0.99%) covered.`

You can always disable spring to get the real test coverage:
`DISABLE_SPRING=1  bin/rails test`

`Coverage report generated for Minitest to /Users/e88838/Code/jupiter/coverage. 2568 / 3164 LOC (81.16%) covered.`

### DIFF
--- a/bin/rails
+++ b/bin/rails
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby
+begin
+  load File.expand_path('spring', __dir__)
+rescue LoadError => e
+  raise unless e.message.include?('spring')
+end
 APP_PATH = File.expand_path('../config/application', __dir__)
 require_relative '../config/boot'
 require 'rails/commands'
--- a/Gemfile
+++ b/Gemfile
@@ -97,7 +97,11 @@ group :development do
   gem 'binding_of_caller'
 
   gem 'brakeman'
+
   gem 'listen', '>= 3.0.5', '< 3.3'
+  gem 'spring'
+  gem 'spring-watcher-listen', '~> 2.0.0'
+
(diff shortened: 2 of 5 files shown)


bin/rails
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby
+begin
```
murny, line 2: Interesting, looks like latest rails has shoved this in `boot.rb` instead:

https://github.com/rails/rails/commit/a6c958b3a7d851d6609188dd4faa5f98fa23f381

PR #1928, opened 2020-10-07
Subdomains

murny:
See https://github.com/ualbertalib/jupiter/pull/1786 for original PR and comments.

This PR wraps all current routes under an "ERA" subdomain. Which sets up the foundation for allowing multiple front doors going forward.


mbarnett (approved):
I think based on the original PR this is good to merge, now that we have somewhere for it to land

mbarnett (approved):

PR #1929, opened 2020-10-07
Solr highlighting

murny:
See https://github.com/ualbertalib/jupiter/pull/1895 for original PR and comments/feedback

#1800 

Adds solr highlighting to item description/thesis abstract fields on search results (for search, profile and collection search pages)

Highlighting in action:

Search page:
image
image

Profile page:
image

Collection show page:
image

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -8,7 +8,11 @@ class CollectionsController < ApplicationController
       base_restriction_key: Item.solr_exporter_class.solr_name_for(:member_of_paths, role: :pathing),
       value: @collection.path,
       params: params,
-      current_user: current_user
+      current_user: current_user,
+      highlight_fields: [
+        Item.solr_exporter_class.solr_name_for(:description, role: :search),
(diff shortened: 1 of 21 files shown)


app/controllers/collections_controller.rb
```diff
@@ -8,7 +8,8 @@ def show
       base_restriction_key: Item.solr_exporter_class.solr_name_for(:member_of_paths, role: :pathing),
       value: @collection.path,
       params: params,
-      current_user: current_user
+      current_user: current_user,
+      fulltext: true
```
mbarnett, line 12: Should be false for all current controllers so-as to maintain the current search experience on ERA for the time being. We'll want to start using this on the Digitization controllers.

mbarnett (changes requested):
Great Job

Looks really really good!

Only needed change is that we disable it on all of the current controllers, as we're not looking to roll this out on ERA for now, just leverage it in the Digitization work.

app/controllers/collections_controller.rb
murny: Makes sense 👍. This now has been updated.

mbarnett (approved):

PR #1930, opened 2020-10-07
Cds/resetable doi

ConnorSheremeta:
Original PR here: https://github.com/ualbertalib/jupiter/pull/1891.

Still need to finish addressing most recent feedback.


app/helpers/items_helper.rb
```diff
@@ -54,4 +54,12 @@ def humanize_date(dateish)
     6 => I18n.t('items.thesis.graduation_terms.spring'),
     11 => I18n.t('items.thesis.graduation_terms.fall')
   }.freeze
+
+  def display_admin_sidebar(item)
+    aasm_not_available = item.not_available?
+    doi_blank = item.doi.blank?
+    item_doi_initialized = item.created_at < Time.zone.now - 1.minute
+
+    item_doi_initialized && (aasm_not_available || doi_blank)
```
lagoan, line 63: Sometimes we have argued against using variables when they are just used once. What do you think of something like:

```
 def display_admin_sidebar(item)
    item_doi_initialized = item.created_at < Time.zone.now - 1.minute
    item_doi_initialized && (item.not_available? || item.doi.blank?)
 end
```
You still have ``item_doi_initialized`` but I think that can be left as it is not just a single method call

test/integration/doi_reset_test.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  setup do
+    @admin = users(:admin)
+    @user = users(:regular)
+    # A community with two collections
+    @community = communities(:books)
+    @collection1 = Collection.create!(community_id: @community.id,
+                                      title: 'Nice collection', owner_id: @admin.id)
+    @collection2 = Collection.create!(community_id: @community.id,
+                                      title: 'Another collection', owner_id: @admin.id)
```
lagoan, line 13: Would it be possible to use the collections already described on fixtures?

app/helpers/items_helper.rb
murny: That's a good point Omar 👍. 

Just to add, `Time.zone.now - 1.minute`  is equivalent to  `1.minute.ago` if you want to shorten that up further

murny (dismissed):
LGTM! Nice work @ConnorSheremeta  👍

murny (approved):

Issue #1897, opened 2020-09-28
Feature Flags


murny:
From Matt:
> we’re probably going to want some kind of feature flagging solution so that we can land stuff like this into master instead of having things sit in some long running branch

We want to look at a way to introduce Feature flags into Jupiter. What is a feature flag? Wikipedia defines this as such (https://en.wikipedia.org/wiki/Feature_toggle):

> A feature toggle (also feature switch, feature flag, feature flipper, conditional feature, etc.) is a technique in software development that attempts to provide an alternative to maintaining multiple branches in source code (known as feature branches), such that a software feature can be tested even before it is completed and ready for release. Feature toggle is used to hide, enable or disable the feature during runtime. For example, during the development process, a developer can enable the feature for testing and disable it for other users.[1] In this way, feature flags can be used to implement targeted rollouts of features to specific subsets of users, a practice known as feature gating.[2]
> 
> Continuous release and continuous deployment provide developers with rapid feedback about their coding. This requires the integration of their code changes as early as possible. Feature branches introduce a bypass to this process.[3] Feature toggles are an important technique used for the implementation of continuous delivery.
> 
> The technique allows developers to release a version of a product that has unfinished features. These unfinished features are hidden (toggled) so they do not appear in the user interface. This allows many small incremental versions of software to be delivered without the cost of constant branching and merging. Feature toggles may allow shorter software integration cycles.[4] A team working on a project can use feature toggle to speed up the process of development, that can include the incomplete code as well.

There is various solutions that exist in the ruby community:
https://github.com/jnunemaker/flipper 
https://github.com/fetlife/rollout
https://github.com/voormedia/flipflop
https://github.com/Unleash/unleash

Let's look into these solutions and find the best one that we can use for feature flags in Jupiter

murny:
# Solutions

## https://github.com/Unleash/unleash

> Unleash is a feature toggle system, that gives you a great overview of all feature toggles across all your applications and services. It comes with official client implementations for Java, Node.js, Go, Ruby, Python and .Net.
> 
> The main motivation for doing feature toggling is to decouple the process for deploying code to production and releasing new features. This helps reducing risk, and allow us to easily manage which features to enable

- More of a Feature flags as-a-service approach. Would be overkill for Jupiter as we would either need to pay or self host our own feature flag service which is would be overkill for Jupiter

## https://github.com/fetlife/rollout

> Fast feature flags based on Redis.

- Very simple global API around redis to check feature flag state

- Quite active

[...]

PR #1948, opened 2020-10-21
Feature flags

murny:
https://github.com/ualbertalib/jupiter/issues/1897

As an example of what feature flag looks like, I have put our Fulltext/highlighting searching functionality behind feature flags. So how can we control this feature via feature flags? 

New Tab in Admin called "Feature Flags":
image

Click Add Feature:
image

Name your feature flag the same as what your referring to in your code:
image

Now setup the enabling of your feature flag. 

Many options here such as enabling by actor (e.g: `User:2`), by group (e.g: Staff), by percentage of actors (e.g: A/B testing), by percentage of time (e.g: load testing), or fully enabled all the time:

image

As an example if I just want Admin users to have access to full text search for now, I can set it up as such:

image

Now when searching as an admin, I see fulltext highlighting:
image

and if user is logged out or not an admin, I see no fulltext highlighting in search results as shown:
image


app/views/layouts/admin.html.erb
```diff
@@ -23,14 +21,10 @@
         <%= active_link_to t('admin.items.index.header'), admin_items_path, class: 'nav-link', active: :exclusive %>
       </li>
       <li class="nav-item">
-        <%= active_link_to "(TODO) #{t('admin.pages.index.header')}", '#', class: 'nav-link', active: :exclusive %>
```
murny, line 26: See no use of leaving these "under construction" links exposed on the admin dashboard. If we ever build them, then we can show it. So removing it for now

config/initializers/sidekiq.rb
```diff
@@ -8,3 +8,6 @@
 Sidekiq.configure_client do |config|
   config.redis = { url: Rails.application.secrets.redis_url }
 end
+
+require 'sidekiq/web'
+Sidekiq::Web.app_url = '/admin'
```
murny, line 13: Kind of cool thing I learned. If you add this configuration, then this exposes a button in the sidekiq dashboard that "takes you back to the App" as shown here:

image

test/controllers/search_controller_test.rb
```diff
@@ -62,12 +62,22 @@ class SearchControllerTest < ActionDispatch::IntegrationTest
...
-  test 'should NOT render highlights on search result page' do
+  test 'should NOT render highlights on search result page by default' do
     get search_url, params: { search: 'French' }
     assert_response :success
 
     assert_no_match(/<mark>French<\/mark>/, response.body)
   end
 
+  test 'should render highlights on search result page when feature flag is on' do
+    Flipper.enable(:fulltext_search)
```
murny, line 73: Example of easy way to turn on/off feature flags for testing

Gemfile
```diff
@@ -33,13 +33,14 @@ gem 'bcrypt', '>= 3.1.13'
...
 gem 'omniauth-saml'
+
 # Authorization
 gem 'pundit', '1.1.0'
 
 # Background tasks
 gem 'sidekiq', '~> 5.2'
 gem 'sidekiq-unique-jobs'
-gem 'sinatra', '~> 2.1.0' # used by sidekiq/web
+gem 'sinatra', '~> 2.0.8.1' # used by sidekiq/web
```
murny, line 43: Had to backport sinatra for flipper-ui as they have locked `rack-protection` gem to `< 2.1.0`

app/controllers/collections_controller.rb
```diff
@@ -9,7 +9,7 @@ def show
       value: @collection.path,
       params: params,
       current_user: current_user,
-      fulltext: false
+      fulltext: Flipper.enabled?(:fulltext_search, current_user)
```
murny, line 12: If you want the control of enabling feature flags by "actor" or "group" you need to pass in the `current_user` to this flipper enabled call.

app/models/user.rb
```diff
@@ -63,4 +63,8 @@ def id_as_hash
     Digest::SHA2.hexdigest("#{Rails.application.secrets.secret_key_base}_#{id}")
   end
 
+  def flipper_id
```
murny, line 66: This method is used for comparing against the actor list. So if I want to turn a feature flag on just for a single actor (admin@ualberta.ca for example), then I would add this ID to the list (for admin@ualberta.ca, they have an ID of 2, so we would add `User:2` to the list). Now this feature flag is turned on only for the users in the list.

config/initializers/flipper.rb
```diff
@@ -0,0 +1,29 @@
...
+end
+
+Flipper::UI.configure do |config|
+  # When no feature flags, turn off random video
+  config.fun = false
+end
+
+unless Rails.env.test?
+  require 'flipper/middleware/memoizer'
+  Rails.application.config.middleware.use Flipper::Middleware::Memoizer
```
murny, line 28: Seems like this works pretty well. It will memoize the first call to a feature flag in the request and any new calls to the same feature flag will use that memorized value instead of hitting the DB again. 

As an example if I was listing out N number of posts, and each post was checking a feature flag, with this enabled, we would only make one trip to the DB. The first post feature flag check will get memorized and the rest of the posts feature flag checks will use the memorized value.

You could maybe argue this doesn't actually vastly increase performance considering ActiveRecord will "cache" repeated calls to the database. But it helps us be explicit and will clean up our logs with the AR calls, so doesn't hurt for us to enable this. Also if we ever move our feature flag storage to redis then this would definitely be a nice to have.

mbarnett (approved):

PR #1950, opened 2020-10-23
Bring in rubocop-minitest gem and lint and fix any violations in our tests

murny:
Learned about `rubocop-minitest` gem while reading Rubocop 1.0 announcement. 

Seemed pretty cool as I always get things wrong with minitest's API (for example assert args order.... is it `assert(expected, actual)` or `assert(actual, expected)` 🤔 ?) . Now this gem will autocorrect it for you!

Gem github and style guide is here:

- https://github.com/rubocop-hq/rubocop-minitest
- https://docs.rubocop.org/rubocop-minitest/

### DIFF
--- a/test/helpers/page_layout_helper_test.rb
+++ b/test/helpers/page_layout_helper_test.rb
@@ -97,16 +97,11 @@ class PageLayoutHelperTest < ActionView::TestCase
 
   # thumbnail_path
 
-  def mupdf_exists?
-    ActiveStorage::Previewer::MuPDFPreviewer.mutool_exists?
-  end
-
-  def poppler_exists?
-    ActiveStorage::Previewer::PopplerPDFPreviewer.pdftoppm_exists?
-  end
-
   test 'thumbnail_path should return preview for pdf (Invariable but Previewable)' do
-    skip 'this test requires the dependency MuPDF or Poppler to be installed' unless mupdf_exists? || poppler_exists?
+    unless ActiveStorage::Previewer::PopplerPDFPreviewer.pdftoppm_exists? ||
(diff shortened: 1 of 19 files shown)


test/models/community_test.rb
```diff
@@ -25,28 +25,28 @@ class CommunityTest < ActiveSupport::TestCase
...
+    assert_equal('GxpIjJsC4KnRoBKNjWnkJA==', community.logo.blob.checksum)
   end
 
   test 'an updated logo replaces the old one' do
     community = communities(:books)
     community.logo.attach io: File.open(file_fixture('image-sample.jpeg')),
                           filename: 'sample1.jpeg', content_type: 'image/jpeg'
 
-    assert_equal community.logo.filename, 'sample1.jpeg'
+    assert_equal('sample1.jpeg', community.logo.filename.to_s)
```
murny, line 45: Interesting side effect of confusing actual vs expected args. Looks like passing in an object into "actual" will `.to_s` it, which makes sense 🤔 .

test/helpers/page_layout_helper_test.rb
```diff
@@ -97,16 +97,11 @@ def base_url
 
   # thumbnail_path
 
-  def mupdf_exists?
```
murny, line 100: rubocop-minitest didn't like having non-test methods inside the test class (wanted to `test_` prefix them). Suppose we could have defined these helper methods outside of the class to get around this.

Alternatively since we only use these methods once, I just moved them to where they being used below

lagoan (approved):
Nice code cleanup!
LGTM! :+1:

mbarnett (approved):
I like it!

PR #1961, opened 2020-10-27
UAT VIRTUAL_HOSTS configuration on just the containers that need it

pgwillia:
## Context

Sometimes when containers were updated by watchtower the VIRTUAL_HOSTS values in the .env_deployment files would apply to several containers.  This mean that sometimes it would try to route to the web container directly instead of nginx (and not work).

There is still a gotcha with solr.era.uat.library.ualberta.ca being redirected to https by HSTS.  To get around this try visiting in a private window or "Clear Cookies and Site Data".

## What's New

apply [VIRTUAL_HOSTS configuration](https://github.com/nginx-proxy/nginx-proxy#multiple-ports) on just the containers that need it.

### DIFF
--- a/.env_deployment_sample
+++ b/.env_deployment_sample
@@ -21,7 +21,3 @@ POSTGRES_USER=jupiter
 # Nginx environment variables
 HOSTNAME=localhost
 SKYLIGHT_AUTHENTICATION=secretauthenticationtoken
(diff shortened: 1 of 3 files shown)


murny (approved):
LGTM 👍

PR #1968, opened 2020-11-05
Add a few updates to Github Actions

murny:
From reviewing Connors Github Action PR on oaisys(https://github.com/ualbertalib/oaisys/pull/49) I saw a couple things we can update here in Jupiter


.github/workflows/push.yml
```diff
@@ -30,7 +26,6 @@ jobs:
 
       - name: Bundle gems
         run: |
-          gem install bundler
```
murny, line 33: Bundler already gets installed from the Ruby Action:
image

.github/workflows/push.yml
```diff
@@ -16,12 +16,8 @@ jobs:
         with:
           ruby-version: 2.6.6
 
-      - name: Install PostgreSQL client
```
murny, line 19: Shouldn't be needed its a default package on ubuntu

pgwillia (approved):
👍 LGTM!

Issue #1769, opened 2020-07-21
Peel Redirects


mbarnett:
# Peel Redirects

- [x] Stub out models identified with bare minimum of attributes.
- [x] Stub out redirect controller.
- [x] routes
- [x] seeds & test data
- [x] redirects and tests

## TLDR;

We will need the following redirects

to `/books/*`
```
http://peel.library.ualberta.ca/bibliography/*.html
http://peel.library.ualberta.ca/bibliography/*/*.html
http://peel.library.ualberta.ca/bibliography/*/*/*.html
http://peel.library.ualberta.ca/bibliography/*/*/*/*.html
```

Then the controller might do something like this:
```
@book = if Book.find_by(peel_id: 3693, run: 2, part_number: 1).present?
  Book.find_by(peel_id: 3693, run: 2, part_number: 1)
elsif Book.find_by(peel_id: 3693, part_number: 1).present?
   Book.find_by(peel_id: 3693, part_number: 1)
else
  Book.find_by(peel_id: 3693)
end
```

to `/newspapers/*`
```
http://peel.library.ualberta.ca/newspapers/*/*/*/*/
http://peel.library.ualberta.ca/newspapers/*/*/*/*/*/
```

Then the controller might do something like this:
```
@newspaper= Newspaper.find_by(publication_code: 'LSV', year: '1967', month: '03', day: '29')
```

to `/images/*`
```
http://peel.library.ualberta.ca/magee/*.html
http://peel.library.ualberta.ca/postcards/*.html
```
Then the controller might do something like this:
```
@image= Image.find_by(peel_image_id: 'MGNGBG0001')
```

to `/maps/*`
```
http://peel.library.ualberta.ca/maps/*.html
```
Then the controller might do something like this:
```
@map= Map.find_by(map_number: 'M000230')
```

# Analysis

http://peel.library.ualberta.ca/

## Books, Pamphlets and Other Materials (for simplicity we'll call the smallest individual unit of these things a `Book`)
/bibliography/`:peel_number`.html
i.e. bibliography/4542.html is just metadata about an item
i.e. bibliography/4062.html is a book

`peel_number`  is the item number
![image](https://user-images.githubusercontent.com/1220762/94199591-4a0a3380-fe76-11ea-9d2d-b9b606d62700.png)
from [Peel's Bibliography of the Prairie Provinces](https://www-deslibris-ca.login.ezproxy.library.ualberta.ca/ID/418355)

### Henderson’s 
[...]

PR #1982, opened 2020-11-11
Peel redirects

pgwillia:
## Context

We will be moving Peel Prairie Province content (Folk Fest Programs first!) into the digitization subdomain.  This will allow us to redirect the old urls to our new structure.  

Not included in this PR is complex or voluminous seeds data.  I think that might be better suited to when we flesh out the book, newspaper, image and maps models with the full compliment of metadata and attachments.

Related to #1769

## What's New

* Stub out models identified with bare minimum of attributes (basically just the old identifiers).
* Stub out redirect controller.
* routes
* redirects and tests


db/schema.rb
```diff
@@ -90,7 +90,6 @@
   end
```
pgwillia, line 91: The changes to this file are unrelated. Needs to be refined.

app/controllers/digitization/books_controller.rb
```diff
@@ -0,0 +1,20 @@
+class Digitization::BooksController < ApplicationController
+
+  before_action :set_digitization_book, only: [:show]
```
murny, line 3: Not sure about the future, but if it's just show pages we have, then I'd throw everything into the show action instead of the before action/etc

e.g: 
```
def show
    @digitization_book = Digitization::Book.find(params[:id])
    authorize @digitization_book
end
```

But if the future is to add other things like Editing/Updating/etc, then this is good 👍

config/routes.rb
```diff
@@ -154,4 +154,22 @@
...
+    get '/newspapers/:publication_code/:year/:month/:day(/*page)', to: 'digitization/redirect#peel_newspaper'
+
+    get '/magee/:peel_image_id', to: 'digitization/redirect#peel_image'
+    get '/postcards/:peel_image_id', to: 'digitization/redirect#peel_image'
+
+    # this pattern is shared with jupiter maps so must be before the other inorder to correctly redirect
+    get '/maps/:peel_map_id', to: 'digitization/redirect#peel_map', peel_map_id: /M[0-9]{6}/
+
+    scope module: 'digitization', only: [:index, :show] do
+      resources :books, :newspapers, :images, :maps
```
murny, line 172: TIL you can compound multiple resources like this. Also using the `only:` on the scope limits all the resources within it. Very cool 👍

murny (dismissed):
All looks really good to me so far 👍 . Just some minor comments on a few things I saw in the code

app/views/digitization/books/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<p id="notice"><%= notice %></p>
```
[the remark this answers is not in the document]
pgwillia: This was pretty close to the initial rails scaffolding.  I can remove the notice part.  The current state of the UI is pretty rough.  I was focused on the redirect behaviour with the hope that this will be more fully built out when we know about the full compliament of metadata.

config/routes.rb
```diff
@@ -154,4 +154,22 @@
...
+
+    get '/newspapers/:publication_code/:year/:month/:day(/*page)', to: 'digitization/redirect#peel_newspaper'
+
+    get '/magee/:peel_image_id', to: 'digitization/redirect#peel_image'
+    get '/postcards/:peel_image_id', to: 'digitization/redirect#peel_image'
+
+    # this pattern is shared with jupiter maps so must be before the other inorder to correctly redirect
+    get '/maps/:peel_map_id', to: 'digitization/redirect#peel_map', peel_map_id: /M[0-9]{6}/
+
+    scope module: 'digitization', only: [:index, :show] do
```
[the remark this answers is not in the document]
pgwillia: I like this suggestion.  I originally had this as a namespace but decided to move to the scope to remove `digitization` from the path.  That had the side effect of changing the path and url helpers.  Your suggestion fixes that. Thanks!

db/migrate/20201030205050_create_digitization_books.rb
```diff
@@ -0,0 +1,12 @@
+class CreateDigitizationBooks < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_books do |t|
+      t.uuid :book_id
+      t.integer :peel_id
+      t.integer :run
```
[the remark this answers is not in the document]
pgwillia: I was wondering the same thing.  `peel_id`, `run` and `part_number` ending up being a complex key where any (or all) parts might be null.  Any suggestions on how to do that?  Adding an index on the database was something I was thinking of doing.

app/controllers/digitization/books_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_digitization_book
+    @digitization_book = Digitization::Book.find(params[:id])
+    authorize @digitization_book
+  end
+
+  # Only allow a list of trusted parameters through.
+  def digitization_book_params
+    params.require(:digitization_book).permit(:peel_id, :run, :part_number)
```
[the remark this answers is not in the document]
pgwillia: This was from the rails scaffolding.  Seemed useful so I left it in, but as you observed not being used.  I'll move them to pundit.

app/controllers/digitization/books_controller.rb
pgwillia: Again rails scaffolding.  I removed those methods because they are not yet, but I imagine someone is going to want the C[R]UD in the future. I did leave in the before_action hook just in case.

app/controllers/digitization/redirect_controller.rb
```diff
@@ -0,0 +1,45 @@
+class Digitization::RedirectController < ApplicationController
+
+  skip_after_action :verify_authorized
+
+  def peel_book
+    @book = if Digitization::Book.find_by(
```
murny, line 6: Yeah it be nice if we can use some SQL magic here to reduce this to one query or something? Just gotta figure out how to do this. This isn't terrible...but worse case here we doing at 3 seperate queries to the DB when we can maybe do one and clean up all the if logic.

It's almost like we want something like ORs or Unions clauses to do the same thing:

```
  Book.where(peel, run, part)
    .or(Book.where(peel, part))
    .or(Book.where(peel)).first
```
Then its just one query (obviously needs lots of testing to make sure its doing the same thing).

db/migrate/20201030205050_create_digitization_books.rb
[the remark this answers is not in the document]
murny: Ah okay, so book will have these `null: true` what about the other tables though? I added a comment on the controller for maybe cleaning up the logic for the books redirect.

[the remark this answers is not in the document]
pgwillia: I just peeked at the schema to see what we did for items and thesis.  We're using the [postgres extension to create uuids](https://www.postgresql.org/docs/9.4/uuid-ossp.html) like this:
https://github.com/ualbertalib/jupiter/blob/7ba41df2ca3e3e2a1fc5e0cadf1d73f1fa45a784/db/schema.rb#L192
We probably want to do the same for these types, right @mbarnett?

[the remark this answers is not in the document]
mbarnett: I don't believe you have to do that with current Rails & Postgres 11 – I think it happens automatically with `create_table ... id: :uuid` . When I did that it was to work around Postgres 9 iirc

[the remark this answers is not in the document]
pgwillia: If I don't specify it gives me `default: -> { "gen_random_uuid()"`  is that the same or close enough?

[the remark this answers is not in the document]
pgwillia: I guess this whole conversation presumes that we're changing the identifiers when we move the Peel content.  Is that a safe assumption? Books, newspapers, images and maps?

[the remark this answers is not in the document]
pgwillia: I guess I'm also presuming that we will eventually have content that doesn't come with the Peel style identifiers as well.  Is that a safe assumption?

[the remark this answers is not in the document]
mbarnett: Yes, we're minting new IDs for these, not retaining the current ones

db/schema.rb
```diff
@@ -89,6 +89,39 @@
     t.index ["owner_id"], name: "index_communities_on_owner_id"
   end
 
+  create_table "digitization_books", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
murny, line 92: Seems like schema is out of date now (missing your new migration changes)

app/models/digitization/book.rb
```diff
@@ -0,0 +1,2 @@
+class Digitization::Book < ApplicationRecord
```
murny, line 1: Only thing I see missing now is Model validations on the columns you just added `null: false` to for your models. 

Then adding tests for checking that validations are working.

Then all looks good to me 👍

config/routes.rb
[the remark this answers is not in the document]
pgwillia: Added the namesapace and updated the url and path helper calls.

app/views/digitization/books/show.html.erb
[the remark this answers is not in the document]
pgwillia: Removed the notice.

app/views/digitization/books/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<p>
```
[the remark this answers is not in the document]
pgwillia: Used the description list html elements.  Thanks for the suggestion!  I expect that this will change and improve when we work up the whole complement of metadata.

app/controllers/digitization/redirect_controller.rb
pgwillia: Was able to reduce this to one query by refactoring the order of the routes.

db/schema.rb
pgwillia: I'm not sure what you mean?  This reflects `id: :uuid` and the indexes I added.

app/models/digitization/book.rb
pgwillia: I added `null: true` is that what you mean?  I can try to validate uniqueness of those keys.  Is that what you mean?

db/schema.rb
murny: Sorry, highlighting a bad line here. But below this you can see `t.intger "peel_id"` among other columns. But in our migrations we added this as `null: false` now.

So this should really be `t.intger "peel_id" null: false` now.

Really simple just have to rerun your migrations and this schema will get updated (might have to drop table and have them run again as you edited the migrations since you ran them)

app/models/digitization/book.rb
murny: For all the models you have, you need to add presence checks for the columns you `null: false` and a validations on the uniqueness indexes. 

Why? Well if you don't and a user submits bad data on the form (like no `peel_id` even though its null: false), the database will bubble up an error and we will return probably a  server error (500) with a bad stack trace from postgres saying column cannot be null. By adding these validations on the Rails models we get a nice set of errors back which we can present this to the user and let them fix their own mistakes.

Then when we add these validations...we should test them with model tests to make sure everything is working correctly. And these will probably be very simple tests like:

```
  validates :username, presence: true
end

class UserTest < ActiveSupport::TestCase
  def setup
    @user = users(:user)
  end

  test 'valid user' do
    assert @user.valid?
  end

  test 'invalid without username' do
    @user.username = nil
    assert_not @user.valid?
    assert_equal("can't be blank", @user.errors[:username].first)
  end
end
```

pgwillia: My understanding is that these ids might be null (hence the `null: true`).  The uuid should be created as Matt said, "we're minting new IDs for these, not retaining the current ones".  The ids **may** be there and we can redirect those.  Others may not and I guess they should 404.

I'm thinking that the digitization content will be a Venn Diagram where only a portion of the content will be from Peel Prairie Provinces.

murny: For uniqueness you want something similar like:

```
class Identity < ApplicationModel
  validates :uid, presence: true, uniqueness: { scope: :provider }
end

class IdentityTest < ActiveSupport::TestCase
  def setup
    @identity = identities(:user_twitter)
  end

  test 'valid identity' do
    assert @identity.valid?
  end

  test 'invalid if uid already taken for same provider' do
    @identity.uid = '1122334455'
    assert_not @identity.valid?
    assert_equal('has already been taken', @identity.errors[:uid].first)
  end
end
```

murny: My bad @pgwillia! 😆  I read your migrations as `null: false` instead of `null: true` 😓 ! 

Above example of testing a unique index might still be valid though!

db/schema.rb
murny: Ignore this, I can't read today 😆 (read null: true as null: false)

app/controllers/digitization/books_controller.rb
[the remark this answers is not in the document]
pgwillia: I changed my mind and removed the abstractions.  If this comes up again (and it probably will) I'll be sure to look to Pundit's strong_parameters.

app/controllers/digitization/books_controller.rb
pgwillia: I changed my mind and removed the abstraction.  Simpler is better for readability.

app/models/digitization/newspaper.rb
```diff
@@ -0,0 +1,6 @@
+class Digitization::Newspaper < ApplicationRecord
+
+  validates :publication_code, uniqueness: { scope: [:year, :month, :day] }
+  validates :year, :month, :day, presence: true, if: :publication_code?
```
pgwillia, line 4: I started going down the path of adding even more specific validation but not sure if it has added value
```suggestion
 validates :publication_code, uniqueness: { scope: [:year, :month, :day] }, format: { with: /[A-Z]{3}/, message: "%{value} should be three letters that represent the newspaper title" }
  validates :year, :month, :day, presence: true, if: :publication_code?
  validates :year, format: { with: /[0-9]{4}/, message: "%{value} should be a four digit year of publication"}
  validates :month, format: { with: /0[1-9]|1[0-2]/, message: "%{value} should be a two digit (zero padded) month"}
  validates :day, format: { with: /0[1-9]|[12]\d|3[01]/, message: "%{value} should be a two digit (zero padded) day of the month"}
```

See also https://ualbertalib-di.slack.com/archives/CSEHP0L0K/p1606408886290400

app/models/digitization/book.rb
pgwillia: I think I've added the Model validations that make the most sense (and tests).  Let me know if you think I'm missing any.

test/models/digitization/newspaper_test.rb
```diff
@@ -1,7 +1,30 @@
...
+  setup do
+    @newspaper = digitization_newspapers(:la_survivance)
+  end
+
+  test 'valid Peel newspaper' do
+    assert @newspaper.valid?
+  end
+
+  test 'unique Peel newspaper' do
+    newspaper = Digitization::Newspaper.create(publication_code: 'LSV',
```
murny, line 14: For these uniqueness tests, if your seeding data like this, make sure you clean up the data after your finished with it (`newspaper.destroy` at the end of the test). 

Or alternatively move this kind of data creation to fixtures (which will cleanup after your done for you)

test/models/digitization/newspaper_test.rb
```diff
@@ -1,7 +1,30 @@
...
+                                               year: '1967',
+                                               month: '03',
+                                               day: '29')
+    assert_not newspaper.valid?
+    assert_equal('has already been taken', newspaper.errors[:publication_code].first)
+  end
+
+  test 'invalid Peel newspaper' do
+    newspaper = Digitization::Newspaper.create(publication_code: 'LSV')
```
murny, line 23: Again you can just reuse the fixture you currently have. Instead of creating a brand new one, just edit the one you have.

```
@newspaper. assign_attributes(year: nil, month: nil, day: nil). # We don't want to save the data quite yet
assert_not @newspaper.valid?  # this will call the validations on @newspaper object
assert_equal("can't be blank", @newspaper.errors[:year].first)
assert_equal("can't be blank", @newspaper.errors[:month].first)
assert_equal("can't be blank", @newspaper.errors[:day].first)
```

app/models/digitization/newspaper.rb
mbarnett: My initial thought is that there isn't a ton of value because whatever the data in Peel is is right by definition, even if we can't construct a valid regex around it (if there's some one-off non-conforming issue or w/e)?

app/models/digitization/book.rb
murny: Yeah looks really good! Just a small comment around data creation

mbarnett (commented):
I'll leave Shane to finish up any reviews around the tests/fixtures, but this is really good, clean work Tricia 🎉

test/models/digitization/newspaper_test.rb
pgwillia: I tried to create duplicate fixtures as an alternate solution but got a database error because of the unique index.  So creating it is!  Is that better/worse/the same as taking another unrelated fixture and using the assign_attributes trick?

test/models/digitization/newspaper_test.rb
pgwillia: I was able to reuse some fixtures for the validity tests rather than creating new.  👍

test/models/digitization/newspaper_test.rb
murny: Sorry should of describe this better with the fixtures

What you could do is create an additional fixture for newspapers (so you have 2)

```
la_survivance:
  publication_code: 'LSV'
  year: '1967'
  month: '03'
  day: '29' 

edmonton_journal:
  publication_code: 'RANDOM'
  year: '1970'
  month: '03'
  day: '29' 
```

Then you can use your @newspaper variable which has been assigned a fixture.

Then try to update @newspaper so it triggers the uniqueness constraint

```
@newspaper.publication_code = 'RANDOM'
@newspaper.valid?
assert_equal('has already been taken', newspaper.errors[:publication_code].first)
```

Which gets you the same test coverage as before just without creating additional data

test/models/digitization/image_test.rb
```diff
@@ -0,0 +1,19 @@
...
+  setup do
+    @image = digitization_images(:magee)
+  end
+
+  test 'valid Peel image' do
+    assert @image.valid?
+  end
+
+  test 'unique Peel image' do
+    @image = Digitization::Image.create(peel_image_id: 'PC015716')
```
murny, line 14: Forgot one here 😄

murny (approved):
LGTM 👍 . Nice work!

gif

PR #1993, opened 2020-11-17
add erd diagram

ConnorSheremeta:
Added ERD diagram to assist myself and others


pgwillia:
👏 This is great idea especially as the application gets more complicated.  One suggestion is to use something that's more readily editable for the team.  I've used [draw.io as a embedded svg](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio) in github readmes before for example.  My other concern is how we'll keep this bit of documentation from falling out of date?

ConnorSheremeta:
>  This is great idea especially as the application gets more complicated. One suggestion is to use something that's more readily editable for the team. I've used [draw.io as a embedded svg](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio) in github readmes before for example. My other concern is how we'll keep this bit of documentation from falling out of date?

I used the gem 'rails-erd' to generate this and it's output is in PDF, it might be possible to have it in an editable format but I don't see that as something we need at the moment. We could probably automate this process and regenerate it for every PR (or maybe just when the models are touched) so it is always up to date.

murny:
Yeah my only concern is keeping this updated as well. I assume you used the https://github.com/voormedia/rails-erd gem. So maybe doesn't hurt to add some instructions on how someone can regenerate this ERD somewhere (ReadMe or new markdown file within Docs folder) 🤔

docs/erd-instructions.md
```diff
@@ -0,0 +1,12 @@
+How to generate new ERD diagram
+---------------
+
+See the [installation instructions](https://voormedia.github.io/rails-erd/install.html) for a complete description of how to install Rails ERD. Here's a summary:
+
+* Install Graphviz 2.22+ ([how?](https://voormedia.github.io/rails-erd/install.html)). On macOS with Homebrew run `brew install graphviz`.
+
+* Add <tt>gem 'rails-erd', group: :development</tt> to your application's Gemfile
```
murny, line 8: I'd keep it as a global gem (instead of application gem). Otherwise your gemfile will be dirty and we might accidentally check this in.

```suggestion
* Add 'rails-erd' gem  to your machine via `gem install 'rails-erd'`
```

🤷

murny (approved):
👍

Issue #1915, opened 2020-10-06
Resolve flapping announcement tests


ConnorSheremeta:
[These tests](https://github.com/ualbertalib/jupiter/blob/integration_postmigration/test/integration/read_only_mode_integration_test.rb) are currently flapping because there are some other announcements that are sometimes created in previous tests. We need to clear all announcements in a setup block.

pgwillia:
Is this [one](https://github.com/ualbertalib/jupiter/pull/2004/checks?check_run_id=1487817386) or is this related?
```
Run options: --seed 51227

# Running:

Capybara starting Puma...
* Version 5.0.4 , codename: Spoony Bard
* Min threads: 0, max threads: 4
* Listening on http://127.0.0.1:37135
......[Screenshot]: /home/runner/work/jupiter/jupiter/tmp/screenshots/failures_test_should_be_able_to_login_as_a_regular_user.png
F

Failure:
AdminUsersShowTest#test_should_be_able_to_login_as_a_regular_user [/home/runner/work/jupiter/jupiter/test/system/admin_users_show_test.rb:119]:
expected to find visible css "h1" with text "Users" but there were no matches. Also found "Dashboard", which matched the selector but not all filters. 

rails test test/system/admin_users_show_test.rb:107

.......................S............

Finished in 116.591905s, 0.3688 runs/s, 3.9282 assertions/s.
43 runs, 458 assertions, 1 failures, 0 errors, 1 skips
```

ConnorSheremeta:
Hmm doesn't look to be related at first glance. I'm curious now what the the capybara screenshot looks like now, might show what happened!

PR #1998, opened 2020-11-26
Clear announcements before read only mode tests

ConnorSheremeta:
#1915

There were leftover announcements causing the read only mode tests to fail


test/integration/read_only_mode_integration_test.rb
```diff
@@ -4,6 +4,10 @@
 
 class ReadOnlyModeIntegrationTest < ActionDispatch::IntegrationTest
 
+  setup do
+    Announcement.destroy_all
```
murny, line 8: Did you find where we are creating these announcements? And destroying them where they are being created instead (or switching them to use fixtures)?

murny: Or does ReadOnly stuff assume there's no current announcement? and this is causing an issue 🤔

murny: Maybe ignore me 😆. It's probably these tests that are creating the announcements and depending on the order they run, could cause issues. So this code setup code makes sense 👍

murny (approved):
👍

test/integration/read_only_mode_integration_test.rb
ConnorSheremeta: I came to the assumption that an announcement was somehow lingering in the test db (not due to another test) because ReadOnlyModeIntegrationTest#test_read_only_mode_creates_and_clears_announcement was consistently failing when this issue was discovered. The ReadOnly tests assume there are no current announcements but it is still functional if there are other announcements.

murny: We will always have 3 announcements in the database from fixtures. One fixture is setup as a "Current" announcement, but this is fine as you can have multiple announcements in Jupiter at the same time. 

All the readonly tests I've seen cleanup after each other (we always enabling...and then disabling). Announcements tests all seem to clean up after themselves too. 

So this is kinda odd. Even if there's multiple announcements on the page, your assert_select 'div.alert' is looking for the Read Only Mode announcement text, so this should be unhindered if there's multiple announcements on the page (from fixtures or elsewhere) 🤔. 

Only thing I can maybe thing is, Rake tasks takes time, but we moving to the next page faster then the rake task can finish and sometimes as a result when the page loads, announcement is still shown or not shown. So we can merge this as is. But it might flake in the future, then you know why?

PR #2005, opened 2020-11-30
Fixed read only mode feature

ConnorSheremeta:
Fixed read only mode feature by adding system account if needed

### DIFF
--- a/db/migrate/20201130183659_add_system_account_if_needed.rb
+++ b/db/migrate/20201130183659_add_system_account_if_needed.rb
@@ -1,0 +1,8 @@
+class AddSystemAccountIfNeeded < ActiveRecord::Migration[5.2]
+  def change
+    User.find_or_create_by(email: 'ditech@ualberta.ca') do |user|
+      user.name ||= 'System user'
+      user.admin = false
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+## [1.2.25] – 2020-11-30
+
+### Fixed
+- Fixed read only mode feature by adding system account if needed [PR#2005](https://github.com/ualbertalib/jupiter/pull/2005)
+
 ## [1.2.24] – 2020-10-13
 
 ### Added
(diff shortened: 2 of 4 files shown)


db/migrate/20201130183659_add_system_account_if_needed.rb
```diff
@@ -0,0 +1,8 @@
+class AddSystemAccountIfNeeded < ActiveRecord::Migration[5.2]
+  def change
+    User.find_or_create_by(email: 'ditech@ualberta.ca') do |user|
+      user.name ||= 'System user'
+      user.admin = false
```
murny, line 5: This isn't a complete 1 to 1 port of the code that's currently in master. Any reason why to leave out certain things (like setting system api key and setting system boolean to true, etc)?

https://github.com/ualbertalib/jupiter/blob/master/db/migrate/20200305210657_add_system_account_if_needed.rb

Is it just cause this branch is a bit behind and missing some of this code (system boolean/api key/etc)? 

Also we will have some wrestling to do when we get production on master. For example, If we run this migration in production as is, then any migration that has been recently added in Master or Integration branches will no longer run (as this migration timestamp is as of today and master/integration new migrations are older, so Rails will think its migrations are up to date.) Easy solution is to back date this migration to an older date which allows master/integration migrations to continue to run when the code is finally brought up to date.

Do we have plans on merging this `old-master-fedora` branch back into Master in the future or just let it die? (if we merging it, then we will have two of these migrations with the exact same thing, doing similar things sitting around which is probably not ideal). 

Finally when we do get production on Master, our system user could be missing the system boolean and api key generated for this from the old migration and we need to remember to add these things onto the system user (this could be resolved by backdating this migration to be older than the one in master. So both get ran. That is if it's okay to have two migrations that are pretty much the same)

But you might have a plan for all this and if so ignore me 😄

mbarnett: Those columns (system api key, and system boolean) don't exist on this branch. So this is kind of the minimal version of what happens over on master, that won't interfere with master, without having to back port a bunch of that API work just to create this account.

I don't really plan on porting this over to master? It just gets current production to a known state that master can then continue forward from without breaking, without having to do a bunch of backporting from master->old master and then old master -> master

murny: Sounds good 👍 . 

Then for timestamps of the migration, seems like master has none, so no conflict here. And integration just has flipper tables. So we have some options to handle that.

murny (approved):

Issue #2009, opened 2020-12-03
Current era-app-prd issues


ConnorSheremeta:
Making one issue to put all the issues I have found from testing era-app-prd.

## 1

- [x]  resolved

Clamby dependencies are  not installed. (will be able to test deposit process once this is done)

## 2

- [x]  resolved

On deletion of an item through the UI I get this notification:
![Screenshot from 2020-12-03 12-15-17](https://user-images.githubusercontent.com/14242562/101077286-de250500-3561-11eb-8d1a-8c4541f1cb92.png)
It looks like the item successfully deletes however.

**Note: this doesn't appear to be reproducible locally, I get the message 'Item was successfully deleted!'... This _may_ have to do with clamav acting up**
## 3

- [x]  resolved

The supervisor and department facets do not appear on the left sidebar. When either a supervisor or department is clicked from an item it appears to just return all items in the search.

Here is the sidebar on current prod:
![Screenshot from 2020-12-03 12-18-51](https://user-images.githubusercontent.com/14242562/101077259-d6656080-3561-11eb-8839-649fe5f23c8d.png)

## 4

- [x]  resolved

It appears that I cannot remove logo from a community. When clicking the 'Remove logo' checkbox and then 'Update Community' the old logo appears to persist. I would assume it would show the standard ERA logo. Setting a new logo with a file works as expected.

**Note: this is reproducible locally. The error I get is 'found unpermitted parameter: :remove_logo' and was able to fix it by making it a permitted param. Now to find the best way to do this.**

## 5

- [x]  resolved

When sorting items by date (either oldest first or newest first) from results you get an error when you switch to either the collections or communities tab:

F, [2020-12-08T11:45:43.220514 #653] FATAL -- : [50d6ad2b-cb3b-429e-8b50-300311d5df4e]   
[50d6ad2b-cb3b-429e-8b50-300311d5df4e] NameError (uninitialized constant JupiterCore::SolrServices::DeferredFacetedSolrQuery::NameManglingError
Did you mean?  JupiterCore::SolrServices::NameManglingError
               JupiterCore::SolrNameManglingError
               JupiterCore::SolrServices::NameMangling):

Looks like this is due to the param 'sort' being 'sort_year' where communities and collections expect it to be 'record_created_at'? It looks to rescue an error that doesn't exist as well.

## 6

- [x]  resolved

[...]

PR #2013, opened 2020-12-04
Add Department and Supervisor to Thesis facets

mbarnett:
## Context

This turned up in testing on new Production environment. Neither field was being indexed for faceting (I must have missed this in the original port off of Fedora)

Related to #2009

## What's New

Marked fields as having the `:facet` role in Theses' SolrExporter


murny (approved):
👍

PR #2017, opened 2020-12-07
Clear logo_id from item/thesis before destroying it

ConnorSheremeta:
2 on #2009

Clear logo_id from item/thesis before destroying it. Error we were getting:

  ActiveStorage::Attachment Destroy (12.4ms)  DELETE FROM "active_storage_attachments" WHERE "active_storage_attachments"."id" = $1  [["id", 225225]]
   (1.6ms)  ROLLBACK
Traceback (most recent call last):
        1: from (irb):8
ActiveRecord::InvalidForeignKey (PG::ForeignKeyViolation: ERROR:  update or delete on table "active_storage_attachments" violates foreign key constraint "fk_rails_63f1378413" on table "theses")
DETAIL:  Key (id)=(225225) is still referenced from table "theses".


mbarnett (approved):
We're going forward with this instead of some more automated ways of enforcing the constraint, temporarily, until we can revist the overall implementation of the logo relation. Just FYI to anyone looking at this

PR #2018, opened 2020-12-08
Fix reindexing task accidental no-op

mbarnett:
## Context

When we tried to reindex Theses on New Prod, nothing happened (and happened very quickly!). Effectively, we tried to rely on re-saving the objects to update the Solr index for them, but Rails is smart enough to know that the objects aren't dirty, and no-ops the save call.

## What's New

Directly call `update_solr` rather than try to rely on it being called-back from the `save` in the reindex task. Also added some output as it runs, just to keep people from getting nervous about it maybe having locked up during long runs.


murny (approved):
Very nice catch 👍. LGTM!

PR #2025, opened 2020-12-10
Fix a mis-named error rescue that resulted in a crash

mbarnett:
## Context

Application 500s if you sort items by date and then switch results tabs to Collections or Communities.

## What's New

We're supposed to handle this by rescuing a NameManglingError if the sort field doesn't have a known Solr mangled index name for the model being searched – but we missed the namespace on the error in the rescue declaration, so it just crashes as an unknown class.

Part of #2009

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 - Make reindex rake task actually reindex all of the objects into Solr, instead of acting as a no-op
+– Fix a mis-named error rescue that resulted in a crash when the sort field wasn't known for a model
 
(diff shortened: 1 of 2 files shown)


murny (approved):
Nice catch 👍 

gif

pgwillia:
Was this related to the rollbar error [#1134 New error: ActionView::Template::Error: uninitialized constant JupiterCore::Search::SearchFailed Did you mean? SearchHelper](https://rollbar.com/ualbertalib/jupiter/items/1134/?utm_campaign=new_item_message&utm_medium=slack&utm_source=rollbar-notification)

mbarnett:
No, that's a different one (probably some kind of transitory Solr failure triggering that raise that I guess we'd never hit much and never noticed references a non-existent error)

PR #2029, opened 2020-12-14
Cds/date facet flash issue fix

### LINKED ISSUE #2009: Current era-app-prd issues

ConnorSheremeta:
#2009 number 7.

No longer 500s. Will put in new ticket to bring this error flash back.


mbarnett (dismissed):

murny (approved):
👍

PR #2038, opened 2020-12-17
try to better resolve issue with logo_id constraints when deleting items

mbarnett:
## Context

1. Items have attachments.
2. Items have a logo_id, which is the id of the attachment that represents the item's "logo".
3. logo_id has a constraint to make sure it points at attachments that actually exist.
4. The logo_id constraint is "on delete nullify", so when you delete an attachment, it should become null.

Nevertheless, when you try to delete an item, it tries to delete its attachments first, and then that fails, because the attachment is referenced by the logo_id on the item – the item that is about to be deleted, and the logo_id that should be nullified by the attachment delete.

I don't know why this happens.

## What's New

Try to make this not happen. Directly bypass everything and write a null on the logo_id column before deleting the item and therefore trigger a delete of the attachments.

### DIFF
--- a/app/controllers/admin/items_controller.rb
+++ b/app/controllers/admin/items_controller.rb
@@ -15,7 +15,20 @@ class Admin::ItemsController < Admin::AdminController
     end
 
     begin
-      @item.logo_id = nil
+      # rubocop:disable Rails/SkipsModelValidations
+      # HACK: bit of a silly hack (write a nil directly to the colum) to get around this hack where destroying
+      # the item tries to destroy the attachments, which then errors out because the attachment id is
+      # still referenced by the item that's about to be destroyed.
+      #
+      # TODO: investigate further why this happens (seems inconsistent with the settings for
+      # the constraint:
+      #  +add_foreign_key "items", "active_storage_attachments", column: "logo_id", on_delete: :nullify+
+      # which should allow this to work?) and consider dropping the constraint
+      # entirely if this is unreliable on PostgreSQL
+
+      @item.update_columns(logo_id: nil)
+
+      # rubocop:enable Rails/SkipsModelValidations
       @item.destroy!
       flash[:notice] = t('.deleted')
     rescue StandardError => e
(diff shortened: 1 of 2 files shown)


app/controllers/admin/items_controller.rb
```diff
@@ -15,7 +15,20 @@ def destroy
     end
 
     begin
-      @item.logo_id = nil
+      # rubocop:disable Rails/SkipsModelValidations
+      # HACK: bit of a silly hack (write a nil directly to the column) to get around this hack where destroying
+      # the item tries to destroy the attachments, which then errors out because the attachment id is
+      # still referenced by the item that's about to be destroyed.
+      #
+      # TODO: investigate further why this happens (seems inconsistent with the settings for
```
murny, line 23: To add to this, maybe it be worth looking into the new Rails 6.1, `dependent: :destroy_async` functionality (https://github.com/rails/rails/pull/40157) and get this working with some refactoring to make this work (but I think your right, might have to change how the foreign key constraint work 🤔 ) .

Also not sure about item model, but we wrote our own code for this on community and had issues with this as well (https://github.com/ualbertalib/jupiter/blob/4e28c2c2914f6630b0626f85bbf62e5e578cb13b/app/models/community.rb#L17

murny (approved):

app/controllers/admin/items_controller.rb
mbarnett: Ah deferring the purge is probably a smarter way to do it than trying to futz around with the constraints. I'm going to rework this a bit

mbarnett: Hmm, maybe not? ActiveStorage seems to hard-code the `dependent: destroy` on `has_many_attached` here: https://github.com/rails/rails/blob/fe76a95b0d252a2d7c25e69498b720c96b243ea2/activestorage/lib/active_storage/attached/model.rb#L49

and even on 6.1
https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/activestorage/lib/active_storage/attached/model.rb#L70

I'm poking around but I can't find any indication that it would skip the destroy if the attachment is already scheduled for a purge later, so this hack or dropping the constraint seem like the only options?

PR #2041, opened 2020-12-17
Fix nil start or end faceting dates error

ConnorSheremeta:
https://rollbar.com/ualbertalib/jupiter/items/1154/

### DIFF
--- a/app/services/user_search_service.rb
+++ b/app/services/user_search_service.rb
@@ -64,7 +64,7 @@ class UserSearchService
     start = range[:begin]
     finish = range[:end]
 
-    return false if start.nil? || finish.nil?
+    return false unless start.present? && finish.present?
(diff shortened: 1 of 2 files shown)


app/services/user_search_service.rb
```diff
@@ -63,6 +63,8 @@ def search_params(params)
   def validate_range(range)
     start = range[:begin]
     finish = range[:end]
+
+    return false if start.nil? || finish.nil?
```
mbarnett, line 67: This is _super_ nitpicky, but `false.nil?` is `false`, meaning it would pass this check, whereas `false.present?` is `false` meaning it would fail `return false unless start.present? && end.present?`. falseyness in ruby is a bit better to rely on than just "isn't nil" in general.

(same deal with empty strings being falsey, but in this case that would be ok just because you can call match on empty strings)

ConnorSheremeta: Fair enough, using `present` definitely seems like the way to go.

mbarnett (approved):

mbarnett (approved):

PR #2049, opened 2020-12-18
Completely disable logging of warnings for unthumbnailable word docs, etc

mbarnett:
## Context

See #2046

After looking at this some more, and reading Rollbar's docs around error filtering more closely, it looks like #2046 isn't enough. We would keep sending this to Rollbar anyways, because a direct call like 

https://github.com/ualbertalib/jupiter/blob/f4297b12fb96d13c0c213d67ee195b6ec087d3ca/app/helpers/page_layout_helper.rb#L53

will bypass the filters set in the Rollbar initializer.

## What's New

 Better to just disable this kind of logging entirely (also pulled it out of the logger since it's a waste of disk space to log every time someone sees a word doc or excel spreadsheet in a search result).


murny (dismissed):

murny (approved):
👍

Issue #2081, opened 2021-01-12
Setup Github Actions to run different versions of rubies


murny:
Docs on how to do this:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

You can follow their example for NodeJS, just instead we going to do ruby for ourselves.

For us, just need to add the following right before `steps` ([on line 7 here]( https://github.com/ualbertalib/jupiter/blob/master/.github/workflows/push.yml#L7)) :
```
strategy:
      matrix:
        ruby: [ 2.6, 2.7, 3.0]
```

Then change the following:

```
      - name: Set up Ruby 2.6
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: 2.6.6
```
to
```
      - name: Set up Ruby
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
```

This will generate 3 jobs for each ruby version we have specified.

Should be it to achieve this (with i'd imagine lots of testing)

murny:
This is now completed. Our CI is now checking both Ruby 2.5 (as production is currently running 2.5) and Ruby 2.7 on master branch.

PR #2082, opened 2021-01-12
#2081: Setup Github Actions to run different versions of rubies

murny:
Quick PR just to test out setup for running different versions of rubies on github actions

UPDATE: Looks like Ruby 3.0 doesn't work with some of our gems. 
First fail is on ezid-client which has ruby locked at `~> 2.1`:
```
ezid-client-1.8.0 requires ruby version ~> 2.1, which is incompatible with the
current version, ruby 3.0.0p0
```

So tweaked it so just 2.6 and 2.7 works for now.


murny:
All seems to be working now 👍 . Just need to update this in our repo settings to allow this to get merged:
image

pgwillia (dismissed):
Nice work! 👍

pgwillia (approved):
Thanks @murny! LGTM 👍

Issue #2086, opened 2021-01-12
Handle 400 from solr properly


ConnorSheremeta:
https://rollbar.com/ualbertalib/jupiter/items/1173/occurrences/146900194021/

Faraday::BadRequestError: the server responded with status 400

When solr gives a 400 we give a 500, we want to give a 400 error code and message.

PR #2088, opened 2021-01-13
Give proper response when solr 400s

ConnorSheremeta:
#2086

When solr gives a 400 we give a 500, we want to give a 400 error code and message.


app/controllers/application_controller.rb
```diff
@@ -103,6 +105,18 @@ def render_404(exception = nil)
     end
   end
 
+  def render_400(exception = nil)
```
murny, line 108: 400 is a generic error (bad request was given). So I would maybe propose we either keep this generic or make this more specific as a result. 

To keep it generic, then Instead of hardcoding the "error message" as "Looks like there is an issue with your search query!" on the page you could do something smarter here like pass error message into the view by setting a instance variable on the controller.

To keep it specific, then rename this to be only for solr bad requests, so rename this method and the 400.html page to be `render_solr_bad_request` and `solr_bad_request.html` or something

app/controllers/application_controller.rb
```diff
@@ -103,6 +105,18 @@ def render_404(exception = nil)
...
+    raise exception if exception && Rails.env.development?
+
+    respond_to do |format|
+      format.html do
+        render file: Rails.root.join('public/400.html'), layout: false, status: :bad_request
```
murny, line 113: Might be out of scope but alternatively, it be really nice if we can just show them the search page again with the "error" highlighted or allow them to recover themselves. Similar to what we do for a bad form submission.

So instead of showing them a generic static 4xx.html  like so which may confuse users:
image

We just re-render the search page again with an error. This way they can enter a new search query again (similar to what we do when no results are found). Much better UX this way
image

test/integration/bad_solr_request_test.rb
```diff
@@ -0,0 +1,23 @@
+require 'test_helper'
+
+class BadSolrRequestTest < ActionDispatch::IntegrationTest
+
+  test 'request in which solr gives 400 gives a proper response' do
+    get "#{root_path}/search?facets[all_subjects_sim][]=Antimicrobial&facets[all_subjects_sim][]=%22blown-pack%22"
```
murny, line 6: Could we try to convert these URLs to use the rails link helpers instead of string concatenation. Think it's easier to read if its just arguments in a method vs a a 3 line long string. Also not sure if this will work in the future once we have subdomains  enabled (link helpers take care of subdomains for us by looking up the subdomain in the routes file).

Might be able to just do the following (needs testing as syntax might not be perfect):

```
get search_url, params: { search: 'test', facets: [all_subjects_sim: ["Antimicrobial", "blown-pack"]], sort: 'title'}
```
etc.

murny (commented):
Looks pretty good. Just some small feedback

app/controllers/application_controller.rb
mbarnett: Do you have a good suggestion on how to kick them back to the general search page without adding a special action or some special case code to the search action, but still make sure we are sending a 400 in the header?

(The backstory here is that Bing has a bad set of URLs for searches from somewhere or other that it keeps hitting over and over again, I assume because it just thinks "hmm 500 better try again later", and it's causing a bunch of noise in Rollbar for no benefit – I'm hoping that seeing the 400 will convince it to drop them from the index finally)

murny: If thats the backstory then this may be fine. As its a bot hitting us, or a user coming from a search engine, and not a user who typed in a search term and got this 400.html page which would be a bit confusing.

No suggestion but you would probably be right. In order to kick them back to the general search page, we would need to change some code (probably simplest is to pass in a `@error_message` that if present, renders it out instead of the normal search results or something). So is this worth it? Not sure. If it's just outside traffic/bots then this is fine 🤔

mbarnett: Yeah I haven't really seen much evidence this is impacting people, other than the occasional curious person messing around with the query params

murny: Let's leave it as is then 👍

test/integration/solr_bad_request_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+
+class SolrBadRequestTest < ActionDispatch::IntegrationTest
+
+  test 'request in which solr gives 400 gives a proper response' do
+    get search_url, params: { facets: { all_subjects_sim: ['Antimicrobial', '"blown-pack"'] } }
```
murny, line 6: Much nicer 👍

murny (approved):
Looks good to me 👍

app/models/jupiter_core/search.rb
```diff
@@ -82,6 +82,10 @@ def self.perform_solr_query(q:, qf: '', fq: '', facet: false, facet_fields: [],
                                                        name: 'solr select',
                                                        query: params) do
       JupiterCore::SolrServices::Client.instance.connection.get('select', params: params)
+      rescue RSolr::Error::Http => e
+        raise JupiterCore::SolrBadRequestError if e.response[:status] == 400
```
murny, line 86: Does `raise` break the execution here? or does it fall through and hit the second raise and you get a double raise 🤔 ?

Not sure if this is a problem but you could restructure the `if` condition here so this would never happen 

```
if e.response[:status] == 400
  raise JupiterCore::SolrBadRequestError
else
  raise
end
```

mbarnett: It _should_ break the execution, unless I'm crazy.

I don't think you can double raise in Ruby. Are you thinking of a double render?

murny: Yeah I think this is fine, so feel free to ignore me 😆

app/controllers/application_controller.rb
```diff
@@ -103,6 +105,18 @@ def render_404(exception = nil)
     end
   end
 
+  def render_solr_bad_request(exception = nil)
```
mbarnett, line 108: Ok, my one and only request here is that, since what we're doing here is 400ing, we make this method name, and the file name, more generically about that status (so that if there are other reasons we also want to return a 400 in the future, we use this same code path, rather than making it seem specific to Solr).

So just like we have `render_404` and `public/404.html` this can just be the 400 version of that method and file.

mbarnett (changes requested):
Just one quick change I'd like to see to the naming to make this a more generic solution to other 400 problems in the future.

mbarnett (approved):

Issue #2010, opened 2020-12-03
Digitization Folk Fest Programs Model


pgwillia:
**Is your feature request related to a problem? Please describe.**
The first set of digitization content will be the Folk Fest Programs.  This set was chosen because it is small (there are 36) and well described.  Before we can ingest them we need to do some modelling.

**Describe the solution you'd like**
@metadata-bot shared "FolkFestTriples - v.1" and "Jupiter II working data dictionary" to inform this exercise

Need to add properties to the [Book model](https://github.com/ualbertalib/jupiter/blob/5228946755e2eb567d952eab7509eb643f4087bf/app/models/digitization/book.rb#L1)

- [ ] https://www.w3.org/TR/rdf-schema/#ch_type
- [ ] https://www.w3.org/TR/rdf-schema/#ch_label
- [ ] http://purl.org/dc/terms/issued
- [ ] https://schema.org/temporalCoverage
- [ ] http://purl.org/dc/terms/title
- [ ] http://purl.org/dc/terms/alternative
- [ ] http://purl.org/dc/terms/identifier
- [ ] http://id.loc.gov/vocabulary/identifiers/local
- [ ] http://id.loc.gov/vocabulary/relators/aut
- [ ] http://purl.org/dc/terms/type
- [ ] http://www.europeana.eu/schemas/edm/hasType
- [ ] http://purl.org/dc/terms/language
- [ ] http://id.loc.gov/vocabulary/relators/pbl
- [ ] http://id.loc.gov/vocabulary/relators/pup
- [ ] http://rdaregistry.info/Elements/u/P60550
- [ ] http://www.w3.org/2004/02/skos/core#note
- [ ] http://purl.org/dc/elements/1.1/coverage
- [ ] http://www.europeana.eu/schemas/edm/rights
- [ ] http://purl.org/dc/elements/1.1/subject
- [ ] http://pcdm.org/models#memberOf

pgwillia:
Tests

- [ ] at least one of the subjects (`temporal_subject`, `geographic_subject`, `topical_subject`)
- [ ] title present

- [ ] date issued encoded as EDTF
- [ ] temporal_subject encoded as EDTF
- [ ] publisher literal or URI
- [ ] place of publication literal or URI
- [ ] geographic_subject literal or URI
- [ ] topical_subject literal or URI

- [ ] resource_type in the controlled vocab and present
- [ ] genre in the controlled_vocab and present
- [ ] language in the controlled vocab and present
[...]

PR #2089, opened 2021-01-14
Digitization books model for Folk Fest items

### LINKED ISSUE #2120: Make the RDAU vocabulary less opaque
We could give this [`::RDF::Vocab::RDAU['/P60550']`] some kind of meaningful name, because as it stands this is incomprehensible, and I don't want to go the route of having these opaque, meaningless codes scattered throughout the code base.

 I see it as kind of the difference between "I should be able to understand what something is by it's name" and kind of the more 'Library-y' approach of a Discovery or similar, which has a bunch of Librarian acronyms and location codes embedded in the code that make onboarding really really hard?

_Originally posted by @mbarnett in https://github.com/ualbertalib/jupiter/pull/2089#discussion_r564129598_

Other resources:
* https://github.com/ruby-rdf/rdf-vocab#adding-new-vocabularies
* http://www.rdaregistry.info/Elements/u/
* 

Q. Why are the IRIs unreadable?

After a great deal of discussion, the JSC (predecessor of the RDA Steering Committee) decided to use opaque IRIs for the element sets. This follows the practice established for the value vocabularies, and reflects the international environment of RDA and its translations. The RSC is also aware of the problems associated with using transparent, human-readable IRIs as a mnemonic device, especially when the properties and values of bibliographic display labels may be further developed. However, English-readable IRIs have been declared owl:sameAs to their canonical, opaque counterparts to alleviate any inconvenience this might cause for English-based linked data applications. The same facility may be developed for other languages, based on translations of RDA, in the future. The mnemonic accuracy of these IRIs is not guaranteed. http://www.rdaregistry.info/rgFAQ#faq3
* https://github.com/ruby-rdf/rdf-vocab/pull/68/files

pgwillia:
## Context

The first set of digitization content will be the Folk Fest Programs. This set was chosen because it is small (there are 36) and well described. Before we can ingest them we need to do some modelling.

Related to #2010
Closes #2120 

## What's New

- added attributes to Digitization::Book model and their rdf annotations
- added validations for the attributes
- added several controlled vocabularies and relevant FolkFest values
- added `edtf` to help parse and validate Extended Date/Time Format
- added rdau to TERMS hash to make an opaque identifier more clear

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,7 @@ gem 'aasm' # state-machine management
 gem 'addressable', '~> 2.7.0' # Replacement for the standard URI implementation
 gem 'differ' # Used to diff two strings
 gem 'draper'
+gem 'edtf', '~> 3.0', '>= 3.0.6' # parsing Extended Date/Time Format
 gem 'ezid-client', '~> 1.8.0'
 gem 'flipper', '~> 0.20.4' # Feature flags for Ruby
 gem 'flipper-active_record', '~> 0.20.3' # Store feature flags in ActiveRecord
--- a/app/validators/edtf_validator.rb
+++ b/app/validators/edtf_validator.rb
@@ -1,0 +1,17 @@
+class EDTFValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attribute, value)
+    return if value.blank?
+
+    value = [value] unless value.is_a?(Array)
+
+    value.each do |date|
+      next if date.blank?
+
+      unless Date.edtf(date)
+        record.errors.add(attribute, (options[:message] || 'does not conform to the Extended Date/Time Format standard'))
+      end
+    end
+  end
+
+end
--- a/app/validators/edtf_validator.rb
+++ b/app/validators/edtf_validator.rb
@@ -8,7 +8,7 @@ class EDTFValidator < ActiveModel::EachValidator
     value.each do |date|
       next if date.blank?
 
-      next if Date.edtf(date)
+      next unless Date.edtf(date).nil?
 
       record.errors.add(attribute, (
         options[:message] || 'does not conform to the Extended Date/Time Format standard'
--- a/app/validators/edtf_validator.rb
+++ b/app/validators/edtf_validator.rb
@@ -8,9 +8,11 @@ class EDTFValidator < ActiveModel::EachValidator
     value.each do |date|
       next if date.blank?
 
-      unless Date.edtf(date)
-        record.errors.add(attribute, (options[:message] || 'does not conform to the Extended Date/Time Format standard'))
-      end
+      next if Date.edtf(date)
+
+      record.errors.add(attribute, (
+        options[:message] || 'does not conform to the Extended Date/Time Format standard'
+      ))
     end
   end
 
--- a/config/controlled_vocabularies/digitization_subject.yml
+++ b/config/controlled_vocabularies/digitization_subject.yml
@@ -1,0 +1,5 @@
+digitization_subject:
+  edmonton: http://id.loc.gov/authorities/names/n79007225
+  edmonton_folk_music_festival: http://viaf.org/viaf/148861489
+  folk_music_festivals: http://id.loc.gov/authorities/subjects/sh2008004058
+  
--- a/config/terms.yml
+++ b/config/terms.yml
@@ -64,11 +64,6 @@
   terms:
     institution: institution
 
-- vocabulary: rdau
-  schema: 'http://rdaregistry.info/Elements/u/'
-  terms:
-    extent_en: extent.en
-
 # The following ontologies will be replaced and for now are just
 # placeholders
 - vocabulary: fits
--- a/config/terms.yml
+++ b/config/terms.yml
@@ -59,6 +59,12 @@
     embargo_history: embargoHistory
     visibility_after_embargo: visibilityAfterEmbargo
 
+# Trying to make ::RDF::Vocab::RDAU.P60550 less opaque
+- vocabulary: rdau
+  schema: 'http://rdaregistry.info/Elements/u/'
+  terms:
+    extent: extent.en
+
 - vocabulary: swrc
   schema: 'http://ontoware.org/swrc/ontology#'
   terms:
--- a/config/terms.yml
+++ b/config/terms.yml
@@ -64,6 +64,11 @@
   terms:
     institution: institution
 
+- vocabulary: rdau
+  schema: 'http://rdaregistry.info/Elements/u/'
+  terms:
+    extent_en: extent.en
+
 # The following ontologies will be replaced and for now are just
 # placeholders
 - vocabulary: fits
--- a/config/terms.yml
+++ b/config/terms.yml
@@ -63,7 +63,7 @@
 - vocabulary: rdau
   schema: 'http://rdaregistry.info/Elements/u/'
   terms:
-    extent: extent.en
+    extent: P60550
 
 - vocabulary: swrc
   schema: 'http://ontoware.org/swrc/ontology#'
--- a/db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
+++ b/db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
@@ -27,7 +27,7 @@ class AddFolkFestAttributesToDigitizationBook < ActiveRecord::Migration[6.0]
       t.language has_predicate: ::RDF::Vocab::DC.language
       t.publisher has_predicate: ::RDF::Vocab::MARCRelators.pbl
       t.place_of_publication has_predicate: ::RDF::Vocab::MARCRelators.pup
-      t.extent has_predicate: ::RDF::Vocab::RDAU.P60550
+      t.extent has_predicate: ::TERMS[:rdau].extent
       t.note has_predicate: ::RDF::Vocab::SKOS.note
       t.geographic_subject has_predicate: ::RDF::Vocab::DC11.coverage
       t.rights has_predicate: ::RDF::Vocab::DC11.rights
--- a/db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
+++ b/db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
@@ -27,7 +27,7 @@ class AddFolkFestAttributesToDigitizationBook < ActiveRecord::Migration[6.0]
       t.language has_predicate: ::RDF::Vocab::DC.language
       t.publisher has_predicate: ::RDF::Vocab::MARCRelators.pbl
       t.place_of_publication has_predicate: ::RDF::Vocab::MARCRelators.pup
-      t.extent has_predicate: ::RDF::Vocab::RDAU['/P60550']
+      t.extent has_predicate: ::RDF::Vocab::RDAU.P60550
       t.note has_predicate: ::RDF::Vocab::SKOS.note
       t.geographic_subject has_predicate: ::RDF::Vocab::DC11.coverage
       t.rights has_predicate: ::RDF::Vocab::DC11.rights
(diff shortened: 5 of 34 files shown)


Gemfile
```diff
@@ -50,6 +50,7 @@ gem 'aasm' # state-machine management
 gem 'addressable', '~> 2.7.0' # Replacement for the standard URI implementation
 gem 'differ' # Used to diff two strings
 gem 'draper'
+gem 'edtf', '~> 3.0', '>= 3.0.6' # parsing Extended Date/Time Format
```
pgwillia, line 53: Added this gem to do the heavy lifting of validating a number of different formats for dates that are less precise (decades, seasons, etc) than Date.

app/validators/edtf_validator.rb
```diff
@@ -0,0 +1,15 @@
+class EDTFValidator < ActiveModel::EachValidator
+
+    def validate_each(record, attribute, value)
+      return if value.blank?
+
+      value = [value] unless value.is_a?(Array)
+    
+      value.each do |date|
+        next if date.blank?
+        unless Date.edtf(date)
```
pgwillia, line 10: This is the one place `edtf` gem is used.

test/controllers/admin/theses/draft_controller_test.rb
```diff
@@ -93,7 +93,7 @@ class Admin::Theses::DraftControllerTest < ActionDispatch::IntegrationTest
     draft_thesis.reload
     assert_equal 'Open to everyone!', draft_thesis.rights
     assert_equal 'embargo', draft_thesis.visibility
-    assert_equal Date.current + 1.year, draft_thesis.embargo_end_date
+    assert_equal (Date.current + 1.year).to_datetime, draft_thesis.embargo_end_date
```
pgwillia, line 96: A side effect of using the `edtf` gem is that the default format for Date objects is less precise that the `embargo_end_date` `datetime` type.

config/controlled_vocabularies/digitization_subject.yml
```diff
@@ -0,0 +1,5 @@
+digitization_subject:
```
pgwillia, line 1: I'm not sure if this approach is going to be able to scale.  The `digitization_subject` vocabulary is used by `geographic_subject` and `topical_subject`.  May be overlap with `place_of_publication` and `publisher`.

db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.date_issued has_predicate: ::RDF::Vocab::DC.issued
+      t.temporal_subject has_predicate: ::RDF::Vocab::SCHEMA.temporalCoverage
+      t.title has_predicate: ::RDF::Vocab::DC.title
+      t.alt_title has_predicate: ::RDF::Vocab::DC.alternative
+      t.resource_type has_predicate: ::RDF::Vocab::DC.type
+      t.genre has_predicate: ::RDF::Vocab::EDM.hasType
+      t.language has_predicate: ::RDF::Vocab::DC.language
+      t.publisher has_predicate: ::RDF::Vocab::MARCRelators.pbl
+      t.place_of_publication has_predicate: ::RDF::Vocab::MARCRelators.pup
+      t.extent has_predicate: ::RDF::Vocab::RDAU['/P60550']
```
pgwillia, line 30: This vocabulary is a bit different from others we've worked with. The uri is `http://rdaregistry.info/Elements/u/P60550` but the property is `rdau:extent.en`.  I think the design of it is to be language agnostic but it ends up being pretty opaque.

db/seeds.rb
```diff
@@ -717,7 +717,26 @@ def file2.original_filename
 end
 
 Digitization::Book.create(peel_id: '4062') # monograph
-Digitization::Book.create(peel_id: '10572', part_number: '1') # Folk Fest
+
+11.times do |i|
+  Digitization::Book.create(peel_id: rand(1..3400), part_number: rand(1..100),
```
pgwillia, line 722: There's probably more variation that I could do here.  There aren't many terms in the controlled vocabulary to iterate on yet.  The biggest improvements that could be made here is to 

- leave blank some of the optional fields
- have more values in the 0+ fields
- choose 1 or more of the subjects

pgwillia (commented):
Would love some feedback.  It's not perfect as I've pointed out but probably good enough.

mbarnett (changes requested):
(just putting a block on the merge to prevent any accidental merging until I've had a chance to fully think through everything)

db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.json :genre, null: false, array: true
+      t.json :language, null: false, array: true
+      t.json :publisher, array: true
+      t.json :place_of_publication, array:true
+      t.string :extent
+      t.json :note, array: true
+      t.json :geographic_subject, array: true
+      t.string :rights
+      t.json :topical_subject, array: true
+    end
```
mbarnett, line 18: We might need to be a bit more liberal in how we approach translating metadata's models – specifically, where possible we want to bias towards making things singular and only emitting multival values when translating back for them, because both the query language and performance implications of having a ton of JSON columns aren't great or will end up being annoying to work with. 

We can certainly use them where we have to, but we should be _sure_ we have to for the data we actually have and that we're not going to be wanting to construct a lot of queries against the data.

So just looking at this and glancing at the triples, date Issued seems to be purely single valued in the data we have from glancing at the triples, so let's run with that. Something like Genre we might consider implementing as a join table if there's a fixed schema of values we'll only ever be referencing (ie. if we can know and preload them all before hand – if I'm looking at the correct triple it looks like that may be the case, but unsure how well that generalizes to other books), otherwise we can go this route. Place of publication doesn't make sense to me that that would routinely have multiple values, but possibly some other data might (and obviously we'll have to reconsider this as more data rolls in over time)...

Just something to think about, because it's very easy for us to go too hard down the JSON rabbit hole and make things excessively hard to write queries against later.

db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
mbarnett: Woof, yeah. This is worth putting some thought into how we could give this some kind of meaningful name, because as it stands this is incomprehensible, and I don't want to go the route of having these opaque, meaningless codes scattered throughout the code base.

 I see it as kind of the difference between "I should be able to understand what something is by it's name" and kind of the more 'Library-y' approach of a Discovery or similar, which has a bunch of Librarian acronyms and location codes embedded in the code that make onboarding really really hard?

mbarnett (changes requested):
There's a couple of small things worth changing I've noted – the TODOs I suggest may even be worth trying to tackle now rather than letting the number of occurrences of the same pattern grow?

Bigger concern I'm wrestling with is just how many JSON columns we're already looking at and whether or not some of that might be better modelled in another way – either as singular values if their multiplicity is more "speculative", or as more of a relational join table design if we know they're not necessarily freeform values but coming from a limited vocab we could just load up in the DB. Some of this might be worth you and I discussing so that we can put together a solid idea of how we want to tackle this going forward.

db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
pgwillia: I was going on the observation that everywhere in ERA that required an array that json was used.  Perhaps these should be a different type?  Arrays of strings?

Maybe I should set up a discussion with the metadata team (Sharon?) about the assumptions that we're making.  I'm sure that they've looked at the data when creating their models.  They might know if the assumptions we're thinking of making will play out.

mbarnett: Yeah, if we need an array, JSON arrays are the way to go unless the things we're trying to have multiple values of are fixed enough that a Join Table might be more natural. So authors are completely freeform and infinite and bad for join tables, but genres might be.

Let's meet and figure out the approach we think we want to take and then develop questions we might have for Metadata based on the outcome of that.

test/fixtures/digitization/books.yml
```diff
@@ -1,18 +1,54 @@
 # Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
 
+# TODO: resource_type, genre, language and a subject are required. These are placeholders.
 peel_monograph:
   peel_id: 4062
+  title: 'Report on a rural survey of the agricultural, educational, social, and religious life'
+  resource_type: <%= CONTROLLED_VOCABULARIES[:digitization][:resource_type].txt %>
```
pgwillia, line 7: @mbarnett I was just thinking that we wouldn't be able to reason like this if there's a direct <label>: <uri> mapping.

mbarnett: How so?

mbarnett: To be slightly clearer, we could just modify the OpenStruct to let us look things up by key/label as well, eg something like

```ruby
    vocab.define_singleton_method(:to_uri) do |key|
      config[vocab_name][key]
    end
```
`in controlled_vocabularies.rb` and then 

```ruby
resource_type: <%= CONTROLLED_VOCABULARIES[:digitization][:resource_type].to_uri('Text') %>
```

The only area of concern here is the "I need to figure out" part, that I'm currently trying to figure out, which is how best to bridge the distinction between different kinds of vocabs (URI->symbol->text indirectly via locale vs URI->text directly), because the text labels are going to not work as method names for the OpenStruct

db/migrate/20201204210653_add_folk_fest_attributes_to_digitization_book.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.date_issued has_predicate: ::RDF::Vocab::DC.issued
+      t.temporal_subject has_predicate: ::RDF::Vocab::SCHEMA.temporalCoverage
+      t.title has_predicate: ::RDF::Vocab::DC.title
+      t.alt_title has_predicate: ::RDF::Vocab::DC.alternative
+      t.resource_type has_predicate: ::RDF::Vocab::DC.type
+      t.genre has_predicate: ::RDF::Vocab::EDM.hasType
+      t.language has_predicate: ::RDF::Vocab::DC.language
+      t.publisher has_predicate: ::RDF::Vocab::MARCRelators.pbl
+      t.place_of_publication has_predicate: ::RDF::Vocab::MARCRelators.pup
+      t.extent has_predicate: ::TERMS[:rdau].extent
```
pgwillia, line 30: This is the same as
```suggestion
      t.extent has_predicate: ::RDF::Vocab::RDAU.P60550.properties[:'http://metadataregistry.org/uri/profile/regap/lexicalAlias']
```
I decided to go with adding this to our TERMS hash so that it's clear what we're talking about.

Another thing I tried was to add accessors for the lexicalAliases in an initializer but wasn't successful.  

```
# config/initializers/rdau_vocabulary.rb
require 'rdf/vocab'

class RDF::Vocab::RDAU
  RDF::Vocab::RDAU.properties.each do |property|\n",
    next unless property.properties[:'http://metadataregistry.org/uri/profile/regap/lexicalAlias']
    define_singleton_method property.properties[:'http://metadataregistry.org/uri/profile/regap/lexicalAlias'].qname.last do
      property.properties[:'http://metadataregistry.org/uri/profile/regap/lexicalAlias']
    end
  end
end
```

With the goal of being able to invoke `RDF::Vocab::RDAU.send(:'extent.en')` and get "http://rdaregistry.info/Elements/u/extent.en"

mbarnett (approved):
This looks really good to me! Do you want to tackle the UI (nothing fancy needed right now, just really basic) or start looking at ingesting real data, as a next step?

config/terms.yml
```diff
@@ -59,6 +59,12 @@
     embargo_history: embargoHistory
     visibility_after_embargo: visibilityAfterEmbargo
 
+# Trying to make ::RDF::Vocab::RDAU.P60550 less opaque
+- vocabulary: rdau
+  schema: 'http://rdaregistry.info/Elements/u/'
+  terms:
+    extent: extent.en
```
pgwillia, line 66: I just looked back at the FolkFestTriples - v.1' and the Data Dictionary.  The URI as specified was 'http://rdaregistry.info/Elements/u/P60550'

```suggestion
    extent: P60550
```

mbarnett (approved):

Issue #2077, opened 2021-01-08
Sort functionality not robust to missing order


pgwillia:
A plagarism checking bot has discovered a bug with our sorting. Can reproduce by visiting http://era.lvh.me:3000/search?sort=sort_year in development.

https://github.com/ualbertalib/jupiter/blob/6fff30c9be66280a434db5ab4fb136215a6878a6/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb#L71-L81

Where 
- `sort_attribute` is "sort_year_isi"
- `order` is `nil`
- `solr_exporter.default_sort_indexes.index(sort_attribute)` is `nil`

Then on line 81 the TypeError is thrown.

View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/1192/](https://rollbar.com/ualbertalib/jupiter/items/1192/)

```

TypeError: no implicit conversion from nil to integer
  File "/var/www/sites/jupiter/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb", line 80, in block in sort
  File "/var/www/sites/jupiter/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb", line 49, in each
  File "/var/www/sites/jupiter/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb", line 49, in each_with_index
  File "/var/www/sites/jupiter/app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb", line 49, in sort
  File "/var/www/sites/jupiter/app/services/user_search_service.rb", line 37, in results
  File "/var/www/sites/jupiter/app/controllers/search_controller.rb", line 24, in index
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/actionpack-6.0.3.4/lib/action_controller/metal/basic_implicit_render.rb", line 6, in send_action
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/actionpack-6.0.3.4/lib/abstract_controller/base.rb", line 195, in process_action
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/actionpack-6.0.3.4/lib/action_controller/metal/rendering.rb", line 30, in process_action
  File "/var/www/sites/jupiter/vendor/ruby/2.5.0/gems/actionpack-6.0.3.4/lib/abstract_controller/callbacks.rb", line 42, in block in process_action
[...]

PR #2095, opened 2021-01-15
Search with sort without default sort direction no longer errors

ConnorSheremeta:
#2077


pgwillia (approved):

pgwillia (dismissed):

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -77,6 +77,7 @@ def sort(attributes, orders = nil)
                                else
                                  # We could not find the order so we switch to default sort direction
                                  sort_direction_index = solr_exporter.default_sort_indexes.index(sort_attribute)
+                                 sort_direction_index = 0 if sort_direction_index.blank?
```
mbarnett, line 80: Might be worth adding a quick comment explaining why this is being done. Other than that this looks good to me

app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
```diff
@@ -77,6 +77,8 @@ def sort(attributes, orders = nil)
                                else
                                  # We could not find the order so we switch to default sort direction
                                  sort_direction_index = solr_exporter.default_sort_indexes.index(sort_attribute)
+                                 # If we cant find the default sort_direction_index for this sort_attribute default it.
+                                 sort_direction_index = 0 if sort_direction_index.blank?
```
pgwillia, line 81: I kinda feel like the comment is redundant to what is already there 
> We could not find the order so we switch to default sort direction

Especially if we make the 0 a constant with a meaningful name.
```suggestion
                                 DEFAULT_SORT_DIRECTION_INDEX = 0  # put this at the top of the class or method
                                 # We could not find the order so we switch to default sort direction
                                 sort_direction_index = solr_exporter.default_sort_indexes.index(sort_attribute) 
                                 sort_direction_index = DEFAULT_SORT_DIRECTION_INDEX if sort_direction_index.blank?
```

Something I didn't think of before is why the solr_exporter.default_sort_indexes.index(sort_attribute) doesn't give the correct value or if it ever would. 
```
solr_exporter.default_sort_indexes
=> ["title_ssi"]
```
This else is assuming that the default index is being used, or no `sort` is specified, which isn't the case here.

mbarnett: Agree that the comment just reiterates what the code does instead of explaining more of the why behind it. I don't feel that the constant significantly improves the situation though. The code as written is awkward and introducing another indirection on top of it just doesn't feel like it improves things.

We should probably take the need for comments and the difficulty of writing them as a sign that the code should be more fundamentally re-worked.

How about something like 

```ruby
       criteria[:sort_order] << case
                                when sort_attribute == :score
                                  # When sorting by score it only makes sense to use :desc order from the user
                                  # perspective so we ignore the order if one is given.
                                  :desc
                                when order.present? && POSSIBLE_SORT_ORDERS.include?(order)
                                  order
                                when (index = solr_exporter.default_sort_indexes.index(sort_attribute)).present?
                                  solr_exporter.default_sort_direction[index]
                                else
                                  # if we have been passed a sort attribute that isn't actually set up for sorting, we'll
                                  # just fall back to the first thing we have on hand 
                                  solr_exporter.default_sort_direction.first
                                end
    end

as a more self-explanatory approach?

mbarnett (approved):
Good job

Issue #2100, opened 2021-01-19
Add read only mode to seeds


pgwillia:
**Describe the bug**
`ReadOnlyMode.create(enabled: false)` should be added to the seeds.  It's in a migration but that doesn't get set in development if you're running `db:prepare` or  `db:setup`.

**To Reproduce**
1. `docker-compose down -v`
2. `docker-compose up -d`
3. `rails db:prepare` (as per the README)

**Expected behavior**
The seeds will prepare the database for development without extra steps.

**Additional context**
This is where it's added in a migration
https://github.com/ualbertalib/jupiter/blob/087a261fa8d745b821fc2b7035b1454ffe897859/db/migrate/20200831182732_add_initial_read_only_modes.rb#L1-L5

PR #2107, opened 2021-01-21
Initialize disabled ReadOnlyMode in seeds

pgwillia:
## Context

`ReadOnlyMode.create(enabled: false)` is in a migration but that doesn't get set in development if you're running `db:prepare` or `db:setup`.

Related to #2100 

## What's New

Initialized disabled ReadOnlyMode in seeds.


pgwillia:
Closing in favor of #2108

PR #2108, opened 2021-01-21
Initialize disabled ReadOnlyMode in seeds

pgwillia:
## Context

`ReadOnlyMode.create(enabled: false)` is in a migration but that doesn't get set in development if you're running `db:prepare` or `db:setup`.

Related to #2100 https://github.com/ualbertalib/jupiter/issues/1838

## What's New

Initialized disabled ReadOnlyMode in seeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ## [Unreleased]
 
+### Added
+- Initialize disabled ReadOnlyMode [#2100](https://github.com/ualbertalib/jupiter/issues/2100)
+
 ### Removed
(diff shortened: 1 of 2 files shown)


murny (approved):
LGTM 👍 .

🤔  Just curious if we should delete the migration for ReadOnlyMode now? Or does it even matter 🤷

pgwillia:
I guess it depends on your development workflow.  I find that I frequently start fresh.  But if your flow is to `rake db:migrate` in development to get the latest (like is done in staging and production) then the migration would have been helpful.

I'm leaning towards it doesn't matter.  It was something that we did at some point so maybe having the history there is helpful.  If we pull it out, maybe there should be something in the README about setting up staging/production environments that mentions setting this value.

murny:
Make sense. Let's leave it 👍

Issue #1261, opened 2019-08-27
[Spike] investigate UI/SOLR for using "OR" instead of "AND"


weiweishi:
**Is your feature request related to a problem? Please describe.**
As a user, I would like to be able to select multiple options in the author facet category, and 
 I would like the search logic to be "OR", so I can select same author with the variation (Jane Doe or J. Doe or Jane M. Doe etc)
Similarly with the department facet - Work done here #1060 

- [ ] Initial investigation on UI options, and Solr configuration that would enable the use of "OR" instead of "AND"

pgwillia:
New egg experience
![Screenshot from 2019-10-18 15-35-56](https://user-images.githubusercontent.com/1220762/67129875-00708d80-f1bd-11e9-9164-707f8b4e6640.png)
![Screenshot from 2019-10-18 15-35-45](https://user-images.githubusercontent.com/1220762/67129876-00708d80-f1bd-11e9-8cdc-78d5c884bdab.png)
![Screenshot from 2019-10-18 15-35-33](https://user-images.githubusercontent.com/1220762/67129877-00708d80-f1bd-11e9-8535-675113c70e48.png)

pgwillia:
From https://articles.uie.com/faceted_search2part2/
>  Multiple Selection from a Facet

> The most common use case for faceted search or navigation is to select at most one value per facet, but there are at least two ways from which a user might select multiple values from the same facet:

>  Disjunctive (OR) selection. Selecting a range (e.g., a price or date range) may be a kind of disjunctive selection, depending how the values are represented.
>  Conjunctive (AND) selection.

> The design challenge is to communicate to users whether selecting multiple values from a particular facet is disjunctive or conjunctive—particularly if the site offers both behaviors. Users are notoriously bad at inferring Boolean logic from subtle cues.

> It is important to use an interface that not only is self-consistent but also adheres to familiar conventions. For example, the check boxes in Figure 7.6 adheres to the convention for disjunctive selection.
results in Help
Figure 7.6: Results for “ribs” in New York at www.yelp.com.
![yelp](https://user-images.githubusercontent.com/1220762/67312191-24361b00-f4be-11e9-9fe1-213779f446a8.jpg)

> There are fewer interfaces that allow conjunctive selection from the same facet, but a convention for those that do is to present the selections as ordinary links, as with the topic facet in Figure 7.7.
results in fcla
Figure 7.7: Results for “tax law” at www.fcla.edu.
[...]

PR #2115, opened 2021-01-25
Labels for active facet badges

pgwillia:
# Context
Applied Filters could show more information. Such as the facet they belong to.  Currently active facets are jumbled together.

Related to #1261

## What's New

Before:
image

After:
image

### DIFF
--- a/app/helpers/presenters_helper.rb
+++ b/app/helpers/presenters_helper.rb
@@ -10,7 +10,8 @@ module PresentersHelper
     # string to class conversion dozens of times during facet rendering
     @presenter_cache ||= {}
     # FacetValues are special insofar as they dynamically specify their own presenter per-attribute-name involved
-    if obj.is_a?(JupiterCore::SolrServices::FacetResult::FacetValue)
+    if obj.is_a?(JupiterCore::SolrServices::FacetResult::FacetValue) ||
+       obj.is_a?(JupiterCore::SolrServices::RangeFacetResult)
       present_facet(obj)
     else
       presenter_for(obj).decorate(obj)
--- a/app/helpers/presenters_helper.rb
+++ b/app/helpers/presenters_helper.rb
@@ -10,6 +10,7 @@ module PresentersHelper
     # string to class conversion dozens of times during facet rendering
     @presenter_cache ||= {}
     # FacetValues are special insofar as they dynamically specify their own presenter per-attribute-name involved
+    # TODO: figure out a better way to do this eg) facet superclass
     if obj.is_a?(JupiterCore::SolrServices::FacetResult::FacetValue) ||
        obj.is_a?(JupiterCore::SolrServices::RangeFacetResult)
       present_facet(obj)
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -113,4 +113,8 @@ module SearchHelper
     last = results.offset_value + results.count
     t('search.page_range', first: first, last: last, total: results.total_count)
   end
+
+  def short_category_name(category_name)
+    category_name.split(' /').first
+  end
 end
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -113,8 +113,4 @@ module SearchHelper
     last = results.offset_value + results.count
     t('search.page_range', first: first, last: last, total: results.total_count)
   end
-
-  def short_category_name(category_name)
-    category_name.split(' /').first
-  end
 end
--- a/app/models/jupiter_core/solr_services/range_facet_result.rb
+++ b/app/models/jupiter_core/solr_services/range_facet_result.rb
@@ -1,6 +1,6 @@
 class JupiterCore::SolrServices::RangeFacetResult
 
-  attr_reader :solr_index, :category_name, :range
+  attr_reader :solr_index, :category_name, :short_category_name, :range
 
   def initialize(facet_map, solr_index, range)
     @solr_index = solr_index
@@ -13,6 +13,11 @@ class JupiterCore::SolrServices::RangeFacetResult
                      else
                        attribute_name.to_s.titleize
                      end
+    @short_category_name = if I18n.exists?("facets.short.#{attribute_name}")
+                             I18n.t("facets.short.#{attribute_name}")
+                           else
+                             @category_name
+                           end
 
     @range = Range.new(range[:begin], range[:end])
   end
--- a/app/models/jupiter_core/solr_services/range_facet_result.rb
+++ b/app/models/jupiter_core/solr_services/range_facet_result.rb
@@ -8,16 +8,8 @@ class JupiterCore::SolrServices::RangeFacetResult
 
     # Allows the user to override the presentation name for the facet category by customizing the
     # facets.<attribute_name> in the locale file.
-    @category_name = if I18n.exists?("facets.#{attribute_name}")
-                       I18n.t("facets.#{attribute_name}")
-                     else
-                       attribute_name.to_s.titleize
-                     end
-    @short_category_name = if I18n.exists?("facets.short.#{attribute_name}")
-                             I18n.t("facets.short.#{attribute_name}")
-                           else
-                             @category_name
-                           end
+    @category_name = I18n.t("facets.#{attribute_name}", default: attribute_name.to_s.titleize)
+    @short_category_name = I18n.t("facets.short.#{attribute_name}", default: @category_name)
 
     @range = Range.new(range[:begin], range[:end])
   end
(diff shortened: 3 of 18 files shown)


app/views/application/_facet_result.html.erb
```diff
@@ -9,6 +11,10 @@
 
       <% if active_facet? facet_value %>
         <% content_for :facet_badges do %>
+          <% if do_header %>
```
pgwillia, line 14: We only want the label to appear once for the active facet.

app/views/application/_facet_result.html.erb
```diff
@@ -9,6 +11,10 @@
...
+            <% do_header = false %>
+            <label class="font-size: 75%"><%= facet_result.category_name %>:</label>
```
murny, line 16: Some tiny comments on UI/UX:

image

- Seems very congested? Probably needs some spacing between facets with something like `ml-3`  (aka `margin-left: 1rem`) or maybe just putting each facet on a separate line? Not sure. Curious what others do (Since we use Newegg as an example in the story, They make, each facet and values is in its own container, so visually its all contained and separated) .
- Not sure if `<label>` is the right HTML element to use (typically you use label with a `for=` attribute to associate it with something like an input tag)? I'd probably just use a `header` or a some text tag instead and nest the `badge` span within it. So if you look at official docs (https://getbootstrap.com/docs/4.6/components/badge/#example), the badges are inside a typography tag (So ours might look like this instead `<h4>Subject/Keyword<span class="badge badge-secondary">Cat</span><span class="badge badge-secondary">Dog</span></h4>`)
- Don't think `class="font-size: 75%"` is valid attribute/html? You probably thinking of `style="font-size: 75%"` instead. However we should be using inline css for styling. I think using a header/typograph element would help with this or look into some font/typograph helpers from bootstrap

Overall though, nice work. Big improvement 👍

app/views/application/_facet_result.html.erb
murny: Makes sense. Be nice if we can find a different way to do this instead of using this `flag` but I'm drawing a blank 🤔

app/views/application/_facet_result.html.erb
pgwillia: Thanks for the comments.
How about using `p class="h6"`?
image

murny: Yeah that looks much better 🎉. Looks pretty good to me for MVP 👍

app/views/application/_facet_result.html.erb
murny: Yeah it's almost like we should ditch the `content_for` stuff and just re-render this html in the facet badges area. Sure your duplicating code, but you will have much better control of it and wouldn't need this `do_header flag`

But this might be out of scope as it will be a bit of a refactor. Curious if Matt has any other ideas to improve this

app/views/application/_facet_result.html.erb
mbarnett: IDK, I feel like this eats up too much vertical space. What about folding the category into the label? Something like

Screen Shot 2021-01-25 at 3 06 02 PM

Would also be worth reaching out to Sarah to discuss, since she might have some ideas around this for Digitization

mbarnett: (would want to have "short-forms" of facets eg) "author" instead of the full category "Author / Creator / Contributor"

murny: I like this  👍. If we did this too then we wouldn't need the `do_header` flag and all that logic that goes with it. As we can just repeat each category with the value. And is probably much closer to what you find in the wild (other ecommerce sites). Code is simpler and looks better (Sure category name is repeated but that's okay)

murny: Yeah heres an example that does this solution: 
image

app/helpers/search_helper.rb
```diff
@@ -113,4 +113,8 @@ def results_range(results)
     last = results.offset_value + results.count
     t('search.page_range', first: first, last: last, total: results.total_count)
   end
+
+  def short_category_name(category_name)
```
pgwillia, line 117: Added a helper to compute the short form.  This is pretty crude.  Maybe we want to be able to specify in the locales instead?

mbarnett: Yeah my first thought is that locales is the right place for this

pgwillia: Fixed using locales!

murny (approved):
Looks really good 👍. Very nice work 🎉

mbarnett (changes requested):
Ok, two more things: this should be merging to integration, not master, and we need to feature flag this off so that we can potentially merge integration and cut a release without deploying this universally

pgwillia:
- [x] feature flags
- [x] integration branch

murny (dismissed):
🎉

app/views/application/_facet_result.html.erb
```diff
@@ -11,7 +11,11 @@
         <% content_for :facet_badges do %>
           <%= link_to query_params_without_facet_value(facet_value.solr_index, facet_value.value) do %>
             <span class="badge badge-secondary p-2 mt-2">
-              <%= present(facet_value).display %>
+              <% if Flipper.enabled?(:facet_badge_category_name) %>
+                <%= "#{facet_result.short_category_name}: #{present(facet_value).display}" %>
+              <% else %>
+                <%= present(facet_value).display %>
+              <% end %>
```
mbarnett, line 18: Suggest putting this logic in the Decorator instead (since containing logic that otherwise has to live in the view is basically what it's there for). Right now `display` here just calls `DefaultFacetDecorator.display` or one of its many overrides in the subclasses in `app/decorators/facets/`, so the simplest way to do this would probably be something like doing `<%= present(facet_value).display_facet_value %>` (or some other method name), add that method to `DefaultFacetDecorator` only, with something like:

```ruby
def display_facet_value
  if Flipper.enabled?(:facet_badge_category_name)
    "{facet_value.short_category_name}: #{display}"
  else
    display
  end
end
```

This way we keep the same logic of kicking display to the customized subclasses, but hide the logic here.

app/views/application/_range_facet_result.html.erb
```diff
@@ -25,7 +25,11 @@
   <% content_for :facet_badges do %>
     <%= link_to query_params_without_range_value(range_facet_result.solr_index) do %>
       <span class="badge badge-secondary p-2 mt-2">
-        <%= range_facet_result.humanized_range %>
+          <% if Flipper.enabled?(:facet_badge_category_name) %>
+            <%= "#{range_facet_result.short_category_name}: #{range_facet_result.humanized_range}" %>
+          <% else %>
+            <%= range_facet_result.humanized_range %>
+          <% end %>
```
mbarnett, line 32: Same basic suggestion as above.

mbarnett (changes requested):
Looks good overall, just a couple of suggestions for making the approach a bit cleaner.

Would also be good to add a couple of tests to check that facets are doing the right thing with/without the flag flipped?

app/views/application/_facet_result.html.erb
mbarnett: (if you're feeling up to the extra refactoring, it might make more sense to rename the existing `display` method in DefaultFacetDecorator and all of the subclassses to something like `display_value` and then call what I called `display_facet_value` above, `display`. Then the views can keep calling `%= present(facet_value).display %>` and the method name in the base class and subclasses is arguably closer to what it actually is)

app/models/jupiter_core/solr_services/range_facet_result.rb
```diff
@@ -21,8 +18,12 @@ def humanized_range
     @range.to_s.gsub('..', ' to ')
   end
 
+  alias value humanized_range
```
mbarnett, line 21: Any reason not to go the extra step and just rename the `humanized_range` method? I don't remember if it has any other uses offhand

mbarnett (approved):
Looks good, just one small question about the method name

app/models/jupiter_core/solr_services/range_facet_result.rb
pgwillia: No any other uses.  That makes too much sense 👍

app/helpers/presenters_helper.rb
```diff
@@ -10,7 +10,8 @@ def present(obj)
     # string to class conversion dozens of times during facet rendering
     @presenter_cache ||= {}
     # FacetValues are special insofar as they dynamically specify their own presenter per-attribute-name involved
-    if obj.is_a?(JupiterCore::SolrServices::FacetResult::FacetValue)
+    if obj.is_a?(JupiterCore::SolrServices::FacetResult::FacetValue) ||
+       obj.is_a?(JupiterCore::SolrServices::RangeFacetResult)
```
mbarnett, line 14: Ok, this is the last bit of feedback, promise: can you add a TODO here to figure out a better way to do this, eg) introduce a facet superclass or something, so that we don't keep pilling on more kinds of things to check for in the future?

mbarnett (approved):

mbarnett (approved):

Issue #1990, opened 2020-11-16
OR facets in the same category


pgwillia:
# Original Goal

> As a user, I would like to be able to select multiple options in the author facet category, and
> I would like the search logic to be "OR", so I can select same author with the variation (Jane Doe or J. Doe or Jane M. Doe etc)

# What we currently have

![image](https://user-images.githubusercontent.com/1930474/72772746-04e48200-3bc2-11ea-82ed-f42c6e1ea53a.png)

- Search results are shown, with facets on left side column. 
- Can click on any "checkbox"  (checkbox in quotes as its actually a link styled with fontawesome to appear like a checkbox) to narrow search results by Facet selection. Similar with limiting by Year, you can hit the "Limit" button. 
- Once facet item clicked, we sent a get request to backend and reload a fresh page
- Fresh page has search results narrowed by facet item. Facets on side left panel have been rearranged, where selected facet is now first facet shown in list. For the selected facet it only shows the one selected item within facet which is now showing as a checked checkbox. Finally above the search results a "selected filter" button showing currently applied facets is shown.  Clicking on this button removes the selected filter and refreshes the page showing the results without this facet being applied.
![image](https://user-images.githubusercontent.com/1930474/72772992-e9c64200-3bc2-11ea-93e8-8340e97c9f5c.png)
Same events happen above on mobile, only instead of a left sidebar, we have a "Filter results" button which pulls out a side tray:
![image](https://user-images.githubusercontent.com/1930474/72773066-272acf80-3bc3-11ea-9692-46f958ce5d03.png)
![image](https://user-images.githubusercontent.com/1930474/72773047-1c703a80-3bc3-11ea-9902-3e848326e81e.png)

# Simplest solution to achieve what we want
 Ultimately we simply just want to be able to select multiple items within the SAME facet. When we do this, these items will be applied as an "OR" query in SOLR.

How to achieve this? Let's use the Author Facet in our example here. 
  - After clicking on a author, "Jane A. Doe" in the author facet, we need to render all the authors in the author facet and make them clickable again. We currently only showing the selected "Jane A. Doe" and hiding the rest. 
- When clicking on an additional author, say "Stephen A. Smith", we need update the results by "OR"ing these two authors together, while "AND"ing any other facets that have been applied. We will need to append "Stephen A. Smith" to the URL as an additional author facet. 
[...]

PR #2116, opened 2021-01-25
Default behaviour within a facet to 'OR'

pgwillia:
## Context

As a user, I would like to be able to select multiple options in the author facet category, and
I would like the search logic to be "OR", so I can select same author with the variation (Jane Doe or J. Doe or Jane M. Doe etc)

Related to #1990

## What's New

The default behaviour within a facet is to 'OR'

### DIFF
--- a/app/models/jupiter_core/search.rb
+++ b/app/models/jupiter_core/search.rb
@@ -37,7 +37,13 @@ class JupiterCore::Search
 
     base_query << q if q.present?
     facets.each do |key, values|
-      fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' ')}\))
+      if Flipper.enabled?(:or_facets)
+        fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' ')}\))
+      else
+        values.each do |value|
+          fq << %Q(#{key}: "#{value}")
+        end
+      end
     end
     ranges.each do |key, value|
       fq << "#{key}:[#{value[:begin]} TO #{value[:end]}]"
--- a/app/models/jupiter_core/search.rb
+++ b/app/models/jupiter_core/search.rb
@@ -38,7 +38,7 @@ class JupiterCore::Search
     base_query << q if q.present?
     facets.each do |key, values|
       if Flipper.enabled?(:or_facets)
-        fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' ')}\))
+        fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' OR ')}\))
       else
         values.each do |value|
           fq << %Q(#{key}: "#{value}")
--- a/app/models/jupiter_core/search.rb
+++ b/app/models/jupiter_core/search.rb
@@ -41,7 +41,7 @@ class JupiterCore::Search
         fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' OR ')}\))
       else
         values.each do |value|
-          fq << %Q(#{key}: "#{value}")
+          fq << %Q(#{key}:"#{value}")
         end
       end
     end
--- a/app/models/jupiter_core/search.rb
+++ b/app/models/jupiter_core/search.rb
@@ -37,9 +37,7 @@ class JupiterCore::Search
 
     base_query << q if q.present?
     facets.each do |key, values|
-      values.each do |value|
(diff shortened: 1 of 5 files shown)


app/models/jupiter_core/search.rb
```diff
@@ -37,9 +37,7 @@ def self.faceted_search(q: '', facets: [], ranges: [], models: [], as: nil)
 
     base_query << q if q.present?
     facets.each do |key, values|
-      values.each do |value|
-        fq << %Q(#{key}: "#{value}")
-      end
+      fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' ')}\))
```
pgwillia, line 40: https://lucene.apache.org/solr/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-BooleanOperatorsSupportedbytheStandardQueryParser

The default is OR so having the terms joined by a space will accomplish this.  Alternately these could be joined by ' OR' with the same result.

test/fixtures/items.yml
```diff
@@ -14,6 +14,22 @@ fancy:
   subject: ['Fancy things']
   member_of_paths: <%= ["#{ActiveRecord::FixtureSet.identify(:fancy_community, :uuid)}/#{ActiveRecord::FixtureSet.identify(:fancy_collection, :uuid)}"] %>
 
+practical:
```
pgwillia, line 17: Wanted another `VISIBILITY_PUBLIC` that I could use to demonstrate OR functionality.

test/integration/sitemap_test.rb
```diff
@@ -52,7 +52,7 @@ def file.original_filename
     assert_select 'loc', item_url(@item)
     assert_select 'lastmod', @item.updated_at.iso8601
     # not show private items
-    assert_select 'url', count: 4
+    assert_select 'url', count: 5
```
pgwillia, line 55: Adding the extra fixture bumps this count.

mbarnett:
Would be worth maybe setting up a meeting with Sarah to discuss the whole context of this, because I don't know off-hand if this is desirable behaviour for the Digitization work, and if it isn't then we need to do something more sophisticated then just a blanket change of the search semantics for everything?

pgwillia:
> Would be worth maybe setting up a meeting with Sarah to discuss the whole context of this, because I don't know off-hand if this is desirable behaviour for the Digitization work, and if it isn't then we need to do something more sophisticated then just a blanket change of the search semantics for everything?

I'll set up that conversation.

mbarnett:
Thanks!

mbarnett (changes requested):
carrying over observations from other PR: integration/feature flags good

pgwillia:
- [x] feature flags
- [x] integration branch

murny (approved):
Looks really good! Nice work 👍

app/models/jupiter_core/search.rb
```diff
@@ -37,8 +37,12 @@ def self.faceted_search(q: '', facets: [], ranges: [], models: [], as: nil, full
 
     base_query << q if q.present?
     facets.each do |key, values|
-      values.each do |value|
-        fq << %Q(#{key}: "#{value}")
+      if Flipper.enabled?(:or_facets)
+        fq << %Q(#{key}:\(#{values.collect { |value| "\"#{value}\"" }.join(' ')}\))
```
mbarnett, line 41: This is super pedantic, but can we join them with an explicit OR rather than just a space? I know the end result is the same, but it might be a bit clearer to the reader (me) who will inevitably forget what meaning a space has in Solr query syntax.

test/models/search_test.rb
```diff
@@ -0,0 +1,26 @@
...
+                                                   'all_contributors_sim': ['Joe Blow']
+                                                 })
+
+    assert_includes results, item1
+    assert_includes results, item2
+
+    assert_includes results.criteria[:fq],
+                    'all_subjects_sim:("Fancy things" "Practical things") AND all_contributors_sim:("Joe Blow")'
+    Flipper.disable(:or_facets)
+  end
```
mbarnett, line 24: This is very close to "test that the universe is sane" but, since we've added the codepath switch, can you do a second test to verify that when the flipper flag is disabled we get Original Flavor search results? Just so we can have some assurance that a change someone is testing against one codepath doesn't inadvertently leak into or otherwise break the other?

mbarnett (changes requested):
This looks really good, I just have a couple of small suggestions about readability and catching a possible issue with another test

mbarnett (approved):
Pretty Good!

PR #2135, opened 2021-02-02
Update Architecture Diagram

mbarnett:
## Context

We've had a lot of big changes under the hood over the last year, so let's update the diagrams to reflect that!

## What's New

- Updated Diagram to Jupiter 2.0 (removed Fedora, added OAISys and ActsAsRDFable, clarified some of the Solr and File storage relationships a bit)
- Added the Draw.io source file so I don't have to hunt it down next time

Issue #1966, opened 2020-11-04
Automate change log link verification


ConnorSheremeta:
Sometimes a PR's change log entry is missing a link to the PR/issue. Sometimes this is caught during review, sometimes it is added later on while cutting a release. It would be nice to automate this with danger on at least Jupiter.

I'm assuming we'll just want to verify the most recent entry as most projects already have missing links (or even better we could add all the missing links and verify the whole change log, will be time consuming.)

PR #2139, opened 2021-02-03
Add links missing from CHANGELOG

pgwillia:
## Context

We've got Danger complaining about dependabot PRs that should be really easy to merge.

Related to #2135 and #2049

## What's New

Added missing links.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ and releases in Jupiter project adheres to [Semantic Versioning](http://semver.o
 
 ### Added
 - Initialize disabled ReadOnlyMode [#2100](https://github.com/ualbertalib/jupiter/issues/2100)
-- Updated Architecture diagrams
+- Updated Architecture diagrams [PR#2135](https://github.com/ualbertalib/jupiter/pull/2135)
 
 ### Removed
 – Remove entirely unnecessary config file. [PR#2044](https://github.com/ualbertalib/jupiter/pull/2044)
-– Completely disable logging of warnings around the "excel spreadsheet" issue
+– Completely disable logging of warnings around the "excel spreadsheet" issue [PR#2049](https://github.com/ualbertalib/jupiter/pull/2049)
 
 ### Changed
 - Added DOI reset feature for admins [#1739](https://github.com/ualbertalib/jupiter/issues/1739)


mbarnett (changes requested):
(one sec typing something long)

mbarnett:
Ok! 

Not to complicate a straightforward PR that solves an issue but – I guess we have to have the conversation somewhere and this might as well be the place (none of this is to pick on you at all Tricia - this is just as good a spot as any for this discussion):

Are we gaining anything from doing this Danger check? What is the value we're gaining here? Is the workflow required to meet this requirement sensible?

Linking to Issues with changes _which have issues associated with them_ I think has value and it makes sense to me.

You could argue that every single change that ever happens in a codebase _should_ have an issue associated with it, but in practice I don't think that's realistic – emergencies happen, sometimes people just notice something that needs to be fixed, a bug is mentioned in passing in a meeting and someone slaps a PR together in-place right then and there.

Maybe these actions aren't theoretically ideal, but I kind of value the flexibility and forward progress more than I value the rote procedure of "now that you've done the work, go create some synthetic paperwork to justify that work in an issue, so that we can put a link in a changelog to make Danger happy". That to me feels like friction that discourages taking the right action quickly in favor of something like bureaucracy. My bias is in favor of reducing friction that impedes improvements.

Long story short, the idea behind this Danger check seems to have become that if a PR doesn't have an issue to link to, the CHANGELOG should link to the PR itself – but unless I'm missing something that's in its own way a frustratingly circular chunk of friction? You have to open the PR to learn what the PR link will be, and then edit the PR you just opened with the link you just created to link to the PR you just opened, while Danger yells at you for not linking to the thing you couldn't know the link to before Danger had a chance to yell at you?

So that leads me to want to deal with the larger questions: 

– Are we just satisfying pointless busywork here to make Danger happy instead of changing Danger's config to eliminate the busywork? 

– Is there concrete value in doing this? It's always possible to find the PR that the change in the CHANGELOG came from just by blaming the line, so the link feels very low-value to me?

pgwillia:
> Are we gaining anything from doing this Danger check? What is the value we're gaining here? Is the workflow required to meet this requirement sensible?

I think we are gaining something from this Danger check.  It reminds us to be disciplined in our creation and completion of work.  Having the reference to the why of something is really important to informing future changes and understanding behaviour of our system.

> You could argue that every single change that ever happens in a codebase _should_ have an issue associated with it, but in practice I don't think that's realistic – emergencies happen, sometimes people just notice something that needs to be fixed, a bug is mentioned in passing in a meeting and someone slaps a PR together in-place right then and there.

I think this is the situation where it is especially important to be disciplined and do all the right things because these are the situations where there is pressure to complete something fast and there is context that is important to the change.  I agree that an issue might not be created, but then the onus should be on the PR to be that much better at explaining the context and is a valuable part of our documentation. 

> Maybe these actions aren't theoretically ideal, but I kind of value the flexibility and forward progress more than I value the rote procedure of "now that you've done the work, go create some synthetic paperwork to justify that work in an issue, so that we can put a link in a changelog to make Danger happy". That to me feels like friction that discourages taking the right action quickly in favor of something like bureaucracy. My bias is in favor of reducing friction that impedes improvements.

I am not in favor of increasing friction, but I think it's tolerable in this case because of the nudge towards discipline. 

> Long story short, the idea behind this Danger check seems to have become that if a PR doesn't have an issue to link to, the CHANGELOG should link to the PR itself – but unless I'm missing something that's in its own way a frustratingly circular chunk of friction? You have to open the PR to learn what the PR link will be, and then edit the PR you just opened with the link you just created to link to the PR you just opened, while Danger yells at you for not linking to the thing you couldn't know the link to before Danger had a chance to yell at you?

I think that opening a PR and then editing the CHANGELOG with the link is not that big of deal.  Especially considering the ease of use of Github for doing so.  I expect this would be the edge case rather than the general case so I'm okay with living with the occasional discomfort.

> 
> So that leads me to want to deal with the larger questions:
> 
> – Are we just satisfying pointless busywork here to make Danger happy instead of changing Danger's config to eliminate the busywork? Is there anything smarter to do here? Could/Should we just add something to not warn if a CHANGELOG entry is marked with something to indicate no issue?

Maybe?  Like a chores section that's excluded?  I'm reluctant to endorse this approach because it erodes the discipline that Danger is supposed to nudge us towards.  Suddenly everything is an exception so we don't have to do documentation.

> 
> – Is there concrete value in doing this? It's always possible to find the PR that the change in the CHANGELOG came from just by blaming the line, so the link to the same thing the blame links to anyways feels very low-value to me?

If I recall one of the reasons we created the Changelog was for the Service Managers and UNIX team who might not be as familiar with git blame as we would be.  

If we put more emphasis on structured commit messages that linked to issues and clearly stated the type of change then I could see us making the Changelog and it's busywork obsolete.

mbarnett:
> It reminds us to be disciplined in our creation and completion of work. Having the reference to the why of something is really important to informing future changes and understanding behaviour of our system.

I agree, but if something is an "important why", containing information we need to know when working on the codebase, it really ought to be, and we ought to be disciplined about making sure it is, living in the codebase, not hiding in a PR description (and I try to raise this whenever I see it happening – if there's valuable info in your PR description or comments, please put those in the code instead).

If we relegate important info to things like PRs, what happens if we have to switch to Gitlab or something for budgetary or ethical reasons in the future? What happens if GitHub goes out of business for one reason or another?

The codebase is the proper location for long-term knowledge; the PR really shouldn't contain information of value beyond providing context to the reviewer _during the review_

> Maybe? Like a chores section that's excluded? I'm reluctant to endorse this approach because it erodes the discipline that Danger is supposed to nudge us towards. Suddenly everything is an exception so we don't have to do documentation.

Again, I'm a bit concerned that it sounds like we're thinking of PRs as a long-term home for important information, and they really really shouldn't be. If Danger is nudging us in the direction of treating them as such, I'm even less in favor of that nudge?

> If I recall one of the reasons we created the Changelog was for the Service Managers and UNIX team who might not be as familiar with git blame as we would be.

For what it's worth, I'm not aware of any Service Manager or Unix Team member who uses the CHANGELOG to routinely follow links to any issue or PR, nor am I convinced that the PRs they find at the end of links would be intelligible to them (not should they be – PRs should be written for a Developer audience, not laypeople)

mbarnett:
I will say that if you're concerned that people would start making _everything_ a "no link" CHANGELOG entry when there's an issue they could or should be linking, that that's still going to be part of the PR and certainly something you can request changes over, just like any other "I feel like there's missing documentation of what's going on here" type feedback that we should always be giving to make sure we're capturing important information in the code?

pgwillia:
Don't confuse anything that I'm saying here as a strong opinion.  I think there are other valid ways to accomplish documentation and change management.  I thought this https://github.com/ualbertalib/jupiter/issues/1966 was a good idea when it was suggested and was happy to endorse and work within those boundaries.  If that's not the consensus of the whole team I'm happy to see it evolve into something more workable.

I think Danger and Rubocop have value exactly because they're nags from not a person on our team. No one wants to be the nagger.

mbarnett (approved):
Summarizing the consensus we reached on Zoom & Slack:

> Just talked to Tricia and landed basically exactly where you are: we think the audience for CHANGELOG is mostly us and maybe it gets used as the basis of an email about what’s in a release that we might send, so we’re thinking: 1) get rid of the requirement for a link, make Danger otherwise required (I dropped it to optional yesterday) and 2) really try to write more detailed entries in the changelog rather than link out to a PR or Issue, so that all the info is in one place. https://github.com/apple/swift/blob/main/CHANGELOG.md was my example of a nice one

> Not all of the swift ones are backed by an RFC, some are just info on their own. It’s super extreme and not something I think we’d fully hit. The main idea was just that the changelog would be more self-contained, summarizing an issue or PR context rather than linking out to it, so that you could basically understand it on its own.

I think we can merge this and then I'll separately open a PR to adjust Danger and maybe but some documentation of the new approach somewhere.

PR #2182, opened 2021-02-19
Revert "Bump ransack from 2.4.1 to 2.4.2"

murny:
Reverts ualbertalib/jupiter#2113

Ransack v2.4.2 drops support for Ruby v2.5.

Production is still running this version so we need to support Ruby 2.5. So reverting this


pgwillia (approved):
Thanks for doing this @murny! 👍

Issue #2119, opened 2021-01-26
Find a sensible way to handle many new URI/literals


pgwillia:
Let's add a TODO to consider more robust ways of doing this (lookup table in DB or some kind of authority query), because I could see this growing quite large.

_Originally posted by @mbarnett in https://github.com/ualbertalib/jupiter/pull/2089#discussion_r564123624_

Peel will introduce many new URIs in it's metadata.  This is different than the Controlled Vocabularies used before where there were a limited/restricted permitted responses that we drew from for submitting items. The part that is similar is the need to convert to/from URI to humanized string.

pgwillia:
# Problem
Peel will introduce many new URIs in it's metadata. This is different than the Controlled Vocabularies used before where there were a limited/restricted permitted responses that we drew from for submitting items. The part that is similar is the need to convert to/from URI to humanized string. 

I think that having an intermediate symbol is desirable in URI translation so that talking about special cases in our code remain readable.  I.e. `CONTROLLED_VOCABULARIES[:era][:item_type].article`

# What are we doing?

We have a number of [config/controlled_vocabularies](https://github.com/ualbertalib/jupiter/tree/master/config/controlled_vocabularies) which are used to set up the app-wide `CONTROLLED_VOCABULARY` hash.

Each of the URIs in the controlled vocabularies has a symbol which matches the same symbol in our [locales](https://github.com/ualbertalib/jupiter/blob/523f6a2aac7e63fade752384222ee632186f3936/config/locales/en.yml#L776-L831) to give a humanized literal
https://github.com/ualbertalib/jupiter/blob/523f6a2aac7e63fade752384222ee632186f3936/app/helpers/application_helper.rb#L4-L13

We use this to 
* [validate that a uri is part of our controlled vocabulary](https://github.com/ualbertalib/jupiter/blob/master/app/validators/uri_validator.rb).
* translate the URI to a human readable [literal for facets](https://github.com/ualbertalib/jupiter/blob/master/app/decorators/facets/uri_value.rb)
* Control Vocab Conversions
  * [licences](https://github.com/ualbertalib/jupiter/blob/523f6a2aac7e63fade752384222ee632186f3936/app/models/draft_item.rb#L200-L221)
[...]

PR #2145, opened 2021-02-08
Namespace for Controlled Vocabularies

pgwillia:
## Context

We're going to be adding digitization which will have it's own controlled vocabulary.  First we move everything ERA to it's own namespace.

Related to #2118

## What's New
1.  Changed API

From `translate_uri(vocab, uri)`

To `translate_uri(namespace, vocab, uri)`

And

From `humanize_uri(vocab, uri)`

To `humanize_uri(namespace, vocab, uri)`

And

From `validates :languages, presence: true, uri: { in_vocabulary: :language }` or   `validates :license, uri: { namespace: :era, in_vocabularies: [:license, :old_license] }`

To `validates :languages, presence: true, uri: { namespace: :era, in_vocabulary: :language }` or   `validates :license, uri: { namespace: :era, in_vocabularies: [:license, :old_license] }`

2. Added `:era` namespace
* config/controlled_vocabularies
* config/locales/en.yml#controlled_vocabularies

PR #2188, opened 2021-02-19
WIP  Vocabulary to manage URIs 

pgwillia:
## Context

Peel will introduce many new URIs in it's metadata. This is different than the Controlled Vocabularies used before where there were a limited/restricted permitted responses that we drew from for submitting items. The part that is similar is the need to convert to/from URI to humanized string.

Related to #2119

## What's New

1. Added the i18n ActiveRecord backend.  
- To new/save or create `I18n::Backend::ActiveRecord::Translation.create(locale: 'en', key: 'vocabulary.digitization.subject.edmonton', value: 'Edmonton (Alta).')`
- Then you can read it `I18n.t 'vocabulary.digitization.subject.edmonton'` You may need to `I18n.backend.reload!` if you try to read a key before it's been saved (because missing is cached).

2. Created a Vocabulary Model
- To create `Vocabulary.create(namespace: 'era', vocab: 'subject', uri: 'http://id.loc.gov/authorities/names/n79007225')`
- To get the `uri` `Vocabulary.namespace(:era).vocab(:subject).code(:edmonton_alta)`
- To get the `code` `Vocabulary.namespace(:era).vocab(:subject).from_uri('http://id.loc.gov/authorities/names/n79007225')`
- To get the label `I18n.t 'vocabulary.era.subject.edmonton_alta'`

## TODO
- [x] changelog
- [ ] some debuging
- [ ] tests with vcr recording of graph collection
- [ ] fixtures


mbarnett:
Alright, I think we’re headed in the wrong direction here. 

1) The problem as I see it is that the URI-> human readable string problem is *similar to* the i18n problem, but it’s not the *same as* the i18n. 

One way I think this shows up is that in trying to make the URI situation better here, we’re making the i18n situation worse: we use i18n translations for all user facing static text throughout the interface, and now each of those uses is a separate database query – ie) a page load now generates dozens of new queries and were drastically increasing database load for no benefit to current uses of i18n, because generally the only reason you’d put those in the DB is to allow people to translate the interface on the fly, which we’re definitely not doing. Sure, we could spend even more time building a bunch of caching around that, but that’s adding complexity on top of complexity just to try to get back to the status quo. 

So our solution to the URI problem should be different from the i18n problem, so that we don’t make the i18n situation and general performance worse than it has to be. 

2) I think we’re overthinking the namespaces. We really just want 2, one for ERA and one for Digitization. If we need a third, it will be as part of a new property build out. We’re never going to want to dynamically generate namespaces, so it doesn’t make sense to add the overhead of a namespace look up query to ever uri query – we should just hardcore this.

mbarnett:
One thing that might simplify things a lot is that where the “URI -> internal symbol, then internal symbol -> human label” lookup made some sense for ERA where we needed to build deposit interfaces around the URIs, I’m not sure it has any value for the Digitization URIs, where we’re not doing unmediated deposit?

All we really want to solve is a direct URI -> label mapping. The more I think about the more I wonder if the whole database idea isn’t overkill at this point – what we could do is just add directory name spacing to the current controlled vocab setup and then, for things like location URIs, use a script to generate direct “label name”: uri yml mappings

pgwillia:
> what we could do is just add directory name spacing to the current controlled vocab setup and then, for things like location URIs, use a script to generate direct “label name”: uri yml mappings. If we see how far we can get with the simple solution first, we’ll have better information for designing a better API if it becomes unmanageable?

Cool.  Let's do some course correction.  For sure we should take the YAGNI approach to not over-engineer this.

Let me see if I understand.  Is this the [directory name spacing](https://github.com/ualbertalib/jupiter/pull/2145) you mean?  

What would the script do? Would it be part of the [controlled vocabularies initializers](https://github.com/ualbertalib/jupiter/blob/748f97543aba984968a0082cd7a9ed666c22ee30/config/initializers/controlled_vocabularies.rb#L19-L26)? Would it create a i18n locale yml file for the vocabulary from the URIs in the `CONTROLLED_VOCABULARY`? Or are we not going to do the same thing as ERA with using i18n for translation?  Would it just be a db table that has uri and label?  Would the labels also be a constant data structure stored in memory?

mbarnett:
> Let me see if I understand. Is this the directory name spacing you mean?

Yep, that's what I had in mind.

> What would the script do? Would it be part of the controlled vocabularies initializers? 

I'm thinking right now we build the controlled vocabularies by hand. The script idea is just that we can automate that for things like Locations where there's a bunch of opaque URIs that you want to look up in SKOS or some other authority -> a quick script could be used to pull the URIs we're given by metadata and set up the yml file. Nothing in the initializer or anything, nothing that runs *in* the web-app, just a simple offline tool to deal with creating these longer files.

> Would it create a i18n locale yml file for the vocabulary from the URIs in the CONTROLLED_VOCABULARY? Or are we not going to do the same thing as ERA with using i18n for translation? 

I need to think more about how or whether to special-case ERA or some vocabs to deal with translations for its URIs if they really do make less sense for Digitization's URIs. Let me get back to you with an answer here, but assume right now you'd just be generating a vocab's yml file and not corresponding translations, so instead of `no_linguistic_content: http://id.loc.gov/vocabulary/iso639-2/zxx` and a corresponding `no_linguistic_content: 'No linguistic content'` you might generate `'No linguistic content': http://id.loc.gov/vocabulary/iso639-2/zxx`.


> Would it just be a db table that has uri and label? 

We _could_ do that, but I'm really doing a 180 and thinking we should just avoid putting this in the DB entirely for now. We're going to need to do tons of URI->label lookups in the hot facet loop, and if we put it in the DB we're committing to building out not just memoization (which would catch the same URI lookup twice in one request) but cross-thread caching (so that multiple requests don't have to keep hitting the DB to look up something that isn't changing) to keep that decently fast.

> Would the labels also be a constant data structure stored in memory?

So right now the URI->symbol mappings live in the CONTROLLED_VOCABULARIES hash at runtime – I'm thinking in this case for at least some vocabs that same hash will simply contain straight URI->label key-value pairs. How to handle the distinction in lookup is the only part I need to put a bit of thought into.

pgwillia (commented):
> I'm thinking right now we build the controlled vocabularies by hand. The script idea is just that we can automate that for things like Locations where there's a bunch of opaque URIs that you want to look up in SKOS or some other authority -> a quick script could be used to pull the URIs we're given by metadata and set up the yml file. Nothing in the initializer or anything, nothing that runs _in_ the web-app, just a simple offline tool to deal with creating these longer files.

Is a spreadsheet in Google drive going to be the main way that we receive metadata?  Can I use the 'FolkFestTriples - v.1' in a prototype?

Issue #1322, opened 2019-10-03
Support HTML code in community/collection description fields


anayram:
Many "Description or abstract" fields (at the Item level), as well as Description fields (Communities and Collections) contain HTML tags. Because these are text fields, HTML is not rendered in the UI and text looks garbled and it's way less readable than ideal.

Support HTML markup, specifically bold/strong text highlighting, italics, line breaks, and paragraphs entered in:
- [x] Description field (Collection and Community levels)
- [x] Description or Abstract field (Item level)

Examples of items with HTML tags:
Ad $ Summary
https://era.library.ualberta.ca/admin/communities/b41cdbfd-6af2-4a13-8ba6-59725565d445/collections/2c1f3d8f-376a-45f8-a422-2afc3363987d

Board Practices
https://era.library.ualberta.ca/admin/communities/b41cdbfd-6af2-4a13-8ba6-59725565d445/collections/71f4b544-22f3-436f-bed0-c215df838be9

Electrical and Computer Engineering, Department of
https://era.library.ualberta.ca/admin/communities/55bafdf5-db7f-48a8-a91e-69b8c01cc9fb

Platts Historical US Domestic Crude Prices
https://era.library.ualberta.ca/admin/communities/b41cdbfd-6af2-4a13-8ba6-59725565d445/collections/b73484ff-194c-4aa5-9011-95410f02ee34

anayram:
Related to #1295

seanluyk:
@mbarnett not sure what our resources are for these kinds of enhancements at the moment, but I'd like to flag this as a very desirable improvement for ERA. Rather than HTML tags for formatting, I think something like Markdown would work really well since that's already used in many of the tools staff working in repositories are familiar with

mbarnett:
Makes sense! HTML is always a bit of a security risk but Markdown ought to be both more doable and fairly low effort. 

I’ll refine this out a bit in the next day or two and then flag it as a good issue for some to pick up when they’re between larger priorities.

seanluyk:
Excellent, thanks, Matt!

mbarnett:
Scope/Notes:

– We convert Markdown->HTML on page render (so don't store the HTML or anything)
– To make this a lot less effort to achieve, no preview functionality initially. Re-edit it if your Markdown isn't great, etc.
– we'll want to do this by building/enhancing decorators for Communities et al using https://github.com/vmg/redcarpet to intercept and render the appropriate field.
[...]

PR #2230, opened 2021-03-11
WIP Use markdown to style some of our descriptions

pgwillia:
## Context

Many "description" or "abstract" fields (at the Item level as well as Communities and Collections) contain HTML tags. Because these are text fields, HTML is not rendered in the UI and text looks garbled and it's way less readable than ideal.

Markdown should work really well for this since that's already used in many of the tools staff working in repositories are familiar with.

Related to #1322

## What's New

- Added the redcarpet gem because it basically does what we want with markdown.  
- add `markdown` and `strip_markdown` helpers
- updated views to use the new `markdown` helper on description and abstract fields
- updated SolrExporters to use the new `strip_markdown` helper to remove content being indexed

### DIFF
--- a/app/helpers/markdown_helper.rb
+++ b/app/helpers/markdown_helper.rb
@@ -1,0 +1,13 @@
+module MarkdownHelper
+  def markdown(text)
+    Jupiter::RenderMarkdown.render(text).html_safe
+  rescue StandardError
+    text.html_safe
+  end
+
+  def strip_markdown(text)
+    Jupiter::StripMarkdown.render(text).html_safe
+  rescue StandardError
+    text.html_safe
+  end
+end
--- a/config/initializers/markdown.rb
+++ b/config/initializers/markdown.rb
@@ -1,0 +1,24 @@
+require 'redcarpet'
+require 'redcarpet/render_strip'
+
+module Jupiter
+  options = {
+    filter_html: true,
+    no_images: true,
+    no_styles: true,
+    hard_wrap: true,
+    link_attributes: { rel: 'noopener noreferrer', target: '_blank' }
+  }
+
+  extensions = {
+    lax_spacing: true,
+    fenced_code_blocks: true,
+    tables: true,
+    autolink: true
+  }
+
+  StripMarkdown = Redcarpet::Markdown.new(Redcarpet::Render::StripDown)
+
+  renderer = Redcarpet::Render::HTML.new(options)
+  RenderMarkdown = Redcarpet::Markdown.new(renderer, extensions)
+end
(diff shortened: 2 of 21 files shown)


app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  rescue StandardError
+    text.html_safe
+  end
+
+  def strip_markdown(text)
+    Jupiter::StripMarkdown.render(text).html_safe
+  rescue StandardError
+    text.html_safe
+  end
+end
```
mbarnett, line 13: I'd like to see `markdown` move into the decorators layer, and `strip_markdown` into the solr exporter layer.

That way, instead of having to remember to wrap a markdown call around all calls to `description` in all of these views, the right thing happens transparently and we can't accidentally not do it in new views.

Generally speaking, any time you find yourself starting to wrap methods around fields in views, what you really want is to wrap that field in the decorator and leave the views alone.

config/initializers/markdown.rb
```diff
@@ -0,0 +1,24 @@
...
+    fenced_code_blocks: true,
+    tables: true,
+    autolink: true
+  }
+
+  StripMarkdown = Redcarpet::Markdown.new(Redcarpet::Render::StripDown)
+
+  renderer = Redcarpet::Render::HTML.new(options)
+  RenderMarkdown = Redcarpet::Markdown.new(renderer, extensions)
+end
```
mbarnett, line 24: `application_decorator` is probably a better location for this, as opposed to setting up a special module in an initializer.

The instinct to reuse the render per their advice is good, but this could just be a constant in the `application_decorator`  for the HTML render, and in base exporter for the Solr. Then `markdown` and `strip_markdown` could be methods in those two places, respectively.

test/helpers/markdown_helper_test.rb
```diff
@@ -0,0 +1,158 @@
...
+            Eos.
+            Facere.
+            Qui.
+            Qui.
+    STRIPPED
+
+    assert_equal stripped_text, strip_markdown(RAW_MARKDOWN)
+  end
+
+end
```
mbarnett, line 158: This really just tests that RedCarpet does what it says on the tin, I think?

Maybe a better approach here would be (at the unit test level) to instantiate an Item or Collection and then either construct a decorator for them and check its output on `description` (and then grab its `solr_exporter` and check that the doc (it's just a hash) that gets generated by `export` has the right output), and then an integration test to check pages/search results seem to work ok for markdown'd items?

Also probably worth adding some markdown'd fixtures.

mbarnett (changes requested):
This is a good first crack, just some feedback about how we can architect this in a way that makes doing the right thing more effortless, and some suggestions for beefing up the testing approach.

PR #2231, opened 2021-03-11
Fix a couple tiny issues in seeds file 

murny:
- Seeds was failing due to tiny mistake on ControlledVocab code. This PR fixes this.
- Fix a depreciation message with `Kernal#open` vs `URI#open`
-  Added some tiny tweaks to Procfile.dev file for quality of life

Created https://github.com/ualbertalib/jupiter/issues/2232 for a few other pain points I saw when running the seeds

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,3 @@
-app: bin/rails s
+app: bin/rails s -p 3000
 frontend: bin/webpack-dev-server
-worker: sidekiq
+worker: bundle exec sidekiq
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -82,7 +82,7 @@ if Rails.env.development? || Rails.env.uat?
         set = (thing == 'cat') ? 'set4' : 'set1'
         url = Faker::Avatar.image(slug: thing, size: "100x100", format: "png", set: set)
         File.open(filename, 'wb') do |fo|
-          fo.write open(url).read
+          fo.write URI.open(url).read
         end
       end
     end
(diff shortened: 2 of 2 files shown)


db/seeds.rb
```diff
@@ -82,7 +82,7 @@
         set = (thing == 'cat') ? 'set4' : 'set1'
         url = Faker::Avatar.image(slug: thing, size: "100x100", format: "png", set: set)
         File.open(filename, 'wb') do |fo|
-          fo.write open(url).read
+          fo.write URI.open(url).read
```
murny, line 85:
```
/Users/murny/Code/ualberta/jupiter/db/seeds.rb:85: warning: calling URI.open via Kernel#open is deprecated, call URI.open directly or use URI#open
```

Was getting this warning message. So let's just specify `URI.open` instead of using `Kernal#open`

db/seeds.rb
```diff
@@ -726,11 +726,11 @@ def file2.original_filename
     resource_type: ControlledVocabulary.digitization.resource_type.from_value("Text"),
     genre: [ControlledVocabulary.digitization.genre.from_value("Programs (Publications)")],
     language: [ControlledVocabulary.digitization.language.from_value("English")],
-    publisher: [ControlledVocabulary.digitization.publisher.from_value("Edmonton Folk Music Festival")],
+    publisher: [ControlledVocabulary.digitization.subject.from_value("Edmonton Folk Music Festival")],
```
murny, line 729: We have no `ControlledVocabulary.digitization.publisher` in our controlled vocab yml files for this? From looking at our Fixtures, I assume we want `subject` instead

Procfile.dev
```diff
@@ -1,3 +1,3 @@
-app: bin/rails s
+app: bin/rails s -p 3000
```
murny, line 1: By default, Foreman uses port 5000, so changed this to be 3000 instead

Procfile.dev
```diff
@@ -1,3 +1,3 @@
...
 frontend: bin/webpack-dev-server
-worker: sidekiq
+worker: bundle exec sidekiq
```
murny, line 3: Depends on how your local is setup but typically want to be explicit and use `bundle exec`

mbarnett (approved):

Issue #2232, opened 2021-03-11
Pay down "db/seeds.rb" technical debt


murny:
Some pain points I saw myself is:

- Currently our seeds are taking a very long time to run. I timed a few runs and it was taking close to 4 minutes for myself. However, we only seeding maybe 200-300 records so this shouldn't be take this long 🤔 ? Perhaps looking into caching/using images from local instead of the internet might be an easy win here.

-  We have almost 800 lines of code in this file. Which is starting to get hard to maintain. As an example we started putting `dev` seeds outside of the `dev/uat` block (https://github.com/ualbertalib/jupiter/blob/master/db/seeds.rb#L9) which is meant for production only seeds. This could be potential issue if these seeds are being ran in production.
   - Could be a good time to look into breaking out seeds out into separate files to help with readability and maintenance in the future (Another alternative is maybe using a gem could help us too like [seed-fu](https://github.com/mbleigh/seed-fu) but I'd probably just break these seeds into separate files first)

- db:seed currently can only be ran once. Running db:seed again will cause a crash.

mbarnett:
Sank a bit of time into this this afternoon:

– runtime about triples just from having Docker running for 2 hours, on my machine, so it's definitely faster with a fresh docker process. Jeff suggested https://engageinteractive.co.uk/blog/making-docker-faster-on-mac#why-is-it-slower might be of value

– there's a hot loop of 5 "THINGS" and in that loop another loop of 20 item/thesis creations, meaning we do a bunch of stuff 100 times that only needs to be done one time, like check if files exist, generate fake strings, create hashes, etc etc etc. This could all be rationalized.

mbarnett:
Jeff's definitely on to something – bouncing docker didn't fix anything, it was still taking almost 3 minutes. I then did a `docker-compose down -v` to clear the volumes, brought it back up fresh, and seeding took 20 seconds flat.

PR #2237, opened 2021-03-15
fix scope for digitization and radioactive thesis

pgwillia:
## Context
 
We started putting dev seeds outside of the dev/uat block (https://github.com/ualbertalib/jupiter/blob/integration/db/seeds.rb#L9) which is meant for production only seeds. This could be potential issue if these seeds are being ran in production.

The better fix will be a refactor for better readability and grouping but this will add a safety guard until we get to that.

Related to #2232

## What's New

* Moved the radioactive thesis and Digitiation seeds in the dev/uat scope and added a comment about the end of the scope.
* Ran rubocop against the seeds

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,14 +7,14 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 if Rails.env.development? || Rails.env.uat?
-  require "open-uri"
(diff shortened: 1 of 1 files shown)


murny (approved):
Nice work! Good job validating this file against rubocop 👍

PR #2238, opened 2021-03-15
Add missing step for project setup in README.md

lagoan:
The documentation for setting up the project was missing the set of
populating the node_modules directory to get jupiter to run for the
first time.

## Context

Missing step for initial setup.

## What's New

An extremely small change in documentation. This description contains
more text that the actual PR :smile:


murny (approved):
Good catch! We definitely need to `yarn install` 😄

PR #2239, opened 2021-03-16
Use markdown to style some of our descriptions

pgwillia:
## Context

Many "description" or "abstract" fields (at the Item level as well as Communities and Collections) contain HTML tags. Because these are text fields, HTML is not rendered in the UI and text looks garbled and it's way less readable than ideal.

Markdown should work really well for this since that's already used in many of the tools staff working in repositories are familiar with.

Related to #1322
Replaces #2230

## What's New
- Added the redcarpet gem because it basically does what we want with markdown.  
- add `markdown` and `strip_markdown` to ApplicationDecorator
- updated SolrExporters to use the new `strip_markdown` helper to remove content being indexed

## Nits
* Faker markdown tables aren't parsed as tables by redcarpet.  Seems that adding a newline before the table is all that's needed as a workaround though.  This means that some text generated by `Faker::Markdown.sandwhich` isn't nice until it's edited to add the extra newline.


app/views/collections/show.html.erb
```diff
@@ -29,7 +29,7 @@
 
   <% if @collection.description.present? %>
     <div class="p-3">
-      <p><%= @collection.description %></p>
+      <p><%= @collection.decorate.description %></p>
```
pgwillia, line 32: I thought I decorated this in the controller, but it didn't seem to take.

app/views/communities/_community.html.erb
```diff
@@ -6,7 +6,7 @@
     <%= render partial: 'thumbnail', locals: { object: community } %>
     <div class="media-body ml-3">
       <h3 class="mt-0"><%= community.title %></h3>
-      <p class="text-muted"><%= community.description %></p>
+      <p class="text-muted"><%= community.decorate.description %></p>
```
pgwillia, line 9: I thought I decorated this in the controller, but again it didn't seem to take.

app/helpers/application_helper.rb
```diff
@@ -59,6 +59,6 @@ def search_link_for(object, attribute, value: nil, facet: :facet, display: nil)
   end
 
   def jupiter_truncate(text, length: TRUNCATE_CHARS_DEFAULT, separator: ' ', omission: '...')
-    truncate text, length: length, separator: separator, omission: omission
+    truncate strip_tags(text), length: length, separator: separator, omission: omission
```
pgwillia, line 62: There are a couple of places where rendering the markdown doesn't make sense.  The meta description of the page and the truncated search results (where we can't ensure matching tags).

pgwillia (commented):
There are a few things I'm stumped by.  Hopefully @mbarnett et al can offer some counsel.

Another thing I'm struggling with is un-decorating the values used in the simple form/ wizard.  Right now the value that is displayed is decorated.  We want the input field in the form to not be decorated by the preview to be decorated.  I can't wrap my head around where that should happen or how. https://github.com/ualbertalib/jupiter/blob/83b42712135ccf59e111e211ea176a10a60ad73a/app/views/items/draft/describe_item.html.erb#L90-L94

app/views/application/_results.html.erb
```diff
@@ -80,7 +80,7 @@
         <ul class="list-group">
           <% results.each_with_fulltext_results do |result, fulltext_hits| %>
             <li class="list-group-item list-group-item-action">
-              <%= render result, highlights: fulltext_hits %>
+              <%= render result.decorate, highlights: fulltext_hits %>
```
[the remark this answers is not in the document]
mbarnett: Hmm, this is a bit of a weird usecase versus "normal" ActiveRecord which Draper has some helpers for, and I think, after reviewing the Draper docs, that decorating it at the call to render the way you have is actually the best approach. Anything earlier forces us into iterating the collection multiple times, and runs counter to their recommendations in a lot of ways.

app/controllers/communities_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
   end
 
   def show
-    @community = Community.find(params[:id])
+    @community = Community.find(params[:id]).decorate
```
mbarnett, line 27: In terms of establishing good patterns, I don't think this is quite what we want. Pulling from the Draper docs:

> 
> Decorators are supposed to behave very much like the models they decorate, and for that reason it is very tempting to just decorate your objects at the start of your controller action and then use the decorators throughout. Don't.
> 
> Because decorators are designed to be consumed by the view, you should only be accessing them there. Manipulate your models to get things ready, then decorate at the last minute, right before you render the view. This avoids many of the common pitfalls that arise from attempting to modify decorators (in particular, collection decorators) after creating them.
> 

What you've done absolutely works, but I think we'd benefit from doing things the way Draper recommends, to establish good decorator hygiene right away and not set ourselves up for future problems down the road. In this case that would look something like:

```ruby
    community = Community.find(params[:id])
    authorize community

    respond_to do |format|
      format.html do
        @collections = community.collections.order(Collection.sort_order(params)).page params[:page]
        @community = community.decorate
      end
      format.js do
        # Used for the collapsable dropdown to show member collections
        @collections = community.collections.order(Collection.sort_order(params))
        @community = community.decorate
      end

      format.json do
        # Used in item_draft.js
        collections = community.collections.order(title: :asc)
        collections = collections.select { |c| c.restricted.blank? } unless current_user&.admin?
        render json: community.decorate.attributes.merge(collections: collections)
      end
```

app/views/collections/show.html.erb
mbarnett: Try sticking a `<% binding.pry %>` right above this, and see what `@collection` is? I show it as 

```ruby
[1] pry(#<#<Class:0x00007fd1eb26f9c0>>)> @collection
=> #<CollectionDecorator:0x00007fd1ac1c8420
 @_routes=nil,
 @context={},
 @object=
  #<Collection:0x00007fd1ac1c8a60
   id: "32157177-bbc8-4e1f-9a09-d4fbea62df
```

etc

app/views/communities/_community.html.erb
mbarnett: I think you missed this one in the controller, actually.

We're getting here via:

https://github.com/ualbertalib/jupiter/blob/7099d30cd11b8d06d96f6a713f8ec55ae52b49a5/app/views/communities/index.html.erb#L38

which means you needed to decorate the collection in `CommunityController#index` via the Draper helper methods for decorating collections of ActiveRecord objects.

mbarnett (changes requested):
This is definitely a better approach with the decorators, and I added some feedback about how we can conform to their suggestions a bit more idiomatically.

I'm not 100% sure what's going on with the Collection show page. We might need to hop on a call Monday if pry doesn't help you figure it out. I did figure out the issue with the community one though.

app/decorators/item_decorator.rb
```diff
@@ -9,4 +9,12 @@ def history
     history.select { |humanized_change_set| humanized_change_set.html_diffs.present? }
   end
 
+  def description
+    markdown(model.description)
+  end
+
+  def plain_description
```
mbarnett, line 16: Maybe "raw" would be more descriptive than "plain"? Or "unrendered" or something?

app/decorators/community_decorator.rb
```diff
@@ -0,0 +1,13 @@
+class CommunityDecorator < ApplicationDecorator
+
+  delegate_all
+
+  def self.collection_decorator_class
+    PaginatingDecorator
+  end
```
mbarnett, line 7: I'm wondering if there's any reason you can think of not to hoist this up into `ApplicationDecorator`? It seems like a generic solution for all such decorators so I'm thinking we could just make it the default, because I can't think of anything obvious it might break.

app/views/application/_item.html.erb
```diff
@@ -48,7 +48,7 @@
 
     <% if item.description.present? %>
       <p>
-        <%= render(highlights) || jupiter_truncate(item.description) %>
+        <%= render(highlights) || jupiter_truncate(item.plain_description) %>
```
mbarnett, line 51: Don't we want rendered markdown here? I'm thinking that this going to lead to some funky looking search results with various raw markdown characters in them?

app/decorators/item_decorator.rb
mbarnett: Further thinking about this, and the `preview_markdown` in Draft*, I'm thinking it would be nice if these were a clear antonym pair, for discoverability? `unrendered_` here and `rendered_` in the Drafts? (or cooked/raw, but that's more of a stretch and less clear?)

app/decorators/application_decorator.rb
```diff
@@ -11,4 +13,29 @@ class ApplicationDecorator < Draper::Decorator
   # URL helpers, we can ensure the decorated methods behave consistently.
   include Rails.application.routes.url_helpers
 
+  options = {
```
mbarnett, line 16: Generally speaking, you don't want to do variables like this directly in the class scope – the semantics are kind of funky, eg) someone will come along and try to do:

```ruby
def foo
   renderer.blah
end
```

and then discover it's not in-scope for anything. The more usual approach is to do it as a method returning the value, or just in this case, make it a constant.

app/decorators/application_decorator.rb
```diff
@@ -11,4 +13,29 @@ class ApplicationDecorator < Draper::Decorator
...
+    filter_html: true,
+    no_images: true,
+    no_styles: true,
+    hard_wrap: true,
+    link_attributes: { rel: 'noopener noreferrer', target: '_blank' }
+  }
+
+  renderer = Redcarpet::Render::HTML.new(options)
+  RenderMarkdown = Redcarpet::Markdown.new(renderer, Rails.configuration.markdown_rendering_extensions)
```
mbarnett, line 25: The ClassCase'd constants here are a bit odd (I think it's just an artifact of the refactorings you've already done?), because these aren't really classes or modules. 

You *could* switch it to constant case – MARKDOWN_RENDERER – but I think a more idiomatic solution in Ruby here is to put this all in memoized methods:

```ruby
private
def html_renderer
  @@html_renderer ||= Redcarpet::Render::HTML.new(options)
end

def markdown_renderer(text)
   @@markdown_renderer ||= Redcarpet::Markdown.new(html_renderer, Rails.configuration.markdown_rendering_extensions)
end
```

etc, and then call those in your public methods.

mbarnett (changes requested):
This is really good and getting really close, just a couple of notes about making things a wee bit more idiomatic, and making some names more guessable – and one concern about how we want items to look in search results?

app/decorators/community_decorator.rb
pgwillia: The only reason I can think of is that the collections of Communities are the only one that needs it right now.  Making it the default collection decorator does make sense but is it a premature optimization?

app/views/application/_item.html.erb
pgwillia: Yeah, I can see what you mean here about names.  We've got rendered, raw and stripped versions of description.  In the Drafts* we want the default description to be raw and the special case to be the rendered version.  In the other cases we want the default description to be the rendered version and the special case to be the stripped version (`plain_description`) here.

pgwillia: This is a special case because of the truncation.  We can't guarantee that the tags will match when the rendered description is truncated.  This would cause some bizarre artifacts in the search results and in the meta page description.

app/decorators/community_decorator.rb
mbarnett: So (to kind of explain what's going on in my head when I see this kind of thing) philosophically I'd draw a distinction between:

a premature optimization – where optimization usually means "making the code really inscrutable and weird because it happens to be faster that way" – being bad because you're degrading the quality of the code but don't actually have the numbers in front of you _proving_ that the performance cost is important enough to offset the readability and understandability cost

and _this_, which we wouldn't be doing for optimization purposes but for what I'm going to call "thinking like a more Senior Developer-y reasons". 

Basically the observation here is that PaginatingDecorator works for all classes, so any Decorator for a collection of a class that needs to be decorated will *at some point* need to add this same method with the same contents. Our job as more Senior people in the codebase is to think ahead and say "well, what I don't want to have happen here is for someone who's maybe a bit newer or more junior to try to figure out how to do this, look at how it's already done, and then copy-and-paste this method into their Decorator to get it to work, and carry on, because that wouldn't be very DRY". 

We can preempt that by making all Decorators behave this way by default, without the loss of readability you'd see in a premature optimization. We're really just thinking ahead and making things more ergonomic for everybody, and it's always good to build your APIs such that they make it effortless for your team to do the right thing with them from the start, whereas this way we're requiring the _next_ person to do the bit of work to solve their problem the right way.

app/views/application/_item.html.erb
mbarnett: Yeah, I missed this the first time around but we don't want to truncate in the middle of a tag, but we ALSO don't want to expose the user to un-rendered markdown just to avoid that.

I think what we want here is either to strip the text, so that it renders plainly without tags to worry about truncating, but doesn't have any weird markdown appearing on the page, or we want to do HTML-aware truncation, where we render the description and then are more careful about how we cut the text. (probably the first way is fine, here, because the second version adds some overhead as you have to parse your own generated HTML just to generate more HTML)

app/controllers/collections_controller.rb
```diff
@@ -13,12 +13,14 @@ def show
...
     @search_models = search_query_index.search_models
+    @collection = @collection.decorate
+    @community = @community.decorate
   end
 
   private
 
   def fetch_and_authorize_community
-    @community = Community.find(params[:community_id])
+    @community = Community.find(params[:community_id]).decorate
```
murny, line 23: Do you need to decorate this twice? Here and on Line 17. Probably can just remove this one?

pgwillia: Good catch.  I'll remove!

app/models/exporters/solr/base_exporter.rb
```diff
@@ -139,6 +142,10 @@ def self.fulltext_searchable?(name)
     name == fulltext_searchable_field
   end
 
+  def self.unrender_markdown(text)
```
mbarnett, line 145: Nit: Rendering markdown means taking markdown and producing HTML, so this doesn't really *unrender* markdown (which implies taking HTML and turning it back into markdown), it just removes the markup, right?

db/seeds.rb
```diff
@@ -74,7 +74,11 @@
     community = Community.create!(
       owner_id: admin.id,
       title: title,
-      description: Faker::Lorem.sentence(word_count: 20, supplemental: false, random_words_to_add: 0).chop
+      description: if idx == THINGS.size-1
```
mbarnett, line 77: Tiny Tiny Nit: there should be spaces around the `-` operator. Surprised Rubocop doesn't, of all things, enforce this.

app/decorators/application_decorator.rb
```diff
@@ -11,4 +13,39 @@ class ApplicationDecorator < Draper::Decorator
...
+  end
+
+  # We trust the redcarpet output which is why we think it's html_safe
+  # rubocop:disable Rails/OutputSafety
+  def render_markdown(text)
+    html_renderer.render(text).html_safe if text.present?
+  end
+  # rubocop:enable Rails/OutputSafety
+
+  def unrender_markdown(text)
```
mbarnett, line 27: See the naming nit on the Solr exporter.

app/decorators/application_decorator.rb
```diff
@@ -1,3 +1,5 @@
+require 'redcarpet'
+require 'redcarpet/render_strip'
```
mbarnett, line 2: These aren't set to `require: false` in the Gemfile, so Bundler automatically requires them at Application boot. You don't need to separately require them here.

mbarnett (changes requested):
Just a few really small nits at this point, but otherwise looks great!

app/decorators/application_decorator.rb
pgwillia: Why do I get `NameError: uninitialized constant Redcarpet::Render::StripDown` in the tests if I remove these?

app/models/exporters/solr/base_exporter.rb
pgwillia: I was trying to get antonym pairs (in this case render/unrender) as we discussed before.  The challenge is still that we want rendered/raw/stripped versions of the text.  If I change this back to `strip_markdown` does it make sense?

app/decorators/application_decorator.rb
mbarnett: That's odd, I specifically pulled this branch down to make sure I was right that this was unnecessary – and I just double-checked that this is passing. 

This is a long shot, but maybe Spring's the issue? You could try doing `bin/spring stop` before re-running the tests, after making this change.

app/models/exporters/solr/base_exporter.rb
mbarnett: I think so, or it makes more sense to me, anyways.

app/decorators/application_decorator.rb
mbarnett: Just for reference:

```
$ bin/spring stop
Spring stopped.
$ rails c        
/Users/matt/.rbenv/versions/2.6.6/gemsets/jupiter/gems/pry-byebug-3.8.0/lib/pry-byebug/control_d_handler.rb:5: warning: control_d_handler's arity of 2 parameters was deprecated (eval_string, pry_instance). Now it gets passed just 1 parameter (pry_instance)
W, [2021-04-22T16:02:29.735234 #41113]  WARN -- Skylight: [SKYLIGHT] [4.3.2] Running Skylight in development mode. No data will be reported until you deploy your app.
(To disable this message for all local apps, run `skylight disable_dev_warning`.)
Running via Spring preloader in process 41140
Loading development environment (Rails 6.0.3.6)
irb(main):001:0> Redcarpet::Render::StripDown
=> Redcarpet::Render::StripDown
```

so it's definitely getting required at boot, for me.

pgwillia: Stopping spring didn't help.  Restarting my machine didn't help.  Hmm.  Wondering what's going on there.

mbarnett: What ruby version are you using?

Also is it possible that you have the gems locally but haven't bundled properly for some reason? (it should prevent you from doing that if the lockfile is stable, but...)

pgwillia:
```
$ ruby -v
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]

$ bundle info redcarpet
  * redcarpet (3.5.1)
        Summary: Markdown that smells nice
        Homepage: http://github.com/vmg/redcarpet
        Path: /home/pgwillia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/redcarpet-3.5.1
```

mbarnett: I'm using 2.6.6p146, let me try it with ruby 2.7...

mbarnett: Fresh installed 2.7.2p137 and completely rebundled:

```
irb(main):001:0> Redcarpet::Render::StripDown
=> Redcarpet::Render::StripDown
```

All tests green.

Try pushing the deletion up to this branch? Let's see what happens on CI.

mbarnett: Ok, I was able to re-create this with Ruby 3.0 and a brand new Rails 6.1 app. Unsure why I can't recreate this in the Jupiter directory.

Poking around the redcarpet gem, it looks like the main gem file doesn't require 'render_strip' by default, which probably explains your error, so you should be able to verify that deleting `require 'redcarpet'` works, because that is the main file and Bundler *does* require it by default (verified in the fresh 3.0 app).

Which leaves us with the optional 'redcarpet/render_strip'. My take is that the right way to this is to keep letting Bundler be the driver of all gem requires, rather than requiring things manually in random files. You can tell bundler to load extra files for you like this:

`gem 'redcarpet', require: ['redcarpet', 'redcarpet/render_strip']` 

(by default, it just infers that it should do `require 'redcarpet'` unless you manually tell it the file(s) to require).

mbarnett: Figured out the jupiter thing – if you look at my diff, I'd deleted the lines from application_decorator.rb but not base_exporter.rb, so 'render_strip' was getting pulled in during boot anyways, creating the confusion. Which kind of illustrates why centralizing requires in Bundler is less error prone I guess 😅

pgwillia: Nice!  Good sleuthing. From Java and C++ includes were pretty common so this ruby way is new to me.  Looks like it works!

mbarnett: Yeah, it works a bit differently than `import` in that there's no scope to it; once something in the process has `require`d a file, everything in it at the top level is visible globally, and not just in the one file that did the `require`. `#include` *kind of* works similarly in that the contents of the file you pull in are globally visible to the entire compilation unit, it's just that in C++ the compilation unit is (generally) a relatively small part of the program, whereas in Ruby the compilation unit effectively is the full program (nature of ahead-of-time compilation of individual files that are eventually linked together vs runtime dynamic languages, more or less)

app/decorators/application_decorator.rb
```diff
@@ -11,4 +11,39 @@ class ApplicationDecorator < Draper::Decorator
...
+
+  private
+
+  REDCARPET_OPTIONS = {
+    filter_html: true,
+    no_images: true,
+    no_styles: true,
+    hard_wrap: true,
+    link_attributes: { rel: 'noopener noreferrer', target: '_blank' }
+  }.freeze
```
mbarnett, line 37: Sorry, I should have thought of this before: I know this is only used here, but can we haul this over to `config/application.rb`? Just so everything that controls what's allowed is centralized in one spot.

mbarnett (commented):
Last thing, I swear

app/decorators/application_decorator.rb
pgwillia: Done.  This makes sense after moving the other part.

mbarnett (approved):
giphy

Issue #2106, opened 2021-01-20
Fix console output in test suite


murny:
Seems like  a recent change has started polluting our test runner (`rails test`). As shown here:

```
ERROR [line 20]      Lexer error: With input '<rs:ln href="/items/e2ec88e3-3266-4e95-8575-8b04fac2a679/view/c7732455-1cff-45c8-b22b-d046a7fd7181/im': Invalid token "<rs:ln": {:production=>:path, :token=>"<rs:ln"}: 
ERROR [line 22]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 23]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 24]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 25]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 26]      Expected objectList: {:production=>:predicateObjectList, :token=>"."}: 
ERROR [line 20]      Lexer error: With input '<rs:ln href="/items/e2ec88e3-3266-4e95-8575-8b04fac2a679/view/c7732455-1cff-45c8-b22b-d046a7fd7181/im': Invalid token "<rs:ln": {:production=>:path, :token=>"<rs:ln"}: 
ERROR [line 22]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 23]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 24]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 25]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 26]      Expected objectList: {:production=>:predicateObjectList, :token=>"."}: 
ERROR [line 20]      Lexer error: With input '<rs:ln href="/items/8e18f37c-dc60-41bb-9459-990586176730/view/f803c425-7424-4f4a-bda0-1c0908702ea1/im': Invalid token "<rs:ln": {:production=>:path, :token=>"<rs:ln"}: 
ERROR [line 22]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 23]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
ERROR [line 24]      Expected objectList: {:production=>:predicateObjectList, :token=>";"}: 
[...]

PR #2242, opened 2021-03-19
Fix rdf-n3 lexer error

lagoan:
This change modifies the n3 graph notation to treat the offending
objects as strings rather than elements.

The problem was likely caused by a change in the lexer on rdf-n3 that
is now more accurate in how it traverses and checks the elements on the
graph.

This error is not caught in all lexers, that is why the error was not
being shown in earlier versions.

After discussing this issue with @anayram, she mentioned that it makes
more sense for the offending object to be treated as a string rather than
an element as it was in the previous version.

## Context

Errors started appearing when parsing n3 files after an update on the 
gem rdf-n3

Related to #2106

## What's New

Updated rdf-gem and changed values for predicate 
http://terms.library.ualberta.ca/sitemapLink To be a string instead of an
element.


app/controllers/aip/v1/entities_controller.rb
```diff
@@ -151,8 +151,8 @@ def file_set
       RDF::Statement(
         subject: self_subject,
         predicate: ::TERMS[:ual].sitemap_link,
-        object: RDF::URI('rs:ln') + %Q( href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}" ) +
-                %Q(length="#{@file.byte_size}" type="#{@file.content_type}")
+        object: %Q(<#{RDF::URI('rs:ln')} href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}") +
+                %Q( length="#{@file.byte_size}" type="#{@file.content_type}">)
```
mbarnett, line 155: Instead of doing a runtime concatentation of two strings with `+`, you can just do:

```ruby
object: %Q(<#{RDF::URI('rs:ln')} href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}"  \
length="#{@file.byte_size}" type="#{@file.content_type}">)
```

also feels like we're calculating a lot "inside" the string interpolation by making so many different method calls to different targets, by idk if temp variables would really improve the situation there much.

mbarnett (approved):
This fixes the bug (yay!)

Outside the scope of this PR and the related bug, there's a few things I think we need to start thinking about:

1. I'm concerned about the long-term health of this controller. It's long in a "fat controllers anti-pattern" way, and it feels like most of that length is dedicated to generating the output format, which is something that feels like it ought to (idiomatically) be handled by a template or Builder pattern rather than programmatically by a method in the controller. I expect that some of the reason it's not handled that way is that there isn't good support for n3 in the standard Ruby libraries, because n3 is odd and not really industry supported.

2. Relatedly, we're living and dying by what the rdf/n3 people decide to do with their API, which frankly I can't make heads or tails out of. I'm super glad that you understand what's going on here, Omar, but I'm concerned if the health of this controller depends on the person who maintains it being married to a metadata specialist 😉

Long story short, I think we should open a ticket in the backlog to start brainstorming ways to clean this thing up. My initial preference here is to do something like completely replace it with something like a JSON builder that outputs JSON-LD on the theory that's a lot closer to a normal industry approach (meaning I can point people at a ton of documentation for how to generate JSON versus generating RDF graphs, and JSON-LD has more mind-share in the community than turtle), and then leverage something on PMPY's end to auto-convert that from JSON-LD to n3 (if n3 is really something Kenton wants to be committed to long-term, but that's a conversation I'll have to raise at some point).

Open to other ideas, but want to capture the discussion in a ticket.

mbarnett (changes requested):
I meant to hit request changes for the multi-line string thing 🙃

app/controllers/aip/v1/entities_controller.rb
lagoan: We can definitely go with your suggestion which avoids the runtime concatenation. The only issue I see with this is that instead of having the value nicely formated with a single space separating attributes:

```
"<rs:ln href=\"/items/e2ec88e3-3266-4e95-8575-8b04fac2a679/view/c6f112b0-4162-4460-9201-a5eacfdfbdd4/image-sample.jpeg\" rel=\"content\" hash=\"md5:GxpIjJsC4KnRoBKNjWnkJA==\" length=\"12086\" type=\"image/jpeg\">";
```

we would now have something like:

```
"<rs:ln href=\"/items/e2ec88e3-3266-4e95-8575-8b04fac2a679/view/60fa3c55-12f5-44e1-8ba6-3c64ef16ce61/image-sample.jpeg\" rel=\"content\" hash=\"md5:GxpIjJsC4KnRoBKNjWnkJA==\"                 length=\"12086\" type=\"image/jpeg\">"
```

This happens if we indent the rest of the string to match the previous declaration (at least the way I am doing it :smiley: ). I don't think the code should affect the output that way.

Let me know what you think about this :smile: 

As for the method calls inside string interpolation, I don't think the temporary variables would change things much if we use meaningful names. The problem really is that the string is rather long and multiple strings interpolated do not make it particularly easy to read.

lagoan:
@mbarnett I totally agree with your comments about simplifying the controller. We've talked in the past about different approaches to do this but we should put everything down in a ticket as you suggest so we can reference back to it when needed.

app/controllers/aip/v1/entities_controller.rb
mbarnett: Whoops, good point. Ruby's got a lot of ways to solve this problem, and I think a squiggly heredoc is probably what we want?

```ruby
object: <<~OBJECT_URI
           <#{RDF::URI('rs:ln')} href="#{file_view_uri.path}" rel="content" hash="md5:#{@file.checksum}" 
             length="#{@file.byte_size}" type="#{@file.content_type}">)
OBJECT_URI
```

It's multiline friendly but should ignore the leading whitespace.

lagoan: Squiggly heredoc keeps the newlines to it won't work unless we remove them. I did some poor man profiling testing value assignment with a long string without breaking it apart, the original concatenation, using squiggly heredoc as you suggested (which unfortunately does not work), and finally, squiggly heredoc removing the newlines.

These are the results:

```
Starting test long single line

real    0m40.407s
user    0m40.376s
sys     0m0.028s

Starting test concatenation

real    0m54.284s
user    0m54.175s
sys     0m0.016s

Starting testing heredoc no substitution (does not produce the desired result)

real    0m58.523s
user    0m58.362s
sys     0m0.012s

Starting heredoc removing newlines through substitution

real    1m39.975s
user    1m39.917s
sys     0m0.040s

```

Why did you suggest no using concatenation?

Sorry to go ask again about this, it is a good learning opportunity, albeit for a small thing. I do not have an opinion on whatever option we use.

mbarnett: Wow there really is no good way to do this without the `+`, huh.

I'm not terribly concerned that this is a performance issue, it's more that it just feels silly that we have to create 2 string objects and then use `+` to combine them into a third when really all we want is the string literal we have at runtime.

Really all we're concerned about here is we have a longer string than we're "supposed" to per-rubocop/common sense/whathaveyou. I'd probably just make it one line and add a disable rubocop directive above it for the warning and add a note that we're breaking style because Ruby syntax limitations don't let us split it over 2 lines without adding undesirable whitespace.

Equally, we could just live with the concatenation. It's an aesthetic issue, not a perf issue unless we start doing it millions of times in a loop.

Either way, not a big deal.

mbarnett (approved):
Flipping the review to Approved per https://github.com/ualbertalib/jupiter/pull/2242#discussion_r598971500 since this is just a style issue that has no perfect work-around. I'll leave it up to you whether or not you want to live with a > 120 char line or keep using the concatenation.

PR #2254, opened 2021-03-29
Various fixes from lighthouse suggestions

ConnorSheremeta:
Now pre-loading main FA fonts, increased link spacing on mobile, and ordered the header tags in a more accessible way.


app/javascript/styles/jupiter.scss
```diff
@@ -25,6 +25,14 @@ body {
   }
 }
 
+// Increase footer link spacing on mobile to improve usability and seo.
```
murny, line 28: You can probably do this all within bootstrap via these utilities classes: https://getbootstrap.com/docs/4.6/utilities/spacing/

So if you want top/bottom padding of around 3px for small devices you can just add the following class to these items:

`py-sm-1`  

(it will actually be 4px instead of 3px, but doubt that makes any difference)

ConnorSheremeta: Neat, those bootstrap classes are very versatile!

murny (approved):
Nice work. LGTM 👍

PR #2258, opened 2021-03-30
Update to Rails 6.0.3.6, removing the mimemagic dep

mbarnett:
## Context

All of the recent mishegoss around mimemagic. All of our CI tests are still broken around this.

## What's New

Update Rails to 6.0.3.6, removing the dependency on mimemagic and fixing CI.


pgwillia (approved):
LGTM 👍

PR #2261, opened 2021-03-30
Move danger ahead so it maybe stops flapping or flaps faster

mbarnett:
## Context

Danger is flapping on the existence of DANGER_GITHUB_API_TOKEN, which seems to sometimes not exist?

## What's New

Kicking this to the front of the build to see if maybe something else is somehow causing it to fail occasionally. Next step is disabling it until we understand the issue better, because flapping builds and quantum env variables are bad.

Also fixed a stray newline in the Danger warning.


pgwillia (approved):
Worth a try.

PR #2275, opened 2021-04-08
Merge master branch into integration branch

murny:
We need @mbarnett fixes he did on master to bump rails to a version without mimemagic dependency. Currently any `bundle install` on integration branches will fail.

This backports master commits onto integration branch


mbarnett:
This: https://github.com/ualbertalib/jupiter/blob/2a8a1d16df94362d9b250744916d4439c088a409/Gemfile.lock#L571 should say 3.1.1 (it's why the tests all bombed out on bundle)

murny:
Yeah just saw that failing on CI. Need your changes for RDF gem. Will add back 👍

mbarnett:
I might have that backwards, I think Tricia said there was something on integration that needed 3.1.2, which Omar fixed a couple of weeks ago on this branch. Either way, there's just a mismatch in the lockfile – it's 3.1.1 in one spot and 3.1.2 in the other, and bundler can't resolve the contradiction.

murny:
Just tossed the Gemfile.lock and recreated it from the Gemfile. Seemed to have fixed the weird RDF dependency issue

mbarnett:
lol I think that jumped Skylight a bit too high. We'd need to update the Normalizer for the Solr query instrumentation to move to 5.0

murny:
K hopefully should all be good now. I just have to change the "required checks" on this branch now we running ruby 2.5/2.7 and then build is green 👍

mbarnett (approved):

Issue #2080, opened 2021-01-11
Digitization Folk Fest Programs Show Views


pgwillia:
**Is your feature request related to a problem? Please describe.**
The first set of digitization content will be the Folk Fest Programs. This set was chosen because it is small (there are 36) and well described. When modelling is complete we should work on presentation of the information.

**Describe the solution you'd like**
Should review https://balsamiq.cloud/s3awdua/pnf6o6n/rAC91 and implement
- [ ] search results
- [ ] book
- [ ] collection?

**Additional context**
#2010 #2011

pgwillia:
This is how it's going:
![image](https://user-images.githubusercontent.com/1220762/115080266-b0f94380-9ebf-11eb-99da-a3292f108527.png)

pgwillia:
need to
    - [ ] rename alt[ernate]_title, langauge[s]
    - [ ] refactor solr_exporter_class out of depositable
    - [ ] add volume [RDF::RDFS.label]
    - [ ] DRY views
    - [ ] DRY helpers

PR #2288, opened 2021-04-22
Corrected missing pluralization in Digitization::Book attributes

pgwillia:
## Context

When I was creating the show views for the Digitization::Book model I found that I was reasoning about arrays but referring to them in the singular.  This meant there wasn't an appropriate name for the singular values from the array.  This PR intends to correct the poor choices I made when originally naming these values.

https://github.com/ualbertalib/jupiter/blob/0bfbb0d9334465630dcdd1cf318b2435db1ba66f/db/schema.rb#L92-L113

Related to #2080 

## What's New
Renamed the attributes in the Digitization::Book model and corrected their use elsewhere in the code as well.

### DIFF
--- a/db/migrate/20210422161911_rename_columns_in_digitization_book.rb
+++ b/db/migrate/20210422161911_rename_columns_in_digitization_book.rb
@@ -12,5 +12,24 @@ class RenameColumnsInDigitizationBook < ActiveRecord::Migration[6.0]
       t.rename :geographic_subject, :geographic_subjects
       t.rename :topical_subject, :topical_subjects
     end
+
+    remove_rdf_table_annotations :digitization_books
+
+    add_rdf_table_annotations for_table: :digitization_books do |t|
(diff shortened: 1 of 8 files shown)


db/migrate/20210422161911_rename_columns_in_digitization_book.rb
```diff
@@ -0,0 +1,16 @@
+class RenameColumnsInDigitizationBook < ActiveRecord::Migration[6.0]
+  def change
```
pgwillia, line 2: Just realized I'll need to change the rdf_annotations names as well.

pgwillia: Fixed!

mbarnett (approved):
Good thinking – it's always much nicer to deal with APIs when plurality correctly reflects the one-or-many nature of the properties themselves.

murny (approved):
👍

PR #2297, opened 2021-04-26
Move `visibility` vocab into `jupiter_core` namespace

pgwillia:
## Context

In reviewing Depositable for making it part of the Digitization object hierarchy noticed that this core class contained ERA specific content. Anything that is deposit-able should have a visibility so we know who to show it to.

Related to #2080

## What's New

Move visibility vocabulary into a `jupiter_core` namespace because it's not an ERA only thing.


murny (approved):
LGTM 👍

mbarnett (dismissed):
This is smart, I should have thought of the need for this initially. Good catch!

mbarnett (approved):

PR #2302, opened 2021-04-27
Test fixture record name refactor

ConnorSheremeta:
It was found that having duplicate fixture record names caused problems. All fixture names have been modified to ensure their uniqueness.


pgwillia (approved):
Nice work.  Adding the namespace to the name of each fixture helps to make it clear which type of fixture we're talking about.

PR #2318, opened 2021-05-05
[WIP] Extract common list items into partials

pgwillia:
## Context

Digitization::Book will look a lot like these views too.  Trying to refactor to make the common parts reusable.  In some cases reuse was made possible by adding a method to the decorator which interprets a Thesis attribute in the way that the Item does.

Related to #2080

## What's New

Trying to refactor to make the common parts reusable.


app/views/items/_item.html.erb
```diff
@@ -41,66 +36,16 @@
...
-        <dd>
-          <ul class="list-unstyled">
-            <% @item.all_subjects.each do |subject| %>
-              <li><%= search_link_for(@item, :all_subjects, value: subject) %></li>
-            <% end %>
-          </ul>
-        </dd>
-      </dl>
-    </li>
-  <% end %>
```
pgwillia, line 57: This structure is pretty common so at one point I was trying to make it a more generic partial with the title and values being passed to the partial using [locals](https://guides.rubyonrails.org/layouts_and_rendering.html#passing-local-variables)

```
    <li class="list-unstyled list-group-item-action">
      <dl>
        <dt><%= title %></dt>
        <dd><%= value %></dd>
      </dl>
    </li>
```

`render partial: 'application/show_metadata/list_item', locals: {title: t('items.show.subject'), value: search_link_for(@item, :all_subjects, value: subject)}`

There are a couple different types of values: 
- the thing
- search_link_for the thing
- humanize_date of the thing
- link_to the the thing

But I backed away from this when a) iterating on things got more complicated b) it became less obvious in the _item.html.erb partial of what was being accomplished.

app/decorators/thesis_decorator.rb
```diff
@@ -17,4 +17,12 @@ def plaintext_abstract
...
   end
 
+  def all_subjects
+    model.subject
+  end
+
+  def languages
+    [model.language]
+  end
+
```
pgwillia, line 27: I was starting to do some funny things in the decorator too.  I had a bunch of nil methods that represented things that we did want in the Item display but not in the Thesis.  I pulled back on this and chose to represent the differences in the specific partials.

Another thing I briefly tried was trying to set out in advance which title/value pairs we wanted to display into a data structure defined here and then iterated on in the view.  That got pretty complicated when the values were from arrays.  I gave up on that idea too.

pgwillia (commented):
I don't know if I'm making this better or worse.  What do you think?

PR #2326, opened 2021-05-07
Use read-only token for Danger

pgwillia:
## Context

Related to #2082 

## What's New

Changed Danger github action to use the read-only GITHUB_TOKEN


pgwillia:
I don't think this worked?!  The tests all returned with a green check mark but the danger check remained yellow dot until I opened another PR from the origin rather than the fork.

PR #1600, opened 2020-04-17
Removed Matomo

ConnorSheremeta:
#1493

## Context
We no longer need Matomo tracking

PR #2327, opened 2021-05-07
Use read-only token for Danger

pgwillia:
## Context

Note that this DOES NOT allow developers the possibility to propose change sets from a fork of the repository. This hasn't been possible since the [policy change in Travis](https://github.com/ualbertalib/jupiter/pull/1600#issuecomment-615385889).

Related to #2082 

## What's New

Changed Danger github action to use the read-only GITHUB_TOKEN

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -35,10 +35,10 @@ jobs:
           bundle config path vendor/bundle
           bundle config set without 'production development'
           bundle install --jobs 4 --retry 3
-      
+
       - name: Run Danger
         env:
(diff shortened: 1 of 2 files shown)


murny (approved):
CI builds look great 👍 

A bit confused on the magic that makes this work?  `DANGER_GITHUB_API_TOKEN` is a token that supplies access to our github Bot Account. So how does replacing it with `GITHUB_TOKEN` (which doesn't seem to be defined in our secrets) do the same thing but better?

Be interesting to see if Danger is still working by seeing it use our bot account to comment on PRs, then I'll accept this magic 😄.

I guess a easy way to test this if danger can still comment is to create a PR with an empty description or change your changelog entry to be less then 10 characters 🤔 .

Otherwise looks good to me, nice work investigating this 👍

PR #2333, opened 2021-05-12
Postgres is now on v12 in production, lets update our code/CI to reflect this

murny:
Matt said production is now using version 12 of postgres, so let's update this for our CI and local docker stuff.


pgwillia (approved):
Nice catch!

PR #2334, opened 2021-05-12
Add GitHub actions to dependabot

murny:
- Adds dependabot for github action packages
- Slight fixes for dependabot config and prettier reformat
- update github actions config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,31 +1,50 @@
 version: 2
 updates:
-- package-ecosystem: bundler
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "10:00"
-  open-pull-requests-limit: 10
-  target-branch: integration
-  ignore:
-  - dependency-name: pundit
-    versions:
-    - "> 1.1.0"
-  - dependency-name: sidekiq
-    versions:
-    - ">= 6.a"
-    - "< 7"
-  - dependency-name: vcr
-    versions:
-    - ">= 5.1.a"
-    - "< 5.2"
-  - dependency-name: vcr
-    versions:
-    - "> 6.0.0"
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -1,31 +1,50 @@
...
+    schedule:
+      interval: daily
+      time: "10:00"
+      timezone: America/Edmonton
+    open-pull-requests-limit: 10
+    target-branch: integration
+    labels:
+      - "dependencies"
+      - "ruby"
+  - package-ecosystem: github-actions
```
murny, line 40: Most of this file is prettier changes.

This is pretty much the change here which adds github actions to dependabot upgrading.

I also explicitly added labels and timezone to the other packages (npm/rubygems)

.github/workflows/push.yml
```diff
@@ -21,21 +21,8 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
```
murny, line 24: This one line basically does what the line 25 to 38 diff was doing.

.github/dependabot.yml
pgwillia: Nice work!  Dependabot will update lines like https://github.com/ualbertalib/jupiter/blob/371be28a31f27847dc3424f21fbe9a06d28b032c/.github/workflows/push.yml#L13

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions

.github/workflows/push.yml
pgwillia: Another nice catch! 

https://github.com/actions/cache/blob/main/examples.md#ruby---bundler

pgwillia (approved):

PR #2335, opened 2021-05-12
Drop logo_id foreign key on item/thesis 

murny:
This foreign key isn't working how it's suppose to be when deleting an Item (or during a roll back on a failed a transaction).

You will receive an error like the following (this was during a roll back of a failed item ingest wrapped in a transaction): 
```
Caused by PG::ForeignKeyViolation: ERROR:  insert or update on table "items" violates foreign key constraint "fk_rails_a9238d43b0"
DETAIL:  Key (logo_id)=(71) is not present in table "active_storage_attachments".
```

Seems like something odd is happening with ActiveStorage associations? When the attachment gets deleted it should null out the logo_id column for Item model via the `on_delete: :nullify` but for whatever reason this isn't happening.

Definitely requires a deeper investigation, but for now, let's just remove this foreign_key.


mbarnett (approved):
I definitely need to read up on Postgres constraint rules and see if I can understand why this wasn't working out the way we expected 👍

mbarnett (approved):

PR #2337, opened 2021-05-12
Refactor to a Depositable view

pgwillia:
## Context

We will be adding other (`Digitization`) things (like `Book`) that have similar properties as Items.  In fact `Item` and `Thesis` are already not that different.  This is a refactor in anticipation that this will continue to be reused for other Depositable things.

Related to #2080

## What's New

Refactored Item and Thesis views into a Depositable view. We assume that Depositable items have
  * title and alternative title
  * creator(s)
  * description
  * creation date
  * subjects
  * item type
  * doi
  * license
  
and will implement `creators_label`, `created_label`, `description_label` in it's decorator.  Any other attributes can be displayed just above or below the fold.

||Before|After|
|---|---|---|
|Item|before_item|after_item|
|Thesis|before_thesis|after_thesis|

Significantly the Graduation Date and Thesis Degree change order to better align with items.


app/views/items/_thesis.html.erb
```diff
@@ -1,54 +1,37 @@
...
   <% end %>
 
-  <li class="list-unstyled list-group-item-action">
-    <dl>
-      <dt><%= t('items.show.item_type_with_status') %></dt>
-      <dd><%= item_type_search_link(@item) %></dd>
-    </dl>
-  </li>
-
   <% if @item.degree.present? %>
```
pgwillia, line 35: This is the most significant change which will have to get more feedback.  This moves the Thesis Degree below the fold.

pgwillia: I'm going to use `content_for` methods to push this above the fold.

app/views/items/show.html.erb
```diff
@@ -88,7 +88,8 @@
 
       <h1 title="<%= t('items.show.title') %>"><%= @item.title %></h1>
 
-      <%= render @item %>
+      <% render @item %>
```
pgwillia, line 91: This loads the `content_for` `above_the_fold` and `below_the_fold` which will be used in the `depositable` partial

pgwillia:
Tagging @sfarnel and @seanluyk for discussion.  Is changing the order of metadata on the Thesis pages still on the table?  I can make the views more reusable by changing some things around.  Take a look at the screen shots in the description.  Is this acceptable?

seanluyk:
This is looking great @pgwillia! I actually think we could do even more re-ordering of the thesis fields, what do you think @sfarnel? For example, keeping all the program completion-related stuff together (e.g. grad date, supervisor, degree, and committee), and then bibliographic metadata together in a logical way too?

How would you like to go about this @pgwillia? Perhaps the three of us could meet to discuss ideas?

pgwillia:
Just spoke with @seanluyk and @sfarnel and came to conclusion that flexibility in ordering attributes by importance in each context is paramount.  So the reuse that I've proposed in this PR isn't exactly what we're looking for.  I'll make some tweaks and put up a revised one.

PR #2338, opened 2021-05-13
Merge master into integration branch, fix existing tests

murny:
Catch integration branch with latest changes in master.

- Also fixed bunch of fixtures/tests to use new fixture names.
- Fix rubocop errors as integration enabled rubocop-minitest


ConnorSheremeta (approved):
LGTM, thank you for this Shane!

.rubocop.yml
```diff
@@ -12,11 +12,11 @@ AllCops:
   DisplayCopNames: true
   DisplayStyleGuide: true
   Exclude:
-    - 'db/**/*'
-    - 'tmp/**/*'
-    - 'vendor/**/*'
-    - 'Dangerfile'
-    - 'node_modules/**/*'
+    - "db/**/*"
```
pgwillia, line 15: For the changes that are just formatting, is this preference in one of our linters?

murny: Think this is a yaml extension I have installed in VSCode (https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) which is auto formatting these... 

Can disable or change the string rule to single quotes if thats what the team decides

mbarnett: I don't think it really matters tbh. The marginal value of rubocop-style pickiness over yaml files is pretty low.

mbarnett (approved):

.rubocop.yml
pgwillia: I agree this isn't a huge deal but it would be nice to prevent churn in these files just because of something like single quotes vs quotations marks.  My understanding is that yaml is pretty specific about spaces and structure so it would be nice if a linter caught anything that might make it invalid or change meaning.

I'm going to install the same extension that Shane to try to help with consistency.

PR #2351, opened 2021-05-19
Use  read-only token for Danger

pgwillia:
## Context

Note that this also allows developers the possibility to propose change sets from a fork of the repository which hasn't been possible since the [policy change in Travis](https://github.com/ualbertalib/jupiter/pull/1600#issuecomment-615385889).

Related to #2082 

## What's New

Changed Danger github action to use the read-only [GITHUB_TOKEN](https://docs.github.com/en/actions/reference/authentication-in-a-workflow)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Run Danger
         env:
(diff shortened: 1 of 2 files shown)


pgwillia:
image

Issue #2011, opened 2020-12-03
Digitization Folk Fest Ingest


pgwillia:
**Is your feature request related to a problem? Please describe.**
The first set of digitization content will be the Folk Fest Programs. This set was chosen because it is small (there are 36) and well described. 

Follows #2010

mbarnett:
– There is a spreadsheet somewhere that maps the folkfest stuff to Swift IDs. I'll try to chase that down, or Sarah knows the details.

– We talked a bit about how we'll want to deal with Preservation/AIPs, and previous archival efforts, and the cheap interim solution I landed on was that we'll just take the tarred (or zipped or whatever the package is) copy of whatever we pull out of Swift, extend the book (and future) models w/ `has_one :historical_archive`, and attached the full thing to the model. The historical archive will never be loaded or made available for UI purposes and we'll have no reason to ever want to load it, so the attachment's only purpose is, in the future, somebody can find the old copy of the data.

– We're just looking to attach one big high quality PDF of the work for things like this and newspapers, not per-page images, and I'm unsure if that's what we'll get out of Swift or if this will require some preprocessing.

– The fulltext is in a metadata file whose name I forget right now. We'll have to find a parser for it if one exists, and figure out how to store the data – I'm thinking as an attachment if the parsing isn't incredibly complex, or in a text field on a not-normally-loaded join table? The reason not to store the full text on the table itself is that that will make reading the row back from the DB for search results and general metadata viewing way to costly – we'll only ever need the fulltext data for (re-)indexing.

sarahseverson:
- Here is [the spreadsheet ](https://docs.google.com/spreadsheets/d/1U8GckSd-tTaGlBk-TllPl1cd80h8PSQqEdKXUuoCNlI/edit?usp=sharing)with the Peel numbers and matching Swift NOIDS. I've asked Tianyu to grab the file packages so we can see what is inside when we meet next Thursday.

mbarnett:
Thanks!

sarahseverson:
FYI: Here is a [Google Drive folder](https://drive.google.com/drive/folders/17ckV-is6Mh_ZA_tPQIHjZPwayxz_T6BN?usp=sharing) with all the Folk Fest items from SWIFT - it's big at 11 GB but we pulled everything so you could see what's there. Have fun with the files!

pgwillia:
> 
> 
[...]

PR #2373, opened 2021-06-03
Add `Digitization::Book` ingest artifacts to model

pgwillia:
## Context

The first set of digitization content will be the Folk Fest Programs. This set was chosen because it is small (there are 36) and well described.  This adds attributes to store the artifacts that we'll need 
- in how we'll want to deal with Preservation/AIPs, and previous archival efforts. The historical archive will never be loaded or made available for UI purposes and we'll have no reason to ever want to load it, so the attachment's only purpose is, in the future, somebody can find the old copy of the data.
- fulltext in an easy to process format. We want this in a join table so that it doesn't penalize us when we frequently request the object but close enough that we can re-index as required.

Related to #2011

## What's New

- Add `has_one_attached :historical_archive` to Book Model which takes the full archive for all directories relating to that object and archive it in case it's needed. Will not appear in the UI
- `fulltext` will be extracted from the ALTO xml. Should be easy to access in a join table but not the main object table so that results stay efficient.
- Attach one high quality PDF

### DIFF
--- a/db/migrate/20210526215307_create_digitization_fulltexts.rb
+++ b/db/migrate/20210526215307_create_digitization_fulltexts.rb
@@ -1,0 +1,10 @@
+class CreateDigitizationFulltexts < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_fulltexts, id: :uuid do |t|
+      t.references :digitization_book, type: :uuid, null: false, index: true, foreign_key: true
+      t.text :text, null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -708,7 +708,7 @@ if Rails.env.development? || Rails.env.uat?
   end
 
   11.times do |_i|
-    Digitization::Book.create(peel_id: rand(1..3400), part_number: rand(1..100),
+    Digitization::Book.new(peel_id: rand(1..3400), part_number: rand(1..100),
                               owner_id: admin.id,
                               dates_issued: [rand(1900..2020).to_s],
                               date_ingested: '2016-12-08T06:00:00.000Z',
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -709,6 +709,7 @@ if Rails.env.development? || Rails.env.uat?
 
   11.times do |_i|
     Digitization::Book.create(peel_id: rand(1..3400), part_number: rand(1..100),
+                              owner_id: admin.id,
                               dates_issued: [rand(1900..2020).to_s],
                               date_ingested: '2016-12-08T06:00:00.000Z',
                               title: "#{Faker::Company.name} #{Faker::WorldCup.city} Music Festival",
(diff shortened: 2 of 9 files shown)


db/seeds.rb
```diff
@@ -709,6 +709,7 @@ def file2.original_filename
 
   11.times do |_i|
     Digitization::Book.create(peel_id: rand(1..3400), part_number: rand(1..100),
```
murny, line 711: Minor Nit: Since we create the book here then right after doing a `.tap` then `save!` inside, we essentially doing two DB trips to the DB. 

Think if you just did a `.new` here instead of `.create` then your only saving the object once (in the `.tap`)

Needs testing though 😄 
```suggestion
    Digitization::Book.new(peel_id: rand(1..3400), part_number: rand(1..100),
```

murny (approved):
LGTM 👍

db/migrate/20210526215307_create_digitization_fulltexts.rb
```diff
@@ -0,0 +1,10 @@
+class CreateDigitizationFulltexts < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_fulltexts, id: :uuid do |t|
```
murny, line 3: Just curious but did you ever look at ActionText instead? Provides a join table and lots of functionality around processing the text etc. Very similar to what your doing here. 

 More info: https://edgeguides.rubyonrails.org/action_text_overview.html

pgwillia: Oh!  I didn't consider this at all.  My impression was that ActionText was for front end functionality.  I'll follow up about this too after some investigation.

pgwillia: After looking at this I think it's best to not make it ActionText for a couple of reasons.  

The fulltext is simple text rather than rich content.  It's not meant to be beautifully formatted and styled text with embedded images, etc.  This text is a derivative of what has been extracted by optical character recognition (OCR) from the digitized work. It's main purpose is as input for a SolrExporter.  

With ActionText out of the box any content that you store is wrapped in a `trix-content` div.  Initially I couldn't find a way to return the original text, but eventually I found `to_plain_text` by reading [the source code](https://github.com/rails/rails/blob/5c09cfd96fadd6080dd2e57ee761f2ccfdbc7d42/actiontext/app/models/action_text/rich_text.rb#L21-L23).
```
[3] pry(main)> book.fulltext = "Some text."
  ActionText::RichText Load (1.0ms)  SELECT "action_text_rich_texts".* FROM "action_text_rich_texts" WHERE "action_text_rich_texts"."record_id" = $1 AND "action_text_rich_texts"."record_type" = $2 AND "action_text_rich_texts"."name" = $3 LIMIT $4  [["record_id", nil], ["record_type", "Digitization::Book"], ["name", "fulltext"], ["LIMIT", 1]]
=> "Some text."
[4] pry(main)> book.fulltext
  Rendered /home/pgwillia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/actiontext-6.0.3.7/app/views/action_text/content/_layout.html.erb (Duration: 4.1ms | Allocations: 402)
=> #<ActionText::RichText:0x0000562db1c80bd8 id: nil, name: "fulltext", body: #<ActionText::Content "<div class=\"trix-conte...">, record_type: "Digitization::Book", record_id: 0, created_at: nil, updated_at: nil>
[5] pry(main)> book.fulltext.to_s
  Rendered /home/pgwillia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/actiontext-6.0.3.7/app/views/action_text/content/_layout.html.erb (Duration: 0.4ms | Allocations: 188)
=> "<div class=\"trix-content\">\n  Some text.\n</div>\n"
[3] pry(main)> book.fulltext.to_plain_text
=> "Some text."
```

There are some features from the action_text_rich_text migration that might be nice here. Similar to the blobs (that we're already using) you add the ActionText table once and use everywhere. 
```
class CreateActionTextTables < ActiveRecord::Migration[6.0]
  def change
    create_table :action_text_rich_texts do |t|
      t.string     :name, null: false
      t.text       :body, size: :long
      t.references :record, null: false, polymorphic: true, index: false

      t.timestamps

      t.index [ :record_type, :record_id, :name ], name: "index_action_text_rich_texts_uniqueness", unique: true
    end
  end
end
```

I like the polymorphic references for future work on other `Digitization` models and the flexibility with attribute naming.

murny (approved):

Issue #1584, opened 2020-04-08
Automate Brakeman


pgwillia:
I have been running brakeman when there is a dependabot version bump.  It would be nice to automate so 
- it's not my job
- we can get feedback on PRs

**Describe the solution you'd like**
Use Github Action [Run ReviewDog with Brakeman](https://github.com/marketplace/actions/run-brakeman-with-reviewdog)

**Describe alternatives you've considered**
There's also just a Github Action [Brakeman linter](https://github.com/marketplace/actions/brakeman-linter) but I think the PR feedback is more valuable if we can keep noise to zero from this tool.

https://github.com/ualbertalib/jupiter/pull/1582
https://github.com/ualbertalib/di_internal/issues/61

PR #2382, opened 2021-06-11
Add brakeman to github workflow

pgwillia:
## Context

Brakeman is a Ruby on Rails static analysis security tool.  We had been running it manually when the gem was bumped but having it in the github workflow will mean that we will have actionable automatic feedback.

There are Github Actions in the marketplace which have more features but this seems like a good first step.

image
The report when failed
image
The report when successful
image

Related to #1584

## What's New

Brakeman is an action in the pr and push github workflow.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -31,6 +31,9 @@ jobs:
       - name: Lint with RuboCop
         run: bundle exec rubocop --parallel
 
+      - name: Run Brakeman
+        run: bundle exec brakeman -A --no-pager
+
       - name: Get yarn cache
         id: yarn-cache
         run: echo "::set-output name=dir::$(yarn cache dir)"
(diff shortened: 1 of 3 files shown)


murny (approved):
Yay 🎉. 

Surprised we had no failing brakeman errors that needed cleaning up. Impressive 👍

pgwillia:
> 
> 
> Yay 🎉.
> 
> Surprised we had no failing brakeman errors that needed cleaning up. Impressive 👍

We've been pretty good with Jupiter.  We did spend some time making sure this was green.  There was [at least one](https://github.com/ualbertalib/jupiter/pull/1517) that we cleaned up and we're ignoring [Unscoped Find](https://github.com/ualbertalib/jupiter/blob/f524110ed5daebf61928f6a9f549b8ab71cb8a0d/config/brakeman.yml#L4).

PR #2396, opened 2021-06-25
Update codeowners :cry:

murny:
Removes `@mbarnett`  so we can still change code within Solr model directories without approvals


mbarnett (approved):
Changing Solr code without me?!?! 

4DD6D05D-B701-4A27-9FDB-6B0A8584B193

PR #2433, opened 2021-07-13
Remove changelog syntax checking and fix bad markdown bullets

murny:
- Remove changelog syntax checking due to false positives like here:
https://github.com/ualbertalib/jupiter/pull/2430#issuecomment-878575859
and here:
https://github.com/ualbertalib/jupiter/pull/2431#issuecomment-878533886

- Fix bad bullets in Changelog that were causing styling issues and no line breaks. A good example is here:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
-– Refactored Controlled Vocabulary support to allow for new, raw vocabs without i18n translations. The motivation here is that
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,8 +8,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
-– Refactored Controlled Vocabulary support to allow for new, raw vocabs without i18n translations. The motivation here is that
```
murny, line 11: Appears to be a different "-" character and markdown can't render it correctly. So updated these to use a real "-" character

pgwillia: Looks like an [en or em dash](https://www.grammarly.com/blog/hyphens-and-dashes/)  Thanks for cleaning that up!

pgwillia (approved):

PR #2435, opened 2021-07-19
Add batch ingest form with google file picker and spreadsheet validation

murny:
This PR will add the batch ingest form so a user will be able to create a new batch ingest. The form includes a file picker that will allow a user to choose which google drive files and spreadsheets to upload.

Form: 
image

image

Error handling:
image

File picker:
image


test/models/batch_ingest_test.rb
```diff
@@ -52,4 +52,30 @@ class BatchIngestTest < ActiveSupport::TestCase
...
+                                                          collection_id: 'BADID',
+                                                          owner_id: 'BADID' }) do
+        batch_ingest = BatchIngest.new(
+          title: @batch_ingest.title,
+          user_id: @batch_ingest.user_id,
+          batch_ingest_files: @batch_ingest.batch_ingest_files,
+          google_spreadsheet_id: @batch_ingest.google_spreadsheet_id,
+          google_spreadsheet_name: @batch_ingest.google_spreadsheet_name,
+          access_token: @batch_ingest.access_token
+        )
```
pgwillia, line 68: Does the fixture not work here? The validation happens when `valid?` is called, right? This test is obviously different because it uses VCR to setup some errors in the spreadsheet so maybe I'm missing something in my understanding.

[edit] Oh, is because the validation happens on `create`?

app/javascript/src/admin/new_batch_ingest.js
```diff
@@ -0,0 +1,131 @@
...
+      + 'image/jpeg,'
+      + 'image/jpg,'
+      + 'audio/mpeg,'
+      + 'video/mpeg,'
+      + 'application/zip,'
+      + 'text/plain,'
+      + 'application/pdf,'
+      + 'application/msword,'
+      + 'application/vnd.ms-excel',
+    );
```
pgwillia, line 72: I think I sorta remember talking about this in the demo.  Is [setMimeTypes](https://developers.google.com/picker/docs/reference#view) a hard limit to which files can be picked? 

We don't put any limits on the draft files picker as far as I can tell ([dropzone](https://www.dropzonejs.com/#config-acceptedFiles)).

pgwillia (commented):
This is looking good 👍

There are a few things I'm wondering about.

app/javascript/src/admin/new_batch_ingest.js
murny: Yeah its a hard limit on which files are shown in the picker.  Only the mime types listed here will be shown in the filepicker. Basically I want the user to be able to choose anything BUT the special google file types (aka choose a word doc file and not a google doc, excel spreadsheet not a google spreadsheet, etc). So I'm more then okay with adding a bunch of mime types here, if I missed any?

test/models/batch_ingest_test.rb
murny: Yeah validation happens only "on create" for the spreadsheet validation. 

Here I mock out the google API call that is called to retrieve the spreadsheet which returns some bad IDs for community_id/collection_id/owner_id. So when we attempt to create a new record, it will take the `google_spreadsheet_id` and fetch the spreadsheet from google API, uses our VCR tape values, runs the spreadsheet validation code and finds out the community/collection and owner ids doesnt exist in the database! Then throws an error which is what I expected, so all is passing!

pgwillia (dismissed):

app/javascript/src/admin/new_batch_ingest.js
pgwillia: I asked around and am still getting feedback:  image/TIFF and image/jp2, image/jpx, image/jpm, video/mj2 for [jpeg2000](https://datatracker.ietf.org/doc/html/rfc3745) were noted by @sarahseverson

murny: Awesome! Will add these 👍

pgwillia: Some more information.  Sean brought up that
> we do have some unusual ones that come in from time to time. This is usually things like [software](https://era.library.ualberta.ca/items/52e1d288-9184-4dfe-95e1-d365efea6863) or data files. Often we've requested that users package things like this up in a zip file but I have my doubts about the stability of that format, since we've had cases where we're not able to open a zip since it wasn't packaged properly. 

These are the content types currently in ERA:
["image/png", "image/tiff", "application/pdf", "image/jpeg", "application/vnd.openxmlformats-officedocument.wordprocessingml.document", "audio/mpeg", "application/msword", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", "application/x-ole-storage", "text/comma-separated-values", "application/vnd.openxmlformats-officedocument.presentationml.presentation", "application/xml", "application/postscript", "video/mp4", "application/zip", "application/octet-stream", "audio/x-wav", "text/plain", "application/vnd.ms-excel", "video/x-msvideo", "application/x-bibtex", "text/html", "application/gzip", "audio/mp4", "video/mpeg", "audio/vnd.wave", "video/quicktime", "application/x-7z-compressed", "application/x-spss-sav", "application/rtf", "text/x-objcsrc", "application/x-rar", "message/rfc822", "application/vnd.ms-asf", "audio/x-wave", "audio/wav", "text/rtf", "application/vnd.ms-excel.sheet.binary.macroenabled.12", "application/vnd.oasis.opendocument.text", "text/x-csrc", "text/csv", "image/gif", "text/x-matlab", "application/vnd.openxmlformats-officedocument.presentationml.slideshow", "application/x-tar", "application/epub+zip", "audio/x-ms-wmv", "application/x-rar-compressed", "text/x-r-source", "application/x-bibtex-text-file", "application/excel", "audio/x-ms-asx", "video/x-m4v", "image/svg+xml", "application/vnd.android.package-archive", "video/x-flv", "application/vnd.ms-access", "application/x-zip-compressed", "audio/x-m4a", "video/x-ms-wmv", "application/vnd.ms-excel.sheet.macroenabled.12", "application/vnd.ms-powerpoint", "text/x-r-sweave", "application/sql", "application/x-xz", "text/x-bibtex", "application/mp4", "application/json", "audio/x-ms-wma", "inode/x-empty", "application/vnd.openxmlformats-officedocument.wordprocessingml.template", "audio/x-mpegurl"]

murny: Added all these mimetypes now 👍

pgwillia (approved):

PR #2441, opened 2021-07-21
Remove rufus-scheduler

pgwillia:
## Context

In version 1.0 of sidekiq-cron they changed to
> use fugit instead of rufus-scheduler - API of cron didn't change (rufus scheduler is using fugit)

https://github.com/ondrejbartas/sidekiq-cron/blob/6a0aeff6c900f3b7246734282f6869c61e1d5b4e/Changes.md#v-100

That means that this dependency is unused so there isn't a good reason to keep it around.

Closes #2434

## What's New

Removes `rufus-scheduler` gem from our dependencies.


murny (approved):
🎉

Issue #2419, opened 2021-07-05
Improve our smelliest tests [Minitest/MultipleAssertions]


pgwillia:
**Is your feature request related to a problem? Please describe.**
Looks like rubocop-minitest was fixed and it's calling us out on too many assertions in our tests.  I moved the max [Minitest/MultipleAssertions](https://docs.rubocop.org/rubocop-minitest/cops_minitest.html#minitestmultipleassertions) to 15 which is kind of smelly.  Additionally I disabled the cop on a number of our worst violations.  This is the ticket to do the work to clean up these tests.

#2416

**Describe the solution you'd like**
Take a good look at these tests and ensure that they are testing one thing.  Also think about the setup of the tests: use the fixtures, change the config of dependencies to focus the test, don't test the dependencies.

**Additional context**
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/integration/oaisys_list_identifiers_test.rb#L18-L111
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/integration/oaisys_list_sets_test.rb#L12-L108
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/models/user_test.rb#L26-L64
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/models/user_test.rb#L66-L102
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/services/doi_service_test.rb#L9-L114
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/services/statistics_test.rb#L13-L68
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/system/admin_users_show_test.rb#L34-L108
- [ ] https://github.com/ualbertalib/jupiter/blob/279205ac05c2ce79535a29b6c21b42b7d6c1bdcd/test/system/collections_pagination_and_sort_test.rb#L15-L110
[...]

Issue #2129, opened 2021-01-28
Occasional Test Suite failures


pgwillia:
**Describe the bug**
Occasionally CI will catch failing tests like the following:

1. 
```
Failure:
DoiResetTest#test_reset_button_displays_as_needed [/home/runner/work/jupiter/jupiter/test/integration/doi_reset_test.rb:19]:
Expected exactly 0 elements matching ".admin-sidebar .btn", found 1..
Expected: 0
  Actual: 1
```
rails test test/integration/doi_reset_test.rb:12
**To Reproduce**
Run options: --seed 41156

**Expected behavior**
Stable and consistent test suite.

**Additional context**
There are others I've seen recently.  I'll update as I experience them.

pgwillia:
`rails system:test ` on WSL2 (so might just be my machine 😅 )

```
Failure:
AdminUsersShowTest#test_should_be_able_to_login_as_a_regular_user [/mnt/c/Users/pgwillia/Code/ualbertalib/jupiter/test/system/admin_users_show_test.rb:116]:
expected to find visible css "h1" with text "Dashboard" but there were no matches. Also found "ERA:\nEducation and Research Archive", which matched the selector but not all filters.


rails test test/system/admin_users_show_test.rb:107
```

```
Error:
EditItemOldLicenseTest#test_can_edit_and_save_an_item_with_an_old_license:
Capybara::ElementNotFound: Unable to find link or button "Save and continue to next step"
    test/system/edit_item_old_license_test.rb:21:in `block in <class:EditItemOldLicenseTest>'


rails test test/system/edit_item_old_license_test.rb:5
```

```
Failure:
AdminUsersIndexTest#test_should_be_able_to_sort_columns [/mnt/c/Users/pgwillia/Code/ualbertalib/jupiter/test/system/admin_users_index_test.rb:25]:
expected to find visible css "tbody tr:last-child th[scope=\"row\"]" with text "administrator@example.com" but there were no matches. Also found "tyrion_lannister@example.com", which matched the selector but not all filters.


rails test test/system/admin_users_index_test.rb:5
```

![failures_test_should_be_able_to_sort_columns](https://user-images.githubusercontent.com/1220762/113048058-5772e400-915f-11eb-9459-00a670bf79e7.png)
![failures_test_can_edit_and_save_an_item_with_an_old_license](https://user-images.githubusercontent.com/1220762/113048061-580b7a80-915f-11eb-8e58-cc646e6ede12.png)

[...]

PR #2460, opened 2021-08-06
Fix some test debt

pgwillia:
## Context

I've notice that the admin users index test will fail intermittently in our dependabot PRs.  Since it is a broken window I made time to fix it.  #2129

While I was investigating that the test to excercise sorting and pagination of communities failed.  I also took some time to address this test as well. #2419

Before:
image
After:
image

## What's New

- Fixed test that exercises sorting

The selector will partially match the page before it's been updated and give a negative result.  First checking for something that has changed on the page will ensure that the following assertions will have a chance to succeed.

- Refactored one of our smelliest tests

Prefer using fixtures to creating new objects.

This test had many assertions  which is contrary to the imperative to test only one thing at a time.


murny (approved):
Awesome work 👏! Thanks so much for taking the time to investigate these 🎉

PR #2468, opened 2021-08-12
Some Adjustments to the `Digitization::Book` Model

pgwillia:
## Context

Faced with real data about the folk fest items demonstrated that I had introduced some errors. 

Choosing to call the attachment `pdf` earlier was limiting.  For Items and Thesis these attachments have been called `files`.  By making this change now we can reuse the `DownloadController` and `files_section_sidebar` partial.

Related to #2011

## What's New

- fix URI and correct predicates
- Making Book act as rdfable will allow us to use the predicate relationships later.
- change `pdf` to `files`

### DIFF
--- a/app/models/digitization/book.rb
+++ b/app/models/digitization/book.rb
@@ -3,7 +3,6 @@ class Digitization::Book < JupiterCore::Depositable
   acts_as_rdfable
 
   has_one_attached :historical_archive
-  has_many_attached :files, dependent: false
   has_one :fulltext, dependent: :destroy, class_name: 'Digitization::Fulltext', inverse_of: :book,
                      foreign_key: :digitization_book_id
 
--- a/app/models/digitization/book.rb
+++ b/app/models/digitization/book.rb
@@ -1,7 +1,9 @@
 class Digitization::Book < JupiterCore::Depositable
 
+  acts_as_rdfable
+
   has_one_attached :historical_archive
-  has_one_attached :pdf, dependent: false
+  has_many_attached :files, dependent: false
   has_one :fulltext, dependent: :destroy, class_name: 'Digitization::Fulltext', inverse_of: :book,
                      foreign_key: :digitization_book_id
 
@@ -30,7 +32,7 @@ class Digitization::Book < JupiterCore::Depositable
   validates :temporal_subjects, edtf: true
 
   def all_subjects
-    topical_subjects + temporal_subjects + geographic_subjects
+    topical_subjects + temporal_subjects.to_a + geographic_subjects.to_a
   end
 
   def all_contributors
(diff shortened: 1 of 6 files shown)


app/models/digitization/book.rb
```diff
@@ -30,7 +32,7 @@ class Digitization::Book < JupiterCore::Depositable
   validates :temporal_subjects, edtf: true
 
   def all_subjects
-    topical_subjects + temporal_subjects + geographic_subjects
+    topical_subjects + temporal_subjects.to_a + geographic_subjects.to_a
```
murny, line 35: Just curious about this change? Was it doing something weird? Having `to_a` casts everything into an array and appends them all?

app/models/digitization/book.rb
```diff
@@ -1,7 +1,9 @@
 class Digitization::Book < JupiterCore::Depositable
 
+  acts_as_rdfable
+
   has_one_attached :historical_archive
-  has_one_attached :pdf, dependent: false
+  has_many_attached :files, dependent: false
```
murny, line 6: theres already a `has_many_attached :files, dependent: false` on Line 14 in this file?

db/migrate/20210422161911_rename_columns_in_digitization_book.rb
```diff
@@ -11,6 +11,7 @@ def change
       t.rename :note, :notes
       t.rename :geographic_subject, :geographic_subjects
       t.rename :topical_subject, :topical_subjects
+      t.references :logo, foreign_key: {to_table: :active_storage_attachments, column: :id, on_delete: :nullify}
```
murny, line 14: Just need to be careful editing migrations like this? Assume we okay here since it's just me and you developing here and none of this integration code has made it out to staging/production etc


Also not seeing any relationship for this in our model? e.g:  `has_one_attached :logo` or `belongs_to :logo` etc

app/models/digitization/book.rb
pgwillia: It does something weird if when `temporal_subjects` or `geographic_subjects` are `nil`
```
> [] + nil
TypeError: no implicit conversion of nil into Array
```

It seems like we've encountered this elsewhere
https://github.com/ualbertalib/jupiter/blob/0f13809f4c6d98eaefbf8207ff77e3bf6553171d/app/models/item.rb#L74-L76

db/migrate/20210422161911_rename_columns_in_digitization_book.rb
pgwillia: I based this on what's happening in Item and Thesis where the schema shows 
https://github.com/ualbertalib/jupiter/blob/0f13809f4c6d98eaefbf8207ff77e3bf6553171d/db/schema.rb#L348

But not showing any relationship in the model.  It's needed for the code in Depositable to work
https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/models/jupiter_core/depositable.rb#L102-L108

I can go back and do a new migration.  I should practice good habits.

murny: Ah, that makes sense. Guess thats an improvement we could make in the future where we can use real database associations now that we are no longer on fedora. This `logo_id` being a proper join and we can walk the association instead of doing code like this:
https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/models/jupiter_core/depositable.rb#L110-L112

app/models/digitization/book.rb
murny: Makes sense! Thanks for explanation 👍

murny (approved):
gif

murny (approved):

PR #2469, opened 2021-08-12
`Digitization::Book` view completed

pgwillia:
## Context

While working through the ingest tasks I wanted the visual feedback that everything was working as expected.  This led to ensuring that the download/view links worked on the page.  Initially they were directing into the era subdomain, which worked but might be confusing to our users.

Related to #2011

## What's New

image (1)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -225,6 +225,8 @@ en:
       download: 'Download'
       download_all: 'Download All'
       header: 'Files'
+      ccid_restricted_item: Item Restricted to University of Alberta Users
+      login_with_ccid_to_view: Log In with CCID to View Item
 
   layouts:
     admin:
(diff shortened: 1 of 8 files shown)


config/routes.rb
```diff
@@ -10,7 +10,14 @@
 
     root to: 'welcome#index'
 
-    resources :items, only: [:show, :edit] do
+    concern :downloadable do
```
murny, line 13: Very cool 👍

config/locales/en.yml
```diff
@@ -212,6 +212,11 @@ en:
       skip_to_results: 'Skip to Search Results'
     range_facet_result:
       to: 'to'
+    files_section_sidebar:
```
murny, line 215: This should be a move diff? I see only an addition but no delete?

We now have this duplicated on Line ~642 in this file, so can probably just cut and paste that one here?

murny (commented):
LGTM, just one comment on duplication of the locales since we now moved the partial file

config/locales/en.yml
pgwillia: Good catch!  That's exactly the correction I've made.  

There is a similar section under `search` which remains because it is actually a separate view.

murny (approved):
🎉

PR #2470, opened 2021-08-12
Add task that will kick off job for  digitization metadata

pgwillia:
## Context

Our digitization content metadata will come to us as exports from a triplestore.  We need to parse this graph and create objects containing the metadata.

Initially this was all just one big rake task.  Moving the logic to a Job allows the inputs to be validated and the logic to be tested.

CSV is part of the ruby standard library.  We've adopted a practice of using bundle to manage the require statements.  Since this is part of the standard library and outside of bundler's perview we prefer to include the require statement once in an initializer.

Related to #2011 

## What's New
- Add task that will kick off job for batch ingestion of digitization metadata from a csv containing triples.  Accompanied by supporting model, relationships and tests.


app/jobs/digitization/batch_metadata_ingestion_job.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  # Most of the time we can directly assign the value to the matching attribute on the model but occasionally we'll need
+  # to do some parsing or preprocessing of this content
+  # use: delegate_transform(book, OpenStruct.new(predicate: 'http://id.loc.gov/vocabulary/identifiers/local', term: 'P010572.1'))
+  # delegates to: parse_identifier(book, 'P010572.1')
+  def delegate_transform(book, attributes)
+    return if attributes.predicate.to_s.blank?
+
+    return if URI_CODE_TO_TRANSFORM_METHOD[attributes.predicate.to_s.to_sym].blank?
```
murny, line 87: Are we expecting the `URI_CODE_TO_TRANSFORM_METHOD` to have more key/values in the future? Otherwise couldnt we just check if it has a predicate of `'http://id.loc.gov/vocabulary/identifiers/local'` and if so just call `parse_identifer(book, attributes.term.to_s)` etc? 

Instead of doing this hash look up and send stuff?

app/jobs/digitization/batch_metadata_ingestion_job.rb
```diff
@@ -0,0 +1,117 @@
...
+      rescue StandardError
+        RDF::Literal.new(row['Value'])
+      end
+      graph << [subject, predicate, object]
+    end
+
+    graph
+  end
+
+  URI_CODE_TO_TRANSFORM_METHOD = {
```
murny, line 47: Could maybe move this to the top of the file? So it lives with the `PEEL_ID_REGEX` constant

config/initializers/stdlib.rb
```diff
@@ -0,0 +1 @@
+require 'csv' # interface to csv (comma seperated value) files and data
```
murny, line 1: Curious if its better to just require this where we using it 🤔 ? instead of requiring it for the entire project? 

(like thinking more from a maintenance side such as if we decide to do something different in the model/job, maybe use Excel/etc etc then good chance we will forget about this initializer )

db/migrate/20210723172409_create_digitization_batch_metadata_ingests.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDigitizationBatchMetadataIngests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_batch_metadata_ingests, id: :uuid do |t|
+
```
murny, line 4: Did you want any kind of identifier when running these ingests? Like giving it a "Title"? Or makes no sense for your use case?

app/models/digitization/batch_metadata_ingest.rb
```diff
@@ -0,0 +1,43 @@
...
+  enum status: { created: 0, processing: 1, completed: 2, failed: 3 }
+
+  belongs_to :user
+
+  has_many :books, dependent: :nullify, class_name: 'Digitization::Book',
+                   foreign_key: :digitization_batch_metadata_ingest_id,
+                   inverse_of: :digitization_batch_metadata_ingest
+  has_one_attached :csvfile
+
+  validates :csvfile, presence: true
```
murny, line 12: Something I forgot on my batch ingest work, but might be a good idea to add a validation on the enum status?

`validates :status, presence: true`

🤔

lib/tasks/digitization_batch_ingest.rake
```diff
@@ -0,0 +1,33 @@
+namespace :digitization do
+  desc 'batch ingest for multiple items metadata from a csv file - used by Admin and Assistants'
```
murny, line 2: Is plan to eventually go to google drive or something? Might not be able to do this in a post-Cloud world?

test/jobs/digitization/batch_metadata_ingestion_job_test.rb
```diff
@@ -0,0 +1,57 @@
+require 'test_helper'
+
+class Digitization::BatchMetadataIngestionJobTest < ActiveJob::TestCase
+
+  setup do
+    RdfAnnotation.create(table: :digitization_books, column: :dates_issued, predicate: ::RDF::Vocab::DC.issued)
```
murny, line 6: Do we need to clean this up this setup code in a `teardown` block?

murny (approved):
All looks really good. Mostly just some questions to better understand the code myself 👍

app/jobs/digitization/batch_metadata_ingestion_job.rb
pgwillia: When I was working through this there were others which is why I have this the way it is.   Turns out that the others were all errors I had made that I fixed in #2468.  

I can see it being more readable if I drop the hash lookup and `send` and then worry about other predicates as they emerge.  Come to think of it this might be more readable as a case statement in the method.

config/initializers/stdlib.rb
pgwillia: Once [upon a time](https://github.com/ualbertalib/jupiter/pull/2239#discussion_r618803144) I learned that it's best to let bundler take care of optional requires.  In ruby it's require once and available everywhere then it follows that it should only be declared once.  Since the ruby standard library isn't managed by bundler I opted to put it in an initializer to centralize it.

pgwillia: At one point I was using this gem: https://rubygems.org/gems/csv not realizing that the same functionality was in the Ruby standard library.

lib/tasks/digitization_batch_ingest.rake
pgwillia: Seeing the work that you'd done that's exactly what I started thinking.  Any admin interface for digitization is outside the MVP so that might be a future enhancement.

Any on server tasks will be more difficult in the cloud.  I think there might be some way to stage data from blob storage?  But I'm not really sure what that means.

test/jobs/digitization/batch_metadata_ingestion_job_test.rb
pgwillia: I don't feel like I've found the best solution for this setup.  We've been building this structure in migrations, for example https://github.com/ualbertalib/jupiter/blob/d25903a8e91996ca871bfe3253cd3b23ab52f57a/db/migrate/20190926213733_annotate_tables_with_rdf.rb#L71-L79

For the AIP tests we've got similar code in a helper https://github.com/ualbertalib/jupiter/blob/86e42f83d86859ae3a1d92407a5963ca9f861397/test/support/aip_helper.rb#L149-L169

Because our rdf annotations is basically part of our database schema we need a way to sync the rdf annotations between the migrations, seeds and tests. I don't think it should be setup and teardown for each test, it should be available and unchanging for all the tests -- but I don't know what that looks like.

app/jobs/digitization/batch_metadata_ingestion_job.rb
murny: Yeah think a simple case statement would be easier to read/maintain in the future. I like that idea (could just be an if statement now, and can grow into a case too)

config/initializers/stdlib.rb
murny: Make sense, let's leave this then 👍

lib/tasks/digitization_batch_ingest.rake
murny: Yeah needs some research, but glad your thinking about this 👍

db/migrate/20210723172409_create_digitization_batch_metadata_ingests.rb
pgwillia: I added the title. Good idea!

test/jobs/digitization/batch_metadata_ingestion_job_test.rb
pgwillia: I'm thinking about modifying [acts_as_rdfable](https://github.com/ualbertalib/acts_as_rdfable/) to help us with this.

Any of the add/remove/delete methods in [ActsAsRdfable::MigrationAnnotations::RdfConfig](https://github.com/ualbertalib/acts_as_rdfable/blob/master/lib/acts_as_rdfable/migration_annotations.rb) could create/update a `db/seeds/rdf_annotations.rb` file in our application (maybe [seed_dump](https://rubygems.org/gems/seed_dump)) with the current dump of the database.
```
# lib/acts_as_rdfable/migration_annotations.rb
def dump_rdf_annotations
  SeedDump.dump(RdfAnnotation, file: 'db/seeds/rdf_annotations.rb')
end
```
```
# db/seeds/rdf_annotations.rb
RdfAnnotation.create!([
  {table: "items", column: "title", predicate: "http://purl.org/dc/terms/title"},
  {table: "items", column: "fedora3_uuid", predicate: "http://terms.library.ualberta.ca/fedora3UUID"},
  {table: "items", column: "depositor", predicate: "http://terms.library.ualberta.ca/depositor"},
...
  {table: "items", column: "alternative_title", predicate: "http://purl.org/dc/terms/alternative"}
])
```
We could check this in along with our `db/schema.rb`.  Then load it from our `db/seeds.rb`
```
# db/seeds.rb
load('db/seeds/rdf_annotations.rb')
```
And for our tests, There's `Rails.application.load_seed` which isn't exactly what we want but we could add
```
class ActiveSupport::TestCase

  # Setup all fixtures in test/fixtures/*.yml for all tests in alphabetical order.
  fixtures :all
  load('db/seeds/rdf_annotations.rb')
```

lagoan: This is a much better solution than what we have right now as it will actually represent what is on the system, rather than repeating the code with a high likelihood that we will miss something.

app/models/digitization/batch_metadata_ingest.rb
murny, line 12: Title is now null false, might want a to check that at the Rails level as well

app/jobs/digitization/batch_metadata_ingestion_job.rb
```diff
@@ -0,0 +1,98 @@
...
+      graph.query(query_for_this_items_attributes) do |attributes|
+        if RDF::Vocab::Identifiers.local == attributes.predicate
+          # There isn't a single local identifier attribute in our model so we will need to parse this value
+          parse_identifier(book, attributes.term.to_s)
+        else
+          assign_attribute(book, book.attribute_for_rdf_annotation(attributes.predicate.to_s)&.first&.column,
+                           attributes.term.to_s)
+        end
+      end
+      book.save!
```
murny, line 72: is there any concern of code failing and sidekiq constantly retrying the job resulting in duplicate records or anything of that nature? 

Do we need to wrap this in a DB transaction?

pgwillia: This is wrapped in a DB transaction but it's further up

https://github.com/ualbertalib/jupiter/blob/971c8466a53e0763f16fff122d82b8ecacd5d38c/app/jobs/digitization/batch_metadata_ingestion_job.rb#L16-L21

The validations will also catch duplicate records.  

I think this scenario is covered by 
https://github.com/ualbertalib/jupiter/blob/971c8466a53e0763f16fff122d82b8ecacd5d38c/test/jobs/digitization/batch_metadata_ingestion_job_test.rb#L26-L41
Is there something I could add?

murny: Oh nice, missed that. LGTM 👍 .

Test is fine (its a bit hard to test that since it gets into sidekiq's retry mechanisms/etc)

murny (approved):

PR #2471, opened 2021-08-12
Add task that will kick of setup job for artifacts

pgwillia:
## Context

Initially this was all just one big rake task.  Moving the logic to a Job allows the inputs to be validated and the logic to be tested.

This is just the setup for the jobs that will actually do the work.  Thinking that each object should have it's own job so that the transactions can stay small and we can get feedback on them independently.

Related to #2011


app/models/digitization/batch_artifact_setup_ingest.rb
```diff
@@ -0,0 +1,52 @@
+class Digitization::BatchArtifactSetupIngest < ApplicationRecord
+
+  enum status: { created: 0, processing: 1, completed: 2, failed: 3 }
+
+  belongs_to :user
+
+  has_many :processing, dependent: :nullify, class_name: 'Digitization::Book',
```
murny, line 7: How do these associations work? Is there some magic happening to filter them by `processing`/`completed`/`failed`?  As all three just look like a `has_many :books`?

app/models/digitization/batch_artifact_setup_ingest.rb
```diff
@@ -0,0 +1,52 @@
...
+  validate :spreadsheet_has_required_data
+
+  def spreadsheet_has_required_data
+    return unless csvfile.attached? && attachment_changes['csvfile']
+
+    CSV.foreach(attachment_changes['csvfile'].attachable[:io].path, headers: true) do |row|
+      # Check if required fields are filled out
+      if row['Code'].blank?
+        errors.add(:csvfile,
+                   "Local Identifier (Code) not found for row #{$INPUT_LINE_NUMBER} of spreadsheet")
```
murny, line 29: Minor but could move these error messages to i18n

app/models/digitization/batch_artifact_setup_ingest.rb
murny: Think my confusion is from the database level, its just a join on the book table with this table and they use the same `digitization_batch_artifact_setup_ingests_id` foreign key. So at the DB level, and when rails loads the record from the DB, it has no idea if its `completed/failed/ or processing`? as each three of these associations will be populated from the same records? 

It's almost like you want the `book` model to have the `status` enum, and here you can use a has many with scopes/where query like so:

```ruby
has_many :processing_books, -> { where(batch_ingest_status: :processing) } , dependent: :nullify, class_name: 'Digitization::Book',
                       foreign_key: :digitization_batch_artifact_setup_ingests_id,
                       inverse_of: false
has_many :completed_books, -> {  where(batch_ingest_status: :completed) } , dependent: :nullify, class_name: 'Digitization::Book',
                       foreign_key: :digitization_batch_artifact_setup_ingests_id,
                       inverse_of: false
has_many :failed_books, -> {  where(batch_ingest_status: :failed) } , dependent: :nullify, class_name: 'Digitization::Book',
                       foreign_key: :digitization_batch_artifact_setup_ingests_id,
                       inverse_of: false
```

app/models/digitization/batch_artifact_setup_ingest.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+      errors.add(:csvfile, "Noid not found for row #{$INPUT_LINE_NUMBER} of spreadsheet") if row['Noid'].blank?
+    end
+  end
+
+  def processing!(book)
```
murny, line 35: Would maybe rename these methods as they are very similar to the enum methods  (ails.org/v5.2.6/classes/ActiveRecord/Enum.html). Enum provides a bang method  for each status. Here we have the same method just with allowing an extra argument for book

`processing!` vs `processing!(book)`

So maybe we can differentiate the two somehow?

app/models/digitization/batch_artifact_setup_ingest.rb
```diff
@@ -0,0 +1,52 @@
...
+    update(status: :processing) if failed.blank?
```
murny, line 36: If you like you can use that enum method i mentioned in last comment instead of using update:

`processing!` vs `update(status: :processing)`

app/jobs/digitization/batch_artifacts_setup_ingestion_job.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  queue_as :default
+
+  rescue_from(StandardError) do |exception|
+    batch_ingest = arguments.first
+    batch_ingest.update(error_message: exception.message, status: :failed)
+    raise exception
+  end
+
+  def perform(batch_ingest)
```
murny, line 11: wonder if we be explicit about the model here? Would help with searching the code, and give instant clarity about what model we talking about here.

Currently I assume this `batch_ingest` record is a BatchIngest model or maybe a BatchMetadataIngest model? 

But here we talking about a `batch_artifact_setup_ingest` record

db/migrate/20210723183112_create_digitization_batch_artifact_setup_ingests.rb
```diff
@@ -0,0 +1,15 @@
+class CreateDigitizationBatchArtifactSetupIngests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_batch_artifact_setup_ingests, id: :uuid do |t|
```
murny, line 3: Was there any reason for adding `setup` to this name? 
Noticed the metadata was just called `batch_metadata_ingests`? Should this just be `batch_artifact_ingests`? Or `setup` is needed here?

test/models/digitization/batch_artifact_setup_ingest_test.rb
```diff
@@ -0,0 +1,73 @@
...
+    @batch_ingest.csvfile.attach(io: csv_content,
+                                 filename: 'artifacts_manifest.csv', content_type: 'text/csv')
+
+    assert_not @batch_ingest.valid?
+
+    errors = ['Local Identifier (Code) not found for row 1 of spreadsheet', 'Noid not found for row 1 of spreadsheet']
+    assert_equal(errors, @batch_ingest.errors[:csvfile])
+  end
+
+  test 'processing and completed status' do
```
murny, line 34: Yeah really curious if you added these books to these has_many associations via these helper methods, then called a `reload!` from the DB. Would you still have these associations populated 🤔  Or is everything just being represented in memory?

murny (commented):
Looking pretty good. Just some comments about those status has_many associations and naming

db/migrate/20210723183112_create_digitization_batch_artifact_setup_ingests.rb
pgwillia: `BatchArtifactSetupIngest` is intentional. In the [next set of work on this](https://github.com/ualbertalib/jupiter/pull/2472) I introduce the `BatchArtifactIngestJob` which attaches the artifacts to the Book object that they belong to.
https://github.com/ualbertalib/jupiter/blob/2197af6b16961bd5a724c105cedb7f0c9c8ee1c6/app/jobs/digitization/batch_artifacts_setup_ingestion_job.rb#L11-L23

Thanks for your questions.  It makes it clear that the way I've organized this is difficult to follow. Perhaps there's a better way. Let me think more about this.

PR #2498, opened 2021-08-31
Change subdomain for digitalcollections

pgwillia:
## Context

I had erroneously conflated the namespace that we created for the difference between ERA models and Digitization models was also the subdomain for the digitalcollections url.  This decision was made in October 2020 and the ip address was allocated for digitalcollections.library.ualberta.ca in December 2020.

## What's New

Changed the subdomain and references from `digitization` to `digitalcollections`


murny (approved):
LGTM 👍

Issue #2298, opened 2021-04-26
OAI: change mappings for oai_etdms date values


anayram:
**Request**
Change mappings for `oai_etdms` `date`. Instead of using `dc:dateAccepted values`, map from `ual:graduationDate` to meet [LAC requirements](https://www.bac-lac.gc.ca/eng/services/theses/Pages/universities.aspx). This change would meet LAC's requirements (to deliver an _issuance_ date, rather than a _processing_ date, e.g. institutional acceptance date). Additionally, I understand this change would also meet FGSRs preferences but I don't have any details on this.

**Background**
LAC requires "Date on title page" of theses to be captured in `etd-ms:date` OAI elements. Dates in theses' title pages are currently captured in `ual:graduationDate` properties. I understand that FGSR has also expressed a preference in using `ual:graduationDate` as the main date, although I don't have exact details of this conversation.

Currently, our OAI mappings use values from `dc:dateAccepted`. Changing mappings to use `ual:graduationDate` would meet both LAC requirements and FGSR's preferences.

We have updated mappings in metadata documentation files at:
[Jupiter Oai_Etdms Application Profile](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/profile_oai_etdms.md
)
and
[Jupiter OAI mapping](https://github.com/ualbertalib/metadata/blob/master/data_dictionary/Jupiter%20OAI%20mapping.xlsx)

**Context**
From the metadata point of view, this change is crucial before LAC's harvest, but we understand that service and development work priorities will inform work on this request. Thank you!

@sfarnel @ConnorSheremeta

jefferya:
Notes from investigations:

* acts_as_rdfable serializes OAI. Details [here](https://github.com/ualbertalib/acts_as_rdfable/blob/ad213eae97d8166a1a190702f3e0eb9a5014b808/lib/acts_as_rdfable/serializers/oai_etdms.rb#L14)
 ```
      oai.etd_ms(:date, object.date)
```
* `date` defined in the thesis decorators. Details [here](https://github.com/ualbertalib/jupiter/blob/d25903a8e91996ca871bfe3253cd3b23ab52f57a/app/decorators/metadata/oai_etdms/thesis_decorator.rb#L21-L24) 

ToDo:

[...]

PR #2510, opened 2021-09-10
oaisys: change etdms date source to graduation date as per LAC spec #2298

jefferya:
## Context

Update decorator for theses that provides the specified [LAC etdms date and format](https://www.bac-lac.gc.ca/eng/services/theses/Pages/universities.aspx) replacing the the current date_accepted and created_at dates that don't meet the LAC requirements

Related to #2298 

## What's New

* updates etdms thesis decorator to use graduation date as per #2298 
* fixes: thesis seeds created with improper graduation_date
    * format correct in prior commit (perhaps rebase mistake): https://github.com/ualbertalib/jupiter/commit/b0f53cba7a70d6dbb8209cdeda32d913971435bd
    * revert format to prior commit https://github.com/ualbertalib/jupiter/blob/fa44471da977863283ad9b6179ba23525f95c373/db/seeds.rb#L199


config/locales/en.yml
```diff
@@ -551,6 +551,8 @@ en:
       graduation_terms:
         spring: 'Spring'
         fall: 'Fall'
+        spring_num: '06'
+        fall_num: '11'
```
pgwillia, line 555: Seems like it might be redundant with https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/models/draft_thesis.rb#L5-L12

Can we consolidate to use one or the other?

pgwillia (commented):
I think you're on the right track with this.

config/locales/en.yml
jefferya: My struggle here is developing a mental model how `Thesis` and `DraftThesis` objects relate. Is there a document that describes the two at a high level contrasting the differences? My naive first assumption was that the `Thesis` and `DraftThesis` would be very similar models with only minor differences owning to the draftiness of a DraftThesis. Looking through the codebase, I see multiple areas that transition dates (e.g., graduation date) from a text representation to a number (string or numerical datatype) version, for example, the conversions ( 6 => Spring, `'06 (Spring)' => '06'`, `'Spring' => '06'`, `'06' => 'Spring'`). The above code in the context of `DraftThesis` is using the text `06 (Spring)` whereas the data in the `Thesis` context is `Spring` (no round brackets etc.) so I unsure if a consolidation can happen due to differences in the string.  I like the idea of consolidation but I not confident I have an understanding of the nuances in the aforementioned and following locations spread across different models.  

https://github.com/ualbertalib/jupiter/blob/a82bdd866e86f508a45e6d48010178547ce45565/app/helpers/items_helper.rb#L53-L56

https://github.com/ualbertalib/jupiter/blob/d1c8fb1d059e8191062bbcaff1f83a0fd4fa16d8/lib/tasks/batch_ingest.rake#L202-L203

https://github.com/ualbertalib/jupiter/blob/d1c8fb1d059e8191062bbcaff1f83a0fd4fa16d8/config/locales/en.yml#L353-L354

pgwillia: Effort was made so that the underlying data was consistent: https://github.com/ualbertalib/metadata/issues/383 and I believe this is still the case.  I reviewed the graduation_date on staging and can confirm that all graduation_dates are YYYY or YYYY-MM.  So now I'm curious where the 'Fall YYYY` or `Spring YYYY` came from.  It might be possible to use the graduation date directly.


The public presentation of these dates (Fall 2017) was discussed https://github.com/ualbertalib/jupiter/issues/1003 where OAI was out of scope.
The admin deposit interface should show the number (06 for spring and 11 for fall) https://github.com/ualbertalib/jupiter/pull/804 https://github.com/ualbertalib/jupiter/issues/709

pgwillia: I can see that our seeds data is spitting out Fall.  That looks like a bad rebase https://github.com/ualbertalib/jupiter/commit/b0f53cba7a70d6dbb8209cdeda32d913971435bd.  It was fixed before that commit.

jefferya: Thx for checking the underlying DB. I can confirm, the `graduation_date` appears to work as the DB contents conforms to the LAC format. The `Fall/Spring YYYY` was from me looking at the DB created from the seeds.rb file and a conversation with Mariana on Theses inputs which I'd falsely assumed aligned with the DB storage format. I'm updating the seeds.rb

app/decorators/metadata/oai_etdms/thesis_decorator.rb
```diff
@@ -19,8 +19,7 @@ def type
   end
 
   def date
-    # TODO: need to talk to metadata about fallback if date accepted is not present, which it isn't for legacy theses
-    object.date_accepted || object.created_at
+    object.graduation_date
```
pgwillia, line 22: This should be present so feel pretty good about using `graduation_date`
https://github.com/ualbertalib/jupiter/blob/d1c8fb1d059e8191062bbcaff1f83a0fd4fa16d8/app/models/thesis.rb#L27

pgwillia (approved):
👍

PR #2511, opened 2021-09-13
Add various fixes and improvements to batch ingestion work

murny:
This PR adds a few fixes to batch ingest work:

- Use Global ID for batch ingestion job, instead of passing in batch ingest id, and refetching inside the job
- There was a bug with using Tempfiles for the file and how they override the filename. Tempfiles take a file name and appends a random hash to it to ensure the tempfile is unique on the file system. However, we want the file to continue to have the same file name that the user provided us in the batch ingestion manifest. 

We can solve this by using Rails UploadedFile class which wraps a Tempfile to give it an "original filename" attribute. Then in our file ingestion/attachment code we can try to use the original filename attribute if its available, otherwise just use the filename from the file's path.

- Add a limit to number of files a batch ingestion can have. This is currently set at 50 files. If a user goes over this, they will be prevented to submit and be given the following error: 

image

Why 50? From stress testing with large files (anywhere from 50 MB to 800 MB....which should be rare? but should highlight the worse case scenario), we may be running into issues where we could have extremely long running jobs (in the hours...which can cause lots of problems). So we need to put some kind of limit here. A user can always submit multiple batch ingest jobs. And this is something we can revisit in the future where we can improve our workflow to allow many large files.


app/services/google_drive_client_service.rb
```diff
@@ -55,9 +55,16 @@ def download_spreadsheet(spreadsheet_id)
...
+    base_name = File.basename(file_name, '.*')
+    extension_name = File.extname(file_name)
+    file = service.get_file(file_id, download_dest: Tempfile.new([base_name, extension_name], binmode: true))
     file.rewind
-    file
+
+    # When you create a Tempfile object, it will create a temporary file with a unique filename.
+    # However we want to keep the original name that the user provide to us.
+    # To do this, we can create an UploadedFile with the original name and the tempfile
+    ActionDispatch::Http::UploadedFile.new(filename: file_name, tempfile: file)
```
pgwillia, line 67: This is pretty clever and gets us what we want without any fuss or mess.  But [this](https://guides.rubyonrails.org/form_helpers.html#what-gets-uploaded) isn't exactly the designed intention of the class.  Do we incur any risk with this approach?

test/models/batch_ingest_test.rb
```diff
@@ -22,6 +22,18 @@ class BatchIngestTest < ActiveSupport::TestCase
...
+    51.times do |i|
+      @batch_ingest.batch_ingest_files.new(google_file_id: "file id #{i}", google_file_name: "file name #{i}")
+    end
+
+    assert_not @batch_ingest.valid?
+    assert_equal(
+      'too many files added (maximum is 50 files). If you need more files, try submitting multiple batch ingests',
+      @batch_ingest.errors[:batch_ingest_files].first
+    )
+  end
```
pgwillia, line 35: Thinking out loud about making this easier to test.  Maybe we could use a class instance variable
```
# app/models/batch_ingest.rb 
class <<self
  attr_accessor :maximum_files
end

validates :batch_ingest_files, presence: true, length: { maximum: maximum_files }
```
Then for this test
```suggestion
  test 'invalid when files exceed maximum length' do
    BatchIngest.maximum_files = 0

    assert_not @batch_ingest.valid?
    assert_equal(
      'too many files added (maximum is 50 files). If you need more files, try submitting multiple batch ingests',
      @batch_ingest.errors[:batch_ingest_files].first
    )
    BatchIngest.maximum_files = 50

  end
```

pgwillia (dismissed):
I think this is great.  I think it's fine as is but have one suggestion that you can take or leave.

test/models/batch_ingest_test.rb
murny: Not a big fan of changing code just for the sake to make tests happy like this. But honestly not sure a better way 🤔 

On the bright side all this record creation is done in memory and is very fast (nothing in this test is hitting the DB (other then the original fixture))

So because of this, this test is just as fast as any of the other tests in this file which is pretty surprising honestly when we looping 50+ times and creating objects in memory, etc.


Total runtime for this entire file is around ~2 secs:

```
➜  rails t test/models/batch_ingest_test.rb
Running via Spring preloader in process 33413
Run options: --seed 38351

# Running:

..........

Finished in 1.413805s, 7.0731 runs/s, 14.8535 assertions/s.
10 runs, 21 assertions, 0 failures, 0 errors, 0 skips
```

Runtime for this offending test: 

```
➜  rails t test/models/batch_ingest_test.rb:25
Running via Spring preloader in process 33562
Run options: --seed 63477

# Running:

.

Finished in 0.930582s, 1.0746 runs/s, 2.1492 assertions/s.
1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

Runtime for one of the other tests like checking presence validation on title:

```
➜  rails t test/models/batch_ingest_test.rb:49
Running via Spring preloader in process 34039
Run options: --seed 31949

# Running:

.

Finished in 0.931902s, 1.0731 runs/s, 2.1461 assertions/s.
1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

So think this is fine to just leave this code alone considering its just as fast as any of the other tests here and entire file takes a few seconds to run.

app/services/google_drive_client_service.rb
murny: I think we be okay here since although this `UploadedFile` is meant for form submissions in the request/response lifecycle for a file input, we essentially doing the same thing...just outside of the request/response lifecycle. And I can't imagine this `UploadedFile` ever drastically changing in the future since it does a very simple thing of just wrapping Tempfile with a few extra getter/setters: https://api.rubyonrails.org/v6.1.4/classes/ActionDispatch/Http/UploadedFile.html

On a file input submission, rails will stream the data into a Tempfile, then wrap that Tempfile with this UploadedFile so it can preserve the "original filename" the user gave in the file input. Once the request ends, it is thrown away.

Here we basically want to achieve the same thing just slightly different. A user will give us a file (via a file id which references a file inside google drive), we download it and stream it into a Tempfile, we then wrap this Tempfile with this UploadedFile class so we can continue to preserve the "original filename" the user gave us in the file input. Then once we done with this object, it gets garbage collected and thrown away.

So pretty much the same kinda workflow, therefore think the risks are very minor. Alternatively we can just create our own  class or OpenStruct to do something similar 🤔

pgwillia: That makes sense.  I can see that the risk here is pretty low.  It also looks pretty stable from [it's history](https://github.com/rails/rails/commits/32b2942946edba6a288c6b992109598eeb626964/actionpack/lib/action_dispatch/http/upload.rb).

test/models/batch_ingest_test.rb
pgwillia: Thanks for the analysis of test runtime.  I can see that it's not really a concern.

pgwillia (approved):

PR #2523, opened 2021-09-22
Keep rdf annotations in git history

pgwillia:
## Context

Then we can load them once/in one place for tests or use them to db:setup an environment from scratch.

This also replaces some redundant test setup that could easily get out of date.

Related to https://github.com/ualbertalib/acts_as_rdfable/issues 
Depends on https://github.com/ualbertalib/acts_as_rdfable/pull/13

## What's New

A change to `acts_as_rdfable` creates `db/seeds/rdf_annotations.rb` from the migrations.  Load these in seeds and tests.  Any future changes to `db/seeds/rdf_annotations.rb` should also get checked in.


Gemfile
```diff
@@ -17,7 +17,7 @@ gem 'active_link_to'
 gem 'simple_form'
 
 # RDF stuff
-gem 'acts_as_rdfable', github: 'ualbertalib/acts_as_rdfable', tag: '0.3.0'
+gem 'acts_as_rdfable', github: 'ualbertalib/acts_as_rdfable', branch: 'dump_rdf_annotations'
```
murny, line 20: I assume this just temporary? Will cut a new version I assume?

murny (approved):
Awesome cleanup! Looks great 👍

Gemfile
pgwillia: Yeah, exactly.  I'll wait for the acts_as_rdfable review and then update this.

pgwillia: Updated with the latest tagged release!

murny (approved):
👏

Issue #2527, opened 2021-09-23
Jupiter is not communicating with pushmi_pullyu


lagoan:
**Describe the bug**
When an item or thesis is saved, a new entry is added to the Redis queue but pushmi_pullyu crashes when trying to read it. Errors on Rollbar are not helpful at this time. This happens because jupiter is not entering all the information that pushmi_pullyu needs.

**To Reproduce**
1. Save an item or thesis
2. pushmi_pullyu will output an error an not pull the information from jupiter.

**Expected behavior**
When saving an item or thesis on jupiter an entry should be added to redis with the correct format.

PR #2528, opened 2021-09-23
Fix/communication pmpy

lagoan:
## Context

pushmi_pullyu has not been communicating with jupiter because the entries on the Redis queue did not have the proper format. This PR simply adds the required information from jupiter to Redis.

Related to #2527

## What's New

The json values for Redis entered when a new item or thesis is saved now include the id and the type of entity that was saved.


murny (approved):
LGTM 👍 . Nice investigation into the problem

app/models/jupiter_core/depositable.rb
```diff
@@ -159,19 +159,21 @@ def add_to_embargo_history
...
-  def push_item_id_for_preservation
+
+  def push_entity_for_preservation
     queue_name = Rails.application.secrets.preservation_queue_name
 
     $queue ||= ConnectionPool.new(size: 1, timeout: 5) { Redis.current }
 
     $queue.with do |connection|
-      connection.zadd queue_name, Time.now.to_f, id
+      entity = { uuid: id, type: self.class.table_name }
```
pgwillia, line 169: Should we add a comment that this needs to match [pushmi_pullyu](https://github.com/ualbertalib/pushmi_pullyu)?

pgwillia (approved):
Thanks again for your investigation of this.

pgwillia (approved):
👍

Issue #2079, opened 2021-01-11
Upgrade Rails to version 6.1.x


murny:
Upgrade should be pretty straight forward. Here is the release notes and upgrade guide:

- https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1
- https://edgeguides.rubyonrails.org/6_1_release_notes.html

However we should go back and use rails stuff instead of our homegrown code. As an example we wrote our own permanent URLs for activestorage blobs code a long time ago. Now this is included in Rails for free, so we should make the switch to use their version.

Also there may be other nice features included in this release we could look into. One example is lazy loaded images: https://github.com/rails/rails/pull/38452

PR #2552, opened 2021-10-12
Upgrade rails to v6.1.X

murny:
https://github.com/ualbertalib/jupiter/issues/2079

Upgrade was pretty minor aside from changes in activestorage.

Activestorage added two new db migrations, with some heavy refactoring of it's internal code  (which we were overriding in certain places). ActiveStorage now also has the ability to create `public` urls (urls that do no expire). I've refactor our code to take advantage of this and work with the new changes.

QAing new changes of activestorage:
Thumbnails still being served correctly:

image
image

Able still download/view files correctly:
image

Tested if thumbnail urls are correctly being permanent (private urls expire in 5 minutes, this isn't happening)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -97,7 +97,7 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'pry-rails'
 
-  gem 'rubocop', '~> 1.22.2', require: false
+  gem 'rubocop', '~> 1.22.3', require: false
   gem 'rubocop-minitest', require: false
   gem 'rubocop-performance', require: false
   gem 'rubocop-rails', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -147,7 +147,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
-    childprocess (3.0.0)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -1,16 +1,18 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
+ruby '2.7.2'
```
murny, line 4: Default rails generator adds this now. Think its smart to lock our ruby version for the app down but I need to make this more flexible for us since we need to support > v2.6 < v2.8

app/controllers/downloads_controller.rb
```diff
@@ -1,13 +1,5 @@
 class DownloadsController < ApplicationController
 
-  include ActiveStorage::SetBlob
```
murny, line 3: This was the biggest trouble with the rails upgrade. Think we just copied this stuff over?

This `include` sets a `@blob` instance variable via looking up the `params[:signed_id]` in the blob table. Here I don't think we ever used this ever. And now in 6.1 is no longer returns a `nil` object but will throw an exception rendering a 404 error page. 

Same with the below `before_action` 🤔 ? 

Automated tests are passing now. But need to do some through testing here to make sure everything is still working correctly.

PR #2558, opened 2021-10-13
bring_back_invalid_date_filtering_flash

ConnorSheremeta:
Bring back the flash message if the filtering dates are invalid (like when the begin date is after the end date.)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -129,7 +129,7 @@ GEM
     bindex (0.8.1)
     binding_of_caller (1.0.0)
       debug_inspector (>= 0.0.1)
-    bootsnap (1.9.3)
+    bootsnap (1.9.4)
       msgpack (~> 1.0)
     brakeman (5.2.0)
     brpoplpush-redis_script (0.1.2)


test/integration/search_invalid_date_range_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_helper'
+
+class SearchInvalidDateRange < ActionDispatch::IntegrationTest
+
+  test 'search with invalid date range shows alert' do
```
murny, line 5: Two suggestions here. Do we need an integration test to test this? Or can we use a controller test? As it's not really doing an "Integration" (in the sense of testing multiple things or a workflow)

Regardless of if using an integration test or not,  we should group this test with another search integration test, instead of having a different file for each test

app/controllers/search_controller.rb
```diff
@@ -24,6 +24,7 @@ def index
     )
     @results = search_query_index.results
     @search_models = search_query_index.search_models
+    flash.now[:alert] = t('search.invalid_date_range_flash') if search_query_index.invalid_date_range
```
pgwillia, line 27: This used to be set
https://github.com/ualbertalib/jupiter/blob/fb8cca3af9f3016d16581a4f4e95ca1fd5cd7e1b/app/services/user_search_service.rb#L63-L70
Before the errors there were fixed.

What is the advantage of moving where we set the error to the controller?

pgwillia, line 27: Good thinking moving this to the locales!

ConnorSheremeta: When calling flash[:alert] inside of the service an error was thrown. Not sure if this is from something recent or if it has always been this way but it's fixed now.

pgwillia (approved):

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ Further discussion of the context can be found at [#2119](https://github.com/ual
 - Brakeman linting to Github Actions workflow
 - Make autocomplete explicit [PR#2449](https://github.com/ualbertalib/jupiter/pull/2449)
 - RdfAnnotation changes will be output to a file to facilitate testing and database setup [acts_as_rdfable#12](https://github.com/ualbertalib/acts_as_rdfable/issues/12)
+- Bring back illogical date range faceting flash message [#2030](https://github.com/ualbertalib/jupiter/issues/2030)
```
pgwillia, line 76: This should be up in the unreleased section.

pgwillia (changes requested):
LGTM!  Just a nit about the Changelog

pgwillia (approved):

PR #2563, opened 2021-10-15
Cds/reduce assertions per test

ConnorSheremeta:
Some of them could not be reduced as it would make the test less comprehensive.


test/services/statistics_test.rb
```diff
@@ -10,61 +15,54 @@ class StatisticsTest < ActiveSupport::TestCase
...
-  # TODO: our tests are quite smelly.  This one needs work!
-  test 'counts increment correctly' do
-    obj_id = generate_random_string
-    test_ip = '192.168.0.1'
-
+  test 'view counts increment correctly' do
     freeze_time do
-      Statistics.increment_view_count_for(item_id: obj_id, ip: test_ip)
-      assert_equal 1, Statistics.views_for(item_id: obj_id)
-      assert_equal [1, 0], Statistics.for(item_id: obj_id)
```
pgwillia, line 22: Looks like we're not exercising `Statistics.for` anymore?

test/services/statistics_test.rb
```diff
@@ -10,61 +15,54 @@ class StatisticsTest < ActiveSupport::TestCase
     assert_equal [0, 0], Statistics.for(item_id: obj_id)
   end
 
-  # rubocop:disable Minitest/MultipleAssertions
-  # TODO: our tests are quite smelly.  This one needs work!
-  test 'counts increment correctly' do
-    obj_id = generate_random_string
-    test_ip = '192.168.0.1'
-
+  test 'view counts increment correctly' do
```
pgwillia, line 18: I like how you broke this up into view and download.  What do you think about breaking each of these up further?
- testing two increments in the same time period
- testing after expiring the key

test/system/admin_communities_index_test.rb
```diff
@@ -10,23 +10,27 @@ class AdminCommunitiesIndexTest < ApplicationSystemTestCase
...
-  # TODO: add more tests
-  test 'should be able to expand the collection for a community in the list' do
     admin = users(:user_admin)
 
     login_user(admin)
 
     click_link admin.name # opens user dropdown which has the admin link
     click_link I18n.t('application.navbar.links.admin')
     click_link I18n.t('admin.communities.index.header')
+  end
```
pgwillia, line 21: Smart way to break this up!  This is [one of the problem tests](https://github.com/ualbertalib/jupiter/issues/2129#issuecomment-905744278).  Perhaps this will reduce it's flakiness!

test/system/collections_pagination_and_sort_test.rb
```diff
@@ -87,17 +77,34 @@ class CollectionsPaginationAndSortTest < ApplicationSystemTestCase
...
-    click_button 'Date (newest first)'
+    click_button 'Sort by'
     click_link 'Date (oldest first)'
     assert_equal URI.parse(current_url).request_uri, community_path(@community, sort: 'record_created_at',
                                                                                 direction: 'asc')
     assert_selector 'button', text: 'Date (oldest first)'
     assert_selector 'div', text: '1 - 10 of 11'
     assert_selector 'ul.list-group li:first-child a', text: 'Fancy Collection 00'
     assert_selector 'ul.list-group li:nth-child(2) a', text: 'Nice Collection 01'
-    assert_selector 'ul.list-group li:nth-child(9) a', text: 'Fancy Collection 08'
```
pgwillia, line 100: Did something happen to the second last item?

test/system/collections_pagination_and_sort_test.rb
```diff
@@ -55,29 +55,19 @@ class CollectionsPaginationAndSortTest < ApplicationSystemTestCase
...
-    click_button 'Title (A-Z)'
+    click_button 'Sort by'
     click_link 'Date (newest first)'
     assert_equal URI.parse(current_url).request_uri, community_path(@community, sort: 'record_created_at',
                                                                                 direction: 'desc')
     assert_selector 'button', text: 'Date (newest first)'
     assert_selector 'div', text: '1 - 10 of 11'
     assert_selector 'ul.list-group li:first-child a', text: 'Fancy Collection 10'
     assert_selector 'ul.list-group li:nth-child(2) a', text: 'Nice Collection 09'
-    assert_selector 'ul.list-group li:nth-child(9) a', text: 'Fancy Collection 02'
```
pgwillia, line 80: Why not test this?

test/system/collections_pagination_and_sort_test.rb
```diff
@@ -30,8 +28,10 @@ class CollectionsPaginationAndSortTest < ApplicationSystemTestCase
     assert_equal URI.parse(current_url).request_uri, community_path(@community, page: '2')
     assert_selector 'div', text: '11 - 11 of 11'
     assert_selector 'ul.list-group li:first-child a', text: 'Nice Collection 09'
+  end
```
pgwillia, line 31: Yass!  This is so much better in organization, description and size of tests.

test/system/search_test.rb
```diff
@@ -109,9 +109,7 @@ class SearchTest < ApplicationSystemTestCase
     end
   end
 
-  # rubocop:disable Minitest/MultipleAssertions
-  # TODO: our tests are quite smelly.  This one needs work!
-  test 'anybody should be able to filter the public items' do
+  test 'anybody should be able to view search results' do
```
pgwillia, line 112: This is great.  Another big improvement.  Hopefully will help with the slow tests too.

pgwillia (changes requested):
This is good stuff!  I have a concern about further relaxing the Max Minitest/MultipleAssertions, a nit about the Changelog, some concerns about assertions that seem to be missing, and some suggestions.

.rubocop.yml
```diff
@@ -73,9 +73,8 @@ Metrics/ParameterLists:
     - app/models/jupiter_core/search.rb
     - app/models/jupiter_core/solr_services/deferred_faceted_solr_query.rb
 
-# TODO: our tests are quite smelly.  We should be working to get this lower!
 Minitest/MultipleAssertions:
-  Max: 15 # default 3
+  Max: 35 # default 3
```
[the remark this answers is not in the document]
murny: Think 15 is a good default. If your over 15 expectations then you should be asking yourself why and if you can do this better.

But we shouldn't strive to get this down to 10 or 5 or.....the default of 3. There could easily be valid reasons to write tests with 10 expectations. A good integration test testing a complicated workflow would easily get close to 10 expectations and this is probably more then fine.

Think the key is keeping code readable and maintainable for the right reasons and not just cause an arbitrary linter rule is telling you to break your code into smaller chunks (which could make it far less readable or maintainable) for no real reason other then we have to keep it under a magic number. (It's the same reason why we disabled MethodLength, ClassLength, BlockLength, etc for similar reasons)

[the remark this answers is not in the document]
pgwillia: :100:  Good default, readable/maintainable code and testing the right things.

pgwillia (changes requested):
This is looking good! I have a question about the job test directory.

pgwillia (approved):

Issue #2566, opened 2021-10-20
Preserving the backlog of items created in the last year


pgwillia:
- [x] @pgwillia is going to investigate if there is a queue for pmpy ready to go or how the back log will be resolved.

Kenton was asking if once we release our pushmi_pullyu fix if the workflow will just work again or if there is anything we have to do to get it to preserve the backlog?  I'm going to take a look at what is currently in the production queue, and if the format is correct.  I'll also check how we might kick off proper jobs for anything that hasn't been preserved if that turns out to be the case.

_Originally posted by @pgwillia in https://github.com/ualbertalib/jupiter/issues/2524#issuecomment-945970088_

pgwillia:
The queue will have just the ids on it
![image](https://user-images.githubusercontent.com/1220762/138167913-c9c3bcab-a6a1-4c3b-83d3-ea4fee64d87f.png)
Which is the cause for our issues with pushmi_pullyu
```
$ pushmi_pullyu
Loading PushmiPullyu 2.0.1
Running in ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux]
Starting processing, hit Ctrl-C to stop
Traceback (most recent call last):
        8: from /home/pjenkins/.asdf/installs/ruby/2.7.4/bin/pushmi_pullyu:23:in `<main>'
        7: from /home/pjenkins/.asdf/installs/ruby/2.7.4/bin/pushmi_pullyu:23:in `load'
        6: from /home/pjenkins/.asdf/installs/ruby/2.7.4/lib/ruby/gems/2.7.0/gems/pushmi_pullyu-2.0.1/exe/pushmi_pullyu:7:in `<top (required)>'
        5: from /home/pjenkins/.asdf/installs/ruby/2.7.4/lib/ruby/gems/2.7.0/gems/pushmi_pullyu-2.0.1/lib/pushmi_pullyu/cli.rb:38:in `run'
        4: from /home/pjenkins/.asdf/installs/ruby/2.7.4/lib/ruby/gems/2.7.0/gems/pushmi_pullyu-2.0.1/lib/pushmi_pullyu/cli.rb:54:in `start_server'
        3: from /home/pjenkins/.asdf/installs/ruby/2.7.4/lib/ruby/gems/2.7.0/gems/pushmi_pullyu-2.0.1/lib/pushmi_pullyu/cli.rb:214:in `run_tick_loop'
[...]

PR #2567, opened 2021-10-20
Tasks for Jupiter Preservation

pgwillia:
## Context

Kenton was asking if once we release our pushmi_pullyu fix if the workflow will just work again or if there is anything we have to do to get it to preserve the backlog? Turns out we're going to have to clear the current queue with inoperable data and try again.  These tasks will help us.

Related to #2566

## What's New

- Fixed the existing preservation task
- Added a task to clear the preservation queue

image
image

### DIFF
--- a/lib/tasks/jupiter.rake
+++ b/lib/tasks/jupiter.rake
@@ -55,16 +55,33 @@ namespace :jupiter do
   end
 
   desc 'queue all items and theses in the system for preservation'
-  task :preserve_all_items_and_theses, [:batch_size] => :environment do |_, args|
+  task :preserve_all_items_and_theses, [:before_date, :batch_size] => :environment do |_, args|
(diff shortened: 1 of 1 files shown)


lib/tasks/jupiter.rake
```diff
@@ -56,13 +56,33 @@ namespace :jupiter do
 
   desc 'queue all items and theses in the system for preservation'
   task :preserve_all_items_and_theses, [:batch_size] => :environment do |_, args|
-    desired_batch_size = args.batch_size.to_i ||= 1000
+    desired_batch_size = if args.batch_size.present?
```
murny, line 59: Haha nice catch (` args.batch_size.to_i` will be zero if nil/blank and we will never set it to 1000)  very subtle 👍

PR #2609, opened 2021-11-01
Fix test deprecation and driver path failure

### LINKED ISSUE #2604: Deprecation Warnings in System Tests
**Describe the bug**
```
DEPRECATION WARNING: ActionDispatch::SystemTestCase#host! is deprecated with no replacement. Set Capybara.app_host directly or rely on Capybara's default host. (called from block in <class:ApplicationSystemTestCase> at /mnt/c/Users/pgwillia/Code/ualbertalib/jupiter/test/application_system_test_case.rb:9)
2021-10-28 11:14:16 WARN Selenium [DEPRECATION] [:browser_options] :options as a parameter for driver initialization is deprecated. Use :capabilities with an Array of value capabilities/options if necessary instead.
```
**To Reproduce**
1. `rails test:system`
4. See warnings

**Expected behavior**
Our codebase is free of warnings and errors.

**Related**
https://github.com/ualbertalib/jupiter/pull/2600
https://github.com/ualbertalib/jupiter/pull/2552

pgwillia:
## Context

A recent upgrade to our application created many deprecation warnings in our application.  It also changed the behaviour of `driver_path`

Closes #2604

## What's New

- fix `host!` deprecation
- make `driver_path` a `Proc` so that it can be called

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Chores
 - bump sidekiq-unique-jobs from 7.0.12 to 7.1.8 and fix a long missed deprecation
+- fixed deprecation warning on tests [#2604](https://github.com/ualbertalib/jupiter/issues/2604)
 
 ## [2.2.0] - 2021-10-21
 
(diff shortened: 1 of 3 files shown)


murny (approved):
Awesome! Thanks for investigating and fixing these 🎉

Issue #2610, opened 2021-11-01
Docker Demo fails to startup: bad URI error


jefferya:
**Describe the bug**
Following the Demo instructions with Docker `Docker version 20.10.10, build b485636` and docker-compose `version 1.29.2` startup of the `web` container fails with `bad URI(is not URI?): "'redis://redis/1'" (URI::InvalidURIError)`

**To Reproduce**
1. Install latest Docker: version 20.10.10
2. Install latest docker-compose: version 1.29.2
3. clone repo 5c8bfce (head of the main branch)
4. `docker-compose -f docker-compose.yml up -d` fails to start `web` container with error  `bad URI(is not URI?): "'redis://redis/1'" (URI::InvalidURIError)`

**Expected behavior**
`web` container starts normally with connection to redis

**Desktop (please complete the following information):**
Ubuntu 20.04

**Additional context**

https://github.com/ualbertalib/jupiter/blob/5c8bfce347e951bb99c0026c21ec87a946b3f1d1/docker-compose.demo.yml#L37

jefferya:
New versions of docker-compose, if I understand some of the reports, became subtly more strict when interpreting yaml.

murny:
Believe this can be closed now, should be fixed by #2611

PR #2611, opened 2021-11-01
Fixes: Docker demo Redis bad URI error #2610

jefferya:
## Context

Docker demo fails to start web container due to error: Redis bad URI: #2610 

## What's New

Update the docker-compose file for the demo


murny (approved):
Good catch! 👏

PR #2612, opened 2021-11-01
Task to run reports for digitization batch ingest

pgwillia:
## Context

When doing batch ingestion from tasks we will need feedback about what went well and the cause of any failures.

image
image
image

Related to #2011 

## What's New

- Task to run reports for digitization batch ingest
- Fix a situation with error handling 
- Add a helper method to quickly display the peel number (peel_id, run, part_number) and custom error message for the validation.


app/jobs/digitization/batch_metadata_ingestion_job.rb
```diff
@@ -53,7 +53,7 @@ def create_items_from_graph(graph, batch_ingest)
 
     graph.query(query_for_local_identifiers) do |statement|
       # Each of the discrete items will become its own object
-      book = batch_ingest.books.new(owner_id: batch_ingest.user_id)
+      book = Digitization::Book.new(owner_id: batch_ingest.user_id)
```
murny, line 56: Think we can just do something similar as before here and save an extra call to the DB below (`batch_ingest.books << book` does DB call) by just setting the batch_ingest id that we want?

```suggestion
      book = Digitization::Book.new(owner_id: batch_ingest.user_id, digitization_batch_metadata_ingest_id: batch_ingest.id)
```

app/models/digitization/book.rb
```diff
@@ -16,7 +16,13 @@ class Digitization::Book < JupiterCore::Depositable
 
   has_paper_trail
 
-  validates :peel_id, uniqueness: { scope: [:run, :part_number] }, presence: true, if: :part_number?
+  validates :peel_id,
+            uniqueness: {
+              scope: [:run, :part_number],
+              message: lambda do |object, _data|
+                "number #{object.peel_number} has already been taken"
```
murny, line 23: Should maybe be i18n if we ever intend to show this to end users (even admin users) in ERA?

lib/tasks/digitization_batch_ingest.rake
```diff
@@ -32,6 +32,26 @@ namespace :digitization do
     log 'FINISH: Batch ingest enqueued!'
   end
 
+  desc 'report on batch_ingest progress - used by Admin and Assistants'
+  task :batch_ingest_report, [:id, :limit] => :environment do |_t, args|
+    limit = args.limit.presence || 10
+    if args.id.blank?
+      puts "id,\ttitle,\tstatus,\tsize"
```
murny, line 39: Tab chars (`\t`) I don't think are consistent? They depend wildly on the Operating system/Command line/etc and can be anything really (but typically they range from 2 to 4 to 8 spaces wide)? Is this okay? If not we should be just using spaces instead 😄

murny (dismissed):
Looks pretty good. Just some minor comments for potential improvements

lib/tasks/digitization_batch_ingest.rake
```diff
@@ -32,6 +32,27 @@ namespace :digitization do
     log 'FINISH: Batch ingest enqueued!'
   end
 
+  desc 'report on batch_ingest progress - used by Admin and Assistants'
+  task :batch_ingest_report, [:id, :limit] => :environment do |_t, args|
+    limit = args.limit.presence || 10
+    if args.id.blank?
+      puts format '%38s,%80s,%12s,%7s', 'id', 'title', 'status', 'size'
```
murny, line 39: Maybe a comment here to describe this (As I would have no idea what this is doing and sure other future devs would feel the same)? Looks like you added good detail in your commit comment so could maybe just take some of this and put that here.

murny (dismissed):

lagoan (approved):
Good work @pgwillia !

jefferya (approved):
Looks like others reviewed before merge... only took a cursory look. Looks good.

PR #2554, opened 2021-10-12
Integration with datacite-client

pgwillia:
## Context

The EZID Compatibility API is sunsetting at the end of this year, per https://blog.datacite.org/sunsetting-of-the-ez-api/

We'll need to switch over to their normal API.  I did consider https://rubygems.org/gems/datacite but there were some oddities about their API.  I wrote a wrapper of the API which would also appreciate a review.

Related to #2268

## What's New

Updated our DOIService to use the [datacite-client gem](https://github.com/pgwillia/datacite-client/).  This is protected by a feature flag so  that it can be turned on/off from the admin panel.

Exercised and sanitized the tests.

Issue #2268, opened 2021-04-01
Replace DataCite EZID API Calls W/ DataCite REST API


mbarnett:
The EZID Compatibility API is sunsetting at the end of this year, per https://blog.datacite.org/sunsetting-of-the-ez-api/

We'll need to switch over to their normal API. A cursory glance at GitHub didn't turn out any useful gems (the only thing I saw was one project that hadn't seen an update in over 6 years), so we may need to wrap this ourselves.

pgwillia:
https://rubygems.org/gems/datacite seems to do most of what we need at this point.
```
client = Datacite::Client.new(username: "foo",
                              password: "bar",
                              host: "api.test.datacite.org")
```

https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/services/doi_service.rb#L27-L49
Looks like this will be two steps to 1. create and 2. add [attributes](https://support.datacite.org/docs/api-create-dois#create-a-findable-doi)
```
result = client.autogenerate_doi(prefix: '10.0001')
attributes = {
      "event": "publish",
      "doi": "10.5438/0012",
      "creators": [{
        "name": "DataCite Metadata Working Group"
      }],
      "titles": [{
        "title": "DataCite Metadata Schema Documentation for the Publication and Citation of Research Data v4.0"
      }],
      "publisher": "DataCite e.V.",
      "publicationYear": 2016,
      "types": {
        "resourceTypeGeneral": "Text"
      }
result = client.update(id: '10.0001/bc123df4567', attributes: attributes)
```

https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/services/doi_service.rb#L51-L76
```
attributes = {
      "event": "publish",
      "doi": "10.5438/0012",
      "creators": [{
        "name": "DataCite Metadata Working Group"
      }],
      "titles": [{
        "title": "DataCite Metadata Schema Documentation for the Publication and Citation of Research Data v4.0"
      }],
      "publisher": "DataCite e.V.",
      "publicationYear": 2016,
      "types": {
        "resourceTypeGeneral": "Text"
      }
result = client.update(id: '10.0001/bc123df4567', attributes: attributes)
```
https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/services/doi_service.rb#L78-L80
[...]

PR #2615, opened 2021-11-05
Preserve the `doi:` namespace 

pgwillia:
## Context

All our existing doi's have a `doi:` prefix.  This is the output we use for identifiers in OAI.

Rather than make a change that has unknown repercussions we will continue the same DOI structure.

I investigated if the Datacite API recognized DOIs in the format `doi:<prefix>/<id>`.  It does not.

I investigated if the Datacite API recognized our existing DOIs in the `<prefix>/<id>` format and it does. We can be assured that this change will ensure continuity of the existing data.
Related to #2554 and #2268

## What's New

- Item and Thesis doi retains the `doi:` prefix and the DOIService will strip the prefix before making requests to Datacite.

### DIFF
--- a/test/fixtures/items.yml
+++ b/test/fixtures/items.yml
@@ -35,7 +35,7 @@ item_practical:
 item_admin:
   id: "e2ec88e3-3266-4e95-8575-8b04fac2a679"
   owner: user_admin
-  doi: "10.80243/99dh-v584"
+  doi: "doi:10.80243/99dh-v584"
   visibility: <%= JupiterCore::VISIBILITY_PUBLIC %>
   creators: ["dc:creator1$ Doe, Jane", "dc:creator2$ Doe, John"]
   contributors:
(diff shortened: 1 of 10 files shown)


test/fixtures/items.yml
```diff
@@ -35,7 +35,7 @@ item_practical:
 item_admin:
   id: "e2ec88e3-3266-4e95-8575-8b04fac2a679"
   owner: user_admin
-  doi: "10.80243/99dh-v584"
+  doi: "doi:10.80243/99dh-v584"
```
pgwillia, line 38: This and the changes to seeds.rb and *.n3 files is reverting a decision I made in https://github.com/ualbertalib/jupiter/pull/2554/commits/f5278ead77395d42fb3b8648b3ceef8c7f9b4bf0

test/services/datacite_doi_service_test.rb
```diff
@@ -4,8 +4,7 @@ class DataciteDoiServiceTest < ActiveSupport::TestCase
 
   include ActiveJob::TestHelper
 
-  # If you need to re-record the vcr cassettes uncomment this line so you get them in order
-  # ActiveSupport::TestCase.test_order = :sorted
+  # If you need to re-record the vcr cassettes use SEED=20 to record the tests in order
```
pgwillia, line 7: Found a better way to run the tests in order. Now we can use more natural language to describe the tests.

murny (approved):
Looks great. Nice work 👍

PR #2617, opened 2021-11-10
Change `byte_size` in `ActiveStorage::Blob`

pgwillia:
## Context

We had a request by a researcher to attach several 2.8 Gb zip files to an existing object.  We couldn't fulfill this request because of the way we were storing metadata about the file.  Using `Integer` put an artificial limitation of 2,147,483,647 (2^32-1) on the size of files we could attach.  We migrate the `byte_size` of blobs to use `BigInt`, 9,223,372,036,854,775,807 (2^64-1), instead.

Looks like rails uses [bigint](https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/actionmailbox/test/dummy/db/migrate/20180212164506_create_active_storage_tables.active_storage.rb#L10) now.  They [changed it](https://github.com/rails/rails/commit/9de6c6c91c715b2c7b4a566b5d350e25936bdc78#diff-a99d3a53b13fcab0bb710064be91c98c5929b3ecc8a000d6b9af47ec74a4c6c4) back in 2018.

Related to https://github.com/ualbertalib/digital-preservation/issues/22

## What's New

- migration to convert the `byte_size` column to `bigint`


murny (dismissed):
Looks good. 

This kind of migration will cause a "lock" on the table, which could prevent any reads/writes for a few seconds. Think we are okay since we deploy after hours and the number of users we have is low but it's something to be aware of.

So makes me curious if we should bring in this `strong_migrations` gem for us to write better, safer migrations. For example for this kinda change it says to do this to make it safer: https://github.com/ankane/strong_migrations#changing-the-type-of-a-column

murny (approved):

pgwillia:
@weiweishi @leahvanderjagt and @seanluyk all accepted this solution and the potential for downtime.

Issue #2621, opened 2021-11-15
Change `byte_size` in `ActiveStorage::Blob`


pgwillia:
**Describe the bug**
We had a request by a researcher to attach several 2.8 Gb zip files to an existing object. We couldn't fulfill this request because of the way we were storing metadata about the file. We try to store the byte_size of the file (2,968,164,521) in an Integer (2,147,483,647), and it doesn't fit.

**To Reproduce**
1. Go to the rails console
2. Create a variable `item` for the `Item` to attach the zip files.
3. Attempt to add and ingest one file:
```
item.add_and_ingest_files([File.open('/home/pjenkins/Downloads/ym2ym_absement-004.zip')])
  ActiveStorage::Blob Load (1.5ms)  SELECT "active_storage_blobs".* FROM "active_storage_blobs" INNER JOIN "active_storage_attachments" ON "active_storage_blobs"."id" = "active_storage_attachments"."blob_id" WHERE "active_storage_attachments"."record_id" = $1 AND "active_storage_attachments"."record_type" = $2 AND "active_storage_attachments"."name" = $3  [["record_id", "b470bdac-540b-4356-a796-eba1de754e6c"], ["record_type", "Item"], ["name", "files"]]
  TRANSACTION (0.9ms)  BEGIN
  User Load (0.7ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Community Load (0.5ms)  SELECT "communities".* FROM "communities" WHERE "communities"."id" = $1 LIMIT $2  [["id", "46972899-2299-4913-b288-745889a8ee17"], ["LIMIT", 1]]
  Collection Load (0.6ms)  SELECT "collections".* FROM "collections" WHERE "collections"."id" = $1 LIMIT $2  [["id", "1ac2a45a-d684-463a-94f1-674112489536"], ["LIMIT", 1]]
  ActiveStorage::Attachment Load (0.8ms)  SELECT "active_storage_attachments".* FROM "active_storage_attachments" WHERE "active_storage_attachments"."record_id" = $1 AND "active_storage_attachments"."record_type" = $2 AND "active_storage_attachments"."name" = $3  [["record_id", "b470bdac-540b-4356-a796-eba1de754e6c"], ["record_type", "Item"], ["name", "files"]]
  TRANSACTION (0.5ms)  ROLLBACK
[...]

PR #2628, opened 2021-11-17
Catch unsafe migrations in development

pgwillia:
## Context

When/if we move this application to the cloud we'll need to be especially careful about database migrations that cause downtime.  We're trying to move to a practice where deployments are easy and can be done multiple times a day.  During deployments we will have some traffic going to the old deployment and some to the new.  We will need to ensure that all changes are compatible with the code that is already running.

Related to #2621

## What's New

Use the `strong_migrations` gem:
  ✓  Detects potentially dangerous operations
  ✓  Prevents them from running by default
  ✓  Provides instructions on safer ways to do what you want

https://github.com/ankane/strong_migrations

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -175,3 +175,8 @@ Naming/VariableNumber:
 Style/OpenStructUse:
   Exclude: 
     - config/initializers/controlled_vocabularies.rb
+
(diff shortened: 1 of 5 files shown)


murny (approved):
LGTM 👍

PR #2638, opened 2021-11-25
Add task that will kick off jobs for artifacts

pgwillia:
## Context

Initially this was all just one big rake task. Moving the logic to a Job allows the inputs to be validated and the logic to be tested.  The job of attaching artifacts can take some time even per object so prefer to have separate jobs.

Related to #2011 #2471

## What's New

Add task that will kick off setup job for attaching artifacts to digitization objects which will kick off the job to attach artifacts per object. Accompanied by model and tests.  Addressed the feedback from the previous iteration.

image
image


app/jobs/digitization/batch_artifact_ingest_job.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    book = Digitization::Book.find_by(peel_id: arguments[1], part_number: arguments[2])
+    book.failed! if book.present?
+
+    raise exception
+  end
+
+  def perform(batch_artifact_ingest, peel_id, part_number, noid)
+    book = Digitization::Book.find_by!(peel_id: peel_id, part_number: part_number)
+    book.processing!
```
murny, line 17: Minor but could maybe save a DB call here? We do Book.save here and in the  add_link_to_preservation_storage. could just combine these

app/jobs/digitization/batch_artifacts_setup_ingestion_job.rb
```diff
@@ -0,0 +1,27 @@
+class Digitization::BatchArtifactsSetupIngestionJob < ApplicationJob
+
+  PEEL_ID_REGEX = /P0*(\d+).(\d*)/.freeze
```
murny, line 3: Minor nit: A code comment could be nice here to explain what this regex is looking for

app/jobs/digitization/batch_artifact_ingest_job.rb
```diff
@@ -0,0 +1,74 @@
+class Digitization::BatchArtifactIngestJob < ApplicationJob
+
+  queue_as :default
+
+  rescue_from(StandardError) do |exception|
+    batch_artifact_ingest = arguments.first
+    batch_artifact_ingest.update(error_message: exception.message)
+
+    book = Digitization::Book.find_by(peel_id: arguments[1], part_number: arguments[2])
+    book.failed! if book.present?
```
murny, line 10: Why in this case we only want to set `status` to failed if the book exists?

app/models/digitization/batch_artifact_ingest.rb
```diff
@@ -0,0 +1,48 @@
+class Digitization::BatchArtifactIngest < ApplicationRecord
+
+  belongs_to :user
+
+  has_many :processing_books, lambda {
```
murny, line 5: Should these just be scopes instead? Or could you give some more details with these has_many's associations that are scoped like this?

db/migrate/20210723183112_create_digitization_batch_artifact_ingests.rb
```diff
@@ -0,0 +1,14 @@
+class CreateDigitizationBatchArtifactIngests < ActiveRecord::Migration[6.0]
+  def change
+    create_table :digitization_batch_artifact_ingests, id: :uuid do |t|
+
+      t.integer :status, default: 0, null: false
+      t.string :error_message
+      t.string :archival_information_package_path
```
murny, line 7: should this be `null: false`? I see we check in the rake task, but should be guarteeded at the DB/model level as well

test/fixtures/digitization/batch_artifact_ingests.yml
```diff
@@ -0,0 +1,3 @@
+digitization_batch_ingest_with_one_item:
+  user: user_admin
+  archival_information_package_path: <%= File.join(Rails.root, "test/fixtures/files/") %>
```
murny, line 3: The below fixture files are quite large? Do they need to be this big? If they do then this is fine. 

Just asking since for example the pdf is 43MB (both of these files will completely diff if ever changed, resulting in our repo size ballooning in size. With that said, these shouldn't change often, but something to keep in mind)

app/models/digitization/batch_artifact_ingest.rb
pgwillia: I thought about this for a bit and it still doesn't feel right to put this in the Book.  The scope is only used in the ingest so I think it should probably live there.

test/fixtures/digitization/batch_artifact_ingests.yml
pgwillia: I'm torn about this.  I think there is value to using a real life example here.  I did crop the pdf to 1 page to decrease its size.

Issue #2645, opened 2021-12-01
Digitization ACN Newspaper Model


pgwillia:
The first set of digitization content will be the ACN Newspapers. This set was chosen as a representative sample of newspapers. Before we can ingest them we need to do some modelling.

@metadata-bot shared 
https://drive.google.com/drive/folders/1sFiHYzPs8r8pUUKz47CgWWd_jh3Ngd2l
and
https://docs.google.com/spreadsheets/d/1cjo9pFRauqPe7d9CtQatSImC6DzcjSJGdpiUnb24Qfg/edit

Need to add properties to the Newspapers model
- [ ]  <http://purl.org/dc/terms/type>
- [ ]  <https://www.w3.org/TR/rdf-schema/#ch_label>
- [ ]  <http://id.loc.gov/vocabulary/identifiers/local> 
- [ ]  <http://id.loc.gov/ontologies/bibframe/editionStatement> 
- [ ]  <http://pcdm.org/models#memberOf> 
- [ ]  <http://pcdm.org/models#memberOf> 
- [ ]  <http://www.europeana.eu/schemas/edm/rights> 
- [ ]  <http://purl.org/dc/elements/1.1/Coverage> 
- [ ]  <http://purl.org/dc/terms/alternative> 
- [ ]  <https://www.loc.gov/marc/relators/relaterm/pup> 
- [ ]  <http://purl.org/dc/terms/dateIssued> 
- [ ]  <http://www.rdaregistry.info/Elements/u/extent.en> 
- [ ]  <http://www.w3.org/2004/02/skos/note> 
- [ ]  <http://www.europeana.eu/schemas/edm/hasType> 
- [ ]  <http://purl.org/dc/terms/language> 
- [ ]  <http://purl.org/dc/terms/title>

We'll also need to do
- [ ] validations
- [ ] tests
- [ ] seeds (synthetic)

pgwillia:
# Tests

## Presence
- [ ] subject
- [ ] title present

## Encoding
- [ ] date issued encoded as EDTF

## Controlled Vocab

PR #2648, opened 2021-12-03
`Digitization:Newspaper` model from ACN triples

pgwillia:
## Context

The second set of digitization content will be the The advertiser and central Alberta news. This set was chosen because it is relatively small (there are 47) and well described. Before we can ingest them we need to do some modelling.

Related to #2645

## What's New

- added attributes to `Digitization::Newspaper` model and their rdf annotations
- added validations for the attributes
- added several controlled vocabularies and relevant ACN values

### DIFF
--- a/db/migrate/20211203185004_add_acn_attributes_to_digitzation_newspaper.rb
+++ b/db/migrate/20211203185004_add_acn_attributes_to_digitzation_newspaper.rb
@@ -12,7 +12,8 @@ class AddAcnAttributesToDigitzationNewspaper < ActiveRecord::Migration[6.1]
         t.text :notes, array: true
         t.string :geographic_subjects, array: true
         t.string :rights
-        t.string :volume_label
+        t.string :volume
+        t.string :issue
         t.string :editions, array: true
       end
     }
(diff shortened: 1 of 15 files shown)


db/migrate/20211203185004_add_acn_attributes_to_digitzation_newspaper.rb
```diff
@@ -0,0 +1,38 @@
...
+        t.string :languages, null: false, array: true
+        t.string :places_of_publication, array:true
+        t.string :extent
+        t.text :notes, array: true
+        t.string :geographic_subjects, array: true
+        t.string :rights
+        t.string :volume
+        t.string :issue
+        t.string :editions, array: true
+      end
```
lagoan, line 18: Just a question, how will the local identifier be added (from #2645) ?

lagoan (approved):
Everything looks good! I just have a question about the migration for table `digitization_newspapers`

db/migrate/20211203185004_add_acn_attributes_to_digitzation_newspaper.rb
pgwillia: This is an open question, not one that I've taken on yet.  I opened https://github.com/ualbertalib/jupiter/issues/2672 so that we're tracking it.

lagoan (approved):
LGTM!

Issue #2655, opened 2021-12-08
NoMethodError: undefined method `collections' for nil:NilClass


pgwillia:
Bug was introduced in https://github.com/ualbertalib/jupiter/commit/c1648f28a941b59b0dc9e9b86d073da3d83e0e42

In the js format section we missed a change from `@community` to `community`

You can trigger the error by clicking any of the Collections drop down on the Communities page
![image](https://user-images.githubusercontent.com/1220762/145244245-050fd26d-7a9e-4f66-ae04-4abbb8d4d322.png)

View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/1327/](https://rollbar.com/ualbertalib/jupiter/items/1327/)

```

NoMethodError: undefined method `collections' for nil:NilClass
  File "/var/www/sites/jupiter/app/controllers/communities_controller.rb", line 37, in block (2 levels) in show
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/action_controller/metal/mime_responds.rb", line 214, in respond_to
  File "/var/www/sites/jupiter/app/controllers/communities_controller.rb", line 30, in show
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/action_controller/metal/basic_implicit_render.rb", line 6, in send_action
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/abstract_controller/base.rb", line 195, in process_action
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/action_controller/metal/rendering.rb", line 30, in process_action
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/abstract_controller/callbacks.rb", line 42, in block in process_action
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/activesupport-6.0.4.1/lib/active_support/callbacks.rb", line 135, in run_callbacks
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/actionpack-6.0.4.1/lib/abstract_controller/callbacks.rb", line 41, in process_action
[...]

PR #2656, opened 2021-12-08
Fixed nil error when viewing Collection drop down

pgwillia:
## Context

Missed change from `@community` to `community` in a refactor.  Began in 2.2.0.

Related to https://github.com/ualbertalib/jupiter/issues/2655

## What's New
Collections drop down on Communities page is fixed
image

### DIFF
--- a/test/controllers/communities_controller_test.rb
+++ b/test/controllers/communities_controller_test.rb
@@ -16,4 +16,14 @@ class CommunitiesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test 'should show community js' do
+    get community_url(@community, format: :js), xhr: true
+    assert_response :success
+  end
+
+  test 'should show community json' do
+    get community_url(@community, format: :json)
+    assert_response :success
+  end
+
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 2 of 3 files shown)


murny (commented):
Good catch 👍 . 

I'll approve, but curious if its worth adding some test coverage for this part of the application?f this was 500ing then we must not have any 🤔? Can just write a simple controller test that given `js format` it returns a 200 etc

murny (dismissed):

pgwillia:
> Good catch +1 .
> 
> I'll approve, but curious if its worth adding some test coverage for this part of the application?f this was 500ing then we must not have any thinking? Can just write a simple controller test that given `js format` it returns a 200 etc

I was thinking the same thing, but I wasn't getting very far fast with the system test I was trying to write.  The controller test is much more elegant.  I also added one for json.

test/controllers/communities_controller_test.rb
```diff
@@ -16,4 +16,14 @@ class CommunitiesControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test 'should show community js' do
+    get community_url(@community, format: :js), xhr: true
```
pgwillia, line 20: Use the same mechanism that the browser would for an AJAX request, a XMLHttpRequest. You can do this in rails testing with xhr.  Fixed the error I was getting

`ActionController::InvalidCrossOriginRequest: Security warning: an embedded <script> tag on another site requested protected JavaScript. If you know what you're doing, go ahead and disable forgery protection on this action to permit cross-origin JavaScript embedding.`

murny (approved):
Awesome! Thanks for adding test coverage 🙏

Issue #2707, opened 2022-01-13
Theses are not being assigned DOIs


pgwillia:
There are three theses so far that have not been assigned a doi:
* https://era.library.ualberta.ca/items/c6fa8448-9907-4308-a406-b3757392c4fe
* https://era.library.ualberta.ca/items/6d93878f-c349-4db9-8224-978eff6b7a71
* https://era.library.ualberta.ca/items/8de0cd4a-90a3-4e06-96c8-513531c0497c

It looks like Thesis does not have a `description` when building the metadata hash to create the doi.  Instead Thesis has an `abstract`.

https://github.com/ualbertalib/jupiter/blob/7e1ce54bc5897fbf25bbef77e988f8f210015683/app/services/doi_service.rb#L129-L131

https://github.com/ualbertalib/jupiter/blob/7e1ce54bc5897fbf25bbef77e988f8f210015683/db/schema.rb#L411-L455

There are two ways I can think of to fix this
1. If a description is not present try the abstract
```
      descriptions: [{
        description: @item.description || @item.abstract
      }],
```
2.  Decorate the Thesis item and add a method `description` which returns the abstract

View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/1339/](https://rollbar.com/ualbertalib/jupiter/items/1339/)

```

ActiveRecord::RecordNotFound: Couldn't find Item with 'id'=6d93878f-c349-4db9-8224-978eff6b7a71
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/activerecord-6.1.4.1/lib/active_record/core.rb", line 338, in find
  File "/var/www/sites/jupiter/app/jobs/doi_create_job.rb", line 6, in perform
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/activejob-6.1.4.1/lib/active_job/execution.rb", line 48, in block in perform_now
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/activesupport-6.1.4.1/lib/active_support/callbacks.rb", line 117, in block in run_callbacks
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/i18n-1.8.11/lib/i18n.rb", line 314, in with_locale
  File "/var/www/sites/jupiter/vendor/ruby/2.6.0/gems/activejob-6.1.4.1/lib/active_job/translation.rb", line 9, in block (2 levels) in <module:Translation>
[...]

PR #2720, opened 2022-01-19
Decorate Thesis so that there is a description

pgwillia:
## Context

It looks like Thesis does not have a description when building the metadata hash to create the DOI. Instead Thesis has an abstract.

Additionally a large number of [legacy theses](https://github.com/ualbertalib/digital-preservation/issues/11) were ingested recently which were not assigned DOIs

Related to #2707

## What's New
Use the decorated version of the object that is Doiable.  This also ensures that any markdown is rendered as html.  Adds a `description` to the Thesis decorator.


murny (approved):

murny (approved):
👍

PR #2733, opened 2022-01-30
Terraform provisioning for Jupiter into Azure Cloud

murny:
TODO: will add more here


.env_deployment_sample
```diff
@@ -10,16 +9,13 @@ SAML_CERTIFICATE=
 ROLLBAR_TOKEN=
 GOOGLE_ANALYTICS_TOKEN=
 RAILS_LOG_TO_STDOUT=true
-# Comma delimited string of rack attack safelisted IPs
```
murny, line 13: Rack attack gem was removed long time ago

.env_deployment_sample
```diff
@@ -1,7 +1,6 @@
 #Rails application env variables
 RAILS_ENV=uat
 DATABASE_URL=postgresql://jupiter:mysecretpassword@postgres:5432/
-FCREPO_URL=http://fcrepo:8080/fcrepo/rest
```
murny, line 4: Fedora was removed long time ago...update all references I could find here

config/puma.rb
```diff
@@ -39,14 +39,5 @@
 #
 # preload_app!
 
-if ENV['RAILS_ENV'] == 'uat'
```
murny, line 42: No need to fork puma workers

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+    SKYLIGHT_AUTHENTICATION = "secretauthenticationtoken"
+    ACTIVE_STORAGE_SERVICE = "microsoft"
+    AZURE_STORAGE_ACCOUNT_NAME = azurerm_storage_account.blob_account.name
+    AZURE_STORAGE_ACCESS_KEY = azurerm_storage_account.blob_account.primary_access_key
+    AZURE_STORAGE_CONTAINER = azurerm_storage_container.storage_container.name
+  }
+}
+
+
+resource "kubernetes_config_map" "solr-config" {
```
murny, line 51: Solr helm charts were mostly around the idea of creating a cluster (multiple solrcloud instances and zookeeper/operator instances etc) which is probably overkill for us. Created a simple solr pod myself which mirrors the work we did in docker compose with old UAT.

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+  metadata {
+    name = "${var.app-name}-solr-config"
+    namespace = "${var.app-name}"
+  }
+
+  data = {
+    "schema.xml" = "${file("${path.module}/../solr/config/schema.xml")}"
```
murny, line 58: I think `solrconfig.xml` is required but not sure about `schema.xml`? Also we have a ton of junk in the `solr/config` and `solr` directory in Jupiter that we should delete if we not using it...as its just confusion.

In Docker compose we would mount this entire `solr/config` directory into the image, but honestly think we just need these 1 or 2 files.

In Kubernetes world, you create a ConfigMap with the file data, then mount this as a volume to do a similar thing

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+              cpu    = "250m"
+              memory = "512Mi"
+            }
+            requests = {
+              cpu    = "150m"
+              memory = "128Mi"
+            }
+          }
+        }
+        volume {
```
murny, line 113: TODO: We have no persistence layer here. So this be a good improvement where we want to look into storing the solr data in a persistent storage so things like restarting this solr pod doesn't delete all the data it currently has

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+            name = "${var.app-name}-solr-config"
+          }
+        }
+        restart_policy = "Always"
+      }
+    }
+  }
+}
+
+resource "kubernetes_service" "solr-service" {
```
murny, line 125: We need to expose solr pod to rest of the pods, so that Rails can talk to it. 

This allows traffic from Rails to Solr via `"http://jupiter-solr:8983/solr/jupiter-uat"` to work

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+    namespace = "${var.app-name}"
+  }
+
+# Rails application env variables
+  data = {
+    RAILS_ENV = "uat"
+    RAILS_LOG_TO_STDOUT = "true"
+    RAILS_SERVE_STATIC_FILES = "true"
+    DATABASE_URL = "postgresql://${urlencode("${var.postgresql-admin-login}@${azurerm_postgresql_server.db.name}")}:${urlencode(var.postgresql-admin-password)}@${azurerm_postgresql_server.db.fqdn}:5432/${azurerm_postgresql_database.postgresql-db.name}"
+    SOLR_URL = "http://${var.app-name}-solr:8983/solr/jupiter-uat"
```
murny, line 29: Just noticed this name could be better, should be driven off the solr pod name instead:
`"http://${kubernetes_deployment.solr.name}:8983/solr/jupiter-uat"`

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+    name = "${var.app-name}-ingress"
+    annotations = {
+      "kubernetes.io/ingress.class" = "nginx"
+      "nginx.ingress.kubernetes.io/proxy-body-size" = "16m"
+    }
+    namespace = "${var.app-name}"
+  }
+  spec {
+    rule {
+      # TODO: Figure out how we will use DNS/etc
```
murny, line 320: As TODO states this is left as a manual task. Where either someone has to manually create an A Record in the DNS after we provisioned UAT environment... or for self local testing you can set your `/etc/hosts` to the correct ip mapping.

But this should get automated! Terraform can do this for us in a very trivial manner.

terraform/kubernetes.tf
```diff
@@ -0,0 +1,333 @@
...
+  }
+}
+
+resource "kubernetes_ingress" "ingress" {
+  wait_for_load_balancer = true
+  metadata {
+    name = "${var.app-name}-ingress"
+    annotations = {
+      "kubernetes.io/ingress.class" = "nginx"
+      "nginx.ingress.kubernetes.io/proxy-body-size" = "16m"
```
murny, line 314: This is an example of how you can configure NGIX via annotations. For example by default you can only upload images to nginx that I believe less then 1 MB in size, this makes it 16 MBs. Obviously we want to increase this and this is how you can do this.

There is a ton of other options you can configure NGINX to your liking. Could look into adding SSL certs, having nginx serve assets instead of Rails, etc.

terraform/main.tf
```diff
@@ -0,0 +1,159 @@
...
+# Create a PostgreSQL Database
+resource "azurerm_postgresql_database" "postgresql-db" {
+  name                = "${var.app-name}_production"
+  resource_group_name = azurerm_resource_group.rg.name
+  server_name         = azurerm_postgresql_server.db.name
+  charset             = "utf8"
+  collation           = "English_United States.1252"
+}
+
+# TODO: Fix this, Currently we allow everything for now
```
murny, line 76: This and the firewall rules for Redis needs to be played with and tightened. 

At the end of the day we only want internal private traffic to flow from Rails/Sidekiq to Redis/Postgres. We ourselves have no business needing access to these resources.

Perhaps we can change 
```
  start_ip_address    = "10.0.0.0"
  end_ip_address      = "10.255.255.255"
```
For an easy fix. But this needs some investigation

terraform/main.tf
```diff
@@ -0,0 +1,159 @@
...
+
+resource "azurerm_redis_cache" "redis" {
+  name                = "${var.app-name}-cache"
+  location            = azurerm_resource_group.rg.location
+  resource_group_name = azurerm_resource_group.rg.name
+  capacity            = 2
+  family              = "C"
+  sku_name            = "Standard"
+
+  # TODO: It possible to use SSL here?
```
murny, line 93: For Postgres and Redis they provide SSL options but this is probably overkill especially if keep this traffic to just inside our pods. But something that might be worth looking into

terraform/main.tf
```diff
@@ -0,0 +1,159 @@
...
+}
+
+resource "azurerm_storage_account" "blob_account" {
+  name                     = "${var.app-name}account${random_string.random.result}"
+  resource_group_name      = azurerm_resource_group.rg.name
+  location                 = azurerm_resource_group.rg.location
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+
+  # TODO: Probably not needed for Jupiter?
```
murny, line 127: This can probably be removed. We only do file uploads from Rails server but I suppose if we ever do something with ActionText or using ActiveStorage direct upload then this might be useful

terraform/terraform.tfvars.sample
```diff
@@ -0,0 +1,24 @@
+####################
```
murny, line 1: Sample file for `terraform.tfvars` This is an easy way to set the variables in `variables.tf` for terraform without having to pass them on the command line. The variables in terraform.tfvars should be kept secret as they will contain sensitive information (DB password, etc)  and as a result are gitignored

.dockerignore
```diff
@@ -11,6 +11,12 @@ coverage/
 .ruby-version
 vendor/bundle
 
+/storage/*
+
+/public/packs
+/public/packs-test
+/node_modules
```
murny, line 18: Probably shouldn't check all this stuff into our docker images 😆. This saves a good 100 MB or so from my images I was building

Gemfile
```diff
@@ -80,6 +80,10 @@ gem 'google-api-client',
 gem 'builder_deferred_tagging', github: 'ualbertalib/builder_deferred_tagging', tag: 'v0.01'
 gem 'oaisys', github: 'ualbertalib/oaisys', tag: 'v1.0.3'
 
+group :uat do
+  gem 'azure-storage-blob', require: false
```
murny, line 84: Required for Activestorage to use Azure  blob storage as a bucket

terraform/kubernetes.tf
```diff
@@ -0,0 +1,335 @@
...
+  metadata {
+    name = "${var.app-name}-config"
+    namespace = "${var.app-name}"
+  }
+
+# Rails application env variables
+  data = {
+    RAILS_ENV = "uat"
+    RAILS_LOG_TO_STDOUT = "true"
+    RAILS_SERVE_STATIC_FILES = "true"
```
murny, line 27: Having Rails serve our assets currently. Nginx could easily do this instead. Could be nice improvement here to investigate this in the future.

terraform/kubernetes.tf
```diff
@@ -0,0 +1,335 @@
...
+          name = "${var.app-name}"
+          port {
+            container_port = 3000
+          }
+          env_from {
+            config_map_ref {
+              name = "${var.app-name}-config"
+            }
+          }
+          # TODO: Figure healthcheck out, seems to be failing on HTTP/HTTPS issue
```
murny, line 195: Not sure why this is failing. Needs more investigation.

We need to white list calls host I believe via:
`config.hosts << IPAddr.new('10.0.0.0/8')`

Apparently you could also whitelist this route instead  (but couldn't get this working locally 🤔) via
```ruby
# Exclude requests for the /healthcheck/ path from host checking
Rails.application.config.host_configuration = {
  exclude: ->(request) { request.path =~ /healthcheck/ }
}
```

More info: https://guides.rubyonrails.org/configuring.html#actiondispatch-hostauthorization

But even after this I am seeing the following error:
```
2022-01-30 20:01:43 +0000 HTTP parse error, malformed request: #<Puma::HttpParserError: Invalid HTTP format, parsing fails. Are you trying to open an SSL connection to a non-SSL Puma?
```

Apparently this can happen when the site keep redirecting http to https when attempting to hit localhost (https://github.com/puma/puma/issues/1128#issuecomment-558455489).

So I set the `scheme = "HTTPS"` so the probe should happen with HTTPS instead but still no luck.

So anyways needs some playing around to figure out how we can get this to pass without issues.

terraform/kubernetes.tf
```diff
@@ -0,0 +1,335 @@
...
+        name = "${var.app-name}-workers"
+        labels = {
+          app = "${var.app-name}-workers"
+        }
+      }
+
+      spec {
+        container {
+          name = "${var.app-name}-workers"
+          image = "murny/jupiter:latest"
```
murny, line 251: Using my own image I build instead of overwriting the ualberta stuff. So this will just need to updated to point to the correct one.

My own image does nothing special...just `docker build -f Dockerfile.deployment`  and `docker push`

Issue #2189, opened 2021-02-19
Upgrade Sidekiq to version 6.x


murny:
We currently have sidekiq pinned on v5.X. There appears to be some breaking changes and cleanup to get us on this version.

Should also make sure existing add-on gems like sidekiq-cron and sidekiq-unique-jobs are still working correctly on this new version!

Here's some more info: 
https://github.com/ualbertalib/jupiter/pull/1279

pgwillia:
- [ ] WARN: DEPRECATION WARNING: Initialization autoloaded the constants ApplicationJob, EmbargoExpiryJob, GarbageCollectBlobsJob, and RemoveInactiveDraftsJob.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload ApplicationJob, for example,
the expected changes won't be reflected in that stale Class object.

These autoloaded constants have been unloaded.

In order to autoload safely at boot time, please wrap your code in a reloader
callback this way:

    Rails.application.reloader.to_prepare do
      # Autoload classes and modules needed at boot time here.
    end

That block runs when the application boots, and every time there is a reload.
For historical reasons, it may run twice, so it has to be idempotent.

Check the "Autoloading and Reloading Constants" guide to learn more about how
Rails autoloads and reloads.
 (called from <top (required)> at /home/pjenkins/Code/ualbertalib/jupiter/config/environment.rb:5)

pgwillia:
- [ ] Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
```
redis.pipelined do
  redis.get("key")
end
```
should be replaced by
```
redis.pipelined do |pipeline|
  pipeline.get("key")
end
```

pgwillia:
- [ ] `Redis.current=` is deprecated and will be removed in 5.0. (called from: jupiter/config/initializers/redis.rb

pgwillia:
- [ ] logging changed

pgwillia:
In production

PR #2745, opened 2022-02-08
Bump sidekiq 6.4

pgwillia:
## Context

Address potential security vulnerabilities in dependency.  This upgrade has been in our backlog for awhile because of the major version bump required more careful checking.

Related to https://github.com/ualbertalib/jupiter/issues/2189

## What's New
- removed sinatra because it wasn't being used
- use the after_initialze block to avoid deprecation warning triggered by zeitwerk for these Sidekiq::Cron
- Automatic configuration of the sidekiq middleware is no longer done so made it explicit
- `Redis.current=` is deprecated and will be removed in 5.0 https://github.com/redis/redis-rb/issues/1064 so added our own accessor
- prepare for next release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Security
 - Bump Sidekiq from 5.2.9 to 6.4.1 [#2189](https://github.com/ualbertalib/jupiter/issues/2189)
+- Bump follow-redirects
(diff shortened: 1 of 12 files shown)


jefferya (dismissed):
Looks good. I read through the related documentation and as far as I can tell looks good.

One question, the `module Cache`, I'm curious as to the name. Is redis used as a cache in Jupiter?

jefferya (approved):

PR #2748, opened 2022-02-10
Update codeowners :cry:

pgwillia:
So long for now, but not goodbye.


pgwillia:
Should we add @jefferya ?

lagoan:
I think adding @jefferya  is a good idea @pgwillia

lagoan (approved):
LGTM!

PR #2798, opened 2022-03-25
Fixed file upload error

pgwillia:
## Context

https://rollbar.com/ualbertalib/jupiter/items/1365/ is the same error as I'm experiencing with File uploads in the staging environment (era.test.library.ualberta.ca).  In production it only happens to select users.  I continue to have no issues with file upload in the production environment but I get 500 errors in staging when the files_are_virus_free code runs.

When I was working on Digitization content ingestion I observed that ActiveStorage files would not be available during some callbacks for things like validation.  Basically before save.  So maybe that's at play here too, but can't account for the occassional failure in production.

## What's New

Use undocumented `attachment_changes` api to access the `ActionDispatch::Http::UploadedFile` tempfile path for virus checking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+### Fixed
+- File Upload error
+
 ## [2.3.4] - 2022-02-08
 ### Added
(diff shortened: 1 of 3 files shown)


jefferya (dismissed):
Nice catch! Great way to end a week... hope you celebrate a bit

pgwillia:
Tried this in staging

Before:
image
After:
image

So clamby not able to scan the file.  Wondering if it's permissions?

Issue #2951, opened 2022-08-22
Date created field on DraftItem creation view fails to save the defined partial dates.


lagoan:
**Describe the bug**
Creating a draft item should allow to add the descriptive metadata of 'Date created' in formats 'YYYY', 'YYYY-MM', and 'YYYY-MM-DD'. It currently adds month and day information if it is not selected from the user interface.

**To Reproduce**
1. Go to 'Deposit item' view
2. Fill in the 'Date created' descriptive information with the format as described of at least year 'YYYY', year and month 'YYYY-MM'
3. Save the item draft and continue to next step
4. Go back to 'Describe Item' tab and see that if only a year incorrect information was added when month or day was missing.

**Expected behavior**
The 'Date created' field should be able to save dates in formats 'YYYY', 'YYYY-MM', and 'YYYY-MM-DD'. This error happens because the date in the DraftItem model handles that value as a date field and it does not allow for partial dates. Formats 'YYYY' and 'YYYY-MM-DD' are handled correctly in the back end.

**Screenshots**

![image](https://user-images.githubusercontent.com/1537909/185985266-cff24a89-993e-47f9-9d55-d04cb1d2870a.png)

Date added in 'YYYY' format

![image](https://user-images.githubusercontent.com/1537909/185985422-4a23c84f-7861-483c-b47a-ed3daf56e7b6.png)

Incorrect month and day added.

![image](https://user-images.githubusercontent.com/1537909/185984723-6bfd0d04-46e8-47ad-b2b7-ac049d4b95a8.png)

The date is selected in 'YYYY-MM' format.

![image](https://user-images.githubusercontent.com/1537909/185984766-13053892-80bc-4ae1-91bd-4e0a87151b64.png)

Incorrect day '01' was added.

Th Item model handles this information as a string type. The DraftItem model could use the string type as well and we could add validation checks on the UI and back end for the required formats.

This bug is related to #2931

pgwillia:
@leahvanderjagt this bug **_does_** have an impact on any user (including administrative users) who edit an item using the user interface.

Suppose an item was created in a batch and assigned a created date of '2016'.  '2016' would be the created date
![image](https://user-images.githubusercontent.com/1220762/186277347-2ec28acd-80d2-4f42-89b6-c49c328a8426.png)

Now suppose an administrative user edits that item.  The item is converted to a DraftItem, and is given more precision (January 1, 2016)
[...]

Issue #2931, opened 2022-08-09
Changes to (created) date field:


anayram:
Format changed from MM/DD/YYYY to YYYY-MM-DD
Allow partial dates (YYYY-MM or YYYY).

PR #2952, opened 2022-08-22
[WIP] Change DraftItem date_created value to string type

lagoan:
## Context

This is an initial migration to change the date_created value type from date to string. The strong_migrations gem complained about a simple type change so I followed the suggested process:

1. Create a new column
2. Write to both columns
3. Backfill data from the old column to the new column
4. Move reads from the old column to the new column
5. Stop writing to the old column
6. Drop the old column

At this point, it would be beneficial to get input on the migration process to see if we need this approach or stick to something simpler. Any feedback is greatly appreciated.

Related to #2931 , #2951 

## What's New

Initial migration changing type from date to string.

### DIFF
--- a/app/models/draft_item.rb
+++ b/app/models/draft_item.rb
@@ -96,6 +96,8 @@ class DraftItem < ApplicationRecord
   validates :license_text_area, presence: true, if: :validate_if_license_is_text?
   validate :license_not_unselected, if: :validate_choose_license_and_visibility?
 
+  validates :date_created, :partial_date, presence: true
+
   def update_from_fedora_item(item, for_user)
     # I suspect this will become some kind of string field, but for now, using UTC
     # HACK: temporarily falling back to a parsable date when editing unparsable data to fix a crasher in Prod,
--- a/app/validators/partial_date_validator.rb
+++ b/app/validators/partial_date_validator.rb
@@ -1,0 +1,29 @@
+class PartialDateValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attribute, value)
+    # validate formats 'YYYY-MM-DD', 'YYYY-MM', and 'YYYY'
+
+    if value.blank?
+      record.errors.add(attribute, "must be format 'YYYY-MM-DD', 'YYYY-MM', or 'YYYY': '#{value}'")
+    else
+      # Note that are adding the month and day values when they are not
+      # presentwe because we use the Date.parse method for convenience.
+      # However, we do not enforce the full date value to have the format
+      # 'YYYY-MM-DD'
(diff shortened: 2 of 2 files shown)


pengyin-shan:
Hi @lagoan sorry I don't know much about Jupiter...sounds like we want to change a column type from timestamp to string? Your plans sound good to me. I was thinking in the last step, instead of dropping the column, probably just rename it to some other name in case we need it again sometime. However, I totally trust your judgment because this way could create legacy data at the same time.

Also do you need me to approve this PR ASAP? or wait for others' feedback?

lagoan:
Thanks for your comment @pengyin-shan. That is right, I need to change the column type from date to string. I am conflicted about dropping the column. If we remove it the bad side would be problems with the migration and loosing some of that data, as well as the legacy data. The good side would be a cleaner database with no noise in the data. This is the outline suggested by the developers of the strong_migrations gem but it would help a lot if I can get more eyes on it just to make sure.

At this time I am just looking for feedback, there is a lot of more work to do before it can ger properly merged

pengyin-shan:
> Thanks for your comment @pengyin-shan. That is right, I need to change the column type from date to string. I am conflicted about dropping the column. If we remove it the bad side would be problems with the migration and loosing some of that data, as well as the legacy data. The good side would be a cleaner database with no noise in the data. This is the outline suggested by the developers of the strong_migrations gem but it would help a lot if I can get more eyes on it just to make sure.
> 
> At this time I am just looking for feedback, there is a lot of more work to do before it can ger properly merged

Sounds good!👍

pgwillia:
My reading of the strong_migration suggestions is that ideally each of these steps should be in a separate deployment.  Would it work or otherwise cause problems to have one app server run the migration while another is still reading the database? My reading of this migration is that the other app server would not apply the `DraftItem.class_eval` blocks the way this is written.

It is possible that following the strong_migrations practice is overkill for us and we should remove/ignore it?

https://ualbertalib-di.slack.com/archives/CSEHP0L0K/p1636567438054800
https://ualbertalib-di.slack.com/archives/CSEHP0L0K/p1637180734087300?thread_ts=1636567438.054800&cid=CSEHP0L0K

pgwillia:
I'd recommend using [edtf-ruby](https://github.com/inukshuk/edtf-ruby) because it comprises a parser and an API implementation of the [Extended Date/Time Format standard](http://www.loc.gov/standards/datetime/).

It was introduced for Digitization content where edtf was a requirement from the metadata team.  You can see an example of a [validator](https://github.com/ualbertalib/jupiter/blob/master/app/validators/edtf_validator.rb) I wrote.

Issue #2941, opened 2022-08-16
List of template headers that have been changed in the template to match the Jupiter code:


lagoan:
- [x] List of template headers that have been changed in the template to match the Jupiter code:
  - alternative_title
  - item_type
  - owner_id - removed
  - subject
  - created
  - spatial_subjects
  - temporal_subjects
  - is_version_of
  - related_link
  - rights

Issue #2853, opened 2022-05-19
Accommodate template changes in new batch ingest workflow


ConnorSheremeta:
Changes to the new batch ingest workflow are needed in order to accommodate the new template and to have the header names used match more closely with the template’s header names. Changes to be done are listed below:

High priority
- [ ] #2931
  - Format  changed from MM/DD/YYYY to YYYY-MM-DD
  - Allow partial dates (YYYY-MM or YYYY).
- [x] #2943

Also requested
- [x] List of template headers that have been changed in the template to match the Jupiter code  (Work tracked in #2941):
  - alternative_title
  - item_type
  - owner_id - removed
  - subject
  - created
  - spatial_subjects
  - temporal_subjects
  - is_version_of
  - related_link
  - rights
  - 
- [ ] License/Rights: previously either License or Rights were required, no need to specify license_text in License. For instance, the user could just enter a rights statement and nothing in the License field. If this is too hard to change we can go with what we have right now, so this change is not a priority. Working tracked in #2942 

Related to: https://github.com/ualbertalib/jupiter/issues/1986

[Instructions to configure Google secrets](https://github.com/ualbertalib/di_internal/blob/master/Applications/Jupiter/Batch_Ingest/README.md)
[ERA Batch Ingest Templates (alfresco, item, legacy_thesis)](https://docs.google.com/spreadsheets/d/1RslH-4roDdbFER_Wa_oHB7mLTaMEmSYTfe6PcmxwG8s/edit#gid=0)

pgwillia:
From @anayram 
> If it helps to split the request, the two priority items are:
- [ ] Dates
    Changes to (created) date field:
    Format changed from MM/DD/YYYY to YYYY-MM-DD
    Allow partial dates (YYYY-MM or YYYY).
- [ ] Need support for multiple files per item in new workflow. Previously done by delimiting multiple files in the same field with a |

PR #2957, opened 2022-08-30
Change/accomodate template changes new batch ingest template headers

lagoan:
## Context

We are changing the spreadsheet headers from the new batch ingest workflow to match the variable names used in jupiter to represent the model

Related to #2853 , #2941

## What's New

+ Changes in batch ingest job to get the data from differently named columns.
+ Changes to the range of data fetched from the spreadsheet
+ Changes in validation to make sure the data is there.

The changed headers are:

| Old header | New header |
| --------------- | -----------------|
| alternate_title | alternative_title |
| type	 | item_type |
| owner_id | removed |
| subjects | subject |
| date_created | created |
| places | spatial_subjects |
| time_periods | temporal_subjects |
| citations | is_version_of |
| related_item | related_link |
| license_text | rights |


app/jobs/batch_ingestion_job.rb
```diff
@@ -46,16 +46,16 @@ def perform(batch_ingest)
   def item_ingest(batch_ingest, item_data)
     item = batch_ingest.items.new
     item.tap do |unlocked_obj|
-      unlocked_obj.owner_id = item_data['owner_id']
+      unlocked_obj.owner = User.first
```
pgwillia, line 49: Could we add a comment documenting this decision?

app/services/google_drive_client_service.rb
```diff
@@ -38,7 +38,7 @@ def download_spreadsheet(spreadsheet_id)
     service.authorization = google_credentials
     service.key = Rails.application.secrets.google_developer_key
 
-    range = "'Data'!A:X"
+    range = "'Data'!A:W"
```
pgwillia, line 41: It took me a moment to figure out why this change.  I realized it's because we removed the owner column.

pgwillia (changes requested):
Looks good!  Seems like a pretty straightforward replacement of the headers you mention.  I think we could document the change to using `User.first` as the owner for any items ingested this way.

app/services/google_drive_client_service.rb
lagoan: Yes, wanted to only fetch the data we needed as we were doing before this change.

app/jobs/batch_ingestion_job.rb
pgwillia: I can see the decision documented in the commit history.  Maybe a comment in the code isn't needed.

pgwillia (approved):

app/jobs/batch_ingestion_job.rb
lagoan: It won't hurt to have it.

pgwillia (approved):

Issue #2946, opened 2022-08-17
Ordering of files in ERA objects


seanluyk:
**Describe the bug**
Not sure if this is a bug or expected behaviour, but for multi-file objects deposited in ERA, files don't seem to be listed in an expected order (in the order they were deposited). A staff member noticed this when depositing this item: https://doi.org/10.7939/r3-n262-a586. Apparently they were depositing them in alphabetic order, but the order was not maintained once the deposit was completed

**Expected behavior**
Files deposited appear in the order the user deposited them or default if alphabetic order

seanluyk:
Thanks, I agree that option 1 is the most intuitive so that’s my preference!

PR #2963, opened 2022-09-12
Change/file order to alphabetical in item view

lagoan:
## Context

When attaching files to an item that is being created the file order is set by the time the files are finished uploading which can be misleading to the user. This change addresses this by ordering the files when shown alphabetically.

Related to #2946 

## What's New

Added a method to `Item` and `Thesis` models to get files ordered alphabetically as well as tests to make sure the end user sees this change in all views.


app/views/admin/theses/draft/review_and_deposit_thesis.html.erb
```diff
@@ -19,9 +19,9 @@
           </div>
           <div class="card-body">
             <%= render partial: 'feature_image', locals: { object: @draft } %>
-            <ul class="list-group mt-3">
-              <% @draft.files.each do |file| %>
-                <li class="list-group-item">
+            <ul class="list-group mt-3 thesis-files">
```
jefferya, line 22: `thesis-files`: I don't see the class used elsewhere in the code. I might be missing something.

test/system/deposit_thesis_test.rb
```diff
@@ -128,6 +128,86 @@ class DepositThesisTest < ApplicationSystemTestCase
...
+
+    # 4. Review and Deposit Form
+    click_on I18n.t('admin.theses.draft.header')
+
+    # Success! Deposit Successful
+
+    assert_selector '.item-files > div', count: 3
+    assert_selector '.item-files > div:nth-child(1) .item-filename', text: 'image-sample.jpeg'
+    assert_selector '.item-files > div:nth-child(2) .item-filename', text: 'pdf-sample.pdf'
+    assert_selector '.item-files > div:nth-child(3) .item-filename', text: 'text-sample.txt'
```
jefferya, line 206: A check based on my comment above about the class `thesis-files`. I'm often confused between `item` and `thesis` and I'm unsure if this should be `thesis-files` and `thesis-filename`? The test passes. Feel free to resolve this comment.

jefferya (commented):
Looks good (though my Ruby is rusty).

I added a couple of minor inquires. One question, if I understand correctly, impacts the view meaning that if a pre-existing item is edited (e.g., metadata change), no change would occur to the underlying files or thumbnails. Also, if a new file is uploaded to a pre-existing item, the back-end file handling will remain as before.

app/views/admin/theses/draft/review_and_deposit_thesis.html.erb
lagoan: Good catch! I was hesitant about this class. I added it mainly to have a selector to simplify the tests. A similar CSS class is used in other parts of the code for items `item-files`. I used the `thesis-files` class here because the rest of the file referenced thesis directly instead of items.

What are your thoughts on adding classes to help with selectors in tests?

test/system/deposit_thesis_test.rb
lagoan: Ah yes, this part of the code uses a partial that is shared for items and theses. Maybe I should consider changing the class you mentioned in the comment above to `item-class` if it does not change the view.

lagoan:
You are right @jefferya, this should impact only the view, a metadata change would not change the underlying files or thumbnails.

I added the tests though because I did change the view and wanted to make sure the new method is used and I did not forget to change it in the view, or that the method is changed somehow and the files are no longer sorted alphabetically. Do you think this is overkill?

> Looks good (though my Ruby is rusty).
> 
> I added a couple of minor inquires. One question, if I understand correctly, impacts the view meaning that if a pre-existing item is edited (e.g., metadata change), no change would occur to the underlying files or thumbnails. Also, if a new file is uploaded to a pre-existing item, the back-end file handling will remain as before.

test/system/deposit_item_test.rb
```diff
@@ -132,6 +133,86 @@ class DepositItemTest < ApplicationSystemTestCase
     logout_user
   end
 
+  # This test follows a very similar behaviour to the successfully deposit an
+  # item making sure all required fields are present
```
pgwillia, line 137: Just curious why this is a mostly repeated test?  Could we combine the two?

test/system/deposit_thesis_test.rb
```diff
@@ -128,6 +128,86 @@ class DepositThesisTest < ApplicationSystemTestCase
     logout_user
   end
 
+  # This test follows a very similar behaviour to the successfully deposit a
+  # thesis making sure all required fields are present
```
pgwillia, line 132: Same question about combining the alpha check with the existing test.

pgwillia (commented):
This is probably a me not seeing something obvious thing.  I tried out your branch. I appears to me that the "Screenshot" filenames are out of order?  But at least it's consistent.

image
image
image

I'm glad that you made the single download from search results and google scholar metadata consistent with this change.

test/system/deposit_item_test.rb
lagoan: I did that initially and rubocop complained about having too many assertions in one test. I was adding more complexity to an already long test.

Following comments made in the past about being ok with having dumb controllers/tests for the sake of clarity, I went with repeating a portion of the test.

lagoan:
Let me take a closer look at this.

test/system/deposit_item_test.rb
pgwillia: Would we still be testing the right thing if we edited an existing item and used the wizard to `attach_file_in_dropzone` and assert the selectors in order, rather than repeating the whole deposit test?

lagoan: That sounds like a good plan 😄

app/views/admin/theses/draft/review_and_deposit_thesis.html.erb
pgwillia: I was looking at this to.  Adding classes to help with selectors in tests can be helpful.  The code doesn't end up using this selector until the last page of the wizard editing the thesis and we aren't using it in the test.  We're reusing the  `app/views/application/_files_section_sidebar.html.erb` for the theses show page which has the selector `item-files`.

image

test/system/deposit_thesis_test.rb
pgwillia: I think the item test is looking good where it is exercising just the files behaviour.  Will there be a similar change for this thesis test?

lagoan: Yes, sorry I missed adding it before I requested the review again. The test is up now.

app/views/admin/theses/draft/review_and_deposit_thesis.html.erb
lagoan: The test that uses de `thesis-filename` selector is now in the code. Sorry for missing it.

lagoan: https://github.com/ualbertalib/jupiter/pull/2963/files#diff-684c236d2bd463dbb936858cbd004dff1ee67675dcfd4809382b58d52e0f5497

pgwillia (approved):
I think this looks good.

PR #2968, opened 2022-09-13
Draft PR Test/increase capybara wait for asynchronous process

lagoan:
## Context

This is only a test PR to review the flapping errors coming up from uploading files.

Tests are failing due to items taking too long to upload and the test suite drops them after 5 seconds. By increasing the amount of time we wait for the change we should stop seeing these errors.

## What's New

This test PR should only increase the value `Capybara.default_max_wait_time` to see how it affects the tests results.

Remember to request a reviewer when you submit!


pgwillia (approved):
Let's try it.

Issue #2969, opened 2022-09-15
Rendering of markdown in collection browse view


seanluyk:
**Describe the bug**
Markdown appears to be rendering differently in the [collection browse view](https://era.library.ualberta.ca/communities?page=2) from how it does on the [collection landing page](https://era.library.ualberta.ca/communities/8eecb737-9733-4319-bd57-f93665d7176d) 

**To Reproduce**
1. Compare the two views from "Describe the bug" above
2. See error - link styling in collection browse view 

**Expected behavior**
Styling is consistent across views

**Additional context**
Issue is low priority for a fix

pgwillia:
I'm just leaving some screenshots to illustrate
![image](https://user-images.githubusercontent.com/1220762/190453791-bab9759e-0dbb-4958-ab84-c3e0e68544c4.png)
![image](https://user-images.githubusercontent.com/1220762/190453879-36be84e8-20fa-48fa-a54c-796a8049c7e7.png)

I think the difference is a line break before the Executive Summary and where the link starts in the first paragraph?

In fact it looks like all the descriptions are wrapped in a link to the collection.

seanluyk:
@pgwillia thanks, this is a helpful addition! P.S. it appears that the styling is "bleeding" onto other collection descriptions across the board, if that helps address the issue

pgwillia:
@seanluyk this was a really small and obvious fix.  You'll see it in the next release.

seanluyk:
@pgwillia awesome, thank you!

PR #2970, opened 2022-09-15
Fixed rendering of markdown in collection browse view.

pgwillia:
Before:
image

After:
image

## Context
Markdown appears to be rendering differently in the [collection browse view](https://era.library.ualberta.ca/communities?page=2) from how it does on the [collection landing page](https://era.library.ualberta.ca/communities/8eecb737-9733-4319-bd57-f93665d7176d)

If the description also included a link this would cause problems. This was caused by wrapping the whole community block of content (thumbnail, title, description) in a link tag.

Related to #2969 

## What's New

No longer wrap community description in link tag.


lagoan (approved):
Nice clean fix, good work!

Issue #2980, opened 2022-10-07
Error in Google workflow batch ingest when adding items with duplicate files


lagoan:
**Describe the bug**
When creating a batch ingest using the Google workflow and we have multiple items (described in rows) with the same file name, the system will create the batch ingest job and will finish, but only one item will in fact have the file attached to it and the user will not know about this error.

**To Reproduce**
1. Log in as admin
2. Click on admin from user drop down menu
3. Click on 'Batch ingests'
4. Click on 'New batch ingest'
5. Enter title
6. Add spreadsheet with duplicate file names in different rows (see attached image)
7. Add required files
8. Click on 'Create batch ingest'
9. Go to each item from the ingest and see how an error occurs when trying to access the duplicate entry of the file.

**Expected behavior**
New batch ingest should stop and indicate an error when multiple items list the same file name.

**Screenshots**

![image](https://user-images.githubusercontent.com/1537909/194595927-384f3e4d-b6db-47c4-93d5-c00cc9b73cfd.png)

PR #2981, opened 2022-10-07
Fix/duplicate file entries batch ingest

lagoan:
## Context

A bug was discovered where a batch ingest from the google workflow would be executed when a spreadsheet describing the ingest included the same file in multiple rows, but the items would not have the file in the system.

We should not allow for multiple items to have the same file so this is considered an error and the batch ingest should not proceed.

Related to #2980

## What's New

Add error checking when items include the same file name on batch ingest workflow. Fix tests to address the change in behaviour.


test/vcr/google_fetch_spreadsheet.yml
```diff
@@ -61,7 +61,7 @@ http_interactions:
             [
               "file_name",
               "title",
-              "alternate_title",
+              "alternative_title",
```
pgwillia, line 64: Oops.  I should have caught this when we made [that change for `alternative_title` in the job](https://github.com/ualbertalib/jupiter/commit/aa331d193d0157888446a556ed9abc6fcee18efe#diff-5277491872a0bded29c05ad45ba5f116c07d99a667c8c5172614be150e7154aa).  I guess we're not verifying metadata there, just that the validation passes and the item is saved.

pgwillia (approved):
LGTM!  Easy to read changes to the validator. Nice to update some spreadsheets cassettes that had old template headers.

pengyin-shan (approved):

pgwillia (approved):
👍

PR #2985, opened 2022-10-19
Add missed release information

lagoan:
## Context

Add missing release information to CHANGELOG


pgwillia (approved):
Should we keep the unreleased section?  It's probably the same amount of work as to add it back with the next change.

lagoan:
Good point, will add it on the next change.

Issue #2990, opened 2022-10-31
Wrong number of files attached to items ingested with google workflow


lagoan:
**Describe the bug**
When doing bast ingestion with multiple files for each item, only the last item is copied to storage, although the files' metadata is still attached to the items

**Expected behavior**
After an item is ingested, all of the files should be available in the filesystem.

lagoan:
This was fixed on #2991

PR #2991, opened 2022-10-31
Change depositable's behaviour to attach files

lagoan:
This change allows having a multi-file attachment even inside a base transaction block.

Before, the behaviour for attaching files in the depositable model allowed attaching a single file at a time. Now we make sure all of the copy callbacks are run inside a transaction block, actually copying the files, and not just the last one.

## Context

Multifile batch ingest was correctly adding the files' metadata but the file itself was not being copied to the storage area.

Related to #2990

## What's New

We changed the attaching method to attach multiple files at the same time, to allow their copy callbacks to be run after a transaction block.

### DIFF
--- a/app/models/jupiter_core/depositable.rb
+++ b/app/models/jupiter_core/depositable.rb
@@ -111,13 +111,16 @@ class JupiterCore::Depositable < ApplicationRecord
     return if file_handles.blank?
     raise 'Item not yet saved!' if id.nil?
 
-    file_handles.each do |fileio|
-      file_name = fileio.try(:original_filename) || File.basename(fileio.path)
-      attached = files.attach(io: fileio, filename: file_name)
-      # TODO: Do something smarter here if not attached
-      next unless attached
+    attachables = file_handles.map do |fileio|
+      filename = fileio.try(:original_filename) || File.basename(fileio.path)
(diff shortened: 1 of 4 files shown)


app/models/jupiter_core/depositable.rb
```diff
@@ -111,13 +111,16 @@ def add_and_ingest_files(file_handles = [])
...
-      # TODO: Do something smarter here if not attached
-      next unless attached
+    attachables = file_handles.map do |fileio|
+      filename = fileio.try(:original_filename) || File.basename(fileio.path)
+      { io: fileio, filename: filename }
+    end
+
+    # We need to attach all the files at the same time to make sure their
+    # callbacks are run when they are wrapped in a base transaction block
+    files.attach(attachables)
```
pgwillia, line 121: > If the record is persisted and unchanged, the attachments are saved to the database immediately. Otherwise, they'll be saved to the DB when the record is next saved.

https://devdocs.io/rails~6.1/activestorage/attached/many#method-i-attach

We weren't doing anything about the `#TODO` on line 117 before so I agree with this change that we not worry about doing something if not attached.

pgwillia (approved):
Great job!  This not only fixes the bug but I think it's easier to read than before.

PR #2999, opened 2022-11-16
Revert "Bump redis from 4.7.1 to 4.8.0" due to seeding issues

ConnorSheremeta:
This reverts commit ef1bcff5feb19b86cec285c906537510cacb311d.

## Context

Bumped redis from 4.7.1 to 4.8.0 and saw that it caused some issues while seeding the db so I'm reverting it here

Related to # (issue)

https://github.com/ualbertalib/jupiter/pull/2962

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ gem 'rdf-vocab', '~> 3.2.1' # this gem changed predicate names without warning, 
 # Database stuff
 gem 'connection_pool'
 gem 'pg', '~> 1.4.3'
-gem 'redis', '~> 4.8'
+gem 'redis', '~> 4.7'
 gem 'rsolr'
 gem 'strong_migrations', '~> 1.4.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -441,7 +441,7 @@ GEM
       rexml (~> 3.2)
     rdoc (6.3.3)
     redcarpet (3.5.1)
-    redis (4.8.0)
+    redis (4.7.1)
     regexp_parser (2.6.0)
     representable (3.1.1)
       declarative (< 0.1.0)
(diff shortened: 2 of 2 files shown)


lagoan (approved):
LGTM!

pgwillia (changes requested):
Could you tell me more about the seeding issue?

What I'm seeing is a deprecation warning but it looks to me like the seeds and jupiter continue to function normally.
```
Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /home/pjenkins/.asdf/installs/ruby/3.0.4/lib/ruby/gems/3.0.0/gems/sidekiq-6.4.1/lib/sidekiq/client.rb:227:in `atomic_push')
```
and 
```Passing the timeout as a positional argument is deprecated, it should be passed as a keyword argument:
  redis.brpop("queue:critical", "queue:default", "queue:active_storage_analysis", "queue:active_storage_purge", "queue:mailers", "queue:low", timeout: 2)(called from: /home/pjenkins/.asdf/installs/ruby/3.0.4/lib/ruby/gems/3.0.0/gems/sidekiq-6.4.1/lib/sidekiq/fetch.rb:47:in `block in retrieve_work')
```

I think this is a deprecation warning showing up from sidekiq.  I think they fixed it in 6.5.5 (.https://github.com/mperham/sidekiq/issues/5484).  It looks like we haven't been getting sidekiq bumps because of 
https://github.com/ualbertalib/jupiter/blob/0613069831dcdd4f4e9627636a410093d0243f3d/.github/dependabot.yml#L15-L18

I think this could have been cleared after https://github.com/ualbertalib/jupiter/pull/2745 so that we continue to get sidekiq updates.  I'll create a PR which removes that block.  Then we won't have to revert this.

ConnorSheremeta:
Yes, it looks like the bump to sidekiq resolves these warnings. Closing in favour of https://github.com/ualbertalib/jupiter/pull/3008

Issue #3023, opened 2022-12-20
Remove Skylight


pgwillia:
**Is your feature request related to a problem? Please describe.**
I get weekly emails from skylight about problem performance that we don't have the capacity to act on.  Some of the information given seems to just be noise in the system.
![image](https://user-images.githubusercontent.com/1220762/208774558-73d07f9f-22b5-4b15-b761-7c7cf7ecd5ad.png)
![image](https://user-images.githubusercontent.com/1220762/208774610-cce43c81-243c-437e-b64e-6feced769987.png)

**Describe the solution you'd like**
Remove skylight.  The opposite of https://github.com/ualbertalib/jupiter/pull/310/files
- [x] https://github.com/ualbertalib/jupiter/blob/76d33b0a6f5e8994268a682cf68efecbeffb8a0f/.env_deployment_sample#L30
- [x] https://github.com/ualbertalib/jupiter/blob/76d33b0a6f5e8994268a682cf68efecbeffb8a0f/Gemfile#L67-L68
- [x] https://github.com/ualbertalib/jupiter/blob/76d33b0a6f5e8994268a682cf68efecbeffb8a0f/config/application.rb#L42-L43
- [ ] Communicate decision to these collaborators and ask @weiweishi who's skylight owner to close the account/connection with github 
![image](https://user-images.githubusercontent.com/1220762/208775582-e45c2145-facc-45af-8562-55c801f73c5d.png)

**Describe alternatives you've considered**
Could leave it in.  However we have extra steps as part of deployments to accommodate this dependency. 

**Additional context**
We've decided not to bump it until cloud https://github.com/ualbertalib/jupiter/issues/2260
This is the original ticket that adds it https://github.com/ualbertalib/jupiter/pull/310
This adds it to staging https://github.com/ualbertalib/jupiter/pull/1865

We have additional workaround steps in order to deploy jupiter because of skylight: https://github.com/ualbertalib/jupiter-deployment/blob/e6e8e01e8e16d1c0bde289483170a594ae9a4a12/deployment/Henry_Deployment_Notes.txt#L62-L64

PR #3028, opened 2022-12-23
Remove skylight

pgwillia:
## Context

We installed this to get more observability about jupiter performance.  I've been getting these weekly emails for years without acting on them.  We now have netdata for some performance statistics.

I've left in ActiveSupport::Notifications instrument (https://edgeguides.rubyonrails.org/active_support_instrumentation.html) in case it comes in handy in the future.  I have removed the normalizer for Skylight.

Related to #3023 

## What's New

Remove Skylight.io

### DIFF
--- a/.env_deployment_sample
+++ b/.env_deployment_sample
@@ -27,4 +27,3 @@ POSTGRES_USER=jupiter
 
 # Nginx environment variables
 HOSTNAME=localhost
-SKYLIGHT_AUTHENTICATION=secretauthenticationtoken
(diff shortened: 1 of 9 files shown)


lagoan (commented):
It all looks good for removing skylight from the project.

I have a question about keeping the ActiveSupport::Notifications instrument. It is indeed interesting, but I don't know if we should keep it if we are not using it at all. Also, we can add it later easily when needed.

pgwillia:
> It all looks good for removing skylight from the project.
> 
> I have a question about keeping the ActiveSupport::Notifications instrument. It is indeed interesting, but I don't know if we should keep it if we are not using it at all. Also, we can add it later easily when needed.

If I'm being honest I saw it integrated in search and balked.  I can try again, you're right that we shouldn't keep unused code.

lagoan (approved):
Good work!

PR #3040, opened 2023-01-19
Wip add community and collection to aips

ConnorSheremeta:
Communities and collections get pushed to preservation queue on save along with there being a rake task for adding them to the queue

### DIFF
--- a/test/integration/add_to_preservation_queue_task_test.rb
+++ b/test/integration/add_to_preservation_queue_task_test.rb
@@ -7,11 +7,10 @@ Rails.application.load_tasks
 class AddToPreservationQueueTaskTest < ApplicationSystemTestCase
 
   setup do
-    Rake::Task['jupiter:preserve_all_collections_and_communities'].reenable
-    Rake::Task['jupiter:preserve_all_items_and_theses'].reenable
-    RedisClient.current.del Rails.application.secrets.preservation_queue_name
     @admin = users(:user_admin)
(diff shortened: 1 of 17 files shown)


test/integration/add_to_preservation_queue_task_test.rb
```diff
@@ -52,16 +51,22 @@ class AddToPreservationQueueTaskTest < ApplicationSystemTestCase
...
     travel 1.week do
-      Rake::Task['jupiter:preserve_all_items_and_theses'].invoke(Date.current.to_s)
+      Rake::Task['jupiter:preserve_all_items_and_theses'].execute(after_date_arguments)
       assert_equal 0, RedisClient.current.zcard(Rails.application.secrets.preservation_queue_name)
     end
   end
 
+  private
+
+  def after_date_arguments
```
lagoan, line 68: Did you consider using a class variable instead of a method?

```ruby
@after_date_arguments =  Rake::TaskArguments.new([:after_date], [Date.current.to_s])
```

and use ``@after_date_arguments`` in the rest of the code.

It would not change much, only that we would not be creating an object for every test. No need to change the code, just curious about why you followed this approach.

ConnorSheremeta: The reason I went down this route is because I need Date.current.to_s to be called inside of the `travel` blocks in order to test the after_date parameter properly

PR #3079, opened 2023-02-17
Version update to 2.4.4

lagoan:
## Context

Preparing release for 2.4.4


pgwillia (changes requested):
You'll also need to bump in [version.rb](https://github.com/ualbertalib/jupiter/blob/master/lib/jupiter/version.rb). Like this: https://github.com/ualbertalib/jupiter/commit/7120abab758678a6b2bf5e97fd509cdb4269f8a3

pgwillia (approved):

PR #3091, opened 2023-03-15
WIP ruby upgrade from 2.7

ConnorSheremeta:
WIP

PR #3096, opened 2023-03-23
Cds/prepare for ruby 3.1 upgrade

ConnorSheremeta:
This PR is to prepare for the ruby 3.1 upgrade which will be done in [this PR](https://github.com/ualbertalib/jupiter/pull/3091)


app/models/jupiter_core/solr_services/client.rb
```diff
@@ -3,7 +3,7 @@ class JupiterCore::SolrServices::Client
   include Singleton
 
   SOLR_CONFIG = YAML.safe_load(ERB.new(Rails.root.join('config/solr.yml').read).result,
-                               [], [], true)[Rails.env].symbolize_keys
+                               permitted_classes: [], permitted_symbols: [], aliases: true)[Rails.env].symbolize_keys
```
pgwillia, line 6: This looks left over from when the rubocop target changed to 3.1.  I'm assuming it's compatible.

pgwillia (approved):
👍 Looks good to me.  Just noted my assumption

PR #3150, opened 2023-07-05
Override unmaintained word-wrap package with a currently maintained fork to resolve CVE

ConnorSheremeta:
Override unmaintained word-wrap node package with a currently maintained fork ([aashutoshrathi/word-wrap](https://www.npmjs.com/package/@aashutoshrathi/word-wrap/v/1.2.6)) to resolve CVE

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Security
 - Bump rails from 6.1.7.2 to 6.1.7.4 [PR#3149](https://github.com/ualbertalib/jupiter/pull/3149)
-- Override unmaintained word-wrap package with a currently maintained fork to resolve CVE
+- Override unmaintained word-wrap package with a currently maintained fork to resolve CVE [PR#3150](https://github.com/ualbertalib/jupiter/pull/3150)
 
 ### Fixed 
 - Resolved new rubocop offenses [PR#3109](https://github.com/ualbertalib/jupiter/pull/3109)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Security
 - Bump rails from 6.1.7.2 to 6.1.7.4 [PR#3149](https://github.com/ualbertalib/jupiter/pull/3149)
-- Override unmaintained word-wrap package with a currently maintained fork to resolve CVE [PR#3150](https://github.com/ualbertalib/jupiter/pull/3150)
+- Override unmaintained word-wrap package with a currently maintained fork through eslint package upgrade [PR#3150](https://github.com/ualbertalib/jupiter/pull/3150)
 
 ### Fixed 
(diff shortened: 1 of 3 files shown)


pgwillia:
My reading of this particular vulnerability is that it doesn't impact us.
```
pjenkins@NITROGEN:~/Code/ualbertalib/jupiter$ yarn why word-wrap
yarn why v1.22.19
[1/4] Why do we have the module "word-wrap"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "word-wrap@1.2.3"
info Reasons this module exists
   - "eslint#optionator" depends on it
   - Hoisted from "eslint#optionator#word-wrap"
info Disk size without dependencies: "16KB"
info Disk size with unique dependencies: "16KB"
info Disk size with transitive dependencies: "16KB"
info Number of shared dependencies: 0
Done in 0.56s.
```

We only have this in our project because of eslint, so the only opportunity for exploitation would be us on ourselves in our development and test environments.  On that basis, I'm inclined to dismiss the security alert.  However, there is [a fix](https://github.com/eslint/eslint/issues/17117) in eslint.  Is there a way we can take advantage of that work?

ConnorSheremeta:
> My reading of this particular vulnerability is that it doesn't impact us.
> 
> ```
> pjenkins@NITROGEN:~/Code/ualbertalib/jupiter$ yarn why word-wrap
> yarn why v1.22.19
> [1/4] Why do we have the module "word-wrap"...?
> [2/4] Initialising dependency graph...
> [3/4] Finding dependency...
> [4/4] Calculating file sizes...
> => Found "word-wrap@1.2.3"
> info Reasons this module exists
>    - "eslint#optionator" depends on it
>    - Hoisted from "eslint#optionator#word-wrap"
> info Disk size without dependencies: "16KB"
> info Disk size with unique dependencies: "16KB"
> info Disk size with transitive dependencies: "16KB"
> info Number of shared dependencies: 0
> Done in 0.56s.
> ```
> 
> We only have this in our project because of eslint, so the only opportunity for exploitation would be us on ourselves in our development and test environments. On that basis, I'm inclined to dismiss the security alert. However, there is [a fix](https://github.com/eslint/eslint/issues/17117) in eslint. Is there a way we can take advantage of that work?

Good catch, it looks like upgrading to v8.44.0 of eslint would resolve this as well as it has switched to the same fork (https://github.com/gkz/optionator/blob/04e773185572b1c37fe469354e101f2ea3cd9ce5/package.json#L34) which would be a cleaner solution. Although we are not currently affected by this there is a chance that this package could be a dependency of something used in prod in the future; I think this fix is worthwhile and will be pushing up the changes soon.

ConnorSheremeta:
Upgrading esling to v8.44.0 is giving me issues surrounding unmaintained packages. They do not seem easily resolvable therefore I am becoming more inclined to dismiss this alert, or maybe go for the first solution potentially. It is doubtful this will cause impact in the future especially considering eslint is explicitly defined in the 'devDependencies' group.

It may be worthwhile putting in the effort to resolve these unmaintained packages with maintained ones however, I suppose that it could be done if we keep running into issues with those unmaintained packages down the road .

PR #3184, opened 2023-08-30
Update dependabot.yml for npm updates to use correct `javascript` label

murny:
## Context

image

As the screenshot above shows, our javascript dependabot PRs are being opened with a `ruby` label, this isn't correct.

We have a `javascript` label already, let's use this instead of `ruby`


pgwillia (approved):
👍  LGTM

PR #3185, opened 2023-08-31
Branch to play with trying to get jupiter running on local

murny:
DRAFT MR to document some of the issues I ran into while running Jupiter locally. 

Most of the pain is due to trying to running old migrations and Solr docker not running on M1 

For migrations, I ran into a few issues:

- StrongMigration, even though we told it to only look at "recent" migrations via `StrongMigrations.start_after = 20211117160533`

Ends up failing on `db/migrate/20200320164053_change_is_published_in_era_column_to_not_null.rb` due to StrongMigrations not supporting the ability to handle adding a `default` onto the column after turning it `non null`

https://github.com/ankane/strong_migrations/blob/fa4356456bc5994698f39b159d7614e9c9a763c6/lib/strong_migrations/safe_methods.rb#L85C1-L87

Not sure any way to fix this other than, just not doing this? 2 migrations later, we drop this column anyways so I think this is safe to do? 

image

- We had a weird hash `null => false` syntax, this was actually updating the schema.rb in a really odd way. Think we can just fix this to use `, null: false` like we do everywhere else.
image


docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
 
   solr:
     image: solr:6.6
+    platform: linux/amd64
```
murny, line 18: Required for me to run Solr on M1 MacBook, but not sure if this works for other people 🤔. Believe this is only for local development, but hesitant to bring this in without further testing. Will try running Jupiter locally on an intel Macbook and see if this still works.

db/schema.rb
```diff
@@ -498,7 +498,7 @@
   add_foreign_key "collections", "users", column: "owner_id"
   add_foreign_key "communities", "users", column: "owner_id"
   add_foreign_key "digitization_batch_metadata_ingests", "users"
-  add_foreign_key "digitization_books", "active_storage_attachments", column: "logo_id", on_delete: :nullify
+  add_foreign_key "digitization_books", "active_storage_attachments", column: "logo_id"
```
murny, line 501: Not sure why this is getting dropped 🤔 Will have to investigate this further

murny:
Some other problems: 

- Lots of deprecation warnings
- reassigning Constants warnings in batch_ingest task
-  VCR appears to be crashing after running tests, can see this in the CI runner here:
image

docker-compose.yml
murny: Seemed to work fine on my Intel Macbook as well. So, think we should probably do this. Essentially the problem is docker will try to install the image that matches your OS, so for M1 it will try to install `linux/arm64/v8` version of Solr dockers image, but this does NOT exist for this old image:

image

So we need to tell docker to always use `linux/amd64` instead via the platform key here.

https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos

murny:
Closing this in favor of 2 MRs that focus on these issues

PR #3193, opened 2023-09-01
Set platform for linux/amd64 for our Solr image

murny:
## Context

Docker will attempt to grab the default docker image for your Operating System. However, if a docker image does not have a version for your system, you will get the following error:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested.
```

This is mainly a problem for new Apple Silicon machines (M1 macbooks for example) as they use ARM64 artitecture. 

A simple solution is to force our docker compose to use `platform: linux/amd64`. Which is the only version Solr 6.6 comes in anyway: 

image

More info here from official docs: https://docs.docker.com/desktop/troubleshoot/known-issues/

```
Some images do not support the ARM64 architecture. You can add --platform linux/amd64 to run (or build) an Intel image using emulation.

However, attempts to run Intel-based containers on Apple silicon machines under emulation can crash as qemu sometimes fails to run the container. In addition, filesystem change notification APIs (inotify) do not work under qemu emulation. Even when the containers do run correctly under emulation, they will be slower and use more memory than the native equivalent.

In summary, running Intel-based containers on Arm-based machines should be regarded as "best effort" only. We recommend running arm64 containers on Apple silicon machines whenever possible, and encouraging container authors to produce arm64, or multi-arch, versions of their containers. This issue should become less common over time, as more and more images are rebuilt [supporting multiple architecturesopen_in_new](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/).
```

So basically they saying you can force the platform to `linux/amd64` which allows containers to work in M1. But it's not ideal since its emulation and docker image creators should be supporting other platforms.

Most popular libraries and their latest versions already do this. Even latest versions of Solr provide multiple different images for different platforms as shown below:
image

So once we get on a later version of Solr this can be removed.


murny:
> I didn't encounter the ARM64 architecture issue, but I can confirm that this change continues to work for me. Another issue that reinforces that we need to move on replacing Solr 6.6.

If you have a M1/M2 macbook you may be using Rosetta (intel emulation which would use linux/amd64)? If you are using M1 without Rosetta, and not getting this error, then I'm not sure 🤔?

pgwillia:
> > I didn't encounter the ARM64 architecture issue, but I can confirm that this change continues to work for me. Another issue that reinforces that we need to move on replacing Solr 6.6.
> 
> If you have a M1/M2 macbook you may be using Rosetta (intel emulation which would use linux/amd64)? If you are using M1 without Rosetta, and not getting this error, then I'm not sure 🤔?

Oh, I didn't encounter the issue because I don't have a MacBook or use that hardware!

Issue #2491, opened 2021-08-25
Use real database associatations for logos in `Depositable` 


pgwillia:
An improvement we could make in the future where we can use real database associations now that we are no longer on fedora. This `logo_id` being a proper join and we can walk the association instead of doing code like this when accessing this "logo/thumbnail" object:
https://github.com/ualbertalib/jupiter/blob/3e2d4ae5cd3bc3b1f227b843a725dc1b34d5f10a/app/models/jupiter_core/depositable.rb#L110-L112

_Originally posted by @murny in https://github.com/ualbertalib/jupiter/pull/2468#r688148493_

May be related to #2340

PR #3194, opened 2023-09-01
Squash old migrations and RDF annotations

murny:
## Context

This uses a gem called https://github.com/jalkoby/squasher to squash old migrations (everything before 2020/05/29) into a single migration.

RDF Annotations are something unique, had to try to squash theses as well. I believe I preserved them all, just the order of rdf annotations inside`db/seeds/rdf_annotations.rb` have changed a tad (for the better?). I assume this is fine, though.

Why we doing this?

There is a problem with `db/migrate/20200320164053_change_is_published_in_era_column_to_not_null.rb` migration not being supported in StrongMigrations gem. Even though we tell StrongMigrations gem to only look at "recent" migrations via `StrongMigrations.start_after = 20211117160533`.  This is still an issue as StrongMigrations attempts to load all migration regardless of this setting.

The problem is StrongMigrations code does not support the ability to handle adding a default onto the column after turning it non null as seen here:

https://github.com/ankane/strong_migrations/blob/fa4356456bc5994698f39b159d7614e9c9a763c6/lib/strong_migrations/safe_methods.rb#L85C1-L87

This crashes running migrations with the following error:
image

Another issue is we use a weird syntax for setting `null: false` in the Versions table here:

https://github.com/ualbertalib/jupiter/blob/master/db/migrate/20200127185942_create_versions.rb#L13
`      t.string   :item_type, {:null=>false}`

This ends up resulting in a weird update to our schema.rb as shown:

image

So we would have had to edit these two migrations anyway to allow developers to be able to migrate their DBs from a fresh start. 

While most users would most likely seed their database via `schema.rb` instead, it's still a good idea to have your migrations healthy and being able to migrate from a fresh instance.

Rails official stance is `schema.rb` is the official snapshot of your schema and not your migration files.  So it's okay to delete or prune old migration files: https://edgeguides.rubyonrails.org/active_record_migrations.html#old-migrations


db/migrate/20200127185943_add_object_changes_to_versions.rb
```diff
@@ -1,12 +0,0 @@
-# This migration adds the optional `object_changes` column, in which PaperTrail
-# will store the `changes` diff for each update event. See the readme for
-# details.
-class AddObjectChangesToVersions < ActiveRecord::Migration[5.2]
-  # The largest text column available in all supported RDBMS.
-  # See `create_versions.rb` for details.
-  TEXT_BYTES = 1_073_741_823
```
murny, line 7: The limits on object/object_changes seems to be dropped in schema.rb 🤔 (my squashed migration does the same thing). Wonder if this wasn't actually doing anything

db/migrate/20200305210657_add_system_account_if_needed.rb
```diff
@@ -1,18 +0,0 @@
-class AddSystemAccountIfNeeded < ActiveRecord::Migration[6.0]
-
-  def up
```
murny, line 3: This appears to be have been dropped, but this in our seeds.rb anyway, so we will be good here

db/schema.rb
```diff
@@ -498,7 +498,7 @@
   add_foreign_key "collections", "users", column: "owner_id"
   add_foreign_key "communities", "users", column: "owner_id"
   add_foreign_key "digitization_batch_metadata_ingests", "users"
-  add_foreign_key "digitization_books", "active_storage_attachments", column: "logo_id", on_delete: :nullify
+  add_foreign_key "digitization_books", "active_storage_attachments", column: "logo_id"
```
murny, line 501: Have to debug this, not sure why this is getting dropped?

db/seeds/rdf_annotations.rb
```diff
@@ -1,4 +1,8 @@
 RdfAnnotation.create!([
+  {table: "active_storage_blobs", column: "filename", predicate: "http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#resourceFilename"},
```
murny, line 2: The diffs in here are just reordering, everything still exists as before

db/schema.rb
murny: My guess from looking at the git history, we used `t.references` to make the schema change, but then later edited this to use a later migration that used `add_reference` but maybe never updated the schema again? These should be similar, but may have differences involved.

https://github.com/ualbertalib/jupiter/pull/2468/commits/5e8c9f44361e93763626af4c581ae309efcc96f8
https://github.com/ualbertalib/jupiter/pull/2468/commits/70f718d99e04c3f4f6ad1783245bf095db11ef4b#r688126020

So running migrations from scratch now, won't put the `on_delete: :nullify` as the original migration was edited/updated. 

Is this a big deal? Not sure, my feeling is we should be doing this at the model level if we care about it:
https://github.com/ualbertalib/jupiter/issues/2491 

Currently, this `logo` concept doesn't exist at the model level at all? And if we did want to `nullify` it on delete, do this at the ActiveRecord model instead of the database level (we do these things currently all at the model level, why we doing this one differently). As an example: https://github.com/ualbertalib/jupiter/blob/2de2947cc9b2b93019670479dbcd81f06d27e5fb/app/models/type.rb#L3

pgwillia: We decided not to add digitization to Jupiter.  Would it be helpful to remove all that code as it is not going to see use in production?

db/migrate/20200305210657_add_system_account_if_needed.rb
pgwillia: Good catch.  We're good.

db/migrate/20200127185943_add_object_changes_to_versions.rb
pgwillia: I suspect it wasn't actually doing anything as we use PostgreSQL.  This came from a papertrail generator.  The initial discussion was .https://github.com/paper-trail-gem/paper_trail/issues/248

pgwillia (approved):
I'm in favour of healthy migrations.  Thanks for sharing the Rails guidance on this. Because we have the git history and the database schema, I don't think we've lost any of the storytelling/documentation that the migrations provide.

Works for me!  I can use `rails db:setup` or 
`rails db:create && rails db:migrate && rails db:seed` after these changes.

db/schema.rb
murny: Less code, less problems 😄. Yeah, I think it would be trivial to remove it all? Basically, if I recall, it was just some CRUD scaffolds to some new models. So we could just remove it all, then write a migration to drop the tables (so we drop them in production). Then we should be good 🤔

PR #3210, opened 2023-09-13
Fix warning about already initialized constants in rake tests within tests

murny:
## Context

Our test output was polluting the console with the following warnings:
```
/jupiter/lib/tasks/batch_ingest.rake:1: warning: already initialized constant INGEST_REPORTS_LOCATION
/jupiter/lib/tasks/batch_ingest.rake:1: warning: previous definition of INGEST_REPORTS_LOCATION was here
/jupiter/lib/tasks/batch_ingest.rake:2: warning: already initialized constant INDEX_OFFSET
/jupiter/lib/tasks/batch_ingest.rake:2: warning: previous definition of INDEX_OFFSET was here
```

Appears we were loading the same rake tasks multiple times, which attempts to redefine the CONSTANTS in batch_ingest.rake file.

We can simply just load our rake tasks in our tests in a smarter fashion. Load them once, and don't load them again if they are already loaded.

### DIFF
--- a/test/integration/add_to_preservation_queue_task_test.rb
+++ b/test/integration/add_to_preservation_queue_task_test.rb
@@ -2,11 +2,11 @@ require 'test_helper'
 require 'rake'
 require 'application_system_test_case'
 
-Rails.application.load_tasks
-
 class AddToPreservationQueueTaskTest < ApplicationSystemTestCase
 
   setup do
+    Jupiter::Application.load_tasks if Rake::Task.tasks.empty?
+
     @admin = users(:user_admin)
     login_user(@admin)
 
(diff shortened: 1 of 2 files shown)


test/integration/add_to_preservation_queue_task_test.rb
```diff
@@ -2,11 +2,11 @@
 require 'rake'
 require 'application_system_test_case'
 
-Rails.application.load_tasks
-
 class AddToPreservationQueueTaskTest < ApplicationSystemTestCase
 
   setup do
+    Jupiter::Application.load_tasks if Rake::Task.tasks.empty?
```
murny, line 8: Same code as before, just moved it into the `setup` block and only load the tasks if they haven't been loaded before.

This makes the warnings go away as we not loading the tasks multiple times.

pgwillia (approved):
:+1: LGTM!  An elegant solution.

As an aside, the branch wasn't working on my machine initially but it had nothing to do with these particular changes.  It had a conflict in the Chromedriver versions.  I deleted `/usr/bin/chromedriver ` and then reinstalled with `sudo snap refresh chromium`

Issue #3014, opened 2022-12-12
RemoveInactiveDraftsJob triggers an ArgumentError after sidekiq bump


pgwillia:
View details in Rollbar: [https://rollbar.com/ualbertalib/jupiter/items/1514/](https://rollbar.com/ualbertalib/jupiter/items/1514/)

https://github.com/ualbertalib/jupiter/blob/46f1c99e64cd893735d1d612e472e53746b95643/app/jobs/remove_inactive_drafts_job.rb#L13-L23

Looks like this may be fixed in sidekiq?  Not sure which version .https://github.com/mperham/sidekiq/issues/5689

```

ArgumentError: wrong number of arguments (given 1, expected 0)
  File "/var/www/sites/jupiter/app/jobs/remove_inactive_drafts_job.rb", line 13, in perform
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activejob-6.1.6.1/lib/active_job/execution.rb", line 48, in block in perform_now
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activesupport-6.1.6.1/lib/active_support/callbacks.rb", line 117, in block in run_callbacks
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/i18n-1.12.0/lib/i18n.rb", line 322, in with_locale
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activejob-6.1.6.1/lib/active_job/translation.rb", line 9, in block (2 levels) in <module:Translation>
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activesupport-6.1.6.1/lib/active_support/callbacks.rb", line 126, in instance_exec
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activesupport-6.1.6.1/lib/active_support/callbacks.rb", line 126, in block in run_callbacks
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activesupport-6.1.6.1/lib/active_support/core_ext/time/zones.rb", line 66, in use_zone
  File "/var/www/sites/jupiter/vendor/ruby/2.7.0/gems/activejob-6.1.6.1/lib/active_job/timezones.rb", line 9, in block (2 levels) in <module:Timezones>
[...]

PR #3217, opened 2023-09-14
Update Sidekiq and Sidekiq cron/unique job gems to latest versions

murny:
## Context

This PR upgrades Sidekiq and sidekiq-cron/sidekiq-unique-job gems to the latest version.

Dependabot has been attempting to do this in the past, like this latest one here: https://github.com/ualbertalib/jupiter/pull/3214 (not sure why for some reason its descriptions are saying it's updating oaysis/rdf/etc gems as well 🤔). This PR was also failing due to not updating sidekiq-cron. So this new PR will do this as well.

Another issue was we now using `redis-client` gem, which also uses the `RedisClient` constant name (they use a class, we used a module). We are now clashing on this. Judging from the PR that brought in our version of `RedisClient` https://github.com/ualbertalib/jupiter/pull/2745, the intention wasn't to shadow this gem at all, so we can just rename it to something better.

So just renamed this to be `Jupiter::Redis` and also wrapped it in the connection pool (we already have the gem in our application)? I assume this is fine? Not sure if there was a reason not to use connection pool in the past 🤔. Then with the new renamed module, I updated all the usages of this.

Sidekiq seems to be working fine locally (all tests are passing) and any job I ran seem to run fine:

image

Ran all the cron jobs we have as well and these all succeed fine:

image

So hopefully with being on Sidekiq 7.0.4+ this issue https://github.com/ualbertalib/jupiter/issues/3014 should be fixed now as well

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -494,7 +494,7 @@ GEM
       parser (>= 3.2.1.0)
     rubocop-minitest (0.31.1)
       rubocop (>= 1.39, < 2.0)
-    rubocop-performance (1.19.0)
+    rubocop-performance (1.19.1)
       rubocop (>= 1.7.0, < 2.0)
       rubocop-ast (>= 0.4.0)
     rubocop-rails (2.21.1)


app/models/jupiter_core/depositable.rb
```diff
@@ -161,14 +161,10 @@ def add_to_embargo_history
...
 
-  # rubocop:disable Style/GlobalVars
-
   def push_entity_for_preservation
     queue_name = Rails.application.secrets.preservation_queue_name
 
-    $queue ||= ConnectionPool.new(size: 1, timeout: 5) { RedisClient.current }
-
-    $queue.with do |connection|
+    Jupiter::Redis.current.with do |connection|
```
pgwillia, line 167: @lagoan this is probably a stretch but... do you think the redundant connection pool could have been a source for any of the errors we were tracking down with PMPY?

pgwillia:
> Dependabot has been attempting to do this in the past, like this latest one here: #3214 (not sure why for some reason its descriptions are saying it's updating oaysis/rdf/etc gems as well 🤔). This PR was also failing due to not updating sidekiq-cron. So this new PR will do this as well.

Yeah, Dependabot had an odd dependency resolution thing with flipper recently too https://github.com/ualbertalib/jupiter/pull/3171#issuecomment-1661086283  I couldn't find an issue that matched in the dependabot-core repo on a quick search, though Grouped Updates sounds interesting.

> 
> Another issue was we now using `redis-client` gem, which also uses the `RedisClient` constant name (they use a class, we used a module). We are now clashing on this. Judging from the PR that brought in our version of `RedisClient` #2745, the intention wasn't to shadow this gem at all, so we can just rename it to something better.
> 
> So just renamed this to be `Jupiter::Redis` and also wrapped it in the connection pool (we already have the gem in our application)? I assume this is fine? Not sure if there was a reason not to use connection pool in the past 🤔. Then with the new renamed module, I updated all the usages of this.

Looks good.  This will remove some redundant initialization of the ConnectionPool too!

pgwillia (approved):

app/models/jupiter_core/depositable.rb
pgwillia: There were the errors: https://github.com/ualbertalib/pushmi_pullyu/issues/311#issuecomment-1504196933

PR #3227, opened 2023-09-15
Fix Non system tests from running capybara/selenium

murny:
## Context

I noticed our non system tests were running capybara/selenium. 

image

It appears we had an integration test inheriting the wrong test helper.

Fixed this and updated any login helpers to the non system test way.  Appears like everything is passing correctly now.


pgwillia (approved):
Nice catch!

Issue #965, opened 2019-01-08
Gem: Upgrade Pundit to latest (2.0.0)


murny:
From #964, We noticed that we couldn't upgrade pundit to the latest version. Looks like it has a few breaking changes with how we are using it and will break tests.

Changelog here: https://github.com/varvet/pundit/blob/master/CHANGELOG.md

Perhaps we can go and figure out the breaking changes and fix them so we can still receive updates for this gem.

ConnorSheremeta:
Reopen then close #1248 when this is completed so that dependabot watches this dependency again

PR #3237, opened 2023-09-21
Upgrade pundit to latest

murny:
## Context

Bump Pundit from v1.1.0 to v2.3.1

The big breaking change with this upgrade was due to this: https://github.com/varvet/pundit/pull/529

Previously with a namespace authorize like `authorize [:admin, @user]` which we do lots for our Admin policies like `app/policies/admin/user_policy.rb` the "record" would be an array containing `[:admin, @user]`. Now after v2.0.0 of pundit, it's the last element of the array. So the "record" is now `@user` in our example.

This is what we expected anyway, we just manually did `record.last` in places. But now we can just drop the `.last` call everywhere. 

Cleaned up any places we were doing this. Everything appears to be working now.

Related to https://github.com/ualbertalib/jupiter/issues/965

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -495,7 +495,7 @@ GEM
       unicode-display_width (>= 2.4.0, < 3.0)
     rubocop-ast (1.29.0)
       parser (>= 3.2.1.0)
-    rubocop-minitest (0.32.1)
+    rubocop-minitest (0.32.2)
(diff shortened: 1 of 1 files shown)


pgwillia (approved):
:+1:  LGTM!  It's nice when breaking changes make things on our end more readable.

PR #3255, opened 2023-10-05
v2.6.0

murny:
## Context

Cutting a v2.6.0 release for Jupiter. Added any missing changelog entries.

Will write a description in the release notes for the Paper Trial migration. For how it needs a good vetting in staging (staging should get data synced from production, then deploy release and run migrations, watching for any complications or if this migration is taking far too long, etc.)

### Added

- Add dev/setup binstubs to more easily setup and dev on local [PR#3238](https:://github.com/ualbertalib/jupiter/pull/3238)
- Upgrade pundit to latest [PR#3237](https:://github.com/ualbertalib/jupiter/pull/3237)
- Bump paper_trail from 13.0.0 to 15.0.0 [PR#3236](https:://github.com/ualbertalib/jupiter/pull/3236)
- Update Sidekiq and Sidekiq cron/unique job gems to latest versions [PR#3217](https:://github.com/ualbertalib/jupiter/pull/3217)
- Set platform for linux/amd64 for our Solr image [PR#3193](https:://github.com/ualbertalib/jupiter/pull/3193)
- add @murny to CODEOWNERS [PR#3183](https:://github.com/ualbertalib/jupiter/pull/3183)

### Removed

- Rails no longer recommends Spring by default. Let's remove it [PR#3231](https:://github.com/ualbertalib/jupiter/pull/3231)

### Fixed

- Squash old migrations and RDF annotations [PR#3194](https:://github.com/ualbertalib/jupiter/pull/3194)
- Fix Non system tests from running capybara/selenium [PR#3227](https:://github.com/ualbertalib/jupiter/pull/3227)
- Add patch for VCR for fixing VCR crashing in Ruby v3.1 [PR#3216](https:://github.com/ualbertalib/jupiter/pull/3216)
- Fix warning about already initialized constants in rake tests within tests [PR#3210](https:://github.com/ualbertalib/jupiter/pull/3210)
- Update dependabot.yml for npm updates to use correct javascript label [PR#3184](https:://github.com/ualbertalib/jupiter/pull/3184)

### Chores

- Bump rubocop-rails from 2.21.1 to 2.21.2 [PR#3254](https:://github.com/ualbertalib/jupiter/pull/3254)
- Bump sidekiq from 7.1.4 to 7.1.5 [PR#3253](https:://github.com/ualbertalib/jupiter/pull/3253)
- Bump core-js from 3.32.2 to 3.33.0  dependencies javascript [PR#3252](https:://github.com/ualbertalib/jupiter/pull/3252)
[...]

PR #3256, opened 2023-10-07
Remove bump infavor of a simple guide doc

murny:
Remove `bump` gem as it appears unused and/or doesn't work with our current process for cutting versions. For example, we don't create tags from `bump` and `bump` doesn't seem to work with our changelog format as of today.

I asked a question around `bump` gem here: https://github.com/ualbertalib/jupiter/pull/3255#issuecomment-1749460480

> Question to team:
> 
> Does anyone use the bump gem we have in our gemfile for cutting releases? I noticed we had:
> 
> No README/docs/guide on how to cut a release with bump.
> using bump had issues with "bumping" our changelog, perhaps due to a different formatting style
> My assumption is bump is probably not that useful, and no one probably uses it? So we should probably just remove it, in favour of a simple guide doc on how to cut a release?

So think we can just remove `bump` as it is not serving a useful purpose in our workflow. Instead, let's add a simple doc to help developers in the future for how to cut a new version of the application.


pgwillia:
> > This seems to duplicate https://github.com/ualbertalib/di_internal/blob/master/System-Administration/Release-Management.md#jupiter. Should we remove and just reference this doc?
> 
> Think this doc is fine from a release point of view. But this doc is missing the "how to" of actually cutting a release (no mention of bumping the `version.rb` for example). I thought I could use the `bump` gem, but it appears this wasn't the case. So I'm okay with updating this doc instead to go into a bit more detail and removing the doc in the PR? No strong feelings either way.

:sweat_smile: I actually meant that we keep this change, and don't repeat ourselves in the internal documentation by just linking to here.  What do you think about that?

PR #3266, opened 2023-10-16
Update bundler to v2.4.18 as this is what we use in staging/production

murny:
Neil mentioned he was getting the following error:

`There was an error installing the locked bundler version (2.3.12), rerun with the `--verbose` flag for more details. Going on using bundler 2.4.18.`

It appears Jupiter has been using v2.3.12 for over a year now (This was last updated here: https://github.com/ualbertalib/jupiter/commit/d1071c3a098eb3aebf4cdf3d0c39f213c04695c3)

This issue is staging/production are on v2.4.18 of bundler now. We not sure why these versions are different.

So let's update Jupiter to use v2.4.18 so everything is on the same version going forward.

 (NOTE: None of our projects are using this version. Doing a search in our organization, the only reference I see to this version was our last release, so perhaps it got updated in v2.5.2: https://github.com/ualbertalib/Unix/blob/f57eb7c280c6a8fea7d228e260e9c5063cdf6ecd/Change/2023/Q3/CR00001145.html#L357).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,19 @@
 # yarn lockfile v1
 
 
+"@aashutoshrathi/word-wrap@^1.2.3":
+  version "1.2.6"
+  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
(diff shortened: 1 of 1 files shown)


pgwillia (approved):
:+1: LGTM, but :thinking: wondering what's up with that.  Bundler is a topic that comes up too often being out of sync.

https://github.com/ualbertalib/library-cms/blob/41252f676825da18fa47645e31a0e23c65886578/Gemfile.lock#L302 and https://github.com/portagenetwork/roadmap/blob/48f391ec39f35885ca298b72d2229f19b884198c/Gemfile.lock#L686 are more recent, but not 2.4.18.

PR #3318, opened 2023-12-20
fix "include Pundit is deprecated. Please use include Pundit::Authorization instead." warnings

murny:
## Context

Fix pundit deprecation messages. 

You will see the following in our code once in a while:
image


ConnorSheremeta (approved):
LGTM!

Issue #3330, opened 2024-01-04
Hardcode "2023" is failing tests in master due to us being in 2024


murny:
It appears we got a failing CI on master right now. Looks like it's a dropdown for the Year that expects it to be in 2023, but now that the current year has changed to 2024, this test is failing consistently. 

![image](https://github.com/ualbertalib/jupiter/assets/1930474/4235b793-ceb4-4347-afcb-d7a7c36136fb)

Will investigate later this week/tonight unless someone beats me to it.

PR #3331, opened 2024-01-05
#3330: Fix new year sensitive tests

murny:
Embargo dropdowns only have current year + 5 years:

image

Tests are attempting to set this dropdown to 2023, which is causing failures.

Simplify the tests and make this never fail again.

#3330


ConnorSheremeta (approved):
:+1:

PR #3346, opened 2024-01-18
Upgrade rails to v7.1.3

murny:
Upgrades Jupiter Rails version from 7.0.8 to 7.1.3

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-0-to-rails-7-1

The upgrade itself is trivial, no real changes required.

However, Rails 7.1 allows the `lib` directory to be autoloaded by default. This causes some issues with our two files `jupiter.rb` and `jupiter/version.rb` as they don't conform to Zeitwerk standards.

- `jupiter.rb` tries to redefine creating a `Jupiter` constant that already exists. `application.rb` creates it before we start autoloading code, so Zeitwerk doesn't try to load it again which isn't ideal.
- `jupiter/version.rb` Zeitwerk expects this to define a `Jupiter::Version` constant which it doesn't. So need to refactor this. 

Note: Rails 7.1 gives one big deprecation warning whenever we use Rails.secrets. Created an issue in the backlog for this here: https://github.com/ualbertalib/jupiter/issues/3347

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,7 @@ gem 'connection_pool'
 gem 'pg', '~> 1.5.6'
 gem 'redis', '~> 4.8'
 gem 'rsolr'
-gem 'strong_migrations', '~> 1.7.0'
+gem 'strong_migrations', '~> 1.8.0'
 
 # Authentication
 gem 'bcrypt', '>= 3.1.13'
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -1,7 +1,5 @@
 require 'sidekiq/web'
 require 'sidekiq/cron/web'
-require 'admin_constraint'
```
ConnorSheremeta, line 3: I am curious about the change here removing these two lines, I can't seem to find anything relevant to this in the upgrade guide

ConnorSheremeta (commented):
Looks good, just a question regarding one of the changes

config/routes.rb
ConnorSheremeta: Makes sense, thank you for the clarification!

ConnorSheremeta (approved):
LGTM! :+1:

PR #3358, opened 2024-01-24
Bump version of ffi to 1.16.3

murny:
Neil was having C compiling errors when bulding the release image which was due to FFI

Appears this is a known issue in FFI on centos: https://github.com/ffi/ffi/issues/1052

It is apparently fixed in v1.16.3


pgwillia (approved):
:+1:

Issue #3277, opened 2023-10-23
Remove feature flags which have been fully rolled out for sometime


murny:
We should be removing feature flags which have been fully rolled out, and we have no intention of turning off again.

Batch Ingest is a good example, but there may be others:
![image](https://github.com/ualbertalib/jupiter/assets/1930474/faa8ed1a-6125-42a6-8562-bd20acfe8b50)

PR #3375, opened 2024-02-11
Remove fully enabled feature flags that have been enabled for years

murny:
https://github.com/ualbertalib/jupiter/issues/3277

This PR removes all the fully enabled feature flags that have been enabled for years:

image

`datacite_api` flag was removed here: https://github.com/ualbertalib/jupiter/issues/3277
`or_facts feature flag` I think is a duplicate and probably never worked

Note: There may be good reason to keep these flags around, such as being able to quickly switch back to old behaviour for some reason? So if so, let me know and we can unremoved these. 

After this is merged, we will have to remove these from Flipper UI in production as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Removed
 - Remove fully enabled feature flags that have been enabled for years [PR#3375](https://github.com/ualbertalib/jupiter/pull/3375)
-- 
+
 ### Chores
 - Update Bundler to v2.5.5 to match production [PR#3374](https://github.com/ualbertalib/jupiter/pull/3374)
 


ConnorSheremeta (dismissed):
LGTM. I wonder if keeping the batch_ingest flag would be worth it to restrict that feature to certain actors (a subset of admins)... however, I don't think that there's much benefit to that.

pgwillia (dismissed):
I spoke with @leahvanderjagt yesterday about this change.  She agrees that these are all behaviours we want to keep long term. :+1: 

Really good work to keep our code clean so what we're maintaining is as simple as possible!

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+### Removed
+- Remove fully enabled feature flags that have been enabled for years [PR#3375](https://github.com/ualbertalib/jupiter/pull/3375)
+-
```
ConnorSheremeta, line 14: Small formatting error from the rebase

ConnorSheremeta (approved):
:+1: Just a small formatting error from rebasing, could be resolved in the upcoming release PR

pgwillia (dismissed):

ConnorSheremeta (approved):
:+1:

PR #3393, opened 2024-02-27
Cut v2.8.0 release and update changelog

murny:
## Context

Bump v2.8.0 release and update changelog with latest changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
-## [2.8.0] - 2024-02-26
+## [2.8.0] - 2024-02-28
 
 ### Added
 * Migrate Webpacker to Esbuild [PR#3320](https://github.com/ualbertalib/jupiter/pull/3320)


CHANGELOG.md
```diff
@@ -9,14 +9,36 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+## [2.8.0] - 2024-02-28
+
 ### Added
-- Migrate Webpacker to Esbuild [PR#3320](https://github.com/ualbertalib/jupiter/pull/3320)
+* Migrate Webpacker to Esbuild [PR#3320](https://github.com/ualbertalib/jupiter/pull/3320)
```
ConnorSheremeta, line 15: I'm curious here about the switch to markdown bullets as `*` rather than `-` like we have further down in the changelog, maybe this was unintentional?

ConnorSheremeta: I support this change, making it easier to backfill through the “Generate release notes” feature and resolving the issue with similar bullets being used would be ideal. I think if we’re sticking to the use of `*` then it should all match for for consistency

ConnorSheremeta (approved):

Issue #3311, opened 2023-12-14
[EPIC] 


pgwillia:
# Epic / Very High Level User Story

## Step 1: Describe The Problem We're Here to Solve

We have 9 servers (SolrCloud nodes, Zookeeper, etc) supporting a end of life version of Solr (6.6) for just ERA now.  It would be a big win to be able to simplify this.  Coming up with a plan and estimates for that work would be a good first step.

New information enforcing this as a priority is that yesterday we had a power supply failure for one of the Solr servers.
Henry had the foresight to order extras the last time they were replaced so Neil could quickly resolve the issue. Because Solr is redundant there was no impact to ERA. We may have more difficulty finding replacement power supplies in the future. Wendy will see if she can find another spare.

Obviously, we're good in the short term, we can borrow parts from staging servers or let nodes go dark.  Longer term we should simplify this and retire the ageing hardware.

### Additional information:

[Any other information you can think of related to this! Links to discussion, other details that would be helpful to know, etc, etc](https://github.com/ualbertalib/jupiter/issues/2406)

### Elevator Pitch

(more information: https://agilewarrior.wordpress.com/2010/11/06/the-agile-inception-deck/)

A feature-oriented proposal for a solution in user-understandable language (and related artifacts like mockups, if those might be helpful), generated by Developers & Metadata specialists based on the problem description, for
approval by Product Owner & relevant stakeholders. Should address the following:

* What are we doing?

* What are we NOT doing?

* What resources will be needed to implement this?

### Refinement

(as much or as little as makes sense for the pitch)

##### Metadata

What metadata is involved? What is changing in terms of metadata tracked for specific entities? Links to documentation etc.

#### Backend

Things like whiteboard sketches of architectural changes, Rails model component interactions, pictures of the outcome of event storm sessions, etc

#### Frontend

Things like wireframes, sketches of potential UI approaches, etc

## Step 3: Tickets

Links to specific, achievable, estimable tickets with clear "Done" criteria, describing the work needed to implement the solution outlined above. By the time these are added to this document, they should be in a state ready for estimation.

pgwillia:
Right sizing the project will be important because it's sounding like we'll be moving to DSpace sometime in 2025 (I've heard December, July and Q1 in the last few weeks).  I think the driving requirements of the project are
* Solr Cloud is supported by several bare metal servers that are end of life and in the ECHA datacenter that will be decommissioned in the next year
[...]

PR #3416, opened 2024-03-17
#3311: Upgrade solr to v7.x

murny:
## Context

I upgraded to 7.0 first to verify everything was working. Was encountering the following error:

```
Can't load schema /opt/solr/server/solr/mycores/development/schema.xml: Setting default operator in schema (solrQueryParser/@defaultOperator) not supported
```

According to the Solr changelog:

> The defaultOperator parameter in the schema is no longer supported. Use the q.op parameter instead. This option had been deprecated for several releases. See the section [Standard Query Parser Parameters](https://solr.apache.org/guide/solr/latest/query-guide/standard-query-parser.html#standard-query-parser-parameters) for more information.

So apparently this has been deprecated for awhile.

I peaked at what Hyrax/Hyku/Samvera did, and they basically just removed the one setting that was using it:

https://github.com/samvera/hyku/pull/1440/commits/17d9a63e384264eafb9887b3d52ec5ad28677850?diff=unified&w=1#diff-c92f6a2e77d4bf367e25e88d6c2c477ea502ca0c819531d927d4c863025e4a9eL337-L342

Where they just removed this configuration:
```
 <!-- SolrQueryParser configuration: defaultOperator="AND|OR" -->
 <solrQueryParser defaultOperator="OR"/>
 ```
 
 This seems to unblock any troubles I had, Jupiter ran just fine on Solr v7.0 after this. I then bumped it up to the lastest 7.X version (v7.7) and this also worked fine.
 
 So this might be easier then I thought. 
 
 According to https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-7.html#upgrade-planning
 It sounds like our current indexed data for solr v6 will also work with v7. But we will eventually want to reindex our data when we try to upgrade to v8.
 
 NOTE: There was a 2-3 things that drifted in the hyrax/samvera schema.xml. Not sure if we should investigate these things or not?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,49 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+## [2.9.1] - 2024-05-16
+
+### Fixed
+* Improves performance of Active Storage lookups via adding an table index.
+  Requires a db:migration that creates the index concurrently thus does not
+  lock the DB table [PR#3428](https://github.com/ualbertalib/jupiter/pull/3428)
+* Fixed edge case on batch ingest front end tool where metadata is invalid [#3148](https://github.com/ualbertalib/jupiter/issues/3148)
+* Fix unsafe redirect errors by using allow_other_host [PR#3429](https://github.com/ualbertalib/jupiter/pull/3429)
+
 ### Changed
-* #3420: Update bundler to match version in production [PR#3433](https://github.com/ualbertalib/jupiter/pull/3433)
+* Update bundler to match version in production [PR#3433](https://github.com/ualbertalib/jupiter/pull/3433)
 
 ### Chores
 * Bump rubocop from 1.62.1 to 1.63.1 [PR#3439](https://github.com/ualbertalib/jupiter/pull/3439)
+* Bump json-schema from 4.1.1 to 4.2.0 by @dependabot in https://github.com/ualbertalib/jupiter/pull/3414
+* Bump esbuild from 0.20.1 to 0.20.2 by @dependabot in https://github.com/ualbertalib/jupiter/pull/3415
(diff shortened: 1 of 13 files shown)


jefferya:
>It sounds like our current indexed data for solr v6 will also work with v7. But we will eventually want to reindex our data when we try to upgrade to v8.

A word of caution on this point from my CWRC experiences: verify that the Solr v6 servers are not running a v5 index before updating production to v7.

pgwillia (dismissed):
LGTM for the necessary parts to upgrade in our dev/test environments.

I doubt that we need to look into the 2-3 things that drifted in the hyrax/samvera schema.xml, but what are they?

I agree that the v6 and v7 indexes are compatible and after that, we can reindex to prepare for v8. Would be good to coordinate with @nmacgreg and perhaps @henryzhang87 on how the process to upgrade Solr would look in our environment.  They'll probably want to be familiar with
* https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-7.html#upgrade-planning
* https://solr.apache.org/guide/7_0/upgrading-a-solr-cluster.html#upgrading-a-solr-cluster
* https://solr.apache.org/guide/7_0/taking-solr-to-production.html#taking-solr-to-production
@murny will you make sure that @nmacgreg is in the loop, doing as much of the investigation as possible for a smooth hand-off?

pgwillia:
> > I doubt that we need to look into the 2-3 things that drifted in the hyrax/samvera schema.xml, but what are they?
> 
> There were three things I noticed in their Solr 7 version of schema.xml:
> 
> They do not have the multiterm analyzer : image

I don't think we use this anywhere.

> They use "string" instead of "alphaSort" for type on `*_ssi` field: image
> 

We use `*_ssi` for our title field.  Might need more investigation.
 
> They do not have Hierarchical pathing support: image
>

We use `*_dpsim` for our `member_of_paths` community/collection field.  Would also need more investigation.
 
> This is based on Hyku (but also see the same changes in Hyrax)

pgwillia (approved):
:+1:

Issue #2826, opened 2022-04-19
Problem Performance (very busy database)


pgwillia:
**Describe the bug**
Neil observed a very busy database via Jupiter starting April 17th, it is intermittent and ongoing.  Skylight observes the same trend https://www.skylight.io/app/applications/Q2pLKz0AwRat/recent/1d/endpoints which gives us some things to look into from our end. Beau and I deployed Jupiter 2.3.5 on April 11th with a new version of pg-client bumped from 9 to 12 is the other thing that might be relevant.   Coincident with a user depositing their journal articles (>100) one at a time.
>We had trouble this afternoon, with the primary database server for Jupiter.  Nagios clearly reports the database load average was high, ~12:30 - 14:30 (and other incidents today). Netdata on the database server reports a larger than normal  number of database sessions open (60), and a lot of queries. I went grepping through the incoming web queries on the HAProxy front-end, but found just a handful of MSNBot IP addresses, each clocking up just a few low-hundreds of queries each.
(I reached a dead-end -- it seems that normal bot activity makes the DB go crazy, so now I'm throwing around speculation!)

![image](https://user-images.githubusercontent.com/1220762/164068779-7c3764cc-bcd3-498a-b33a-10158b362427.png)
![image](https://user-images.githubusercontent.com/1220762/164068826-a39376fd-6515-4865-b16c-458c0ba7131f.png)

![image](https://user-images.githubusercontent.com/1220762/164068868-e676e8e0-d5bd-4eca-b7b4-08d0a683f895.png)
![image](https://user-images.githubusercontent.com/1220762/164068899-bad0f6be-1191-4bae-8df0-814d27dd1dca.png)

nmacgreg:
This problem occurred repeatedly over Christmas, 2023, making it a little less cheerful. 
Bump!

nmacgreg:
Occurance Sept, 2023: https://helpdesk.library.ualberta.ca/otrs/index.pl?Action=AgentTicketZoom;TicketID=67376

nmacgreg:
Occurrence Xmas '23 + Jan. '24: https://helpdesk.library.ualberta.ca/otrs/index.pl?Action=AgentTicketZoom;TicketID=68358

[...]

PR #3428, opened 2024-03-27
Improve performance of Active Storage lookups #2826

jefferya:
Address, partially, #2826 

Active Storage attachments database table appearing in the slow query log.

Requires: a `bin/rails db:migration`

The change adds an index on the `record_id` of the `active_storage_attachments` table to improve performance. The index is added via a DB migration that concurrently added the index while not locking the DB table. Quoting the PostgreSQL documentation:
"CONCURRENTLY When this option is used, PostgreSQL will build the index without taking any locks that prevent concurrent inserts, updates, or deletes on the table; whereas a standard index build locks out writes (but not reads) on the table until it's done. There are several caveats to be aware of when using this option — see [Building Indexes Concurrently](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)"

When the active_storage_attachments contains a significant number of records (10k), queries on the table by record_id are causing a sequential DB scan on the table. By adding an index, performance improves on queries like the following from the PostgreSQL slow query log:

```
explain analyze 
  SELECT "active_storage_attachments".* 
  FROM "active_storage_attachments" 
  WHERE
    "active_storage_attachments"."record_type" = 'Item'
    AND "active_storage_attachments"."name" = 'files'
    AND "active_storage_attachments"."record_id" = '0625726d-cc85-40cd-8c54-7e7e65d06df6';
```

https://github.com/ualbertalib/jupiter/issues/2826#issuecomment-1934429981


db/migrate/20240327203605_active_storage_attachments.rb
```diff
@@ -0,0 +1,6 @@
+class ActiveStorageAttachments < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
```
jefferya, line 2: Disable transaction as incompatible with "concurrent" index creation.

db/migrate/20240327203605_active_storage_attachments.rb
```diff
@@ -0,0 +1,6 @@
...
+  def change
+    add_index :active_storage_attachments, :record_id, algorithm: :concurrently, if_not_exists: true
+  end
```
jefferya, line 5: Uses "CREATE INDEX CONCURRENTLY" to avoid locking the DB table. For details: https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

ConnorSheremeta (dismissed):
LGTM! I added the schema changes from the migration and did some reformatting in the changelog. I can confirm that the addition of this index has had an impact on queries that were previously slow to complete. I ran a batch ingest locally with 1000 theses and logged sql queries that were not performant through the auto_explain module setting the min_duration threshold, I then added this index and noted that there were no queries in the logs from auto_explain using the same min_duration threshold after the batch ingest had completed a second time (ensured caches were cleared.) Will leave this up for a couple more days just in case there is another review on this and then will merge

lagoan (dismissed):
LGTM

Issue #3423, opened 2024-03-25
ActionController::Redirecting::UnsafeRedirectError: Unsafe redirect to "https://era.library.ualberta.ca/items/3bcc9b7f-5b22-4cf4-ba4c-ede791eb3d1c/view/e35dcade-8076-487...", pass allow_other_host: true to redirect anyway.


pgwillia:
View details in Rollbar: [https://app.rollbar.com/a/ualbertalib/fix/item/jupiter/1849](https://app.rollbar.com/a/ualbertalib/fix/item/jupiter/1849)

```

ActionController::Redirecting::UnsafeRedirectError: Unsafe redirect to "https://era.library.ualberta.ca/items/3bcc9b7f-5b22-4cf4-ba4c-ede791eb3d1c/view/e35dcade-8076-487...", pass allow_other_host: true to redirect anyway.
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/actionpack-7.1.3/lib/action_controller/metal/redirecting.rb", line 199, in _enforce_open_redirect_protection
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/actionpack-7.1.3/lib/action_controller/metal/redirecting.rb", line 95, in redirect_to
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/actionpack-7.1.3/lib/action_controller/metal/flash.rb", line 62, in redirect_to
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/actionpack-7.1.3/lib/action_controller/metal/instrumentation.rb", line 49, in block in redirect_to
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/activesupport-7.1.3/lib/active_support/notifications.rb", line 206, in block in instrument
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/activesupport-7.1.3/lib/active_support/notifications/instrumenter.rb", line 58, in instrument
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/activesupport-7.1.3/lib/active_support/notifications.rb", line 206, in instrument
  File "/var/www/sites/jupiter/vendor/ruby/3.1.0/gems/actionpack-7.1.3/lib/action_controller/metal/instrumentation.rb", line 48, in redirect_to
  File "/var/www/sites/jupiter/app/controllers/application_controller.rb", line 121, in redirect_back_to
[...]

Issue #3419, opened 2024-03-19
Fix unsafe redirect errors from Rollbar


murny:
We are getting a few of these:

![image](https://github.com/ualbertalib/jupiter/assets/1930474/bd0ce74c-422d-424a-bfdd-234ca6f4c936)

We are redirecting to a generic `request.referer` which in Rails 7 is now an error since we are redirecting to an external URL.

We should look into improving this logic. Either approvelist the possible URLs here or just allow this behavior via `allow_other_host: true`

Maybe look into `redirect_back_or_to` method which might simplify this logic as well.

PR #3429, opened 2024-03-29
#3419: Fix unsafe redirect errors by using `allow_other_host`

murny:
This should be a quick fix to the following issues:

- https://github.com/ualbertalib/jupiter/issues/3423
- https://github.com/ualbertalib/jupiter/issues/3419

Both having to do with Rails 7 new logic to prevent "unsafe redirects" to external websites.

In the first issue, the external website appears to be ERA's domain such as `https://era.library.ualberta.ca/items/3bcc9b7f-5b22-4cf4-ba4c-ede791eb3d1c/view/e35dcade-8076-487...`, so perhaps maybe just need to make this recognized as a internal URL instead?  

In the second case it's our OAuth flow so the website is redirecting to `https://login.ualberta.ca/`

Both of these seem safe (since we are setting these urls and its not coming from user input). So `allow_other_host` seems like a quick fix to resolve these issues.

But, curious if there is a better way to handle this? Like, could you "approvelist" these domains/urls? (Doesn't appear to be a way to do this from first glance). 

Also, maybe we should more gracefully handle these errors in the future: https://github.com/rails/rails/commit/c3758a71af949db849d5b7f176677653e4e4fae9
But these seem to be the last `redirect_to` that could be "external" urls so maybe don't need this.

UPDATE: We ended up refactoring both use cases to just "redirect" better, so we shouldn't need the `allow_other_host` argument anymore. Also added a catch all that would just redirect back to root page if we somehow missed something.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 * Bump rubocop from 1.62.1 to 1.63.1 [PR#3439](https://github.com/ualbertalib/jupiter/pull/3439)
 
 ### Fixed
-* #3419: Fix unsafe redirect errors by using allow_other_host [PR#3429](https://github.com/ualbertalib/jupiter/pull/3429)
+* #3429: Fix unsafe redirect errors by using allow_other_host [PR#3429](https://github.com/ualbertalib/jupiter/pull/3429)
 
 ## [2.9.0] - 2024-03-14
 


app/controllers/application_controller.rb
```diff
@@ -83,7 +83,7 @@ def user_not_authorized
       # referer gets funky with omniauth and all the redirects it does,
       # so handle this sanely by ignoring any referer coming from omniauth (/auth/) path
       if request.referer&.exclude?('auth')
-        redirect_to request.referer
+        redirect_to(request.referer, allow_other_host: true)
```
pgwillia, line 86: `request.referer` could be user input.  Do you have concerns about this with `allow_other_host: true`?

pgwillia (commented):
This is a really elegant solution to a puzzle that had me stumped last week.  Thanks for noticing the change in rails and following up on this! 🙏

app/controllers/application_controller.rb
pgwillia: I think you're right and `redirect_back_or_to(root_path)` is what we want to use.

I think the `allow_other_host: true` solution has some risk to it, especially in combination with `request.referrer`, which is why the default changed in Rails.

app/controllers/application_controller.rb
```diff
@@ -118,7 +111,8 @@ def render_400(exception = nil)
   end
 
   def redirect_back_to
-    redirect_to session.delete(:forwarding_url) || session.delete(:previous_user_location) || root_path
+    redirect_to(session.delete(:forwarding_url) || session.delete(:previous_user_location) || root_path,
```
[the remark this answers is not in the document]
pgwillia: I like that approach.  Perhaps we could do this in combination with your suggestion above for adding
```
rescue_from ActionController::Redirecting::UnsafeRedirectError do
   redirect_to root_url
end
```
The redirect should always be safe so we shouldn't see this, but at least it would avoid the 500 error if we're wrong?🙃

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 ### Chores
 * Bump rubocop from 1.62.1 to 1.63.1 [PR#3439](https://github.com/ualbertalib/jupiter/pull/3439)
 
+### Fixed
+* #3419: Fix unsafe redirect errors by using allow_other_host [PR#3429](https://github.com/ualbertalib/jupiter/pull/3429)
```
pgwillia, line 25:
```suggestion
* #3429: Fix unsafe redirect errors by using allow_other_host [PR#3429](https://github.com/ualbertalib/jupiter/pull/3429)
```

pgwillia (dismissed):
Looks good to me :+1:  I feel good about not going the `allow_other_hosts` route.  Thanks for thinking this through :)

pgwillia (approved):

Issue #3420, opened 2024-03-19
Update bundler (v2.5.5) to match production (v2.5.6)


murny:
We should update our bundler version we are using to create our gemfile.lock so it matches production

PR #3433, opened 2024-04-02
#3420: Update bundler to match version in production

murny:
## Context

https://github.com/ualbertalib/jupiter/issues/3420

Neil mentioned production/staging bundler version is at v2.5.6. So we should update our local version to match this.


ConnorSheremeta (commented):
Missing a changelog entry

pgwillia (approved):
LGTM! :+1:

PR #3442, opened 2024-04-12
Upgrade solr to 8

murny:
## Context
Similar to the work done in https://github.com/ualbertalib/jupiter/pull/3416, I first upgraded Solr to v8.0 off this branch.

There was an error arround `solr.StandardFilterFactory`. This is apparently a no-op and hasn't been doing anything since [Solr 3.1 as mentioned here](https://solr.apache.org/guide/7_0/filter-descriptions.html#standard-filter).
We can safetly remove this as a result.

With this option removed, Solr v8.0 worked just fine. I upgraded then to Solr v8.11.X and everything continued to work as well.

I also compared what Hyrax did to get to Solr v8 and they took the same approach for this. 

[Solr v8.11 is the oldest version to still be supported](https://solr.apache.org/downloads.html#about-versions-and-support) and is not end of life. So this version will get us critical security updates.

NOTE: Leaving this in DRAFT until https://github.com/ualbertalib/jupiter/pull/3416 is merged first and production Solr index has been reindexed on Solr 7.


pgwillia (approved):
:+1:  for whenever this needs merging.

jefferya (approved):

Issue #3495, opened 2024-06-13
Change target community destination for theses ingest.


lagoan:
**Context**

The faculty that receives the ingested theses has changed its name. We need to update the rake task so it deposits them in the correct community called "Graduate and Postdoctoral Studies (GPS), Faculty of".

PR #3496, opened 2024-06-13
Change the target community for theses ingestion

lagoan:
The previous rake task required the target community to be called
"Graduate Studies and Research, Faculty of". This changes the target
community to "Graduate and Postdoctoral Studies (GPS), Faculty of" which is the current community that will receive the theses for ingestion.

## Context

A new community will recevie the ingested theses due to a change in name for the faculty. We need to change the target name in the code to complete this task.

Related to #3495

## What's New

Changed the string for checking community name in rake task.


pgwillia (approved):
LGTM! 👍

PR #3533, opened 2024-08-06
docker compose command has changed, updated to newest command

murny:
Seeing some fails on recent pull requests for our CI pipeline due to:
image

`docker-compose` command has been deprecated for a while, perhaps GitHub has updated their docker on their CI servers to a later version where this command no longer exists.

We should be using `docker compose` instead for now on.


pgwillia (approved):
LGTM! 👍

Issue #3676, opened 2025-01-14
GitHub Actions Change broke CI


pgwillia:
A bunch of breaking changes
* ubuntu-latest label to ubuntu 24 starting on December 5, 2024 and ending on January 17, 2025. The ubuntu 24 image has a different set of tools and packages than ubuntu 22. We have made cuts to the list of packages so that we can maintain our SLA for free disk space. This may break your workflows if you depend on certain packages that have been removed. Please review this [list](https://github.com/actions/runner-images/issues/10636) to see if you are using any affected packages.
* Artifact actions v3 will be closing down by January 30th, 2025
* v1-v2 of actions/cache closing down Feb 1

https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/

This impacted us.  Seems ImageMagick isn't available in the new OS
```
Error:
PageLayoutHelperTest#test_page_image_url_should_return_community/item_logo:
TypeError: no implicit conversion of nil into String
    test/helpers/page_layout_helper_test.rb:94:in `+'
    test/helpers/page_layout_helper_test.rb:94:in `block in <class:PageLayoutHelperTest>'

bin/rails test test/helpers/page_layout_helper_test.rb:88
```
```
Error:
PageLayoutHelperTest#test_thumbnail_path_should_return_preview_for_image_(Variable):
MiniMagick::Error: You must have ImageMagick or GraphicsMagick installed
    test/helpers/page_layout_helper_test.rb:126:in `block in <class:PageLayoutHelperTest>'

bin/rails test test/helpers/page_layout_helper_test.rb:118
```

pgwillia:
Fixed!

PR #3677, opened 2025-01-14
Pin ubuntu to 22.04

pgwillia:
## Context

For stability through to the end of the project.  latest doesn't have ImageMagick.

Related to #3676 

## What's New

Pin GitHub actions to ubuntu 22.04


lagoan (approved):
LGTM

Ubuntu 22.04 reaches EOL in 2027, we should have migrated by then 😄

PR #3697, opened 2025-02-03
Bump ruby 3.3

pgwillia:
## Context

This morning breakman warned us that 
> Support for Ruby 3.1.4 ends on 2025-03-31

Related to https://github.com/ualbertalib/library_applications_development/issues/163

image

## What's New

Bump Ruby to 3.3 in Gemfile, working dev environment, and part of CI strategy (include 3.4 as well).

Not done: docker environment

PR #3698, opened 2025-02-03
Remove brakeman's ruby EOL check

pgwillia:
## Context

Temporary so that CI passes in case we need security release.  Also working on bump to Ruby 3.4

Related to #3697 

## What's New

Instruct brakeman to skip CheckEOLRuby


jefferya (approved):
LGTM!

Issue #3758, opened 2025-03-12
Make Collection and its member objects read-only


pgwillia:
**Is your feature request related to a problem? Please describe.**
@leahvanderjagt asks "can you guys help us stop deposits or edits for collections - can you impose freezes for collections or communities as we go?"

**Describe the solution you'd like**
Service team identifies a collection we'd like to make read only in preparation to move the data to Scholaris dev.
Dev team marks a collection "frozen" from the rails console which updates all the member_objects.
Users visiting the collections or items edit page will be redirected to the view page
Dev team can generate SAF package for that collection and have confidence that there won't be changes

**Describe alternatives you've considered**
We also discussed
1. A collection freeze managed with communication strategies only
2. A collection freeze managed with communication strategies only starting with the most stagnant collections
3. Putting the application in read-only mode for a longer duration. 
4. Generating replacement SAF packages based on delta between package creation and application freeze
 
**Additional context**
The larger the delta, the greater the risk.  New items would be easy, metadata changes present some challenges,and bitstream changes present the biggest challenge.

pgwillia:
![Image](https://github.com/user-attachments/assets/4318c490-6754-4422-81a2-47788579b790)

pgwillia:
![Image](https://github.com/user-attachments/assets/93d4be9c-a584-47c0-9e19-65238b6b9456)

PR #3761, opened 2025-03-13
Make Collection and its member objects read-only

pgwillia:
## Context

As we work through the Scholaris migration, we want the ability to freeze one collection, or batches of collections at a time.

Related to #3758 

## What's New

* Database migrations to add `read_only` column
* Change controller to redirect to show page from edit page when `read_only` is set
* Test of the behavior
* CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 * pin chrome so ci passes [3764](https://github.com/ualbertalib/jupiter/issues/3764)
 
 ### Added
-* read_only behavior to items, thesis and collection [[#3758](https://github.com/ualbertalib/jupiter/issues/3758)]
+* read_only behavior to items, thesis and collection [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
 
 ## 2.10.3 - 2025-03-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 ### Changed
 * pin chrome so ci passes [3764](https://github.com/ualbertalib/jupiter/issues/3764)
 
+### Added
+* read_only behavior to items and thesis [[#3758](https://github.com/ualbertalib/jupiter/issues/3758)]
+
 ## 2.10.3 - 2025-03-06
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Added
 * read_only behavior to items, thesis and collection [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
+* rake task to freeze collection, item or collections [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
 
 ## 2.10.3 - 2025-03-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ### Added
 * read_only behavior to items, thesis and collection [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
-* rake task to freeze collection, item or collections [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
+* rake task to freeze collection, item or collections, or to unfreeze everything [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
 
 ## 2.10.3 - 2025-03-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ New entries in this file should aim to provide a meaningful amount of informatio
 * pin chrome so ci passes [3764](https://github.com/ualbertalib/jupiter/issues/3764)
 
 ### Added
-* read_only behavior to items and thesis [[#3758](https://github.com/ualbertalib/jupiter/issues/3758)]
+* read_only behavior to items, thesis and collection [[#3758](https://github.com/ualbertalib/jupiter/issues/3758)]
 
 ## 2.10.3 - 2025-03-06
 
--- a/test/controllers/admin/collections_controller_test.rb
+++ b/test/controllers/admin/collections_controller_test.rb
@@ -53,7 +53,7 @@ class Admin::CollectionsControllerTest < ActionDispatch::IntegrationTest
     get edit_admin_community_collection_url(@community, @collection)
 
     assert_redirected_to admin_community_collection_url(@community, @collection)
-    assert_equal 'This item is frozen and cannot be edited.', flash[:alert]
+    assert_equal 'This collection is frozen and cannot be edited.', flash[:alert]
 
     @collection.read_only = false
     @collection.save!
--- a/test/controllers/admin/collections_controller_test.rb
+++ b/test/controllers/admin/collections_controller_test.rb
@@ -46,6 +46,19 @@ class Admin::CollectionsControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test 'shouldnt edit read_only collection' do
+    @collection.read_only = true
+    @collection.save!
+
+    get edit_admin_community_collection_url(@community, @collection)
+
+    assert_redirected_to admin_community_collection_url(@community, @collection)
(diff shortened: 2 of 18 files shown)


test/models/draft_item_test.rb
```diff
@@ -210,6 +210,21 @@ class DraftItemTest < ActiveSupport::TestCase
...
+    Collection.create!(title: 'Risque fantasy Books',
+                       owner_id: user.id,
+                       restricted: true,
+                       community_id: @community.id)
+    draft_item.assign_attributes(
+      member_of_paths: { community_id: [@community.id], collection_id: [@collection.id] }
+    )
+
+    assert_not draft_item.valid?
+    assert_equal ['Collection is frozen and not available for deposit'], draft_item.errors.messages[:member_of_paths]
```
pgwillia, line 224: image

test/controllers/items_controller_test.rb
```diff
@@ -83,4 +83,18 @@ class ItemsControllerTest < ActionDispatch::IntegrationTest
...
 
+  test 'shouldnt be able to edit an item if its read_only' do
+    @item.read_only = true
+    @item.save!
+
+    sign_in_as @admin
+    get edit_item_url(@item)
+
+    assert_redirected_to item_path @item
+    assert_equal 'This item is frozen and cannot be edited.', flash[:alert]
```
pgwillia, line 94: image

pgwillia: [Screencast from 2025-03-18 03:08:21 PM.webm](https://github.com/user-attachments/assets/cca905e5-fc6f-43f5-b0d2-1d5425888830)

test/controllers/admin/items_controller_test.rb
```diff
@@ -27,6 +27,19 @@ class Admin::ItemsControllerTest < ActionDispatch::IntegrationTest
...
   end
 
+  test 'shouldnt destroy read_only item' do
+    @item.read_only = true
+    @item.save!
+
+    delete admin_item_url(@item)
+
+    assert_redirected_to item_path @item
+    assert_equal 'This item is frozen and cannot be edited.', flash[:alert]
```
pgwillia, line 37: [Screencast from 2025-03-18 03:13:09 PM.webm](https://github.com/user-attachments/assets/f9cf2d5c-93d5-484d-9407-801c961259a5)

test/controllers/admin/collections_controller_test.rb
```diff
@@ -46,6 +46,19 @@ class Admin::CollectionsControllerTest < ActionDispatch::IntegrationTest
...
   end
 
+  test 'shouldnt edit read_only collection' do
+    @collection.read_only = true
+    @collection.save!
+
+    get edit_admin_community_collection_url(@community, @collection)
+
+    assert_redirected_to admin_community_collection_url(@community, @collection)
+    assert_equal 'This collection is frozen and cannot be edited.', flash[:alert]
```
pgwillia, line 56: image

test/models/draft_item_test.rb
```diff
@@ -210,6 +210,21 @@ class DraftItemTest < ActiveSupport::TestCase
 
     assert_predicate draft_item, :valid?
 
+    @collection.read_only = true
+    @collection.save!
+    Collection.create!(title: 'Risque fantasy Books',
```
ConnorSheremeta, line 215: Unsure what the creation of this collection is for?

test/controllers/admin/collections_controller_test.rb
ConnorSheremeta: Might be worthwhile to reference the i18n entry (like on line 31 of this file) for "This collection is frozen and cannot be edited." here to make it easier to change this string in the future.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 ### Changed
 * pin chrome so ci passes [3764](https://github.com/ualbertalib/jupiter/issues/3764)
 
+### Added
+* read_only behavior to items, thesis and collection [[#3758](https://github.com/ualbertalib/jupiter/issues/3758)]
```
ConnorSheremeta, line 16: Small formatting issue: extra `[` and `]` here.

test/integration/add_to_preservation_queue_task_test.rb
```diff
@@ -29,7 +29,8 @@ class AddToPreservationQueueTaskTest < ActionDispatch::IntegrationTest
         Community.first.save!
         Rake::Task['jupiter:preserve_all_collections_and_communities'].execute(after_date_arguments)
 
-        assert_equal 2, Jupiter::Redis.current.zcard(Rails.application.secrets.preservation_queue_name)
+        # the community, collection and four items all have save actions that trigger the preservation step
+        assert_equal 6, Jupiter::Redis.current.zcard(Rails.application.secrets.preservation_queue_name)
```
ConnorSheremeta, line 33: Does this test assertion change have any bigger implications that go along with it? I assume that it would be valid for the items themselves to trigger the preservation step here?  :thinking:

ConnorSheremeta (commented):
Some small things I noticed/questions. One thing I wonder is if this would be better off handled through a rake task rather than in the console. Rake task(s) to manage what is considered read_only might be less prone to errors compared to working in the console. Looks good overall!

test/integration/add_to_preservation_queue_task_test.rb
pgwillia: https://github.com/ualbertalib/jupiter/pull/3761/files#diff-cd41c1e2f5b95f15b54f7ff160fb7a828f112ae8deef7ba12ee561ee6fe62202R60

This is the change that changed this test.  Now `update` is called on each of the member objects and hence the `after_save :push_entity_for_preservation` call back is triggered.  https://github.com/ualbertalib/jupiter/blob/be01f8623cc8949f2d0623d42b0c7f0c24eacde2/app/models/item.rb#L23

test/models/draft_item_test.rb
pgwillia: I think this is the result of a bad rebase.  I think I didn't end up using this code snippet.  Thanks for noticing!

test/controllers/admin/collections_controller_test.rb
pgwillia: I thought about that and ended up not doing it because I thought that the test shouldn't demonstrate inside knowledge of the application.  But, since that's already how we've done it we should be consistent.  Thanks for noticing!

pgwillia:
> Some small things I noticed/questions. One thing I wonder is if this would be better off handled through a rake task rather than in the console. Rake task(s) to manage what is considered read_only might be less prone to errors compared to working in the console. Looks good overall!

I can write a rake task.  @lagoan how do you think it'll be used?  Give the rake task a list of uuids in a rake file.

test/controllers/admin/collections_controller_test.rb
ConnorSheremeta: That's fair, there has been some back and fourth on this in the past and I feel like our codebase probably has a mixture of both approaches at this point.

ConnorSheremeta (approved):
LGTM, another pair of eyes wouldn't be a bad idea however. Only thing pending is adding rake task(s) surrounding this

ConnorSheremeta (dismissed):
🎊 👍 Looks ready to merge to me!

ConnorSheremeta:
Had a discussion surrounding unfreezing objects/collections in stand-up: if this is something that will be done it may be worthwhile adding it to a rake task, there would be some logic there surrounding unfreezing a collection but ensuring it's member objects remain frozen if they happen to be apart of a frozen collection (if thats how the logic should work.) It might be worthwhile too to have a task to freeze/unfreeze all collections/objects. These may be one-off tasks that could be done in the console however as needed 🤔

pgwillia:
> Had a discussion surrounding unfreezing objects/collections in stand-up: if this is something that will be done it may be worthwhile adding it to a rake task, there would be some logic there surrounding unfreezing a collection but ensuring it's member objects remain frozen if they happen to be apart of a frozen collection (if thats how the logic should work.) It might be worthwhile too to have a task to freeze/unfreeze all collections/objects. These may be one-off tasks that could be done in the console however as needed 🤔

I'm sorry I missed the discussion. In the collection after the save callback, I hadn't considered checking if any of the member items belonged to a frozen collection. I'll fix the callback and add a rake task to unfreeze all.  Hope we don't need the unfreeze functionality.

ConnorSheremeta (approved):
Looks good!

PR #3772, opened 2025-03-27
Updates the freeze collection rake task

pgwillia:
## Context
@ConnorSheremeta noticed deficiencies when testing the 2.11.0 deployment in the staging environment.

Related to # (issue)

## What's New
* Ability to freeze individual theses.
* More performant unfreeze all by focusing on only the objects that are set to read_only and skipping callbacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ New entries in this file should aim to provide a meaningful amount of informatio
 
 ## [Unreleased]
 
+### Changed
+* fixed freeze collection rake tasks [#3758](https://github.com/ualbertalib/jupiter/issues/3758)
+
 ## 2.11.0 - 2025-03-25
 
 ### Changed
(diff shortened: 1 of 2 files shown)


ConnorSheremeta (dismissed):
Force pushing to a branch sometimes throws off CI which is why the rubocop checks are still failing. This PR looks good however! Two nice to haves which have came to mind: console outputs in the rake tasks to confirm that it has started/ended/progressed and doing the multi-item tasks in a transaction, making it all-or-nothing in case it errors out. This change in itself looks good however and should speed up the unfreeze_all task! :clap:

pgwillia:
> Force pushing to a branch sometimes throws off CI which is why the rubocop checks are still failing. This PR looks good however! Two nice to haves which have came to mind: console outputs in the rake tasks to confirm that it has started/ended/progressed and doing the multi-item tasks in a transaction, making it all-or-nothing in case it errors out. This change in itself looks good however and should speed up the unfreeze_all task! 👏

The CI failure was because rubocop was cautioning against the use of `update_columns`.  Disabled that cop for our task.

Also added some console output for the rake tasks and wrapped the tasks in transactions.

ConnorSheremeta (approved):
:+1:

Issue #3347, opened 2024-01-18
Find replacement for `secrets.yml` as it's deprecated


murny:
From recent #3346 we are now getting the following deprecation warning:

```
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2.
```

We should maybe move to credentials? But this requires a potential organizational shift as Jupiter would hold all the secrets and not Ansible, and we will use a MASTER_KEY to decrypt it.

https://edgeguides.rubyonrails.org/security.html#custom-credentials

pgwillia:
We won't fix it.  Jupiter will be decommissioned in January 2025 after moving to Scholaris, the national repository of DSpace instances hosted by Scholars Portal.

pgwillia:
From brakeman
```
Confidence: Weak
Category: Unmaintained Dependency
Check: EOLRails
Message: Support for Rails 7.1.5.1 ends on 2025-10-01
```

PR #3821, opened 2025-08-12
Update brakeman.yml

pgwillia:
## Context

CI is failing on this check.  Ignore it because we know we won't be able to fix it, so we'll decommission it instead.

Related: https://github.com/ualbertalib/jupiter/issues/3347

## What's New

Disable Brakeman EOLRails


jefferya (approved):
Sounds good to me.

